PR #172, opened 2015-01-21
Update LocalizationService (v6+), adding all Methods

st3v3nv:
I've added all Methods to the Umbraco.Core.Services.LocalizationService documentation


Shazwazza:
great!

Shazwazza:
think there's some new ones coming in 7.2.2 and 7.3 but that's ok, we'll fill them in once they're released

PR #173, opened 2015-01-23
Update Tags.md

bootnumlock:
Just adding some reference on how to use tags in v7.15 and later to get a count and list the actual tag text.

### DIFF
--- a/Documentation/Using-Umbraco/Backoffice-Overview/Property-Editors/Built-in-Property-Editors/Tags.md
+++ b/Documentation/Using-Umbraco/Backoffice-Overview/Property-Editors/Built-in-Property-Editors/Tags.md
@@ -115,6 +115,18 @@ In /config/xsltExtensions.config add the following to register the library
 	    </xsl:for-each>
 	  </ul>    
 	</xsl:if>
+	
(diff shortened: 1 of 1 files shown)


Shazwazza:
Hi @bootnumlock !

There's a few things incorrect here:

* There is a doc's page for v7 prop editors here: http://our.umbraco.org/documentation/Using-Umbraco/Backoffice-Overview/Property-Editors/Built-in-Property-Editors-v7/ - so v7 prop editor stuff needs to go there instead 
* You should never really use the Services on ApplicationContext for anything on the front-end since this is business logic stuff that interacts with the database. The front-end should use UmbracoHelper as a read-only way of displaying data. You should use @Umbraco.TagQuery which returns an instance of https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/TagQuery.cs which contains methods to work with tags on the front-end.

hemraker:
Hi @bootnumlock, closing this as the pr hasn't been updated based on feedback or to reflect new docs structure.

Good reminder though that property editors needs an overhaul as eg. *tags* is not documented.

Thanks!

PR #176, opened 2015-02-04
Update ContentService - added all missing methods

st3v3nv:
Added new/missing methods, completing the description of the ContentService (for now).
Sorted all methods, similar to the (order of the) methods in the ContentService.


Shazwazza:
thx!

PR #188, opened 2015-04-08
Update index.md

BEllis:
- Added full namespace to types as some conflict with classes in other namespaces.
- Fixed the MemberAuthoize attribute example as it is incorrect (should be = instead of :)


Shazwazza:
thx!

PR #191, opened 2015-05-11
Fixed the API paths after the breaking change in 7.0.2

IanGrainger:
As discussed here: https://our.umbraco.org/forum/umbraco-7/using-umbraco-7/64690-Simple-package-not-available-TextValue-Dropdown-List?p=1#comment218791 and here: https://our.umbraco.org/forum/umbraco-7/developing-umbraco-7-packages/48668-Creating-an-UmbracoAuthorizedJsonController some of the documentation hasn't been updated after this change to start all API paths with "umbraco/backoffice" in 7.0.2 http://umbraco.com/follow-us/blog-archive/2014/1/17/heads-up,-breaking-change-coming-in-702-and-62.aspx. I've done that here, though not searched the rest of the documentation for any other places it might need to happen.

### DIFF
--- a/Documentation/Extending-Umbraco/Property-Editors/creating-tutorial4-v7.md
+++ b/Documentation/Extending-Umbraco/Property-Editors/creating-tutorial4-v7.md
@@ -84,7 +84,7 @@ Inside the GetAll() method, we now write a bit of code, that connects to the dat
 	//fetch data from DB with the query and map to Person object
 	return db.Fetch<Person>(query);
 
-We are now done with the server-side of things, with the file saved in app_code you can now open the Url: /umbraco/My/PersonApi/GetAll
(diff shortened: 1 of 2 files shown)


Shazwazza:
Thanks for this, but to be complete you also need to add that another attribute is required on the PersonApiController class: [IsBackOffice] in order for it to route to /backoffice and be authenticated correctly.

IanGrainger:
Can you add that then please? I don't know where that attribute needs to
go. I only tested the tutorial, as that's what I was doing when I found the
problem in the documentation.

Shazwazza:
sure, no problem, thanks for submitting this. i'll fix up the attribute.

Shazwazza:
Oops, sorry I was confusing myself, no attribute necessary. Thanks again :)

PR #193, opened 2015-07-08
English Syntax edit

RyKilleen:


Shazwazza:
thanks, we'll pull this in and make further changes.

PR #227, opened 2015-08-27
Info on excluding files from publish

lars-erik:
I banged my head a few hours today to figure out why the standard ExcludeFilesFromPackage Item Group and siblings didn't work. I'm sure I'm neither first nor last, so I guess it could just as well be in the docs.


Shazwazza:
sounds good

lars-erik:
Hey, I wasn't done! :)

sitereactor:
_Web One-Click Publish with Visual Studio_ is that really what its called? Sounds a bit funny. This is referring to the Publish option in the right-click context menu, right?

lars-erik:
The toolbar thingy (right click + customize) is named Web One Click Publish. Might be the main feature is just called Web Publish. Dunno. I think there's 3 different versions of msbuild, mspublish, msdeploy etc. behind the scenes depending on _how_ you publish.

Issue #232, opened 2015-09-02
7.3 load balancing docs


Shazwazza:

Shazwazza:
See PR #257

Shazwazza:
I'll close this issue since we can comment on the PR

PR #257, opened 2015-10-06
7.3 load balancing docs #232

Shazwazza:
Docs are pretty complete, please review and pull in when you are happy with it.


hemraker:
@Shazwazza let me know if you are ok with my changes

Getting-Started/Setup/Server-Setup/Load-Balancing/flexible.md
```diff
@@ -0,0 +1,102 @@
...
+**_Traditional load balancing must be used for Umbraco versions less than 7.3.0._**
+
+**Be sure you read the [Overview](index.md) before you begin!**
+
+##Design
+These instructions make the following assumptions:
+
+* All web servers can communicate with the database where Umbraco data is stored
+* You are running Umbraco 7.3.0 or above
+* _**You will designate a single server to be the back-office server for which your editors will log into for editing content.**_ Umbraco will not work if the back-office is behind the load balancer *(see DNS for more information below)*.
```
Jeavon: I think this should be removed as Umbraco v7.3 works fine behind the load balancing???

Shazwazza: No... it actually doesn't. It might seem to 'work fine' but there are all sort of things that can go wrong. In no way do we support load balancing the back office. If multiple admins are using the back office at the same time and things are being copied and moved and saved, things might end up with incorrect or non existent parents. We cannot guarantee the order that instructions are queued if you are load balancing the back office.

Jeavon: Ah yes, just remove the "(see DNS for more information below)" section then?

Shazwazza: yup making some updates now.

Jeavon: That's something to think about for multiple developers using a shared DB which is in effect the same as a load balanced back office.

PR #262, opened 2015-10-23
Updating documentation for Image-Cropper.

clausjensen:


Jeavon:
@clausjensen I think you may have been having a specific issue when you created these changes as they are not correct..?

hemraker:
@Jeavon The updated version of the Image Processor (v4) handles urls differently. See discussion here http://issues.umbraco.org/issue/U4-7223

Jeavon:
@hemraker no not really, as @clausjensen says in the issue the encoding is just what Razor does as intended and in a img tag it's correct that the ampersands are encoded. The issue is only when used in css as a background, IMO I think it's fine that a developer should know to use Html.Raw for that usage.
That said I don't think the Html helper is a bad addition but I think it should only recommended in the documentation for background-image: url not normal usage. It's much more concise as the extension method as current documented.

Jeavon:
@hemraker @clausjensen please let me know if that makes sense?

I just had to to check for my own sanity with the W3C validator and using the new Html.GetCropUrl makes the rendered HTML invalid.

Try this on the W3C validator:

    <!DOCTYPE html>
    <html lang="en">
    <head>
        <title>Home | Fanoe - Umbraco Starter Kit</title>
    </head>
    <body>
        <pre>@@mediaItem.GetCropUrl(300);</pre>
        
        <pre>@@Html.GetCropUrl(mediaItem, width:300)</pre>
        
    </body>
    </html>

You will see the second one is invalid while the first is valid.

hemraker:
Aha, and the MVC engine automatically encodes html attributes so thats why it works there. Correct? Thanks for the clarification!

Jeavon:
I don't think Razor will encode a IHtmlString which is why the new Html.GetCropUrl method is rendering invalid html when used for the src in a img tag, @Shazwazza ?
Also I'm not sure where this leaves us with dynamics support using CurrentPage or is that not a concern in v7.4+?

Jeavon:
@Shazwazza forgetting the dynamics issue for a moment, the only way I can make this work is to add an additional htmlEncode parameter to the HtmlHelper extension methods, e.g.:

        public static IHtmlString GetCropUrl(this HtmlHelper htmlHelper,
            IPublishedContent mediaItem,
            int? width = null,
            int? height = null,
            string propertyAlias = Constants.Conventions.Media.File,
            string cropAlias = null,
            int? quality = null,
            ImageCropMode? imageCropMode = null,
            ImageCropAnchor? imageCropAnchor = null,
            bool preferFocalPoint = false,
            bool useCropDimensions = false,
            bool cacheBuster = true,
            string furtherOptions = null,
            ImageCropRatioMode? ratioMode = null,
            bool upScale = true,
            bool htmlEncode = true)
        {
            var cropUrl = mediaItem.GetCropUrl(width, height, propertyAlias, cropAlias, quality, imageCropMode,
                imageCropAnchor, preferFocalPoint, useCropDimensions, cacheBuster, furtherOptions, ratioMode,
                upScale);

            return htmlEncode ? new HtmlString(HttpUtility.HtmlEncode(cropUrl)) : new HtmlString(cropUrl);

        }

Then we get the correct output like this

    @Model.value.caption

    <style>
    .htmlHelper {
        background-image: url("@Html.GetCropUrl(mediaItem, width:300, htmlEncode:false));
    }
    </style>

I don't know if this is better than using a UrlHelper extension method and expecting developers to use Html.Raw when using within a style tag...

Shazwazza:
(sorry, no time today, will have a look at this tomorrow)

Jeavon:
@Shazwazza did you get anywhere with this? I just gave it a whirl with Umbraco v7.4 and it's generating invalid HTML for img tags...

Shazwazza:
So let me get this straight, an html image tag requires query strings in the src param to be encoded, but only for that circumstance, otherwise the image url should not be encoded. Is that correct?

Jeavon:
Kinda, certainly the src attribute of a img element value should be encoded. In fact all ampersands and quotes should be encoded when used in html attributes.
When used in CSS, then it should not be encoded as that's not html....

Shazwazza:
Well then it would seem that we need to add the htmlEncode parameter (but we cannot simply add another optional parameter since that is a breaking change if you don't recompile, we'd have to add another overload). Also note that the usage above in your example `HttpUtility.HtmlEncode(cropUrl)` is incorrect, you would use the HtmlHelper to encode: `htmlHelper.Encode(cropUrl)`

I'll make this change in an upcoming 7.3.5 ... does this suit everyone then?

Shazwazza:
Actually... this won't solve the problem because the other overloads cannot just be changed to use encode=true since that will also be an 'unbreaking' change.

If we want an encoded vs unencoded option we'll have to make it a UrlHelper extension and have devs use Html.Raw for unencoded strings. Just means now we'd have a whole bunch of ways to do the same thing... we'll need to obsolete the ones we don't want and update docs.

Will that work for everyone ?

Jeavon:
@Shazwazza are these Html helper extensions already in v7.3 then (I thought they were only coming in v7.4)?
I think your suggestion of changing to a UrlHelper extension and letting the Razor engine encode itselft and documenting that for use in CSS developers should use Html.Raw makes complete sense to me...

Shazwazza:
This is a task for 7.3.1: http://issues.umbraco.org/issue/U4-7223

so yes, they are already there.

Shazwazza:
@Jeavon I guess making this a UrlHelper extension would bring us full circle since this would be the exact same thing as using the GetCropUrl extension methods on IPublishedContent ;) In theory, it makes more sense to be an extension method of UrlHelper because a UrlHelper is the thing that generates URLs, but the IPublishedContent extension methods will still be there too.

So... I think we should do this: 

* Add the extension methods to UrlHelper
* Each method will have an optional `htmlEncode` parameter (default is true)
* This will be the documented way of getting crop URLs and will work for CSS and Html so long as the `htmlEncode` parameter is used accordingly - then there's no reason to use `Html.Raw`
* We obsolete the HtmlHelper extensions (and remove in v8)
* We can keep the IPublishedContent extensions - knowing that they are not encoded URLs

Shazwazza:
This is complete in this revision: fab30d353a26cfef145b2597e0cd72ad6aba5b63

If you can have a quick look and let me know it's ok. I've tested and all works as expected. I think we then need to update our docs to reflect the UrlHelper approach. In razor it would be like:

```
@Url.GetCropUrl("/media/1234/blah.jpg", htmlEncode: true, height: 1234)
```

This will be available in 7.3.5 (within a week or so)

Jeavon:
:boom: Full circle complete! This is now working perfectly!

    @{
        var mediaItem = Umbraco.TypedMedia(1080);
    }

    

    <style>
    .urlHelper {
        background-image: url("@Url.GetCropUrl(mediaItem, width:300, htmlEncode:false));
    }
    </style>

Also we need the IPublishedContent extension methods for those who choose to use dynamics, I assume the documentation will still include examples for dynamics as well as the new method for strongly typed....?

Thanks @Shazwazza

Shazwazza:
How does it work for dynamics now ? AFAIK all GetCropUrl methods take an IPublishedContent item, I don't see any specific ones about dynamics. I would assume if you just use this same method like this it would still work since the dynamic representation of IPublishedContent is still IPublishedContent:

```
@{
    var mediaItem = Umbraco.Media(1080);
}

<img src="@Url.GetCropUrl(mediaItem, 200, 200)"/>

<style>
.urlHelper {
    background-image: url("@Url.GetCropUrl(mediaItem, width:300, htmlEncode:false));
}
</style>
```

hemraker:
This is really great!

Docs on the way!

Jeavon:
@Shazwazza it works as a IPublishedContent extension due to the dark magic of the dynamic ExtensionMethodFinder that I think you reworked from Gareth's original DynamicNode stuff. 
The new UrlHelper "extension methods cannot be dynamically dispatched"

Shazwazza:
This is the problem with dynamics... they are just generally bad.

We cannot do any magic with UrlHelper extensions because we are not in control of UrlHelper. There are only 3 options:

* Use the methods as static methods: `@Umbraco.Web.UrlHelperRenderExtensions.GetCropUrl(Url, CurrentPage, "blah")`
* Use the original `CurrentPage.GetCropUrl` methods
* Cast to IPublishedContent: `@Url.GetCropUrl((IPublishedContent)CurrentPage, "blah")`

Those are the only possibilities due to the nature of dynamics - so i guess we'll need to know what everyone thinks (IMO... don't use dynamics, but i guess that's easy for me to say ;) )

Jeavon:
Absolutely, I think given the planned depreciation of dynamics support in Umbraco we should continue to recommend the usage of the current IPublishedContent extension methods in the documentation when using CurrentPage but the UrlHelper for strongly typed...?

Shazwazza:
Sounds good to me!

hemraker:
ref #299

PR #265, opened 2015-10-26
Adds Security reference doc - for custom back office user auth.

Shazwazza:
See http://issues.umbraco.org/issue/U4-7032

### DIFF
--- a/Reference/Security/index.md
+++ b/Reference/Security/index.md
@@ -50,7 +50,7 @@ Here are the steps to specify your own logic for validating a username and passw
 		    }
 		}
 
-1. Modify the ~/App_Start/UmbracoCustomOwinStartup.cs class
+1. Modify the `~/App_Start/UmbracoCustomOwinStartup.cs` class
 
 	* Replace the `app.ConfigureUserManagerForUmbracoBackOffice` call with a custom overload to specify your custom `IBackOfficeUserPasswordChecker`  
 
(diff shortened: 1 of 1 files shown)


kgiszewski:
:+1: This is great and thanks for this. Is it worth mentioning which version this is for? i.e. 7.3.1+

Shazwazza:
@kgiszewski versions are another story for docs ;) We're working on a solution to that. We don't want to pollute the docs with version numbers everywhere since that is what we had before and it leads to mass confusion. Docs listed by default will indicate latest version, if there are docs for a previous version there will be a drop down ... this is a proof of concept at the moment though.

kgiszewski:
Fair enough. I spent this morning working on using Sandcastle against the core DLL's: https://github.com/kgiszewski/Umbraco-CMS-Docs/blob/2e67ed3dd44ca9ce6d827e89eb8722aee00b5546/Help/P_Umbraco_Core_Models_IPublishedContent_Parent.md

But there are 1000's of broken links :)

Was thinking one branch per version etc. Oh well :(

kgiszewski:
Don't click the link, I removed the repo :)

Shazwazza:
There's already a sandcastle build fike on the core, if you have sandcastle installed it will just work ;) but we need to update more styles so that we can host it.

kgiszewski:
Interesting... I'm just learning GhostDoc and Sandcastle so this is a good thing IMHO :)

PR #267, opened 2015-10-28
More fixes for typos, formatting and wording

greystate:
I was on a roll, so ran through some more files :)

Just let me know if it's too nit-picky :)

### DIFF
--- a/Reference/Config/index.md
+++ b/Reference/Config/index.md
@@ -1,10 +1,10 @@
 #Configuration Files
 
-_This section will define the many configuration options in Umbraco and what they do_
+_This section will explain the many configuration options in Umbraco and what they do_
 
 ###[web.config](webconfig/index.md)
 
-There are quite a few umbraco options in the web.config file defined in appSettings. Click on your version below to view details:
+There are quite a few Umbraco options in the `web.config` file defined in `<appSettings>`. Click on your version below to view details:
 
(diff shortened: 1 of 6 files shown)


Reference/Config/index.md
```diff
@@ -1,10 +1,10 @@
 #Configuration Files
 
-_This section will define the many configuration options in Umbraco and what they do_
+_This section will explain the many configuration options in Umbraco and what they do_
 
 ###[web.config](webconfig/index.md)
 
-There are quite a few umbraco options in the web.config file defined in appSettings. Click on your version below to view details:
+There are quite a few Umbraco options in the `web.config` file defined in `<appSettings>`. Click on your version below to view details:
```
greystate: > "Click on your version below [...]"

What "version" is referred to here - do you know, @hemraker ?

hemraker: Tbh I have no idea. Looks like it is left overs from an older version of docs where there were version specific docs for web.config. Think it slipped through the cracks because there actually is something listed below.

Think that line should just go.

greystate: Alright - I'll commit a nuke :bomb: for it :)

hemraker: bombs away!!

hemraker:
There's no such thing as nit-picking here. Thanks for taking the time to do this :smile:

PR #270, opened 2015-11-08
Update using-ioc.md

blueboxes:
I have found that umbraco 7 uses IOC and needs you re-register the api controllers to work with autofac.

### DIFF
--- a/Reference/using-ioc.md
+++ b/Reference/using-ioc.md
@@ -4,7 +4,7 @@ _This section will show you how to setup Ioc/Dependency Injection with your Umbr
 
 ##Overview
 
-Since Umbraco 7 we use IoC in the Umbraco source code for the backend site. With that said, it is possible to implement whatever IoC engine that you'd like you just need to re-register our api controllers!
+We don't use IoC in the Umbraco source code. This isn't because we don't like it or don't want to use it, it's because we want you as a developer to be able to use whatever IoC framework that you would like to use without jumping through any hoops. With that said, it means it is possible to implement whatever IoC engine that you'd like!
 
 ##Implementation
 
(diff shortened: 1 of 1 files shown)


Reference/using-ioc.md
```diff
@@ -4,7 +4,7 @@ _This section will show you how to setup Ioc/Dependency Injection with your Umbr
 
 ##Overview
 
-We don't use IoC in the Umbraco source code. This isn't because we don't like it or don't want to use it, it's because we want you as a developer to be able to use whatever IoC framework that you would like to use without jumping through any hoops. With that said, it means it is possible to implement whatever IoC engine that you'd like!
+Since Umbraco 7 we use IoC in the Umbraco source code for the backend site. With that said, it is possible to implement whatever IoC engine that you'd like you just need to re-register our api controllers!
```
Shazwazza: This actually isn't true at all... we don't use any Dependency Injection library in the core. You *can* use whatever IoC engine that you want.

Shazwazza: If you revert this description, the other changes below for registering api controllers is fine.

blueboxes:
So just to confirm core does not use IoC but the Umbraco.Web.Ui does hence you have to re-register the api controllers (and any backend plugins) that are used?

Reference/using-ioc.md
blueboxes: So just to confirm core does not use IoC but the Umbraco.Web.Ui does?

The reason I came to this consultation is all the artical and the issue I have had where you have to re-register the api controllers (and any backend plugins) that are used ?

Shazwazza: No, Umbraco does not use IoC whatsoever in it's codebase. 

WebApi and MVC do not require IoC to execute, this is an optional thing for developers if they want that functionality. There's not 're-registering' of api controllers... that is simply required if you choose to use IoC. We don't 'register' controllers anywhere, this is simply handled by WebApi/MVC controller factories. When you implement IoC, these controller factories delegate to your IoC container.

So, if you can update these docs by reverting the description change and also changing your text:

> //re-register umbraco webapi controllers used by the admin site

to simply 

> //register umbraco webapi controllers used by the admin site

We'll pull these docs in asap.

blueboxes: Great, I understand now. I will update the request. Thanks

Shazwazza:
Fantastic thanks!

PR #277, opened 2015-11-20
Add links to Examine videos

charlesroper:
Remove the old broken link and add links to other working videos.


dampee:
Thanks for the catch.  While it or old video's, it's probably better thank a 404 not found.

PR #278, opened 2015-11-20
Link to kudo

kedde:
Add a link to the kudo wiki


hemraker:
Hi Kedde, not a bad idea linking to the technologies. Might be a good idea to a little text to, ie. "if you want to read more about...." and then link to both git, kudu and courrier.

:smiley:

kedde:
We discovered that the supported kudu services is only a subset of the ones available(no ftp etc), I think this would be mentioned as well?

dampee:
Let's close this as there is a section about [kudu in the documentation](https://our.umbraco.org/Documentation/Umbraco-Cloud/Set-Up/Power-Tools/)!

PR #279, opened 2015-11-23
Update incorrect instructions.

JimBobSquarePants:
ImageProcessor.Web does not require disabling the VirtualPathProvider in order for the cache to work. Disabling the provider is strongly discouraged as it will cause Umbraco to persist the media url in absolute format. This causes two issues:

1. ImageProcessor cannot now intercept the image request to the external site without using the /remote.axd prefix.
2. The media cannot now be shared across different urls.

Both the IFileystemProvider implementation and ImageProcessor.Web.Plugins.AzureBlobCache will automatically create containers if they do not exist so you do not have to manually create them.

https://github.com/JimBobSquarePants/ImageProcessor/blob/V2/src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs#L100

### DIFF
--- a/Umbraco-as-a-Service/Set-Up/Media/index.md
+++ b/Umbraco-as-a-Service/Set-Up/Media/index.md
@@ -34,18 +34,7 @@ Update `~/Config/FileSystemProviders.config` replacing the default provider with
 </FileSystemProviders>
 ```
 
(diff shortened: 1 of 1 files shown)


paulsterling:
Thanks for reviewing and updating!

JimBobSquarePants:
Happy to help :smile:

PR #183, opened 2015-02-13
Display success messages added

TempleClause:
I added a section on how to display success messages... I explained two different ways.

PR #285, opened 2015-12-01
Added a section on how to display a success message

TempleClause:


hemraker:
Merged in #183 
Thanks!

nul800sebastiaan:
Be better to not hardcode the message in the co troller though, also for multingual purposes. Best would be do something like `Tempdata["success"] = true;` and in the view you can simply check `if Tempdata["success"]` != null` and display some text / markup.

nul800sebastiaan:
If you make a new commit  you can just submit a new one! Thanks by the way, love documentation contributions!

PR #290, opened 2015-12-07
missing namespace for generic GetCacheItem<> extention methods

dampee:
Not everyone has plugins in visual studio helping them.  Or the plug-ins don't work one day...


hemraker:
Thanks, good point!

PR #303, opened 2016-01-28
Fix typo/syntax in Collections.md, ".Where" section

esn303:
Not a problem for experienced developers, but possibly killing the experience for newbies if they copy and past without further checking.


hemraker:
Thanks!!

PR #313, opened 2016-03-01
Changed 'viola' to 'voila' for the intended meaning

chrisgalloway1:
Just fixing a typo in the documentation.


hemraker:
Viola is such a lovely name though... Thanks!

PR #318, opened 2016-04-11
Update AzureBlobCache config

alindgren:
The `~/config/imageprocessor/cache.config` example had the URL, but in reality, it needs to be the container name for CachedBlobContainer and SourceBlobContainer


paulsterling:
Awesome, thanks!

PR #329, opened 2016-05-06
Added documentation for date picker with time property editor

Aaen:


nul800sebastiaan:
@Aaen @hemraker  This is not linked from the Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/ page, so people will never know it exists :-)

PR #339, opened 2016-05-20
Removed default member props no longer present in 7.4.2

glcheetham:
These default member properties:

* umbracoMemberPassWordRetrievalAnswer
* umbracoMemberPasswordRetrivalQuestion

Have typos and are no longer present for default members in my version of Umbraco (v7.4.2)

capture


hemraker:
Good catch! Thanks 😄

PR #343, opened 2016-05-25
Fixed small typo in the Controller documentation

Cosmin-Parvulescu:
I believe it was a copy mistake and it should be Umbraco Api Controllers.


nul800sebastiaan:
:+1: Thanks!

PR #346, opened 2016-05-31
Change in argument or ".Media()" to use int id instead of sample id.

davidwandar:
Change complies with other headers and samples in this document.


nul800sebastiaan:
Thanks!

PR #354, opened 2016-06-21
Update partial-views.md

pynej:
Added details about controlling CachedPartial caching.


Reference/Templating/Mvc/partial-views.md
```diff
@@ -95,6 +95,26 @@ The above will cache the output of your partial view for one hour (3600 seconds)
 		int cachedSeconds,
 		bool cacheByPage = false,
 		bool cacheByMember = false,
-		ViewDataDictionary viewData = null)
+		ViewDataDictionary viewData = null
```
marcemarc: This is missing a comma after = null 

eg it should be

ViewDataDictionary viewData = null,
Func<object, ViewDataDictionary, string> contextualKeyBuilder = null)

pynej:
Should be resolved.

PR #361, opened 2016-07-11
Updated the text about hostnames

Aaen:
I have updated the text for hostnames, so we tell people to point their sites at (e.g. mysite.s1.umbraco.io) instead of (e.g. mysite.umbraco.io)


hemraker:
👍

PR #363, opened 2016-07-21
Image references seems to be failing ...

mikkelhm:
on our.umbraco.orh when the ?raw=true querystring is contained in the image urls


Jeavon:
It used to be the case that GitHub didn't show the images unless they had the raw=true querystring but that's no longer the case. The new processor lock down in ImageProcessor is blocking them on Our now. I think there will be quite a few more to find and remove!

mikkelhm:
Hi @Jeavon,
I noticed as well, so cloned the repo and replaced it globally, and got this PR updated. It wasn't that bad though, 10'ish files

Jeavon:
Cool!

hemraker:
Awesome @mikkelhm 👏

nul800sebastiaan:
D'oh!

D'oh

Thanks for the update!

PR #367, opened 2016-08-03
Custom backoffice route should start with 'umbraco/backoffice'

ianrandell:
I followed the instructions on this page and got 401s in the backoffice when trying to hit my controller (UmbracoAuthorizedController).
I believe this page is wrong - the custom route should start with 'umbraco/backoffice'  NOT just 'backoffice'.
When I changed my route to use 'umbraco/backoffice' my code started working.
Please correct me if I'm wrong on this.

### DIFF
--- a/Reference/Routing/Authorized/index.md
+++ b/Reference/Routing/Authorized/index.md
@@ -33,7 +33,7 @@ Defining a route is done with the standard ASP.Net MVC routing practices. In Umb
     {
       RouteTable.Routes.MapRoute(
       name: "cats",
-      url: "backoffice/cats/{action}/{id}",
+      url: "umbraco/backoffice/cats/{action}/{id}",
       defaults: new
       {
         controller = "Cats",
@@ -42,7 +42,7 @@ Defining a route is done with the standard ASP.Net MVC routing practices. In Umb
       });
     }
 
-_NOTE the route must be prefixed with `backoffice` in order for Umbraco to check user authentication._
+_NOTE the route must be prefixed with `umbraco/backoffice` in order for Umbraco to check user authentication._
 
 ###What about Surface Controllers?
 Surface Controllers should not be used in the back office.  Surface Controllers are not designed to work with the back office, they are not meant to be used there and will not be supported being used there.


ianrandell:
Great thanks!

Sent from my Windows Phone
________________________________

alindgren:
So if you are writing an UmbracoAuthorizedController, how should the routing be set up so that it will work if the "umbraco" part has been configured to something else?

Shazwazza:
@alindgren routing doesn't come into play for api controllers that inherit from UmbracoApiController - these are auto routed. If you are using these and you want their URLs for angular you should not be hard coding any of these, these URLs should be set server side and injected via the server variables collection: https://our.umbraco.org/Documentation/Extending/version7-assets

alindgren:
@Shazwazza I was building an MVC controller for the back office. The documentation states:

> If you are using MVC in the back office then you would normally inherit from `Umbraco.Web.Mvc.UmbracoAuthorizedController`. This type of controller is not auto-routed like Umbraco Api controllers so will require a custom route declaration to make it work.

The example that shows the routing that follows has `url: "backoffice/cats/{action}/{id}`. What worked for me was `url: "umbraco/backoffice/cats/{action}/{id}`.  But based on your initial comment, I guess it would be better not to hard code `umbraco` so it works if the back office url gets changed.

Shazwazza:
Hi @alindgren ,

Sorry i must have misread! I thought you were referring to `UmbracoAuthorizedApiController` oops! 

So yes, that is not auto-routed and the 'umbraco' token is actually configurable so when creating your route you should take that into account too. The 'Umbraco' part comes from `GlobalSettings.UmbracoMvcArea`

So you'd want to do `GlobalSettings.UmbracoMvcArea + "/backoffice/" ....

I'll update these docs now.

Shazwazza:
I've updated this here: https://github.com/umbraco/UmbracoDocs/commit/96d6056d96745cb0cc26b7deb5acc85acd6fd522

thanks for all your feedback and input!

alindgren:
@Shazwazza: Just tried it out (with Umbraco 7.5 beta 2). I assume it's `Umbraco.Core.Configuration.GlobalSettings` you are referring to. I get an error message that "'GlobalSettings' is inaccessible due to its protection level."

Shazwazza:
I'm updating this now, thanks for the heads up! This stuff was meant to be public for a while now

Shazwazza:
Hi, unfortunately we are too close to release to make changes like this but I have scheduled this task for 7.5.1 here: http://issues.umbraco.org/issue/U4-8861

I'll update the docs too

alindgren:
Thanks @Shazwazza - for me it's not critical but will be good to have.

PR #370, opened 2016-08-14
fix Back Office controller example

alindgren:
In my experience (working on a plugin today that uses an MVC controller), the route should start with `umbraco/backoffice` for it to use User authentication.


alindgren:
Just saw that PR #367 already dealt with this.

PR #373, opened 2016-08-29
Update index.md

ashkansiroos:
I've tried to use it but the IMacroGuiRendering is deprecated and the cmsMacroPropertyType  is not in the database anymore


dampee:
Hi
I'm closing this as the page has been reworked.
Thanks a lot for the suggestion @ashkansiroos . I'm sure futher submissions will be taken up much quicker!

PR #376, opened 2016-09-20
Explicitly mention what version these settings are for

warrenbuckley:


warrenbuckley:
After talking with @nul800sebastiaan as this feature is not available to all versions, explicitly saying what version these features are available from

PR #384, opened 2016-10-20
Created documentation page for fileSystemProviders.config

simonech:
I wrote most of the reference for the configuration of the PhysicalFileSystem provider, but despite this being the way it should work, I cannot get it to work on my machine.
After I change the config I can save and edit new files, but I get an exception when trying to update old files.
Also old files are still pointing to the old url, so there must be a way to easily change the path of pre-existing files. But I don't know how to do it.

Addressed issue #218


simonech:
Ok, should be good to go now :)

hemraker:
Thanks a bunch, nice work h5yr ✋

simonech:
I just noticed that you site doesn't render standard markdown for code block, with 3 tickmarks, but code is with indent...
I'll send a new PR :)

PR #386, opened 2016-10-31
Correct the log's filename in **Git deploy logs**

greystate:
The `log.xml` file is actually a `.log` file (from what I've seen so far...) (?)


nul800sebastiaan:
Ah yes, correct! Thanks!

PR #388, opened 2016-11-08
Flag that the Redirect URL Management dashboard is in the Content Sec…

pauldambra:
…tion

Installing 7.5.3 and spent some time debugging why the *Redirect Url Management* dashboard was not in the *Developer* section. Finally noticed that the dashboard.config file says 

```
 <section alias="RedirectUrlManagement">
    <areas>
      <area>content</area>
    </areas>
    <tab caption="Redirect URL Management">
      <control>
        views/dashboard/developer/redirecturls.html
      </control>
    </tab>
  </section>
```

The URL is in Developer but the tab is in Content


dampee:
Good catch @pauldambra !  Thanks a lot!

PR #393, opened 2016-11-28
Document the do's and don't do's in Umbraco

Shazwazza:
Part of core sprint task - see review board for details U4-9098

### DIFF
--- a/Reference/Common-Pitfalls/index.md
+++ b/Reference/Common-Pitfalls/index.md
@@ -91,10 +91,11 @@ You create a menu on your Home page like:
 
 ```
 <ul>
-@foreach(var node in Model.Content.DescendantsOrSelf().Where(x => x.Level == 1)) 
-{
-    <li><a href="@node.Url">@node.Name</a></li>
-}
+    <li><a href="@Model.Site().Url">@Model.Site().Name</a></li>
+    @foreach(var node in Model.Site().DescendantsOrSelf().Where(x => x.Level == 2)) 
+    {
+        <li><a href="@node.Url">@node.Name</a></li>
+    }
 </ul>
 ```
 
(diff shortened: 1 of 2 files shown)


nul800sebastiaan:
What's the recommendation for `Performing lookups and logic in Examine events`? Use published content lookups instead? I know it's important to highlight potential problems but possible solutions are needed to because people just don't know what to do better.

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+In all Umbraco base classes that you'll normally use, these object are already exposed as properties, so please use these instead!
+For example, all Razor views that Umbraco creates expose an `Umbraco` property which is the UmbracoContext, they expose an `ApplicationContext` 
+property which is Umbraco's `ApplicationContext`. The other base classes that expose all the instances you need are things like `SurfaceController`,
+`UmbracoApiController`, `UmbracoController`, `RenderMvcController`, `UmbracoUserControl`, `UmbracoPage`, `UmbracoHttpHandler`, and the list goes on...
+
+So next time you are using `ApplicationContext.Current` or `UmbracoContext.Current` think "Why am I doing this?", 
+"Is this already exposed as a property of the base class that I'm using?", "I'm using Dependency Injection, I should be injecting this instance into my class."
+
+
+## Static references to web request instances (such as `UmbracoHelper`)
```
warrenbuckley: Ace @Shazwazza you talk about what not to do in the examples, but there is no follow up example of how it should be done, I know you mention it's available in the base classes, but perhaps pick an example base common class probably a SurfaceController & show how to access the membership helper in a simple code snippet.

So ie. don't do this but do this, however the technical explanation reads well & makes sense why you shouldn't do it that way.

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+
+Which just renders out: _Home, Blog, Office Locations, About Us, Contact Us_
+
+BUT! ...  this is going to perform most horribly. This is going to iterate over every single node in Umbraco, all 10,000 of them. Further more, 
+this means it is going to allocate 10,000 `IPublishedContent` instances in memory just in order to check it's `Level` value. 
+
+This can easily be re-written as:
+
+```
+<ul>
```
warrenbuckley: Just clarified & checked the query builder in the template editor does a sensible suggestion which it does by using children & not descendants. 

All good here :)

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+</ul>
+```
+
+The syntax `@Model.Content.Site()` is actually shorthand for doing this:
+`Model.Content.AncestorsOrSelf(1)` which means it is going to traverse up the tree until it reaches an ancestor node
+with a level of one. As mentioned above, traversing costs resources and in this example there is 3x traversals being done
+for the same value. Instead this can be rewritten as:
+
+```
+@{
```
warrenbuckley: Again checked the query builder breaks it out into variable first before iterating over it & it does like so. So all good again :)
```
@{
	var selection = CurrentPage.Site().Children.Where("Visible");
}
<ul>
	@foreach(var item in selection){
		<li>
			<a href="@item.Url">@item.Name</a>
		</li>
	}
</ul>
```

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+How do you know if you are using Dynamics?
+
+* If you are using `@CurrentPage` then __you are__ using dynamics
+* If you are using the UmbracoHelper query methods like `@Umbraco.Content` or `@Umbraco.Media` instead of the typed methods like `@Umbraco.TypedContent` and `@Umbraco.TypedMedia` then __you are__ using dynamics
+
+It is strongly advised that you use the strongly typed `@Model.Content` instead of `@CurrentPage` models in your views,  
+this will actually perform much better and you'll be forward compatible with Umbraco v8+ with regards to querying `IPublishedContent`. 
+
+A large problem with the performance of dynamics is having to parse string syntax such as:
+`@CurrentPage.Children.Where("DocumentTypeAlias == \"DatatypesFolder\" && Visible")` and turn that into something that is compilable when
```
warrenbuckley: OK so the query builder uses `dynamics` then with this produced from it
`var selection = CurrentPage.Site().Children.Where("Visible");`

If we are advising people to not use dynamics, but our tooling is producing them it gives a mixed message. Also have we marked them as obsolete/deprecated so that Razor views get lovely red squigglys & nice message in not to use these & will be removed in V8

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+It is strongly advised that you use the strongly typed `@Model.Content` instead of `@CurrentPage` models in your views,  
+this will actually perform much better and you'll be forward compatible with Umbraco v8+ with regards to querying `IPublishedContent`. 
+
+A large problem with the performance of dynamics is having to parse string syntax such as:
+`@CurrentPage.Children.Where("DocumentTypeAlias == \"DatatypesFolder\" && Visible")` and turn that into something that is compilable when 
+instead it can just be written as something that compiles
+
+## Using the Services layer in your views
+
+The Services layer of Umbraco is for manipulating the business logic of Umbraco directly to/from the database.
```
warrenbuckley: Can I make a suggestion of a simple don't vs do example, as most of the time people are unaware of what the equivalent is in UmbracoHelper.

Perhaps something as simple not using ContentService like so
```
var nope = ApplicationContext.Current.Services.ContentService.GetById(123);
var stillNope = UmbracoContext.ContentCache.GetById(123);

var better = Umbraco.ContentQuery.TypedContent(123);
var evenBetter = Umbraco.TypedContent(123);
```

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+Your views should rely only on the readonly data access of the `UmbracoHelper` and the properties/methods that it exposes. This ensures
+that the data being queried is fast (comes from cache) and that you aren't inadvertently making database changes.
+
+If you are using `Application.Services...` in your views, you should figure out why this is being done and in most cases remove this logic.   
+
+## Using UmbracoContext to access ApplicationContext
+
+You should not access the `ApplicationContext` via the `UmbracoContext`. 
+
+For example: `UmbracoContext.Current.Application` _<-- this is now deprecated/obsolete_
```
warrenbuckley: Should give/explain the reasoning (even thoiugh you kinda cover it the opening of this document) as opposed to just saying, no you shouldn't use this thing...

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+
+This is one of the worst Umbraco anti-patterns and could very well cause your site to perform ultra poorly. 
+
+Umbraco's content should not be used for volatile data, Umbraco's APIs and the way Umbraco's data is persisted
+was never designed for this. If you need to store/write/track data that changes a lot you should use a 
+custom database table or another service but not Umbraco content nodes.
+
+Some examples of what not to do are:
+
+* Hit counters to track the number of times your page has been viewed - use something like Google Analytics for this or a custom database table
```
warrenbuckley: Good examples here of what not to do ;)

Luckily we don't automatically turn on the Umbraco Forms workflow for saving responses as child nodes and you have to explicitly opt into this & create a workflow to create nodes for you.

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+        {
+            //Lazy load the property value and ensure it's not re-resolved once it's loaded            
+            return _related ?? 
+                (_related = GetPropertyValue<IEnumerable<int>>("related").ToList());
+        } 
+    }    
+}
+```
+
+This is slightly better:
```
warrenbuckley: Good to see the bad example & improved example, but this copy of `this is slightly better` indicates to me that this could be improved more, if so then why not expand on it further to the best scenario?

warrenbuckley: Ignore the comment above - after reading the next part I see how you cover this then with XPath selectors.

So as a reader of this, my understanding is then not to use my own custom models, especially if I am doing any logic for getting any properties in constructors, as when used in conjunction with LINQ is gonna create a crazy amount of objects in memory & thus slow perf.

So to improve this its better to write an xPath selector to fetch the 10000 nodes as XML & then use a LINQ query to sort them by votes from the associated XML property.

Reference/Common-Pitfalls/index.md
```diff
@@ -0,0 +1,466 @@
...
+Umbraco allows you to run some initialization code during startup by using `ApplicationEventHandler`, however great
+care should be used to ensure that you are not slowing down application startup. You should be especially careful
+as a Package developer that you are not slowing down application startup since your package may end up being used for
+thousands of websites.
+
+In many cases, [initialization code can be done lazily instead of eagerly](https://msdn.microsoft.com/en-us/library/dd997286(v=vs.110).aspx). 
+Instead of initialization everything you need as soon as the application starts you could execute your initialization code only when it is required. 
+This can be achieved in various ways such as:
+
+* Using [`Lazy<T>`](https://msdn.microsoft.com/en-us/library/dd642331(v=vs.110).aspx) and put the initialization logic in it's callback
```
warrenbuckley: Could a super simple example of how to do this be given. As even Umbraco Forms itself does not follow this pattern fully & initialises some stuff in startup however does leave a marker file to avoid re-checking all the time.

Reference/Common-Pitfalls/index.md
Shazwazza: Good idea, I've added an example

Reference/Common-Pitfalls/index.md
Shazwazza: I'll update these docs about this - there's nothing we can do about it until we change the query builder but we should still be telling people to avoid dynamics by default.

Reference/Common-Pitfalls/index.md
Shazwazza: I'll add this but as a side note, your examples above with `stilNope`, `better`, `evenBetter` are basically all the same. It really depends on what querying you are doing and what services you rely on - and then what you want to mock for unit tests. The `Umbraco.ContentQuery.TypedContent` is actually the preferred when speaking in terms of unit testability because `ContentQuery` is an interface which can be directly mocked whereas if you want to mock `UmbracoContext.ContentCache.GetById` or `UmbracoHelper.TypedContent` - these would be much more difficult to mock.

Reference/Common-Pitfalls/index.md
Shazwazza: Have added the reason

Reference/Common-Pitfalls/index.md
Shazwazza: I'm going to leave an example out for now - I've spent too much time on this for now and it'll be better to just get this released.

warrenbuckley:
Yep all reads great with some good examples added in :)

warrenbuckley:
Re-opened to help with code formatting issues happening on our.umbraco

PR #397, opened 2016-12-01
Fix typo so the example of the configuration is showing.

Aaen:


nul800sebastiaan:
This needs to be fixed by indenting the code block, triple backticks don't work and a single backtick just puts everything on one line. I'll fix it up.

PR #400, opened 2016-12-06
Update 'Too much querying' examples

greystate:
Was using `@Model.Content.Site()` instead of `@Model.Site()`


Shazwazza:
Thanks!

greystate:
Allrighty - I've submitted #449 to revert this and use the `Model.Content` syntax in the samples.

PR #402, opened 2016-12-09
Updates some LB docs, adds FAQs section, changes Traditional to Legac…

Shazwazza:
…y and updates the docs to recommend setting an explicit master scheduler.


Shazwazza:
I'm pulling this in

PR #406, opened 2016-12-18
Update Macro Parameter Documentation for V7

marcemarc:
I know it's just a case of setting isParameterEditor to true on an existing Property Editor, but if you were googling you'd find the old Custom Macro Parameter documentation that applies to V6, which is a little more complicated...


Shazwazza:
Thanks!!

PR #411, opened 2017-01-03
Update MemberService.md

gsidoel:
GetMembersByPropertyValue("city", "Horsens") seems to match any fields containing the first parameter.

### DIFF
--- a/Reference/Management/Services/MemberService.md
+++ b/Reference/Management/Services/MemberService.md
@@ -150,7 +150,7 @@ Returns all members in a given Group - same as "Role".
 Returns all members of a given Type
 
 ###.GetMembersByPropertyValue("city", "Horsens");
-Returns all Members, of any type, with a mathcing value in the property with the given property alias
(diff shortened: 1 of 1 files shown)


Shazwazza:
Hrm, i'll need to look into that, haven't looked at these apis for some time

gsidoel:
Yeah we came across this in a version 7.2.2 assembly: 1.0.5529.18522 instance. In this particular case it worked out for us as we wanted to check two fields for a value, but we were quite perplexed at first. After some testing we concluded that it did a "contains". But I imagine it could be quite confusing for someone looking to test only 1 field. Perhaps we should have an overload to cover both cases?

dampee:
I don't think it is relevant anymore as there is now a matchType.

See https://github.com/umbraco/Umbraco-CMS/blob/7ee510ed386495120666a78c61497f58ff05de8f/src/Umbraco.Core/Services/MemberService.cs#L421

PR #418, opened 2017-01-16
added "mode = 'script'" as it is provided by default

davidwandar:


davidwandar:
Hm, think I probably misunderstood the scripts. The docs might need update tough....

PR #419, opened 2017-01-16
Updated Umbraco Cloud documentation

Aaen:
I have updated Umbraco Cloud documentation so it reflects our currently setup with starter plan and professional plan. I also added documentation about the live environment has initial authorization until it has been paid.


hemraker:
👏 😄

PR #425, opened 2017-02-05
Update value-converters.md

soreng:
fixed code sample to be more likely to compile


dampee:
Hi @soreng 
What is the idea for this change?  It is probably better on elaborating on the switch that to just return.  people should not copy past this code. Right?

Also, it has conflicts in the meantime which should be fixed.

soreng:
@dampee damn, I really don't know why I did this -- just removed some lines to make it more readable IMHO :)

Extending/Property-Editors/value-converters.md
```diff
@@ -137,10 +137,9 @@ If all values should use the same level you can use the short cut below
 
             case PropertyCacheValue.XPath:
                 return PropertyCacheLevel.Content;
-
-            default:
```
dampee: Why removing the default?

soreng: do you think it's easier to read?

switch(...) {
  case ... :
   return ...
 default :
  return
}

vs.

switch(...){
  case ... :
   return ...
}
return ...

dampee: It's not about easier to read.  A default does what it does.  You are able to change the code after the switch and change loose the missing enum. It is a change in future maintainability.  Anyway I reviewed the complete file (not only the change).  And it's good enough for me.  People won't be copying this without thinking anyway.

PR #427, opened 2017-02-14
Update working locally docs

dzordzs:
Made it a bit more clear that we are using Source Tree in this example.


hemraker:
Great 🎉 Thanks 😄

PR #428, opened 2017-02-15
Custom Model Advice

craigs100:
Added example of passing through the normal Umbraco Model when setting up a Custom Model if inherited from RenderModel.

When trying to do this for the first time I found this bit of advice essential as it wasn't obvious and so had to be hunted down ;)

### DIFF
--- a/Reference/Routing/custom-controllers.md
+++ b/Reference/Routing/custom-controllers.md
@@ -62,15 +62,17 @@ Please note that if your template uses a layout that expects the model to be of 
 1. Break the dependency on `Umbraco.Web.Models.RenderModel` in your layout by having it inherit from `Umbraco.Web.Mvc.UmbracoViewPage<dynamic>` (this means `@Model` will be of type `dynamic` in the layout).
 2. Make your custom model inherit from `Umbraco.Web.Models.RenderModel` and ensure you pass through the Umbraco model thus:-
 
-  public class MyNewViewModel : RenderModel
-  {
-      //Standard Model Pass Through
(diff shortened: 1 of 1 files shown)


Reference/Routing/custom-controllers.md
```diff
@@ -60,7 +60,17 @@ If you are returning a custom model, then this directive will need to change bec
...
+
+  public class MyNewViewModel : RenderModel
+  {
+      //Standard Model Pass Through
+      public MyNewViewModel(IPublishedContent content) : base(content) { }
+
+      //Custom properties here…..
+      public string myProperty1 { get; set; }
+      public string myProperty2 { get; set; }
+  } (this means `@Model...` will continue to work in the layouts used by your template).
```
Shazwazza: This class declaration needs to be indented for code formatting to work and the last sentence needs some line breaks after the code block. Also please use the correct c# naming conventions for your properties, these should be MyProperty1 and MyProperty2

Shazwazza:
Thanks!

craigs100:
Sorry, my first time, downloaded a better Markdown editor ;)

Shazwazza:
VSCode works well :)

Shazwazza:
You can also edit directly in GitHub and preview changes, it's shows mark down quite well too

craigs100:
Didn't know that. VSCode doesn't show me the end result though. I used 
http://remarkableapp.github.io/ in the end.

Shazwazza:
Yup it does, VSCode can do split window + real time preview. when editing an md file, click the top right split window button, then click the preview button or you can just toggle between preview and code

PR #429, opened 2017-02-16
If sample markup correction

richarth:
Added the missing @ and closing bracket


hemraker:
`Umbraco.If(@ != true, "This won't work", "This will work")` ... or something to that affect 😉 Thanks!

PR #432, opened 2017-03-02
Updates to Azure best practices.

SayTen:
Removes reference to directory @see https://our.umbraco.org/forum/developers/extending-umbraco/74731-examine-corruption-issues


Shazwazza:
Thanks, will merge in, though we have another option now which i need to document but it's best to change this for now.

PR #434, opened 2017-03-10
Remove reference to GlobalSettings.UmbracoMvcArea in Defining a route section 

marcemarc:
GlobalSettings is an internal class, and so the UmbracoMvcArea property is inaccessible here - anyone following this example to define a route to an UmbracoAuthorizedController will have errors - and confusingly GlobalSettings will resolve to the lowercase u umbraco namespace that does not have an UmbracoMvcArea property! - should we be using Umbraco Path in appsettings here ? it works but seems very messy ? or maybe we can make the UmbracoMvcArea property publically available in some kind of Route Mapping Helper class, for use in this instance, in the meantime though we need to update or remove the documentation to stop devs from getting stuck!

### DIFF
--- a/Getting-Started/Setup/Upgrading/general.md
+++ b/Getting-Started/Setup/Upgrading/general.md
@@ -15,6 +15,8 @@ It is necessary to run the upgrade installer on each environment of your Umbraco
 You can simply open up the **Package Console** and type:
 `Update-Package UmbracoCms`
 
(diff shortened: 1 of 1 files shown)


Shazwazza:
Doh! Yes sorry for the confusion, this was meant to be done ages ago: http://issues.umbraco.org/issue/U4-8861

In the meantime, I'll just create a public property on the 'old' GlobalSettings class 99ed4d6cb9727aaa87a440862221aca9ca8b5bd0

marcemarc:
thanks, yes I had a support client and a forum poster having the same issue in the same week, so thought I'd flag it up, got them working with the hack above, but having some helper as per the docs (wherever it comes from) makes the best sense!

PR #435, opened 2017-03-11
Fix misspelled word

dallastaylor:


hemraker:
Nice catch 😄 Thanks a bunch!

PR #436, opened 2017-03-16
UaaS-CLI doc update

greystate:
This update adds a section about the configuration file, as I couldn't find anything about it — so I thought I'd better put it somewhere that would be easy to find for the next guy or girl :-)


hemraker:
@greystate ❤️ Nice addition, thanks!

PR #439, opened 2017-03-22
Fix formatting of headers

rulrok:


hemraker:
Perfect, thanks 😄

rulrok:
After I noticed that all documents are in this non-github friendly markdown like the main readme file.
I imagine umbraco documentation page uses another markdown syntax parser, because here on github the tags for Headers ( # ) for example would just show up as normal text because there was no space between it and the text which followed it (#header instead of # header).

I think the root readme file does not affect the docs page, then I changed to make it easier to read on github.

PR #441, opened 2017-03-23
fix typo and grammar

rulrok:
I believe the intended meaning was 'and' by the sentence structure.


hemraker:
I believe you are absolutely correct! Thanks 🙌  #h5yr

PR #446, opened 2017-04-18
Correcting ContinouslyUpdateXmlDiskCache description

PeteDuncanson:
Fixes #U4-7101 and a little confusion I had with this one myself. Having checked through the code I believe this setting does very little these days however the existing documentation was miss leading. Feel free to sanity check it :)


Shazwazza:
Yup will agree, this is a strange old legacy setting which i believe previously did other things, thanks pete!

PR #451, opened 2017-05-04
Promote the HtmlHelper methods over the obsolete CurrentPage.GetGridHtml

marcemarc:
…tml and Model.Content.GetGridHtml approaches..

see http://issues.umbraco.org/issue/U4-9850


Shazwazza:
wonderful! thanks :)

PR #456, opened 2017-05-13
Fix the IIS Url Rewrite link on the 7.6.0 Upgrade page.

dallastaylor:
The link to the IIS URL Rewrite page was broken. It was linking to this url https://www_iis_net/downloads/microsoft/url-rewrite. (with underscores in the hostname)

I think it was because of the angle brackets in the url.


Shazwazza:
thanks!

PR #457, opened 2017-05-15
Provide some documentation about Umbraco UDIs

nul800sebastiaan:


Shazwazza:
great start, i'll merge and make some small updates. We should figure out the best place to put this info since it's not specifically to do with published content.

PR #459, opened 2017-05-17
Add example for request

StefanoChiodino:
Took me ages to find out how that works. Was actually about to write a custom workflow!


dampee:
Does this still works?  Can someone double check this?

jmayntzhusen:
I did a super quick test of this, I haven't worked much with Forms, but if this is the only confirmation needed we may as well get this pulled in 🙂

Setup:


Result:


Is this the intended behavior? Can you have another look at this @dampee ?

nul800sebastiaan:
Note that while this might work, I wouldn't want to rely on this as the referer is notoriously easy to spoof, so whatever you do with the information, make sure you clean it and don't show the raw data directly to anyone. 

References: 
https://en.wikipedia.org/wiki/Referer_spoofing
https://www.brokenbrowser.com/referer-spoofing-defeating-xss-filter/

PR #470, opened 2017-06-05
typo on version specific page: "revesed"

dampee:
should be reversed


hemraker:
Closed with #494 Sorry about that @dampee h5is

PR #479, opened 2017-06-06
Documentation about shortcuts in the backoffice

JeffreyPerplex:
Documentation about what has been fixed in here; http://issues.umbraco.org/issue/U4-9991

### DIFF
--- a/Getting-Started/Backoffice/index.md
+++ b/Getting-Started/Backoffice/index.md
@@ -100,7 +100,7 @@ Throughout the whole backoffice you can use several shortcuts. Most of them are 
 ** Ctrl + U: Unpublish the node
 
 * In the content tree:
-** Arrow Down: Go to the next node
+** Arrow Down: Go to the next node 
 ** Arrow Up: Go to the previous node
 ** Arrow Left: Close the node, if there are any children or go to the above node if not.
 ** Arrow Right: Expand the node, if it has any children. If not, it opens the options menu


dampee:
Hi Jeffrey,
This is a great addition.  Could you fix the merge conflicts?

JeffreyPerplex:
Hi @dampee,

it is a great addition, but the documentation that I wrote was also about the shortcut additions we've made during the Umbraco Retreat, but that PR has never merged in / is still pending (https://github.com/umbraco/Umbraco-CMS/pull/1985). I think that should be first finalized, before this could be added to the docs.

dampee:
What?  I would have sworn that it was already merged.  Never tried it to be honest.  Too bad.

JeffreyPerplex:
Hi @dampee.

thank you for bumping this issue. I've created a new smaller pull request with only the addition of a few shortcuts (https://github.com/umbraco/Umbraco-CMS/pull/2426). Once this get accepted, this file can be merged into the documentation.

Only thing that should be removed is the arrow-navigation, because that is not part of the current pull request. I know this would be the most helpful of all, but there were some concerns about the number of event listeners in the tree. For now I've removed that part out of my PR

I'll keep you updated,

Jeffrey

dampee:
Hi @JeffreyPerplex 
Do you create a new pull request or do you rebase (and update) so we can merge this in.
Would you have some time to prepare this?
Kind regards
Damiaan

JeffreyPerplex:
Hi @dampee, thanks for the reminder. I'm currently crazy busy so little to no time to do anything about this.

The initial PR was merged into 7.9, but due to some breaking issues it was removed in 7.9.1 or 7.9.2. so the documentation is not up-to-date.

Sorry for that,

Jeffrey

dampee:
No Problem.  Didn't know it was removed again from the latest versions.  Then there is no need to update the docs currently.  Right?

jmayntzhusen:
Hey @JeffreyPerplex 

This change was reverted and seems to not be worked on? 🙂 (Was this the one that broke the '@' sign on Danish keyboards? 😄 )

If nothing is happening here and the CMS PR seems closed then we should close this too, unless you think otherwise?

JeffreyPerplex:
Hi @jmayntzhusen, the documentation was written for a pull request that was released in 7.9 but reverted in 7.9.1 because it used a shortcut key Alt + <number> and that is also used for "weird" chars.

so the documentation was related to that PR and that one got reverted.

Still we should fix the original PR sometimes (just replace Alt + <number> by something else) but I don't see that happening soon. So close the PR for now I guess...

Cheerio, Jeffrey

PR #481, opened 2017-06-08
Escape generic

stevetemple:


hemraker:
Great catch! Thanks @stevetemple  🎉

PR #482, opened 2017-06-08
Added all methods for PreValues get

koblamic:
Just want to add few lines to docs for PreValues, and just checking process of Pushing :-).


hemraker:
Very nice - Thanks @koblamic

PR #487, opened 2017-06-19
FIPS Tutorial added to support Umbraco 7.6.4

protherj:
For U4-9792 on the YouTrack Issue Tracker

http://issues.umbraco.org/issue/U4-9792

### DIFF
--- a/Tutorials/Setup-Umbraco-for-a-Fips-Server/index.md
+++ b/Tutorials/Setup-Umbraco-for-a-Fips-Server/index.md
@@ -1,0 +1,83 @@
+# Setup Umbraco for a FIPS Compliant Server
+
+_This tutorial walks through configuring Umbraco and Lucene to be FIPS compliant and serve up websites on a server with FIPS enabled._
+
+## What is FIPS?
+
+The Federal Information Processing Standard (FIPS) Publication 140-2, ([FIPS PUB 140-2][1]), is a U.S. government computer security standard used to define approved cryptographic modules. The FIPS 140 standard also sets forth requirements for key generation and for key management.
+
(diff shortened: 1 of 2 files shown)


Shazwazza:
You rock, this is great! @hemraker could you have a look too please?

hemraker:
So awesome @protherj 

Will have a look at it today #h5yr

hemraker:
Very well written @protherj 👏 I fixed a few typos but other than that I have nothing to add on the docs themselves.

I do however think this belongs under reference -> security and not in the tutorial section. Any thoughts on that @Shazwazza ?

hemraker:
Just had an interesting conversation with @nul800sebastiaan about this. Since (Microsoft do not recommend using FIPS)[https://blogs.technet.microsoft.com/secguide/2014/04/07/why-were-not-recommending-fips-mode-anymore/] there should be a disclaimer at the beginning of the docs explaining that this should only be implemented for compliance and not for security reasons. We could do a compliance section as well but I do think it makes sense to have it under security. For now at least.

protherj:
Good idea.  Yea, its only for servers where there is a requirement than it is present.

Thanks,
Jason

Shazwazza:
Yeah, should be moved over to https://our.umbraco.org/documentation/Reference/Security/ including the comments you've listed above

hemraker:
Closed with #490

protherj:
Any chance 7.6.4 could get released next week?  I'm getting a lot of pressure from my client because they have to go through a long-ish testing process and need to start that as soon as possible.

protherj:
Also, #h5yr for reviewing this quickly and getting it ready to go!

PR #495, opened 2017-06-28
Changed reference to CDN from http to https

JeffreyPerplex:
Made the world a bit more secure, and make sure that if people copy-paste the code won't have "Blocked content"-error messages in their console because their website is running https, but they're non-https scripts


hemraker:
Wonderful 😄 Thanks!

PR #496, opened 2017-06-29
Added documentation for limitations on hostnames

Aaen:
I have add documentation for the limitation on hostnames per environments.


sofietoft:
Would it make sense to say something like this instead:
"You can bind a total of 15 hostnames to each of your Umbraco Cloud environments." ?

PR #503, opened 2017-07-18
Update using-ioc.md

gunnim:
Update to the recommended applicationeventhandler usage


Shazwazza:
Changing this is incorrect since you need your container to be registered even if the database is not installed or Umbraco is in an installation state. Otherwise if you inherit from the base class `ApplicationEventHandler` instead of using the interface you'd need to override `ExecuteWhenApplicationNotConfigured` and `ExecuteWhenDatabaseNotConfigured` and return true.

gunnim:
Interesting! @Shazwazza, maybe we should add what you just said to the Events section under reference?
That is that inheriting from the interface has different defaults.
I always just assumed it was the old way of doing things..

PR #504, opened 2017-07-21
Added explanation about need to copy and amend Script.cshtml file whe…

tom-d8:
…n using your own js in a theme

more detail in this forum post https://our.umbraco.org/forum/umbraco-forms//87038-using-custom-js-file-via-themes-and-changing-validation-to-onblur

### DIFF
--- a/Add-ons/UmbracoForms/Developer/Themes/index.md
+++ b/Add-ons/UmbracoForms/Developer/Themes/index.md
@@ -76,4 +76,6 @@ To do this, when inserting the form using the macro ensure the checkbox for the 
         }
     }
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Hi Tom,

I've been doing some testing with Umbraco Forms and custom themes, and it seems you are correct.
You do need to have your own _Script.html_ file, where you mention the theme-specific scripts you want to use.

The Themes documentation already mentions this:
_Copy the explicit files you wish to override in your theme, it may be a single file or all files from the default theme folder._
Nothing specifically about the scripts though.
Perhaps you can move what you wrote closer to this section: https://our.umbraco.com/documentation/Add-ons/umbracoforms/developer/Themes/#creating-a-theme ? Would that make sense?

-Sofie

PR #506, opened 2017-07-25
Updated UmbracoAuthorizeAttribute information

HarveyWilliams:
Added information about how UmbracoAuthorizeAttribute doens't work after Umbraco 7.2.2 onwards.

Honestly, I'm not even sure what the point of this attribute is. I keep accidentally using it only to find that it stops anyone from being able to access the method even if they're logged in to Umbraco!

### DIFF
--- a/Reference/Routing/WebApi/authorization.md
+++ b/Reference/Routing/WebApi/authorization.md
@@ -10,7 +10,9 @@ Probably the easiest way to ensure your controller is secured for only back offi
 
 The `UmbracoAuthorizedApiController` is automatically routed.  Check out the [routing documentation](../Authorized/index.md) for more information on this topic.
 
-###Using UmbracoAuthorizeAttribute
+###Using UmbracoAuthorizeAttribute (<7.2.2)
+
+**Please note, this attribute [will not work](http://issues.umbraco.org/issue/U4-6352) in Umbraco from version 7.2.2. Please get your controller to inherit from the `UmbracoAuthorizedApiController` as is shown above.**
 
 To secure your controller based on back offce membership use the attribute: `Umbraco.Web.WebApi.UmbracoAuthorizeAttribute`. 
 


Shazwazza:
I don't understand what you mean. Maybe you can elaborate on the actual issue you are having?

Also note there are 2x this attribute:

* Umbraco.Web.Mvc.UmbracoAuthorizeAttribute
* Umbraco.Web.WebApi.UmbracoAuthorizeAttribute

which are for each purpose described by it's namespace. 

This attribute does exactly this: If the user is not logged in as a back office user then no access to any attributed controller or action will be allowed.

HarveyWilliams:
@Shazwazza This bit of documentation appears to be specifically about API and not MVC routing.

I linked to an issue on my updated documentation (http://issues.umbraco.org/issue/U4-6352). In this issue you said that:

> your route path MUST be prefixed with:
> 
> /umbraco/backoffice/...

We can also see that in [this issue](http://issues.umbraco.org/issue/U4-9299) someone else had the same issue which prompted me to update this documentation. They attempted to use the `[UmbracoAuthorize]` attribute in a controller which inherits from `UmbracoApiController`.

However, surely any controller which inherits from `UmbracoAuthorizedApiController` will require all of the surface methods in it to have the Umbraco user logged in? This means that the `[UmbracoAuthorize]` attribute practically has no use.

Correct me if I'm wrong at any point here. Thanks!

Shazwazza:
This attribute *does* work so i cannot accept this PR. If it did not work the entire back office would be available to the public.

Let me (try to) explain:

These are 2 different things: Authentication (AuthN) and Authorization (AuthZ). 

AuthN - is the process of assigning a Principal/User to the request which is based purely on the request information and this happens at the very beginning of a request before any webapi or mvc routing or anything happens. Since Umbraco does not want to interfere with how developers might normally setup AuthN/AuthZ on their website (i.e. front-end), Umbraco does not use the built-in FormsAuthentication or default ASP.NET Identity AuthN logic, it uses a customized version of this. In order for Umbraco to know if it should perform the AuthN for the back office, it needs to check the current URL since this is available at AuthN time and this URL must match the pattern for being a back office request. This is why this documentation is here: https://our.umbraco.org/Documentation/Reference/Routing/Authorized/#routing-requirements-for-back-office-authentication 

AuthN - this is the process of validating the Principal/User already assigned to the request has access tot he resource it is trying to access. This can be based on all sorts of logic and is easily extensible. For example, we have other custom authorization logic to check if a user has access to a certain content item based on the user's start nodes. This is Authorization and by this stage the Authentication part is already done.

`UmbracoAuthorizedApiController` is attributed with `[UmbracoAuthorize]`, see here: https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/WebApi/UmbracoAuthorizedApiController.cs#L20

So if you inherit from `UmbracoAuthorizedApiController` then you don't need to apply this attribute since it's already applied. Further more the this controller inherits from `UmbracoApiController` which is auto-routed which means you don't have to worry about adding your own routes on startup. However any controller inheriting from `UmbracoApiController` is auto-routed for a front-end route so AuthN will *not* take place for this controller so a back office user is not assigned to the request and therefore if `[UmbracoAuthorize]` is used on this controller, all requests to it will fail because there is no back office user assigned to the request to validate.

To have an auto-routed umbraco api controller work for the back office, it needs to be attributed with `[IsBackOffice]` as per this documentation: https://our.umbraco.org/Documentation/Reference/Routing/Authorized/#special-back-office-routes-for-user-authentication . If that is applied, the route is auto-routed to a back office pattern URL which means AuthN is applied, an Umbraco back office user is assigned to the request and the AuthZ attribute will verify that it has access.

Now, if you don't use an auto-routed controller, maybe you create your own custom WebApi controller that just inherits from the base class `ApiController` and you want to use this controller for the back office, well then you will need to make sure you route it correctly according to these docs: https://our.umbraco.org/Documentation/Reference/Routing/Authorized/#routing-requirements-for-back-office-authentication 

I hope that clears things up

HarveyWilliams:
@Shazwazza Okay I think I'm starting to get the hang of this. Thanks for the detailed explanation!

Let me try to summarise how this all works:

Inheriting from `UmbracoAuthorizedApiController` routes to `/umbraco/backoffice/api/{controller}/{action}` and causes all methods to require Umbraco User authentication.

Inheriting from `UmbracoApiController` routes to  `/umbraco/api/{controller}/{action}` and does not require any authentication.

Inheriting from `UmbracoApiController` and using the `[IsBackOffice]` attribute  on your controller class will cause your methods to get routed to `/umbraco/backoffice/api/{action}/{method}`, however, no authentication will be done unless your particular method also uses the `[UmbracoAuthorize]` attribute.

#### Here's where a bug seems to lie:

Inheriting from `UmbracoApiController` and **not** using the `[IsBackOffice]` attribute on your class, but trying to use `[UmbracoAuthorize]` on any methods will cause that method to **always** return '401 Unauthorized' regardless of if you are logged into Umbraco. (This is because Umbraco can't do authorisation on Urls which are not routed through ` /umbraco/backoffice/api`.)

I have tested this on Umbraco 7.2.6.

#### Updating the documentation

I think the documentation needs to be updated to include this information about `[UmbracoAuthorize]` only being useful on a controller which inherits from `UmbracoApiController` and uses `[IsBackOffice]`.

Currently the documentation [here](https://our.umbraco.org/documentation/Reference/Routing/WebApi/authorization#using-umbracoauthorizeattribute) says that this will work:

```
public class ProductsApiController : UmbracoApiController
{
    [Umbraco.Web.WebApi.UmbracoAuthorize]
    public IEnumerable<string> GetAllProducts()
    {
        return new[] { "Table", "Chair", "Desk", "Computer", "Beer fridge" };
    }
}
```

But the `GetAllProducts` method here will always return 401 as per my previous testing.

HarveyWilliams:
Just to say, I'm not sure if the `[UmbracoAuthorize]` worked differently before 7.2.2.

Shazwazza:
I've updated your statements with the right terminology, it's very important to know the difference between Authentication (AuthN) and Authorization (AuthZ)

> Inheriting from UmbracoAuthorizedApiController routes to /umbraco/backoffice/api/{controller}/{action} and causes ~~all methods to require Umbraco User authentication.~~  the route to be **authenticated** as a back office user and all actions to be **authorized** as a back office user.

> Inheriting from UmbracoApiController routes to /umbraco/api/{controller}/{action} and does not ~require any authentication.~ **authenticate** the route as a back office user. No **authorization** is performed for any action since there is not authorization attribute or logic applied.

> Inheriting from UmbracoApiController and using the [IsBackOffice] attribute on your controller class will cause your methods to get routed to /umbraco/backoffice/api/{action}/{method}  ~~however, no authentication will be done~~ so the route will be **authenticated** as a back office user, however, no **authorization** will be done unless your particular method also uses the [UmbracoAuthorize] attribute.

Documentation can certainly be updated and the docs for this page should reference more of the info that is talked about on this page: https://our.umbraco.org/Documentation/Reference/Routing/Authorized/ 

> I think the documentation needs to be updated to include this information about [UmbracoAuthorize] only being useful on a controller which **either** inherits from UmbracoApiController and uses [IsBackOffice] **or if the controller inherits from UmbracoAuthorizedApiController**

Regarding how it worked differently before 7.2.2 please read: http://issues.umbraco.org/issue/U4-6352#comment=67-19435 - to reiterate - if your route is not a back office route then **authentication** for a back office user will not be done which means there is no back office user assigned to the request which means if you have `[UmbracoAuthorize]` applied to a controller on a non back office route, there is nothing to authorize and therefore authorization would fail.

HarveyWilliams:
@Shazwazza Thanks, I still don't fully understand AuthN and AuthZ, but I get how Umbraco backoffice authorisation works now.

I was about to submit another pull request, but I have just noticed that `[Umbraco.Web.WebApi.UmbracoAuthorize]` can in fact be used on a controller. It even [says this](https://github.com/umbraco/UmbracoDocs/blob/master/Reference/Routing/WebApi/authorization.md) in the documentation:

> To secure your controller based on back offce membership use the attribute: Umbraco.Web.WebApi.UmbracoAuthorizeAttribute.

However, I feel like this is an accident as in the example below that excerpt we can clearly see the attribute being used on a method.

Do you know what using it on a controller actually does?

Shazwazza:
Authentication (AuthN) - is the process of assigning a Principal/User to the request. This is how something like `HttpContext.User` or `Thread.CurrentPrincipal` (these are the same object) is populated which happens at the beginning of the request. This is done based on both the current Uri and the information in the request (i.e. an authentication cookie)

Authorization (AuthZ) - this is the process of validating the Principal/User already assigned to the request has access to the resource it is trying to access. 

The way WebApi and MVC authorization and filter attributes work - they can be applied at the: Global, Controller or Action level. Global = will execute for all controllers and actions, Controller = will execute for all actions in the controller, Action = will execute for the action

HarveyWilliams:
@Shazwazza Cool. I tested using the attribute on the controller and it worked as you described.

I have created a new pull request here: https://github.com/umbraco/UmbracoDocs/pull/511

Gonna close this pull.

Thanks for the help. I hope the new pull is better than the one I made here!

PR #507, opened 2017-07-25
U4-9792 - Revised the FIPS documentation for new PreAppStart method of setting Lucene.net

protherj:
Revised the FIPS documentation to remove the requirement of re-compiling Lucene.net and instead setting the flag at PreApplicationStart.  This is a much less intrusive method and allows Lucene.net to continue to be pulled in via Nuget.


protherj:
It looks like my code snippets have the "csharp" part in the live documentation.  How should code snippets be formatted?

Shazwazza:
Not sure if the Our MD parser is smart enough for that. I'll pull it in and we'll see :)

Shazwazza:
Nope that didn't work :P i'll update it

PR #508, opened 2017-07-29
Updated with Unity example

gunnim:
Here's my attempt at offering Unity advice :)

### DIFF
--- a/Reference/using-ioc.md
+++ b/Reference/using-ioc.md
@@ -11,7 +11,7 @@ We don't use IoC in the Umbraco source code. This isn't because we don't like it
 In most IoC frameworks you would setup your container in your global.asax class. To do that in Umbraco, you will need to inherit from our global.asax class called: `Umbraco.Web.UmbracoApplication`. You should then override the `OnApplicationStarted` method to build your container and initialize any of the IoC stuff that you require.
 Alternatively you can implement the `Umbraco.Web.IApplicationEventHandler` interface.
 
-##Example
+##Autofac Example
 
 This example will setup Autofac to work with Umbraco (see [their documentation](http://autofac.readthedocs.org/en/latest/) for full details)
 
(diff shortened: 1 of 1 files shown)


Reference/using-ioc.md
```diff
@@ -141,3 +141,103 @@ If you don't register assemblies that contain controllers you may end up with YS
...
+	    public void OnApplicationStarted(
+		UmbracoApplicationBase httpApplication,
+		ApplicationContext applicationContext
+	    )
+	    {
+		var container = UnityConfig.GetConfiguredContainer();
+
+		// If you intend to use DI inside filter attributes
+		FilterProviders.Providers.Remove(FilterProviders.Providers.OfType<FilterAttributeFilterProvider>().First());
+		FilterProviders.Providers.Add(new UnityFilterAttributeFilterProvider(container));
```
Shazwazza: Since this is a global change will this interfere with any Umbraco filters, etc... ?

Reference/using-ioc.md
```diff
@@ -141,3 +141,103 @@ If you don't register assemblies that contain controllers you may end up with YS
...
+		
+		// Some common examples
+		
+		// The commonly used Umbraco ApplicationContext
+		container.RegisterType<ApplicationContext>(
+		    new ContainerControlledLifetimeManager(), 
+		    new InjectionFactory(c => ApplicationContext.Current)
+		);
+
+		// An HttpContext abstraction that you may find useful
```
Shazwazza: I'd rather keep this out of the docs as it's not necessary for Umbraco to operate and I don't want to add any confusion

gunnim: Okey, i'm removing the other examples as well then

Reference/using-ioc.md
gunnim: It works with my umbraco installation but might be better left out of the docs. will remove.

Reference/using-ioc.md
```diff
@@ -141,3 +141,82 @@ If you don't register assemblies that contain controllers you may end up with YS
...
+
+		// The UmbracoContext must be registered so that the umbraco backoffice controllers 
+		// can be successfully resolved
+		container.RegisterType<UmbracoContext>(
+			new PerRequestLifetimeManager(), 
+			new InjectionFactory(c => UmbracoContext.Current)
+		);
+
+		// Instantiate the HealthCheckController using it's default parameterless constructor
+		container.RegisterType<HealthCheckController>(new InjectionConstructor());
```
Shazwazza: Why are these controllers registered singularly? These docs should register all MVC and WebApi controllers found in the `typeof(UmbracoApplication).Assembly` just like the autofac docs.  The code block above does this but I'm unsure if that's for MVC or for WebAPI (or both) ? Similarly the LegacyTreeController doesn't need to be registered singularly, it will be taken care of by registering all MVC and WebApi controllers found in `typeof(UmbracoApplication).Assembly`

gunnim:
The manually registered controllers need the above configuration. They have multiple constructors and unity will attempt to use the wrong constructors and fail.

The above configuration instructs unity to use the default parameterless constructor.

This issue is mentioned here:
http://issues.umbraco.org/issue/U4-9562
and here f.x.
https://our.umbraco.org/forum/extending-umbraco-and-using-the-api/84189-healthcheckcontroller-doesnt-work-with-webapi-and-unity

Shazwazza:
I have commented on that issue http://issues.umbraco.org/issue/U4-9562#comment=67-41707

Reference/using-ioc.md
```diff
@@ -141,3 +141,81 @@ If you don't register assemblies that contain controllers you may end up with YS
...
+
+		// The UmbracoContext must be registered so that the umbraco backoffice controllers 
+		// can be successfully resolved
+		container.RegisterType<UmbracoContext>(
+			new PerRequestLifetimeManager(), 
+			new InjectionFactory(c => UmbracoContext.Current)
+		);
+
+		// Instantiate the HealthCheckController using it's default parameterless constructor
+		container.RegisterType<LegacyTreeController>(new InjectionConstructor());
```
Shazwazza: how come this line is necessary?

gunnim:
Unity by default will attempt to instantiate types using the constructor with the most parameters.
Legacetreecontrollers most populated constructor accepts some strings and other thats probably intended for unit testing right?
Anyway that means we have to instruct unity to instead use the default constructor using a parameterless InjectionConstructor

Shazwazza:
Ah ok makes sense then. Perhaps some comments about why it requires a separate registration would be good .

Shazwazza:
superb! Thanks for your effort, patience and helping us uncover a few coding errors, #h5yr!!

gunnim:
Oops, the formatting is way off.
I assume I have to resubmit a PR since this ones been merged. 

New ones here
https://github.com/umbraco/UmbracoDocs/pull/541

PR #511, opened 2017-08-09
Updated using `UmbracoAuthorizeAttribute`.

Harvzor:
Explained how `UmbracoAuthorizeAttribute` should be used with `IsBackOfficeAttribute` to secure Web API actions.

### DIFF
--- a/Reference/Routing/WebApi/authorization.md
+++ b/Reference/Routing/WebApi/authorization.md
@@ -1,18 +1,22 @@
 #Umbraco Api - Authorization
 
-_This section will describe how to secure your Umbraco Api controllers based on a users membership_ 
+_This section will describe how to secure your Umbraco Api controllers based on a users membership._ 
 
 ##Authorizing for the back office
 
 ###Inheriting from UmbracoAuthorizedApiController
 
-Probably the easiest way to ensure your controller is secured for only back office users is to inherit from `Umbraco.Web.WebApi.UmbracoAuthorizedApiController`. This is essentially the same as applying the `UmbracoAuthorizeAttribute` to your controller (see below). We also expose the Umbraco user property `User UmbracoUser {get;}` on this base controller as well.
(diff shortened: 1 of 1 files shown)


Jeavon:
@Shazwazza could you please have a quick review of this PR?

Reference/Routing/WebApi/authorization.md
```diff
@@ -1,18 +1,22 @@
...
 
-The `UmbracoAuthorizedApiController` is automatically routed.  Check out the [routing documentation](../Authorized/index.md) for more information on this topic.
+Controllers inheriting from `UmbracoAuthorizedApiController`, or ones that have the `IsBackOfficeAttribute` are automatically routed. Check out the [routing documentation](../Authorized/index.md) for more information on this topic.
 
 ###Using UmbracoAuthorizeAttribute
 
-To secure your controller based on back offce membership use the attribute: `Umbraco.Web.WebApi.UmbracoAuthorizeAttribute`. 
+Using this attribute in conjunction with `IsBackOfficeAttribute` will allow you to secure some actions but still leave others open. If you don't use `IsBackOfficeAttribute` on your controller, then `UmbracoAuthorizeAttribute` will always cause your actions to return unauthorised, regardless of if you are logged into the back office.
```
Shazwazza: Again i think there is some confusion as to what IsBackOfficeAttribute is doing so this isn't correct. The `IsBackOfficeAttribute` is for one thing - auto-routed UmbracoApiControllers with this attribute are routed as back office routes so they can be authenticated as back office requests.

If you use `UmbracoAuthorizeAttribute` then it will do an authorization check to make sure that there is a back office user logged in, otherwise the resource cannot be accessed. This can be applied at the controller level which will secure all actions on the controller, or at the action level which can secure individual actions.

Reference/Routing/WebApi/authorization.md
```diff
@@ -1,18 +1,22 @@
...
+
+`Umbraco.Web.WebApi.UmbracoAuthorizeAttribute` can be either used on the controller or on an individual action. Applying it to the controller will ensure all actions require an Umbraco user to be logged in, where as using it on individual action will only secure that action.
+
+To secure your controller actions based on back office membership use the attribute: `Umbraco.Web.WebApi.UmbracoAuthorizeAttribute`.
```
Shazwazza: I think the term 'back office membership' is a bit confusing since the term 'membership' is used on the front end. Perhaps should be 'back office users'

jmayntzhusen:
As we haven't seen any replies from @Harvzor  for a long time I will close this.

PR #513, opened 2017-08-10
Update registrations for treeview assemblies

quiznaut:
Without the registrations I've added, the back office throws null reference errors when clicking on Developer, Users or Forms.  I found these fixes in two forum posts: https://our.umbraco.org/projects/developer-tools/diplo-trace-log-viewer/diplo-trace-log-viewer/75668-tree-error-thrown-after-install-when-entering-developer-section and https://our.umbraco.org/forum/umbraco-7/using-umbraco-7/61745-Failed-to-retrieve-data-for-application-tree-forms-AutoFac-IOC-Umbraco-722-and-Forms.

### DIFF
--- a/Reference/using-ioc.md
+++ b/Reference/using-ioc.md
@@ -43,9 +43,16 @@ Here's an example of a custom global.asax class which initializes the IoC contai
 			//register all controllers found in your assembly
 			builder.RegisterControllers(typeof(MyApplication).Assembly);
 
-			//register umbraco MVC + webapi controllers used by the admin site
+			//register umbraco MVC + webapi controllers used in the back office
 			builder.RegisterControllers(typeof(UmbracoApplication).Assembly);
     			builder.RegisterApiControllers(typeof(UmbracoApplication).Assembly);
+			
+			//register webapi controllers for developer, users and forms
+			//treeviews used in the back office
+			builder.RegisterApiControllers(typeof(Diplo.TraceLogViewer
+				.Controllers.TraceLogController).Assembly);
+			builder.RegisterApiControllers(typeof(Umbraco.Forms.Web
+				.Trees.FormTreeController).Assembly);
 
 			//add custom class to the container as Transient instance
 			builder.RegisterType<MyAwesomeContext>();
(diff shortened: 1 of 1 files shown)


Reference/using-ioc.md
```diff
@@ -43,9 +43,16 @@ Here's an example of a custom global.asax class which initializes the IoC contai
 			//register all controllers found in your assembly
 			builder.RegisterControllers(typeof(MyApplication).Assembly);
 
-			//register umbraco MVC + webapi controllers used by the admin site
+			//register umbraco MVC + webapi controllers used in the back office
 			builder.RegisterControllers(typeof(UmbracoApplication).Assembly);
     			builder.RegisterApiControllers(typeof(UmbracoApplication).Assembly);
+			
+			//register webapi controllers for developer, users and forms
+			//treeviews used in the back office
```
Shazwazza: You don't need to mention 'treeviews' , trees are controllers. This should be re-worded like:

> you will need to register all MVC and webapi controllers found in other assemblies too. For example, if you are running in Umbraco Cloud you will have both Umbraco Forms and Diplo.TraceLogViewer installed and therefore you'll need to register their webapi controllers to, example: 

You should also leave this block of code commented out because people will just copy/paste this whole thing in and if they don't have Forms or Diplo.TraceLogViewer installed they will get compile time exceptions.

Reference/using-ioc.md
```diff
@@ -70,9 +77,16 @@ If you like to use the `IApplicationEventHandler` alternative - here is an examp
 			//register all controllers found in this assembly
 			builder.RegisterControllers(typeof(MyApplication).Assembly);
 
-			//register umbraco MVC + webapi controllers used by the admin site
+			//register umbraco MVC + webapi controllers used in the back office
 			builder.RegisterControllers(typeof(UmbracoApplication).Assembly);
     			builder.RegisterApiControllers(typeof(UmbracoApplication).Assembly);
+			
+			//register webapi controllers for developer, users and forms
```
Shazwazza: As above - please use the same type of wording

Reference/using-ioc.md
```diff
@@ -130,7 +144,7 @@ As another example, you can do the same with SurfaceControllers. Here we are cre
 
 ##What assemblies and controllers do I need to register?
 
-You need to register all assemblies that may contain MVC or WebApi controllers. In Umbraco this is the `umbraco` assembly which you can get a direct assembly reference to using the example syntax used above:
+You need to register all assemblies that may contain MVC or WebApi controllers. In Umbraco, these include the `umbraco` assembly and assemblies containing treeviews used in the back office (see the examples above). You can get a direct assembly reference with the following syntax:
```
Shazwazza: Again the mention of 'treeviews' shouldn't exist since that is not an Umbraco term. This can just be reworded as

> You need to register all assemblies that may contain MVC or WebApi controllers this includes both Umbraco and non Umbraco assemblies. For Umbraco, all MVC and WebApi controllers are in the `umbraco` assembly which you can get a direct assembly reference with the following syntax:

jmayntzhusen:
As we haven't seen any replies from @quiznaut for a long time I will close this.

Issue #518, opened 2017-08-25
A Question of Image Sizes


greystate:
In the README for this repo it's stated that [Images can have a maximum width of **800px**.](https://github.com/umbraco/UmbracoDocs/blob/master/README.md#images) — is this still the case, or are they being "sensibly" sized?

Also, related: Should they really be retina-ready versions?

/cc @hemraker

sofietoft:
Hi there,
Excellent questions! 😁 

I believe that the images are indeed sensibly sized.
I can see that I've added images where the width exceeds the 800px limit 🙈 
Those looks to be fitted well enough to the screen: https://our.umbraco.org/documentation/Umbraco-Cloud/Getting-Started/The-Umbraco-Cloud-Portal/

I can see that your PR was already merged, so I'm closing this issue down!
Thanks for contributing! Keep up the great work 😄 

-Sofie

PR #523, opened 2017-09-07
Update Getting Started screenshots to v7.6

greystate:
_This is just to get the ball rolling — I asked a question on the image sizes (in #518) , but thought it'd be just as good to have that discussion here, so..._

This PR initially adds new screenshots for the Getting Started page with dashboard, tree, sections etc.
Also updates the settings icons with the new background color.

It's a start, but the 7.6 theme feels so much more solid and well thought through, so I figured it'd be worth to bet on updating these as they'll probably be "valid" for quite some time :)

Thoughts?


hemraker:
@greystate my thoughts are that this is awesome 😄 Not only did you add updated screenshot but also cleaned up grammar and markdown. Thanks 🙌

PR #527, opened 2017-09-12
Deleting the folder itself causes the next step to fail, so empty it instead.

tomvanenckevort:
While I was following the guide during a migration of the existing site, I deleted the `data/Revision/` folder completely as step 15 suggests. However, the next step 16 (`echo > deploy-export`) would then fail.
When I put back an empty `data/Revision/` folder the deploy export would work fine.

So I think the instructions should clarify that the folder needs to stay as an empty folder, rather than deleting it.


Shazwazza:
Thanks!! and yes that makes sense to update that statement. perfect.

PR #543, opened 2017-09-20
Code samples in the ContentService

liamlaverty:
Adds code samples for the following methods:

- SaveAndPublishWithStatus
- CreateContent, parent as ID and parent as IContent
- GetById and GetByIds
- GetAncestors


dampee:
I would suggest adding the code examples in the [umbraco cms project](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Core/Services/ContentService.cs) so it ends up in the auto generated documentation.  Otherwise we will have lot's of maintaining to do.

dampee:
As there is no update on this pull request, we are closing the issue.  Thanks a lot for your effort.  If you disagree, please let us know. We are open for your feedback.

PR #545, opened 2017-09-26
Hash tag md fixes

liamlaverty:
Fixes headers for github's markdown parser. Any string in the repo starting with a hashtag but not followed by either whitespace or hash character is updated so that they've got a space following them. Should ignore hashtags in angular URLs, should ignore any hashtag where there's already a space or another hashtag following it

regex to find was ^([#]{1,})(?!#)(?!\s)(.*)

regex to replace was ${1} ${2}

### DIFF
--- a/Add-ons/UmbracoCourier/Architecture/Configuration.md
+++ b/Add-ons/UmbracoCourier/Architecture/Configuration.md
@@ -1,4 +1,4 @@
-#Configuration
+# Configuration
 
 Courier 2.5 comes with sensible defaults, but in special cases, you might need to add or modify core settings.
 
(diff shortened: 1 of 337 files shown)


dampee:
Hi @liamlaverty 

Look great.  I do have 2 questions.

1. Can you rebase your branch on the lastest changes so we don't need to fix everything manually.  There are currently issues which must be resolved before this can be merged in.
2. have you changed everything manually?

liamlaverty:
Hey,

1. I couldn't rebase properly, so I just re-did it here at #691 

2. I did them all using Sublime Text's find & replace. There's a results gist available here which shows all of the matches [here](https://gist.github.com/liamlaverty/5570b5dae145219dc92cb8e77a9af048).

I went through and double checked things like XLST, angular, and CSS where it might have got some false positives, and couldn't find any

dampee:
Closing as #691 is accepted!

PR #550, opened 2017-10-04
`CookComputing.XmlRpcV2` actually removed in 7.7.2

drbldgtl:
It was left in the nuspec per http://issues.umbraco.org/issue/U4-10443


Shazwazza:
Thansk!

PR #557, opened 2017-10-11
Update Image-Cropper.md

gijswijs:
In dynamic you should call Umbraco.Media(1234).GetCropUrl("banner") instead of Umbraco.Content(1234).GetCropUrl("banner")


nul800sebastiaan:
Whoops! Good catch! :-)

PR #566, opened 2017-10-23
Adds modelsbuilder documentation

hemraker:


hemraker:
@Shazwazza don't know if this is the "best" place to put this. It does add the Models part of MVC 😉

Shazwazza:
@hemraker I don't think this should be under MVC, i think it would be better under Templating (beside MVC), you can use these models anywhere so it's not really tied to MVC

We should probably change the MVC section to be "Razor/Mvc"

hemraker:
@Shazwazza yes, good point. Is it even necessarily tied to templating then?

Shazwazza:
I think so, i mean if you are running PureLive, then you can only use them in the templates and i guess the primary point of MB is to have nicer templating syntax.

hemraker:
👍

PR #567, opened 2017-10-24
Updates to the lovely new ModelsBuilder docs

greystate:
So pleased to see the **ModelsBuilder** docs on Our Umbraco!

However, I got the urge to run through them and fix a couple of missing/wrong links.

But then after that, I really wanted to update the code samples, so they'd be just a little easier for someone like myself to grasp :)

<details>

(E.g. instead of `MyDocument` and `MyProperty` — which could pretty much just be `Foo` and `Bar` in halloween costumes — use `TextPage` and `Header` which are actual Umbraco types we typically see on a daily basis).
</details>

---
And then it got out of hand... so I think I've touched all of the ModelsBuilder docs by now, so decided on sending a PR to get some feedback - especially, I'd love for @zpqrtbnk to have look at the updates to the Control-Generation docs, just to ensure I haven't destroyed any of the original intents 🥐 ☕️ 

Cheers!


hemraker:
@greystate absolutely amazing update, thank you so much 🙌 

I do think the examples in control-generation.md get to specific without providing context for the examples. This is perfectly done in **understand and extend**:
> a model for content type `TextPage` ends up in file `TextPage.generated.cs` and looks like:
here it is very clear where textpage comes from. 

So I think we should either preface each example so it's obvious what i.e. eventCalendar and employeeAge are or use more generic examples.

Other than that, it is magnificent! You get ⭐️⭐️⭐️⭐️⭐️ from me 😄

hemraker:
Thinking something along the lines of 27d9031a6e30c733ff8cdbf5da7ea3107221f042

hemraker:
@greystate Loving the movie references 😂 

giphy 3

PR #570, opened 2017-10-31
I have removed the ```

Aaen:


sofietoft:
Hi Dennis,

You should also remove the <head> tags, I believe!

All the Best
Sofie

PR #577, opened 2017-11-07
Fix broken DropDown-List-Multiple samples

matthew-darby:
Current code samples cause an error in later versions of Umbraco


dampee:
Hi Matthew,

Thanks for your suggestion!

  I agree it should reflect the new versions.  On the other hand, your code would cause errors in older versions. But we should take note that you can't ignore older versions without ModelsBuilder just yet.

Could you improve accordingly?

Jeavon:
There are some examples of how to incorporate this type of sample (v7.6+ value converters enabled) in some of the editors, take a look at https://github.com/umbraco/UmbracoDocs/blob/master/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker.md

jmayntzhusen:
As we haven't seen any replies from @matthew-darby  for a long time I will close this.

PR #582, opened 2017-11-22
Add some documentation for the EditorModelEventManager SendingContent…

marcemarc:
…Model event

Think this might be a really useful event for people to be aware of, but it's not really mentioned on any of the official Umbraco Training courses nor in the documentation that I can find, other than the generated API reference, but if you didn't know it existed, you probably wouldn't look for it, and if you did think something might exist, you might not guess/remember it's called the EditorModelEventManager... (When I forget what it's called - I actually remember it because Lee Kelleher tweeted it about once as his 'new favourite #Umbraco thing' - so I google that phrase, and then find the name of it!

Anyway, I've added this amongst the general Events documentation, but part of me thinks it's hidden here, and there should be perhaps something under the heading of 'How to set a default value', as that's probably the angle people will benefit from discovering this extension point from ... perhaps linking through to this page?

### DIFF
--- a/Reference/Events/EditorModel-Events.md
+++ b/Reference/Events/EditorModel-Events.md
@@ -49,45 +49,27 @@ Example usage of the **EditorModelEventManager** '*SendingContentModel*' event -
         <td>SendingContentModel</td>
         <td>(HttpActionExecutedContext sender,  EditorModelEventArgs&ltContentItemDisplay&gt; e)</td>
         <td>
-        Raised when ContentService.Save is called in the API.<br />
-        NOTE: It can be skipped completely if the parameter "raiseEvents" is set to false during the Save method call (true by default).<br />
-        "sender" will be the current IContentService object.<br />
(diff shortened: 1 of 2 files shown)


Jeavon:
First code block formatting seems a little wonky but great otherwise!

marcemarc:
Cool, yes I'll tweak the code block, also have noticed the descriptions are wrong, (cut and paste the table format from content service events) so will update that too

marcemarc:
Updated the Pull Request, how does that look now? @Jeavon @dampee reluctant to merge my own PR seems wrong somehow!

Jeavon:
Merged, fantastic work @marcemarc :thumbsup:

marcemarc:
woot, I'd have thought merging would have automatically closed the issue: https://github.com/umbraco/UmbracoDocs/issues/615 but actually think I created the issue after the PR, closing it now, just another 46 issues to go!

PR #586, opened 2017-12-07
Updated "generate UDA files" section

dzordzs:
After you run echo > deploy-export, you need to run additional echo> deploy in order to generate UDA files. :)


sofietoft:
Hi Georgs,
Thanks a bunch for your contribution!! #h5yr

I just made a minor adjustment.
All the Best
Sofie

PR #594, opened 2017-12-09
Add table of User Permission Codes

marcemarc:
Answered a forum post on what each letter corresponded to for User Permissions, thought this used to be written down somewhere?

If not there is a table of them in this Pull Request, or did this change in the new User Group stuff, @Shazwazza @andybutland?


AndyButland:
Not that I'm aware of - believe they are still there, just assigned to groups instead of users.  And good idea!  I remember struggling to find a "spare" one of these when looking to add a new action before.

PR #604, opened 2018-01-06
Added documentation for UserService

abjerner:
As there wasn't already a documentation page for the user service, I've made one based on the corresponding page for the content service - but with a few differences.

The page for the content service (and the other service pages) lists the methods available in the service, but also has a link to the API docs. So perhaps it is a bit redundant to have the method list in the documentation page as well?

From what I understand, the API docs pages are automatically generated, while the documentation page isn't? If so, my suggestion would be to just have the method list in a single place (the API docs), and skip it for the page in the regular documentation.

The page for the content service also mentions **Please note that this page will be updated with samples and additional information about the methods listed below**, but doesn't actually have any samples (none of the other service pages do either).

Again assuming the documentation pages are manually maintained, I've skipped this line, but then instead introduced a **Samples** section. Right now there is only a single sample, but my thought was to create one or more samples for each method (or whatever is relevant). IMO this explains how to use the user service a lot better.

Anyways, as my page for the user service is structured a bit different than the pages for the other services, I'll look forward to some feedback ;)


Shazwazza:
I agree we shouldn't duplicate the listing of methods, etc... in the docs since the API docs are auto generated with this information anyways. No sense in duplicating all of our work especially since that will be much harder to maintain

abjerner:
Awesome. Thanks for the feedback 👍 

By the way, a little side node for the part about assigning a user to a user group:

https://github.com/abjerner/UmbracoDocs/blob/749b195b5d6fa21bec5686a7cc1179399a5ad4b7/Reference/Management/Services/UserService/Create-a-new-user.md#assigning-the-user-to-a-user-group

The `GetUserGroupByAlias` returns an instance of `IUserGroup`, but the `AddGroup` method takes an instance of `IReadOnlyUserGroup`. I haven't been able to find any documentation why the latter isn't just `IUserGroup`, so my example casts the user group to `UserGroup`, which seems to implement both interfaces. It works, but I guess it's a bit wrong to cast it to the class.

abjerner:
@Shazwazza Any progress on merging this? If you agree with my changes to the document/file structure, I'd he happy to apply this to the files of the other services as well ;)

Reference/Management/Services/UserService/Create-a-new-user.md
```diff
@@ -0,0 +1,85 @@
+# Create a new user
+
+The document illustrates how you can create a new backoffice user from a Razor view. 
+
+
+### Getting a reference to the UserService
+The first step is to get a reference to the `UserService`. This is useful, as we're using in multiple times, and therefore doesn't have to reference it through `ApplicationContext.Current.Services.UserService` each time:
+
+    // Get a reference to the user service
+    IUserService us = ApplicationContext.Current.Services.UserService;
```
Shazwazza: I don't agree with telling people to use Singletons. In fact in 95% of cases there's no reason to use singletons at all. All of these services are exposed on base classes which in most cases people will be using. Even the ServiceContext is exposed as a `Services`  property.  We should be encouraging people to use all of the services we expose on base classes instead since they are all readily available there including: ApplicationContext, ServiceContext, UmbracoHelper, UmbracoContext, etc... and the base classes range from Views, controllers, web services, http handlers, etc...

Reference/Management/Services/UserService/Create-a-new-user.md
```diff
@@ -0,0 +1,85 @@
...
+    // Creates a new user
+    IUser user = us.CreateUserWithIdentity("john-doe@xample.org", "john-doe@xample.org");
+    
+The `CreateUserWithIdentity` method takes two parameters - `username` and `email`. However in newer versions of Umbraco, it's considered best practice to also use the email address as the username. Calling `CreateUserWithIdentity` will create the user in the database, so it is not necessary to call the `Save` method for this (unless you further update the user as shown below).
+
+### Setting a password
+So far we have created the user, but we haven't yet set a password. We can do this through the `SavePassword` method, and Umbraco will then handle all the necessary password hashing for us:
+
+    // Set a new password for the user
+    us.SavePassword(user, "HelloWorld1234");
```
Shazwazza: This is a grey area (but i see you've noted some of this below)

This is not exactly the way to deal with passwords. Passwords policies are all governed by ASP.NET Identity and the user service does not wrap this (and cannot because there would be a circular reference). Currently this method wraps the membership provider that ASP.NET identity talks to but in recent versions of Umbraco (IIRC 7.7), this method is no longer used at all... it in fact is obsoleted and hidden but not at the IUserService interface, only the UserService class (oops!), so those attributes need to be added at the interface level too. 

This could actually cause some problems and the user won't be able to login since if the ASP.NET identity `IPasswordHasher` is replaced with a custom one.  The password hashed here will always be done with the membership provider but if the `IPasswordHasher` is replaced with a custom one

The other caveat to this is that it will work currently if the `IPasswordHasher` is not replaced by a custom one but only in a couple scenarios: 
* the user is new and doesn't have a password set
* during installation to set the admin password

Otherwise this method will fail because to change someone's password with the membership provider you have to pass in the current password and this method doesn't support that which is also why it's obsoleted and no longer used.

Remember, the services are for persisting data and talking to the database. 

So if you really want to create a new user, you'd use ASP.NET identity APIs like we do in core. However if you want to write information directly to the database, you'd use the user service. Both the SavePassword methods on IUserService and IMemberService are or should be obsolete. These methods existed a long time ago but should have really never been created and more importantly will no longer work in many cases since the membership providers  are updated to require an existing password to change them and with users ASP.NET Identity is in charge.

Reference/Management/Services/UserService/Create-a-new-user.md
```diff
@@ -0,0 +1,85 @@
...
+## Full example
+
+    @using Umbraco.Core.Models.Membership
+    @using Umbraco.Core.Services
+    @inherits UmbracoViewPage
+
+    @{
+    
+        // Get a reference to the user service
+        IUserService us = ApplicationContext.Current.Services.UserService;
```
Shazwazza: No need to use singletons here, this is already exposed on the base class!

Reference/Management/Services/UserService/Create-a-new-user.md
```diff
@@ -0,0 +1,85 @@
...
+    
+        // Creates a new user
+        IUser user = us.CreateUserWithIdentity("john-doe@xample.org", "john-doe@xample.org");
+
+        // Set a new password for the user
+        us.SavePassword(user, "HelloWorld1234");
```
Shazwazza: It might be worth an example on doing this with ASP.NET identity APIs instead like we do in core.

Reference/Management/Services/UserService/Get-a-reference-to-a-user.md
```diff
@@ -0,0 +1,52 @@
...
+## Full example
+
+    @using Umbraco.Core.Models.Membership
+    @using Umbraco.Core.Services
+    @inherits UmbracoViewPage
+
+    @{
+
+        // Get a reference to the user service
+        IUserService us = ApplicationContext.Current.Services.UserService;
```
Shazwazza: singleton not needed

Jeavon:
@Shazwazza do you think we could merge this PR now and create a issue for adding examples using ASP.NET Identity APIs?

Reference/Management/Services/UserService/Create-a-new-user.md
abjerner: Good point. The examples should be updated to use either `Services.UserService` or  `ApplicationContext.Services.UserService` ;)

Reference/Management/Services/UserService/Create-a-new-user.md
abjerner: Reference to singletons should be removed now ;)

Reference/Management/Services/UserService/Get-a-reference-to-a-user.md
abjerner: Fixed ;)

abjerner:
@Shazwazza I've now updated the examples to not use `ApplicationContext.Current`, and tried a little better to describe how to use either `Services.UserService` (eg. in controllers) or `ApplicationContext.Services.UserService` (Razor).  🎉 

I'm not 100% sure how to use the ASP.NET Identity APIs, so I'm all for Jeavon's suggestion to merge now, and then rectify it with an issue / a pull request after that.

clausjensen:
I'd like to have this merged in, but I don't think we should do it with the current `SavePassword` part included.
This is not the correct way to set a password for a user (as Shannon mentions it basically works around any restrictions/customizations put in place by the Identity APIs by using direct access to the database).

We can't really merge in something that we don't recommend doing and then have an outstanding task of having it updated to be what we actually recommend. We all know that this task will not be done anytime soon - otherwise this PR would just have been updated with this info before being pulled in 😃 

I've created an issue for getting the UserService interface updated with obsoletes.

In my opinion this cannot be merged until we update the examples with the correct way of updating a user's password to using the Identity API.

abjerner:
@clausjensen fair point ;)

But so far I haven't been able to find out how this is done the *right* way. Googling how to change the user password in Umbraco, I've only been able to find examples with people either changing the password directly in the database, or using the `SavePassword` method as in my example.

I can see that Umbraco internally uses the [PasswordChanger](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/Editors/PasswordChanger.cs) class, but this is an internal class, which we therefore can't use.

Would something like the code below be the correct way?

```C#
UsersMembershipProvider provider = (UsersMembershipProvider) Membership.Providers["UsersMembershipProvider"];
provider.ChangePassword("johndoe@example.com", "1234", "password");
```

or for a new user:

```C#
UsersMembershipProvider prov = (UsersMembershipProvider) Membership.Providers["UsersMembershipProvider"];
string password = prov.ResetPassword("johndoe@example.com", null);
prov.ChangePassword("johndoe@example.com", password, "password");
```

<hr />

Anyways, for the sake of simplicity, and opposed to marking the method as obsolete, couldn't the `SavePassword` method be updated to call the right logic under the hood? This way, managing users will follow the same pattern as the other services in Umbraco where one just have to use the relevant service, and not some other logic.

clausjensen:
@abjerner Take a look at this one https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/Security/MembershipHelper.cs

abjerner:
@clausjensen Thanks. I'll have a look at that ;)

What are you're thoughts on updating the `SavePassword` method to call the correct logic under the hood?

dampee:
@clausjensen Could you pitch in one last time so this can be merged?

clausjensen:
We cannot update the method to do something (very) different than it used to do before without it being a breaking change. Can consider doing it in V8 though.

The PR here will have to be updated according to the above comments before it can be merged in

Shazwazza:
The IUserService.SavePassword is obsolete and gone in v8. We cannot wrap ASP.NET Identity behind the user service. The service is for persisting data to the database, not for managing passwords, etc...

We're stuck in an ugly scenario because we have these old membership providers and ASP.NET competing for the same space and we are magically maintaining compatibility all while trying to phase out membership providers entirely. It is possible for developers to change the password hashing by either manipulating ASP.NET identity and/or the membership providers so we need to cater for this which is why it's not as straight forward as we want. 

The `PasswordChanger.ChangePasswordWithIdentityAsync` is essentially what we use to change passwords since it wraps all the zany logic we need to support including being able to change a user's password without knowing their existing password, but for what you are trying to do, all you need to do is use the `BackOfficeUserManager` to change the password

`BackOfficeUserManager.ChangePasswordAsync(userId, currentPassword, newPassword);`, you can get a `BackOfficeUserManager` from `IOwinContext.GetBackOfficeUserManager()`

dampee:
Let's have a birthday cake.  This PR is a year old.

clausjensen:
Well we can't really just merge in the PR when the issues we've highlighted with the documentation examples hasn't been rewritten yet...

jmayntzhusen:
Hey @Shazwazza 

We have now added versioning where we keep the savePassword method for the pre 7.7 version of the documentation and mention them having to use ASP.NET identity for post 7.7, we do not have an example yet, and will also have to add in the v8 way of doing it later, but this way we can at least get in all of the other great work done by @abjerner !

What do you think? 🙂

jmayntzhusen:
Talked with @nul800sebastiaan, and as @Shazwazza  hasn't had time to review we will just merge it in, and if he has comments later on we will create an issue to have the information here amended 🙂

PR #610, opened 2018-01-10
Add umbracoLocalTempStorage for umb7.7.3+

mistyn8:
http://issues.umbraco.org/issue/U4-10503

https://our.umbraco.org/forum/contributing-to-umbraco-cms/88804-should-umbraco-be-able-to-work-out-that-its-running-on-azure-and-apply-sensible-defaults#comment-280782


Jeavon:
Thanks for this @mistyn8 I'm going to merge and then add a bit more info about umbracoLocalTempStorage as it effects more than only the xml cache file when used

Issue #606, opened 2018-01-09
The documentation on Custom HTML markup is behaving strangely.


Frost117:
In the Forms documentation, under the Developer section, the Custom HTML Email is behaving very odd. 
It is has closed brackets seeming randomly all over, but taking a closer look, it looks like it has not been implemented correctly.

Frost117:
https://our.umbraco.org/documentation/Add-ons/UmbracoForms/Developer/Email-Templates/

PR #611, opened 2018-01-12
Fixes issue #606 (Email templates page not displaying properly in our.umbraco) - I think

liamlaverty:
replaces the three backtick with tabs the whole way down the file. I think that fixes the issue at #606 https://github.com/umbraco/UmbracoDocs/issues/606

Also fixes the heading in github markdown


hemraker:
Awesome!!!! Thanks bud 🎉

PR #617, opened 2018-01-23
Update index.md

pat-flew:
typo: possessive use of it's -> its


dampee:
Great catch! Thanks a lot @pat-flew

PR #619, opened 2018-01-25
Added Markdown Editor

biapar:
Added Markdown Editor


dampee:
Thanks for the suggestion.  

There is a default PropertyValueConverter ([MarkdownEditorValueConverter](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/PropertyEditors/ValueConverters/MarkdownEditorValueConverter.cs)) converting the markdown.  Isn't it easier to point out this will be handeld by the modelsbuilder.  Or if we really need to give some example code, reuse the included Markdown converter (and removing suggestion of using nuget packages)?

dampee:
Hi
This file has been updated already, We are closing the issue. 
Thanks a lot for your suggestion.  If you have other changes, please do not hesitate to submit them!

PR #620, opened 2018-01-26
Update index.md on key umbracoVersionCheckPeriod 

fungybytes:
Added information about how to set the umbracoVersionCheckPeriod key to never let the backoffice check for updates. By default Umbraco ships with a value of '7' is not true because this key is not present in the default installation.

### DIFF
--- a/Reference/Config/webconfig/index.md
+++ b/Reference/Config/webconfig/index.md
@@ -95,4 +95,4 @@ If you are not running a load balanced environment on a central SAN based file s
 The default value is:
 `7`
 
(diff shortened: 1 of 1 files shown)


Jeavon:
Umbraco does use a value of 7 if this appsetting isn't present, see code https://github.com/umbraco/Umbraco-CMS/blob/368fec4ac1c7b2083d80cbbc6b2bc79d783b3729/src/Umbraco.Core/Configuration/GlobalSettings.cs#L519

fungybytes:
Thanks guys for reviewing this

PR #625, opened 2018-01-30
Update Scripts to latest Umbraco Forms version.

janborup:


warrenbuckley:
Hi @janborup thanks for your PR
Can you tell me, if this update was for a specific version of Forms, only this seems to be out of date a little bit than what I was expecting.

Only because the contents of this file has changed in V6/V7 with the introduction of Forms & Themes.

The file `Views/MacroPartials/InsertUmbracoFormWithTheme.cshtml` contains the following where a macro parameter is used to include/exclude the scripts when rendering the form.

Many Thanks,
Warren :)

```
@inherits Umbraco.Web.Macros.PartialViewMacroPage
	
@if (Model.MacroParameters["FormGuid"] != null)
{
    var s = Model.MacroParameters["FormGuid"].ToString();
    var theme = "";
    var includeScripts = true;

    @*
        If a theme is provided as a macro parameter, forms will attempt to use custom theme files
        ex: given the value "test", forms will look up views for the form and fields in a custom directory:

        Form will be rendered with
            /Views/Partials/Forms/Themes/test/form.cshml

        Fields will be rendered with
            /Views/Partials/Forms/Themes/test/FieldType/[fieldtypename].cshtml

        If a file is not available in the custom directory, forms will fallback to the default files in
            /Views/Partials/Forms/Themes/default/
    *@

    if(Model.MacroParameters.ContainsKey("FormTheme"))
    {
        theme = Model.MacroParameters["FormTheme"].ToString();
    }

    if (Model.MacroParameters.ContainsKey("ExcludeScripts"))
    {

        @*
            Inverse the boolean - RenderAction Parameter expects it to be false for exclusion
            But with default value for Macro Paramater boolean/checkbox being false 
            We want to ensure by default we include the scripts next to the form
        *@
        includeScripts = Model.MacroParameters["ExcludeScripts"].ToString() == "0";
    }

    var form = new Guid(s);
    var recordGuid = Guid.Empty;

    if (string.IsNullOrEmpty(Request.QueryString["recordId"]) == false)
    {
        Guid.TryParse(Request.QueryString["recordId"], out recordGuid);
    }

    Html.RenderAction("RenderForm", "UmbracoForms", new {formId = form, recordId = recordGuid, theme = theme, includeScripts = includeScripts });
}
```

OlePc:
I believe this PR i obsolete with the v7 changes that @warrenbuckley mentions. Instead of `mode="form"`, you should now use `ExcludeScripts = true` when rendering the InsertUmbracoFormWithTheme macro.

sofietoft:
Hi @janborup ,

We've released a few new versions of Umbraco Forms since you made this PR.
Unfortunately this means that your PR has become obsolete 🙈 

So sorry for the inconvenience!
Hope to see you back here with more contributions 😁 

-Sofie

PR #626, opened 2018-02-02
Fix grammar (re-use)

umohl:
Reuse is a single word, not a hyphenated compound.


dampee:
According https://en.wiktionary.org/wiki/re-use it is allowed to use a hyphen.  The https://en.oxforddictionaries.com/definition/reuse and https://www.merriam-webster.com/dictionary/reuse both don't mention the hyphen. Therefore I merged.

PR #627, opened 2018-02-05
Add clear example for @section usage

umohl:
As @RenderSection has a code example, I feel that @section should have
one too.
https://our.umbraco.org/documentation/Getting-Started/Design/Templates/#template-sections

### DIFF
--- a/Getting-Started/Design/Templates/index.md
+++ b/Getting-Started/Design/Templates/index.md
@@ -85,8 +85,15 @@ By default, when defining a section it is required. To make the section optional
 
     @RenderSection("Head", required: false)
 
- On your child page call `@section Head {}` and then type your markup that will be pushed into the Master Template.  
+On your child page call `@section Head {}` and then type your markup that will be pushed into the Master Template:
 
+    @section Head {
+        <style>
+            body {
+                background: #ff0000;
+            }
+        </style>
+    }
 
 ## Injecting partial template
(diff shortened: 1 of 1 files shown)


Getting-Started/Design/Templates/index.md
```diff
@@ -85,8 +85,15 @@ By default, when defining a section it is required. To make the section optional
 
     @RenderSection("Head", required: false)
 
- On your child page call `@section Head {}` and then type your markup that will be pushed into the Master Template.  
-
+On your child page call `@section Head {}` and then type your markup that will be pushed into the Master Template:
```
marcemarc: Should we tweak this text at the same time? for clarity.... so it's 

On your child page template call...

or on your inheriting page template call...

marcemarc: That is an excellent observation @umohl and yes we should!, thinking reading the text, should we at the same time tweak 'On your child page' to be clear its an inheriting template where  you place the section... (or is it obvious?)

Let me know what you think...

then we can pull this in, if not I'll pull it in anyway because the example is great, and required here.

umohl: I think "On your child page template call ..." is gives a fine explanation, really. Inheritance is already mentioned multiple times above the section in question, so I think it's obvious.

marcemarc: cool, is it easy for you to update your PR to add the word 'template'? or I can drop it in afters...

PR #631, opened 2018-02-06
Make deploy description match the GIF

umohl:
'Activity Log' and 'View Details' are not highlighted as they are not as important (or required) as "Deploy changes".


sofietoft:
Good point, @umohl !

Thanks a bunch! Keep up the good work 😁

PR #632, opened 2018-02-06
Update IContentFinder.md

fungybytes:
Add an example of how to actually start an implementation of the IContentFinder interface and to make aware of the Is404 property that's available out of the box.


dampee:
Thanks for your submission.  

I agree that we miss an example implementation.  Even if this is a very custom setup per website.

But I am not sure if your example is the best possible:

* A `ContentLastChanceFinder` will add a 404 status code by default.  So, there is no need to use the `Is404` for a ContentLastChanceFinder.
* Also, a fallback ContentFinder which returns false as an example, is not the best idea.  We should return something!

Could you update your example accordingly?

fungybytes:
Hi Dampee, 

I've just made a few changes to the example IContentFinder implementation. Not sure about how to deal with a situation when no possible 404 node / homepage node was found. Perhaps you have any suggestions? Thanks!

Reference/Routing/Request-Pipeline/IContentFinder.md
```diff
@@ -91,6 +91,33 @@ this example shows how to add custom content finder to (and how to remove Conten
...
+    public class My404ContentFinder : IContentFinder {
+    	public bool TryFindContent(PublishedContentRequest contentRequest) {
+            //logic to find your 404 page and set it to contentRequest.PublishedContent
+	     CultureInfo culture = null;
+            if (contentRequest.HasDomain) {
+                culture = CultureInfo.GetCultureInfo(contentRequest.UmbracoDomain.LanguageIsoCode);
+            }
+
+            //replace 'home_doctype_alias' with the alias of your homepage
+            IPublishedContent rootNode = UmbracoContext.Current.ContentCache.GetByXPath("//home_doctype_alias").FirstOrDefault(n => n.GetCulture().ThreeLetterWindowsLanguageName == culture.ThreeLetterWindowsLanguageName);
```
dampee: Don't want to do difficult.  But could we do `GetByXPath("root/home_doctype_alias")`.  That is more performant than a full scan using `//`. It's just to avoid bad practices.

Reference/Routing/Request-Pipeline/IContentFinder.md
```diff
@@ -91,6 +91,33 @@ this example shows how to add custom content finder to (and how to remove Conten
...
+            //replace '404_doctype_alias' with the alias of your 404 page
+            IPublishedContent notFoundNode = UmbracoContext.Current.ContentCache.GetByXPath("//404_doctype_alias").FirstOrDefault(n => n.GetCulture().ThreeLetterWindowsLanguageName == culture.ThreeLetterWindowsLanguageName);
```
dampee: Same remark here : using `//` is just a full content scan which is not really best practice.  This is working code, so we should not let people suggest it.  
In this case I would rather use something like `GetByXpath("root/homeDocType[id=123]/404_doctype_alias") where 123 is the Id of the root node found earlier.  Or just get all the 404 pages and then filter by the correct culture.

Reference/Routing/Request-Pipeline/IContentFinder.md
```diff
@@ -91,6 +91,33 @@ this example shows how to add custom content finder to (and how to remove Conten
...
+
+            if (notFoundNode != null) {
+                contentRequest.PublishedContent = notFoundNode;
+            } else if (rootNode != null) {
+                contentRequest.PublishedContent = rootNode;
+            } else {
+                contentRequest.PublishedContent = UmbracoContext.Current.ContentCache.GetAtRoot().FirstOrDefault(n => n.GetTemplateAlias() != "");
+            }
+
+            return contentRequest.PublishedContent != null;
```
dampee: We should always return something here because it's the notFoundHandler.  We could also replace the code by a comment suggesting the user they should always have something to return.

dampee:
Hi @fungybytes 
This really starts to look good!  Only a few small things and we can merge this.
Thanks a lot for the effort you are doing!  A lot of people will profit from this.
#H5YR!

Reference/Routing/Request-Pipeline/IContentFinder.md
fungybytes: Hi Dampee, I've updated the code according to your input though i'm not sure about the returned result. I thought by checking if a node was assigned to the `contentRequest.PublishedContent ` then it's ok. Or do you mean to always assign a node that actually has a template and can be viewed like a page? Thanks!

dampee: Yes, that is what I meant.

dampee:
Hi @fungybytes 
Thanks again for your suggestion.  H5YR!

PR #633, opened 2018-02-06
Update install to VS Code instead of WebMatrix

jmayntzhusen:


dampee:
This looks awesome!

I am only wondering what the change to Getting-Started/Design/Templates/index.md has to do with it.

jmayntzhusen:
It was me being stupid and messing up a pull somehow so I merged the new commits from the UmbracoDocs into my fork then had them as part of my commit when I submitted this fork request.

It is the merge requests #626 and #627. That have been added to mine.

gareth-d85:
@jmayntzhusen I have gone through the new process to setup iis express and one thing is missing, you have to install iis express from https://www.microsoft.com/en-us/download/details.aspx?id=48264
Just installing the VS code extension will not work. it tries to look for the iisexpress.exe in C:/program files.
I hope this is helpful. ps I am new to this.

jmayntzhusen:
@gareth-d85 I had the same issue on one computer but on another I had Visual Studio installed already which includes IIS Express so I mistakenly just discarded the issue. Thanks for notifying me :)

sofietoft:
@jmayntzhusen This is really f** great!! 😁 

A couple of things:
* Does it make sense, on the /Setup/Install/index.md to have "Download and install IIS Express" **before** the "Download and unzip Umbraco" bullet?
* And perhaps also add to the final bullet something like "Open the Umbraco installation with VS Code and run the ... " ?
* Would be awesome if you could also make sure the syntax is correct on all the headings 😉 - make a space between '#' and the heading it self

Does this all makes sense?

jmayntzhusen:
@sofietoft You are right, the quick start guide was a bit rushed, I updated the things you mention in my latest commit :)

PR #638, opened 2018-02-07
Added VS17-info to installation-documentation

enkelmedia:


marcemarc:
Thanks Markus!

Now that you have your eyes open on this page :-P - what do you think about the .NET framework version recommendation for Umbraco 7...

... do you think we should update that too? the bit about it 'needing' to choose 4.5 or 4.5.1?

nul800sebastiaan:
@marcemarc I think 4.5+ is the default now for at least VS15 and 17, so I would drop it completely

image

dampee:
The .net framework thing is created as a new issue #674.  Merging this change. 

Thanks a lot Markus

PR #652, opened 2018-02-20
My two cents on UaaS-Cli

hartviglarsen:
Just some small changes :)

Including:
- Show Linux some love (works just fine, by the way. I've tested it on Solus and Ubuntu)
- Saying that Git is required seems more correct than a "Git client"

Note to self: Editing files directly on Github is very confusing because of all the red and green hightligts.


sofietoft:
This is great @hartviglarsen 😄 

The small changes gives the article a better flow!

PR #662, opened 2018-02-22
Add Safari to the list of browsers

hartviglarsen:
Some non-tech people seem to have the understanding that Umbraco won't work in Safari.


hartviglarsen:
@dzordzs The requirement page is not about the Cloud Portal, though - but Umbraco in general. 

The page even specifically say Umbraco UI.

PR #663, opened 2018-02-22
Note about directoryFactory setting with auto scale

kmeilander:
I've always found it confusing here that you set the directoryFactory once, and then 3 paragraphs later it has you set it to something else.  I added a note on the 2nd reference that you are using this instead of what was mentioned above.


Jeavon:
@kmeilander you have now earned your contrib badge on our ;) https://our.umbraco.org/member/150245

PR #666, opened 2018-02-23
Update Code overview to match the navigation

hartviglarsen:
https://our.umbraco.org/documentation/Getting-Started/Code/

### DIFF
--- a/Getting-Started/Code/index.md
+++ b/Getting-Started/Code/index.md
@@ -1,8 +1,5 @@
 #Code
 
-###[Debugging your Umbraco site](Debugging/)
-Using Tracing and MiniProfiler to debug your code
-
 ###[Using Umbracos Service APIs](Umbraco-Services/)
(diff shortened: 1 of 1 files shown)


Jeavon:
The slightly strange thing here is that as far as I can tell there isn't anything controlling the order in the left hand navigation on our... They are generated by the folder names in the repo but the order....???

hartviglarsen:
@Jeavon The navigation is defined in this file: https://github.com/umbraco/OurUmbraco/blob/master/OurUmbraco/Documentation/Busineslogic/GithubSourcePull/ZipDownloader.cs#L237

If the article/guide isn't in the file (switch statements), it will be added to the bottom of the list in the navigation tree automatically. 

@nul800sebastiaan is in charge of this. :)

Jeavon:
@hartviglarsen @nul800sebastiaan that's terrifying but very useful to know, thanks!

PR #670, opened 2018-02-25
Add a link to adding a type on Provider model

Matthew-Wise:
When working with forms adding a custom setting was something I ended up not been able to find via the documentation instead I found other peoples examples, adding a link here may help others find this information more readily.

Matt


warrenbuckley:
Thanks @Matthew-Wise

PR #683, opened 2018-02-28
XPATH Examples: Wrong output result

Mivaweb:
Removed the last output row of subject //Track\[@rating >= 4] because the rating is 1 and therefore is not  equal or higher as 4


dampee:
Spot on.  Good find! 

Thanks a lot for your contribution!

PR #686, opened 2018-02-28
Add information about Umbraco namespaces in *.Core projects with VS

hartviglarsen:
I was a bit confused about how I could use Umbraco functionality (SurfaceControllers in particular) in my *.Core project. However, I apparently have to add the references manually, as they're not there.

This is odd, though, as both Umbraco.Forms and Umbraco.Licensing are listed in References, but the rest of the Umbraco DLLs are not. I don't know whether this is a bug or on purpose.

If it's on purpose: Why? It should either reference every Umbraco DLL or none, in my opinion. Otherwise it's extremely confusing. 

The solution was created with UaaS.cmd. :-)


Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
```diff
@@ -32,7 +32,15 @@ Once you have added a new project to the solution right-click on the solution in
 
 ![](images/solution-dependencies.png)  
 
-##Git - what should be committed
+## Using Umbraco namespaces in your *.Core project
```
sofietoft: Could you put *.Core into `` ? Otherwise, the entire section will be italics 😉

sofietoft: Instead of giving this a new heading, I think it should just be a note, descriping the current work-around for the bug (http://issues.umbraco.org/issue/UAAS-527)

PR #687, opened 2018-03-01
Update data type doc page inline with Umbraco v7.9

Jeavon:
Made a quick update to this page to match Umbraco v7.9 default installation as it was very old and contained references to XSLT


sofietoft:
Hi @Jeavon ,

One thought on this one;
Would it make sense to keep a mention of the "old" dataypes? Perhaps mention them as "legacy datatypes" or something? 🤔

Jeavon:
@sofietoft In the case of the default Data types I don't think so... We do however have "Obsolete" on the property editors https://github.com/umbraco/UmbracoDocs/blob/master/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/index.md

sofietoft:
Ah, yes! You are right 😁 Forgot about those docs.

PR #691, opened 2018-03-03
updates PR#545 with up to date files

liamlaverty:
See #545 for details 

Fixes headers for github's markdown parser. Any string in the repo
starting with a hashtag but not followed by either whitespace or hash
character is updated so that they've got a space following them. Should
ignore hashtags in angular URLs, should ignore any hashtag where there's
already a space or another hashtag following it

regex to find was ^([#]{1,})(?!#)(?!\s)(.*)

regex to replace was ${1} ${2}


dampee:
Awesome!  This is amazing work.  As we are working towards github MD compatibility this is really fantastic.  

Thanks a lot for your effort!

PR #692, opened 2018-03-03
Issues/215 dashboardconfigdocumentation

marcemarc:
Rejigging some of the existing dashboard information, trying to find a pattern for this kind of feature, full config in the reference section but introducing the concepts separately - previously the extending/dashboard page contained the reference info, and the reference info was a stub - we don't really introduce what a dashboard is anywhere... except perhaps the tutorial, the extending section may be the wrong place for the 'concept introduction' text in the long run,  but at least prevents broken links.

So in short experiment is

Concept Introduction, and quick reference => leading to separate full Configuration in Reference section


Extending/Dashboards/index.md
```diff
@@ -1,136 +1,57 @@
...
-		<access>
-			<grant>writer</grant>
-			<grant>editor</grant>
-			<grantBySection>content</grantBySection>
-		</access>
-		<control>/usercontrols/dashboard/latestEdits.ascx</control>
-	</tab>
 
-## Customizing
+Checkout the awesome package 'The Dashboard' for provideing great informatino on the editorial status of an Umbraco site.
```
Jeavon: "provideing great informatino" possibly a drunk keyboard?

marcemarc:
@Jeavon that whole paragraph was hanging around from teh frist dratf... (keyboard issues not drunk)

PR #693, opened 2018-03-05
Fixed typo. Changed describtion to description.

Jwheuvel:


Jeavon:
Thanks! Do you have an profile on Our Umbraco (we can give you a Contributor badge if you don't already have one)?

PR #700, opened 2018-03-07
Correct `DateTime.MinDate` to `DateTime.MinValue`

greystate:
Nobody copies those samples anyway, right? :-)

Well, just in case they do...


dampee:
Thank a lot for the fix!  You are right nobody copies this anyway  :-D

Issue #634, opened 2018-02-06
Update "Setup Your Site to use Azure Storage for Media" 


Jeavon:
The security.config example is outdated and needs to be updated to use the container setting

ed-parry:
@Jeavon Happy to PR for this one if you're not looking at it? From what I can see from recently implementing this process, the `Container` is missing from `security.config` as you said, and `CachedCDNTimeout` and `UseCachedContainerInUrl` from `cache.config`?

Jeavon:
@ed-parry yes, that's exactly it. If you want to send in a PR I would love to see it :)

ed-parry:
There you go, @Jeavon! Added in the missing items and updated Azure-based URLs to recommend HTTPS by default (though made an assumption sites on the umbraco.io domain pick up HTTPS by default?)

Also set `UseCachedContainerInUrl` to `false` by default, as the latest [ImageProcessing docs](http://imageprocessor.org/imageprocessor-web/plugins/azure-blob-cache/) suggest newer Azure formats require that. Though we're using a couple of implementation currently with that set to `true` without issue, so you're probably in the best place to confirm it either way.

PR #716, opened 2018-03-17
Update "Setup Your Site to use Azure Storage for Media"

ed-parry:
For Issue #634. Updated the required configuration for Azure Storage (blob/CDN). Updated Azure-based URLs to recommend HTTPS.


Umbraco-Cloud/Set-Up/Media/index.md
```diff
@@ -127,7 +128,9 @@ Update `~/config/imageprocessor/cache.config` by removing the default “DiskCac
           <settings>
             <setting key="CachedStorageAccount" value="DefaultEndpointsProtocol=https;AccountName=;AccountKey=" />
             <setting key="CachedBlobContainer" value="cache" />
-            <setting key="CachedCDNRoot" value="http://<your Azure CDN>.vo.msecnd.net/" />
+            <setting key="CachedCDNRoot" value="https://<your Azure CDN>.vo.msecnd.net/" />
+            <setting key="CachedCDNTimeout" value="2000" />
+            <setting key="UseCachedContainerInUrl" value="false" />
```
Jeavon: I think UseCachedContainerInUrl should be "true" by default. Only if you have used a "Origin path" when setting up the CDN endpoint would you want to set it to "false"?

ed-parry: @Jeavon Makes sense - I mentioned this in the issue as well:

"Also set `UseCachedContainerInUrl` to false by default, as the latest ImageProcessing docs suggest newer Azure formats require that. Though we're using a couple of implementation currently with that set to true without issue, so you're probably in the best place to confirm it either way."

I'll update and push up shortly :)

Jeavon: I've just provisioned a new CDN to double check too

ed-parry: From what I can see in our instances, it should be the recommendation to be `true` by default, unless your tests suggest otherwise?

Jeavon:
Thanks @ed-parry

PR #723, opened 2018-03-22
Added documentation for allowManuallyChangingPassword

JeffreyPerplex:
I thought it did nothing, but I found out that it does something when you try to change a password from code


Jeavon:
Thanks @JeffreyPerplex

PR #726, opened 2018-03-27
Document ncNodeName filter

dawoe:
This angular filter allows to get the name of  item picked with a content picker as name template. This was also in original docs and the filter is present in the current core version

https://github.com/umco/umbraco-nested-content/blob/develop/docs/developers-guide.md#name-template

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Nested-Content.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Nested-Content.md
@@ -17,7 +17,7 @@ You should then be presented with the **Nested Content** property editors data-t
 
 The data-type editor allows you to configure the following properties:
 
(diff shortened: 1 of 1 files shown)


dampee:
I did not know that the filter was added in core too.

https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web.UI.Client/src/common/filters/nestedcontent.filter.js

dampee:
Thanks a lot Dave!  I had to merge this immediatly because this is very valuable information. :-)

Awesome to have a contribution from you. 👍 Let's keep them coming!

dawoe:
There is one coming on this same functionality. I'm extending the filter so it also works for media pickers and possibily member pickers. Once the PR for core is done, I'll create a new one here

dampee:
Cool thanks!

PR #727, opened 2018-03-27
Update Umbraco.Library documentation

marcemarc:
Cutting edge, but here is a list of the methods on Umbraco.library, if this is useful to anyone (it helps us close an 'issue' in github :-)) for 'missing content'


dampee:
@marcemarc we can close the issue now.  :-)

PR #729, opened 2018-03-30
Adding information in the reference section for the trees.config file

marcemarc:
Updating the reference page for trees.config configuration file, to contain details of the trees.config, and linking through to this information from the extending trees section.


dampee:
Thanks again @marcemarc

PR #733, opened 2018-04-05
Updated backoffice rewrite rule

MartijnKooij:
The regular expression for blocking access to the backoffice contains an error which makes it easy to get around it.
The problem is that the rule only prevents access to https://site.com/umbraco but not https://site.com/umbraco/#/somerandombackoffice part.
The updated RegEx should be: (^umbraco/backoffice/(.*)|^umbraco($|/$))


dampee:
Hi @MartijnKooij 
You are right!  Thanks a lot for the suggestion!

Quick question, are you sure about the solution?
I mean, should a forward slash not be escaped?

     (^umbraco/backoffice/(.*)|^umbraco($|\/$))

I can't test this at the moment.  Just wanted to be sure.

Issue #639, opened 2018-02-08
Document the appSetting: umbracoDisableElectionForSingleServer and it's role in flexible load balancing


marcemarc:
We should document this setting and how it affects flexible load balancing:

               //by default we'll use the db server registrar unless the developer has the legacy
            // dist calls enabled, in which case we'll use the config server registrar
            if (UmbracoConfig.For.UmbracoSettings().DistributedCall.Enabled)
            {
                ServerRegistrarResolver.Current = new ServerRegistrarResolver(new ConfigServerRegistrar());
            }
            else if ("true".InvariantEquals(ConfigurationManager.AppSettings["umbracoDisableElectionForSingleServer"]))
            {
                ServerRegistrarResolver.Current = new ServerRegistrarResolver(new SingleServerRegistrar());
            }
            else
            {
                ServerRegistrarResolver.Current = new ServerRegistrarResolver(
                    new DatabaseServerRegistrar(
                        new Lazy<IServerRegistrationService>(() => ApplicationContext.Services.ServerRegistrationService),
                        new DatabaseServerRegistrarOptions()));
            }

So essentially if the setting is set to true, the server won't register itself in the pool of Servers for flexible load balancing, but it's a bit obscure because people won't see 'Election' to be something related to Server Role assignment...

Perhaps too if we have a Troubleshooting Flexible Load Balancing section just confirming

1) Distributed Calls Off
2) umbracoDisableElectionForSingleServer - false
3) umbracoApplicationUrl set and resolvable by all servers... 
etc

Shazwazza:
There's a whole topic in the docs about election here: https://our.umbraco.org/documentation/Getting-Started/Setup/Server-Setup/load-balancing/flexible#scheduling-and-master-election

Enabling this setting simply means that any election will be disabled. We use this on Umbraco Cloud since we are certain there is no load balancing taking place. This can slightly improve startup performance since no DB calls are needed.

marcemarc:
@Shazwazza Yes, that is the page I was going to add it on, helped a client today, who couldn't get flexible load balancing to work, they had that setting, set to true in the web.config, in fact, the site 'used' to be hosted on Umbraco Cloud, which probably explains why it was there. So I think important to mention the setting, somewhere in the docs, and within the context of this page,  and I was thinking it will usually be within the context of troubleshooting... the text on this page is very informative but very dense to read through, if you are trying to work out what you have done wrong... so a quick troubleshooting checklist may help here!

[...]

PR #738, opened 2018-04-06
Added umbracoDisableElectionForSingleServer docs

dawoe:
This value is set on all our Umbraco Cloud projects. So worth documenting


dampee:
Thanks a lot Dave.  I am wondering how it is possible we missed this one! 

H5YR !!

dampee:
Just for reference, this is a part of Issue #639

PR #739, opened 2018-04-06
Update the True/False property editor

Mivaweb:
Added the new definition which contains the Default Value setting from Umbraco version 7.4.2


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/True-False.md
```diff
@@ -6,7 +6,9 @@ True/False is a simple checkbox which saves either 0 or 1, depending on the chec
 
 ## Data Type Definition Example
 
-![True/False Data Type Definition](images/True-False-DataType.png)
+![True/False Data Type Definition](images/True-False-DataType-742.jpg)
+
+From Umbraco version 7.4.2 the True/False property has a setting which allows you to set the default value of the checkbox, either 0 or 1.
```
dampee: I would rather love to see a less "developer" perspective.

1) either 0 or 1 should rather be something like "checked or unchecked".  Or "on or off".  Or why not checked (true) or unchecked (false).
2) Mentioning the version is very good information.  But I would rather add this to the end of the line.  This allows us to "rewrite" it easer at a later time.  E.g.: `This property was added in Umbraco 7.4.2.`  Or `(only Umbraco 7.4.2+)`

Mivaweb: You are correct sounds indeed to much developer language, will change it!

Mivaweb:
New commit pushed to this PR with required changes from your comment. Thanks for the help!

dampee:
Hi @Mivaweb 
Thanks a lot for your contribution (again) and improving the Umbraco Docs.

You rock!

PR #741, opened 2018-04-13
Added info regarding umbracoDisableElectionForSingleServer

dawoe:
Added extra information that this should be removed from the web.config or set to false for flexible loadbalancing to work

### DIFF
--- a/Getting-Started/Setup/Server-Setup/Load-Balancing/flexible.md
+++ b/Getting-Started/Setup/Server-Setup/Load-Balancing/flexible.md
@@ -72,6 +72,8 @@ In many scenarios this is fine, but in case this is not adequate there's a few o
 
 There isn't any _Umbraco_ configuration file changes necessary. **You must not enable the distributed calls flag in the umbracoSettings.config** file for Flexible Load Balancing to work, that is purely for Traditional load balancing. 
 
+Also check if the app-setting [umbracoDisableElectionForSingleServer](../../../../Reference/Config/webconfig/index.md) is set to false or removed in the web.config. Otherwise your servers will all be setup as single servers.
+
 There are some Examine/Logging config file updates needed (see below and the [Overview](index.md))
 
 ## Option #1 : Cloud based auto-scale appliances


nul800sebastiaan:
Closing for now, see comment on https://github.com/umbraco/Umbraco-CMS/pull/2582

dawoe:
And what if this has been set ? And then the environments will be switch to loadbalanced environment. Would be nice to have this information here. There is also a mention that you need to check that traditional loadbalancing is disbalbed ?

dampee:
There is also a reference to this setting if you do a [manual upgrade to Umbraco Cloud](https://github.com/umbraco/UmbracoDocs/blob/master/Umbraco-Cloud/Upgrades/Manual-Upgrades/Manual-CMS-upgrade.md).  Likewise, if you move away from the cloud or you want to check your loadbalancing setup, we should mention to verify this setting.  Therefor I reopen the pull request.

However, I do not know if the wording is correct.  Is it wise to advise people to set it to false?  It's the default value.

I'll check with @Jeavon and @marcemarc on there opinion.  

Can we agree to leave this pull request open until for now?

marcemarc:
Have fallen foul of people moving away from Umbraco Cloud and then trying to setup flexible load balancing and it fails because this setting is still in the web.config from the migration from cloud. So yes we should mention it's existence in reference to flexible load balancing, hence we created this issue: https://github.com/umbraco/UmbracoDocs/issues/639 the conclusion we came to was there should be a common 'Troubleshooting list' for flexible load balancing, including the check for this setting added. That made me think it could just be a healthcheck, which I started here: https://github.com/umbraco/Umbraco-CMS/pull/2443 was going to hopefully get feedback on that, and perhaps get into the core, then create the troubleshooting section text to refer to the healthcheck! ... best laid plans and all that

dampee:
As elaborated in #742 we should should mention to remove the setting.

dampee:
Discussed this with my fellow curators.

Taking the other comments on the other pull request into considaration, we should move this to a separate "troubleshooting" flexibel load balancing section.

@dawoe do you have some time to update the pull request?

dawoe:
Will try to find some time this week. If not remind me 😄

dampee:
@dawoe created an issue ( #1011 ) to be sure we do not forget this one.  :-)

PR #744, opened 2018-04-17
SSL is deprecated and we shouldn't pretend like we support it

nul800sebastiaan:
SSL is not a thing any more, SSL is depracated, at best we could change "SSL" to "TLS" but the technical name for it doesn't matter very much. Let's stick with "certificate" as the regular term to use for this.

### DIFF
--- a/Umbraco-Cloud/Set-Up/Manage-Domains/Security-Certificates/index.md
+++ b/Umbraco-Cloud/Set-Up/Manage-Domains/Security-Certificates/index.md
@@ -2,13 +2,13 @@
 
 ![Manage certificates](images/manage-certificates.png)
 
-On the **Manage domains** page you'll also find the option to upload and configure SSL certificates for your Cloud environments.
+On the **Manage domains** page you'll also find the option to upload and configure certificates for your Cloud environments.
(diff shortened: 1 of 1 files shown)


dampee:
Hi @nul800sebastiaan 

While I know you are right, I am afraid lots of people will still search for SSL certificates.  

Could we add a paragraph telling them we only support TLS and not the deprecated SSL?  Or do you think that is a bad idea?

nul800sebastiaan:
@dampee 
Yes! I would love if we added an explanation of the difference between TLS and SSL. The problem here is that it will sound like we offer SSL while it is in fact completely disabled on our servers.

@sofietoft Hmm .. ;-)
>In Canadian law, a security certificate is a mechanism by which the Government of Canada can detain and deport foreign nationals and all other non-citizens living in Canada.

I think either referring to HTTPS or just "certificate" should be fine good. _HTTPS certificate_ could work.

nul800sebastiaan:
Some people also refer to _"Website security certificate"_ - not sure about that.

dampee:
Let's go with HTTPS certificate.

Issue #712, opened 2018-03-14
Umbraco.Media vs Umbraco.TypedMedia


NikRimington:
Hi Curators of the Umbraco Docs.

As I understand it Dynamics is being phased out as the best approach for accessing content. I was under the impression that this was also true for accessing media. Does the following page need updating to show the TypedMedia way of accessing media? 

https://github.com/umbraco/UmbracoDocs/blob/master/Getting-Started/Design/Rendering-Media/index.md

(If it does I'm happy to look at creating a pull request for it :-)  )

Thanks.

Jeavon:
:thumbsup: Absolutely! Look forward to seeing a PR from you.

marcemarc:
Was about to say exactly the same thing! we should be looking I think to show the TypedMedia approach but also how you'd work with media if using modelsbuilder - also looking very forward to seeing a PR :-P

dampee:
My eyes hurt when I see all these "dynamics" in examples. :-)

ed-parry:
Looks like this one can be closed now?

dampee:
We could indeed close this one.  There are certainly other examples, but we could just create a new issue at that moment!

PR #747, opened 2018-04-19
Updating Render Media file to use TypedMedia and include a Models Builder related example - Issue #712

NikRimington:
Hi Currators, 

Pull request for issue: #712 

I've replaced Media with TypedMedia in the examples.
In addition:
I've added additional copy around the existing examples, 
and added an example that talks about Models Builder as well.


Getting-Started/Design/Rendering-Media/index.md
```diff
@@ -1,40 +1,71 @@
...
+
     @{
-        var mediaItem = Umbraco.Media(1234);
-        var height = mediaItem.UmbracoHeight;
-        var file = mediaItem.UmbracoFile;
+        var mediaItem = Umbraco.TypedMedia(1234);
+        var height = mediaItem.GetPropertyValue<string>("umbracoHeight");
+
+        //There are two ways to get your media item url, this is one way.
+        var file = mediaItem.GetPropertyValue<string>("umbracoFile");
```
Jeavon: This won't work as the default is the image cropper for type "Image" (it would work for type file), can you please remove?

Getting-Started/Design/Rendering-Media/index.md
```diff
@@ -1,40 +1,71 @@
...
+
+        //This is the other
+        var url = mediaItem.Url
     }
 
     <img src="@file" height="@height" />
```
Jeavon: Perhaps rendering the name as alt text might be a better example than height?

Getting-Started/Design/Rendering-Media/index.md
```diff
@@ -1,40 +1,71 @@
...
 
-## Rendering media item with Image Cropper
-If the Image Cropper is used as the upload field (umbracoFile) on a media type you can get the cropped image by calling the `GetCropUrl` function and referencing a crop alias.
+But wait a second, why do we have to access properties in the old IPublishedContent approach when Umbraco now comes with [ModelsBuilder](../../../Reference/Templating/Modelsbuilder/index.md) you may be asking yourself. Well, if you are using Umbraco version 7.4.0 or newer, you don't. You can happily use a nicely TypedModel for your media item if ModelsBuilder is enabled.
```
Jeavon: IPublishedContent is not a old approach, you are referring to the the "stringly typed" approach, I'm not sure what we should call it officially though.... Perhaps just focus on the ModelsBuilder approach...

Getting-Started/Design/Rendering-Media/index.md
NikRimington: Hi Jeavon... I had the same thought, but I checked it on a 7.9.2 build with the starter kit and it did seem to work. 

I added this `@(Umbraco.TypedMedia(Model.Content.Photos.FirstOrDefault().Id).GetPropertyValue<string>("umbracoFile"))` to the Product.cshtml from the starter kit as a test and it did give the URL to image. I think Umbraco must be doing something under the hood because as far as I can tell it is just the standard Image media type with the ImageCropper as the umbracoFile property

I could be wrong and it could be luck that it works, but could I ask you to check?

Getting-Started/Design/Rendering-Media/index.md
NikRimington: Yeah no problem, I'll get on that. I only did the height as that was what the original one did :-)

Getting-Started/Design/Rendering-Media/index.md
NikRimington: I agree the wording isn't quite right looking back on it. But I also think we should keep both examples in. Not everyone is using ModelsBuilder, so having both GetPropertValue examples and Strongly typed examples would help. What about if we re-worded it to the classic/traditional approach when referring to GetPropertyValue?

Getting-Started/Design/Rendering-Media/index.md
Jeavon: @NikRimington I think you might have seen a quirk of how the starter kit installs the media item, try to save the media items and see if it still works

Getting-Started/Design/Rendering-Media/index.md
Jeavon: @NikRimington absolutely we should have both, I wasn't suggesting removing the "stringly" typed examples I was suggesting about the wording of this paragraph to focus on Models Builder,
Something like:
------------------------------------------------------------------
If you are using the ModelsBuilder included with Umbraco v7.4.0+ you use the strongly types models for your media items
------------------------------------------------------------------

Jeavon:
Hi @NikRimington did my last comments make sense? Is there anything I can do to help complete this?

NikRimington:
Hey @Jeavon I think so. I was hoping to get to it this weekend but unfortunately I haven't had a chance. I should be able to update my PR tomorrow evening so hopefully you will have a fresh update to review on Thursday :-)

Getting-Started/Design/Rendering-Media/index.md
NikRimington: @Jeavon you were right the starter kit was doing something odd :-) I've removed that line and just kept the .Url approach.

NikRimington:
@Jeavon I've pushed the updates now, let me know if more changes are need :-)

Jeavon:
Thanks @NikRimington Do you have a profile on Our as I can't find you?

NikRimington:
@Jeavon  Yep, just don't have my surname on it. https://our.umbraco.org/member/146168 👍

Jeavon:
Ah perfect, thanks

PR #749, opened 2018-04-23
Remove install folder text

hartviglarsen:
It's no longer needed, as the folder isn't there. :)


marcemarc:
Yes, there 'used' to be an Install folder, 'a long time ago' so this information may still be useful if someone is doing an upgrade from an older site, to make sure they do actually remove it, but in the context of installing a fresh Umbraco, eg this page of documentation, I agree it should be removed, as it will only confuse... @dampee do you agree?

dampee:
@marcemarc as it is mentioned in [the upgrade procedure](https://our.umbraco.org/Documentation/Getting-Started/Setup/Upgrading/general#post-installation) we can indeed remove this.

dampee:
@hartviglarsen  Thanks a lot for the good find.  It definitely makes documentation less confusing.  Please keep submitting improvments!

PR #752, opened 2018-04-27
BackOfficeUserManager.Create overload obsolete

drpeck:
I've added specified all the dependencies as requested by the obsolete overload. Additionally I've removed the difference in how the method is written in the two locations is appears.


dampee:
Hi @drpeck 
You are completely right.  We should not have any examples to obsolete code in the documentation.  For our own reference i've looked up [the obsolete method](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Core/Security/BackOfficeUserManager.cs#L57).
Thanks a lot for your pull request!

#H5YR

Kind regards
Damiaan

PR #759, opened 2018-05-07
Updated to use the "Current"-property, more clear example of usage.

enkelmedia:


dampee:
Hi Markus (@enkelmedia )
Thanks a lot for your suggestion!  We really appreciate you took time to do a pull request!
I discussed this with my fellow curators and I am afraid we not accept your suggestion.  
Please let me explain why...

First.
Adding the .Current would not work if you are inside e.g. an UmbracoApiController.  At that moment you can you the ApplicationContext property which IS the current. And that `ApplicationContext` has no `Current` property. We know that if you are in a custom class, your example WILL work.  But...

Second
There comes our second argument why we should not suggest to people to use the Current property.  And I am refering here to the docs: [Common Pitfalls § Singletons and statics](https://our.umbraco.org/documentation/Reference/Common-Pitfalls/#usage-of-singletons-and-statics).  
I'll quote this:
> In most cases you shouldn't be using these Singleton accessors, it makes your code very difficult to test but more importantly using Singletons and Statics in your code make it very hard to manage, APIs become leaky and ultimately you'll end up with more problems than when you started.
> 
> In all Umbraco base classes that you'll normally use, these object are already exposed as properties, so please use these instead! 

What do you think?  Can we close the pull request?

Eagerly awaiting your next suggestion for the docs!

Kind regards
Damiaan

enkelmedia:
Thanks for a great answer!

I agree! =D I've used the cache outside the "context" of a regular Umbraco controller a couple of times using the .Current-property (knowing the stuff from the "Common pitfalls"-article), that's why I figured that this should be changed. 

Let's close it =D

Issue #366, opened 2016-08-02
Add info on what to include in source control


Myster:
What should always be in source control
What should never be in source control
What may or may not be in source control depending on your work-flow.

A discussion I started about this is here: https://our.umbraco.org/forum/using-umbraco-and-getting-started/78957-what-should-i-commit-in-the-umbraco-folder 

I nearly just submitted a PR, but I wasn't sure where to put it, perhaps if a stub were created I could put in the info from the forum, then someone else could check it :-)

marcemarc:
We now have a page that we can update with Source Control info!

tlaukkanen:
Adding a direct link here to that page so that next person saves time from searching it 😅 
[What folders and files should I exclude from my source control repository?](https://github.com/umbraco/UmbracoDocs/blob/main/Getting-Started/Code/Source-Control/index.md#what-folders-and-files-should-i-exclude-from-my-source-control-repository)

PR #761, opened 2018-05-07
Hmm Source Control

marcemarc:
making a start on trying to document which files should and shouldn't be source controlled on a typical Umbraco install, as per Issue #366


marcemarc:
@dampee , @Jeavon hmm, ran out of steam on this a bit, could do with seeking your views on whether this is going a bit off track, and also whether it's completely correct in places :-)

I might also be able to review it myself, in a week or so, once I've forgotten I wrote it.

Getting-Started/Code/Source-Control/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+The App_Plugins folder is the home for all third party packages installed on your site, if they are not installed via NuGet then they should be added to source control.
+
+#### Config
+
+Your site's /config folder contains the set of configuration files for your Umbraco site.
+
+#### DocumentType - Backoffice Structure Changes
+
+When you create and edit DocumentTypes, MediaTypes, DataTypes, Macros, DictionaryItems, Languages, MemberTypes & Templates('names' not the actual files) in the Umbraco Backoffice these values are stored in the Umbraco Database, making them difficult to source control. However there is a third party package called uSync - https://our.umbraco.org/projects/developer-tools/usync/ which when installed will serialize these settings to disk, in a folder called /uSync - enabling you to source control these changes too.
```
dampee: We do also take this approach.  But I do have 2 remarks 

1) If we are mentioning a package like usync, shouldn't we say. a package like usync as there are other things out there? (which don't work/not work well/).  Courrier was serializing too right?  And you could say that modelsbuilder is also serializing (into c#).
2) is the default value "import on startup" still active?  Because that can cause serious side effects if you are deploying to another environment?

marcemarc:
closing this in favour of PR #978

PR #782, opened 2018-05-19
Lots of small changes to the docs

readingdancer:
I decided to do a few more tweaks while flying from the USA to the UK on my way to CG18


dampee:
Hi Chris
Thanks for your time to update all these things.
I love to see how the value goes up from the docs every time you submit something.

Hope to see more of you!

PR #783, opened 2018-05-22
Update index.md

tienlx91:
- Add $timeout to make sure the function will run after the DOM is updated


dampee:
Thanks a lot for the pull request.  We'll look into this as soon as possible.

dampee:
@madsrasmussen or @hemraker could you review this?

PR #795, opened 2018-05-26
Cloud computing edits to the docs

readingdancer:
Various changes to some more doc files post codegarden flight #1


greystate:
Some nice catches in there @readingdancer - great stuff 👍 

I'd like to ask about the reasoning behind changing *jQuery* to *JQuery* in most places – the jquery.com site uses *jQuery* everywhere and the object itself is called `jQuery`?

I don't really use jQuery myself, and I'm not english, so I can't tell if this is a "language" thing. Just curious to know...

readingdancer:
Hi both, this was a mistake on my part, I noticed various different casing and I was just trying to make them all the same. I have not changed them to all jQuery.

I have also now fixed a few nodejs -> Node.js

PR #800, opened 2018-05-27
Adding using statement for startup.cs for .Net core static site.

crgrieve:


marcemarc:
@crgrieve thanks! and if there is more you find as you begin to play with headless that would help people get started, that would be ace!

PR #801, opened 2018-05-27
The name is Builder, Models Builder...

abjerner:
The correct name for Models Builder (when not referring to the package alias) appears to be **Models Builder**, although the documentation primarily uses either **Modelsbuilder** (eg as seen here for the entire section about Models Builder) or **ModelsBuilder**:

https://our.umbraco.org/documentation/Reference/Templating/Modelsbuilder/

With this pull request, I've updated the Markdown files so that **ModelsBuilder** is now **Models Builder** (with a space). A lot of places where also referring to *the* Models Builder, so I've removed the `the` where it makes sense as I don't think it's necessary to refer to Models Builder explicitly (as we all know which Models Builder we're talking about).

My pull request does however not fix the the folder name, which should also be **Models Builder** instead of **Modelsbuilder**. Changing the folder name is a little more tricky because that would cause any existing links to the Models Builder pages to break. So what do we do then? Ideally we could change the folder name and create the necessary IIS rewrite rules in the Our project, and then have @nul800sebastiaan deploy them at the same time.


dampee:
Thanks  @abjerner for your hard work. 
I'll take this up with my fellow Curators in the near future.  
Concerning the changes when changing the filename.  We could create an OUR pull request and reference this on a separate pull request on the documentation. So that @nul800sebastiaan knows that both should be accepted at the same time.
But I would not do that for now as it will create merge conflicts anyway.

What do you think?

abjerner:
@dampee If you and the other curators agree to the current changes, we can merge this PR first.

I can then create another PR for the documentation in which I rename the folder, and a PR for Our with the necessary redirects. Then I'll also make sure to includes notes for Sebastiaan so both PRs are merged and deployed at the same time.

PR #803, opened 2018-05-27
Fixed links to NodeJS and Grunt

abjerner:
Figured I'd write a link scanner for the Our documentation. While still in development, this is the first page that it discovered with broken links ;)

### DIFF
--- a/Development-Guidelines/building-angular-project.md
+++ b/Development-Guidelines/building-angular-project.md
@@ -14,10 +14,10 @@ The bottom line is the UI project has zero dependencies on asp.net or Windows. H
 Umbraco 7 needs a couple of things to run:
 
 ### Node.js 
-To compile and run the UI project you need node.js installed, you can get that at [http://nodejs.org](nodejs.org) for both Windows and OSX.
+To compile and run the UI project you need node.js installed, you can get that at [nodejs.org](http://nodejs.org) for both Windows and OSX.
 
 ### Grunt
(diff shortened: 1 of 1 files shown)


Development-Guidelines/building-angular-project.md
```diff
@@ -14,10 +14,10 @@ The bottom line is the UI project has zero dependencies on asp.net or Windows. H
 Umbraco 7 needs a couple of things to run:
 
 ### Node.js 
-To compile and run the UI project you need node.js installed, you can get that at [http://nodejs.org](nodejs.org) for both Windows and OSX.
+To compile and run the UI project you need node.js installed, you can get that at [nodejs.org](http://nodejs.org) for both Windows and OSX.
 
 ### Grunt
-When you have node.js installed, you need to install Grunt. Grunt is a simple JavaScript task runner, basically like Nant, Msbuild or any traditional build system [http://gruntjs.com](more about grunt here).
+When you have node.js installed, you need to install Grunt. Grunt is a simple JavaScript task runner, basically like Nant, Msbuild or any traditional build system. You can [read more about grunt here](http://gruntjs.com).
```
dampee: Can we find something else that "here" as a link text?  Eg same as nodeJs fix above?

abjerner: Could be something like shown below?

image

dampee: Would be better.  Mentioning that you are going to another website in this case is much better!

abjerner: Should be fixed now ;)

dampee:
Thanks a @abjerner 
Very good catch!

PR #804, opened 2018-05-27
Fixed link

abjerner:
Technically the link still works, but the old URL will redirect the user, so we might as well just use the new URL.


dampee:
Thanks a lot Anders!

PR #805, opened 2018-05-27
Headless

dampee:
We should merge the headless branch so the documentation becomes visible in the documentation


dampee:
Thanks @hemraker

PR #806, opened 2018-05-27
Fixed a broken link

abjerner:
+ improved the text a bit as well


dampee:
What do you think of "This setting can be configured in the [Web.config]"

Would that be fine for you?

abjerner:
Probably minus `the` before `Web.config`. But otherwise yes ;)

PR #807, opened 2018-05-27
Not sure what happened here

abjerner:
It seems like this should look like the other links on the page - but it doesn't. Also, the referenced page doesn't appear to exist anymore, so I've removed it.


Jeavon:
Bigger issue is that we have duplicated "Add-Ons" and "Products", I'm working on solving this.

But that doc doesn't exist in products either.... 

https://github.com/umbraco/UmbracoDocs/blob/3264f63133701e706e24e811fc2d36a5c72c5b5b/Products/UmbracoForms/Editor/index.md

sofietoft:
Hi guys,

I've had a look at this, and can't seem to find any article related to this.

Let's remove it for now.

-Sofie

PR #794, opened 2018-05-26
- fixed lots of broken links

prjseal:
- fixed lots of broken links

PR #808, opened 2018-05-27
Another 404

abjerner:
Seems like the page has moved, so fixed the link ;)


abjerner:
@dampee Sure. But I'm not totally sure where I can see the other pull request. Any pointers?

dampee:
It was a massive pull request from @prjseal : #794  :-)

abjerner:
Ahh, I see. I've figured out how to resolve it now, so I merged my change back to master as mine uses `Routing` instead of `routing` (the casing matters for the breadcrumb).

Although compared to the rest of Our, lowercase URLs would be more correct. Does the curators team have any preference for this? Ideally the breadcrumb could look up the names of parent pages - also because name (H1) doesn't always match the URL.

dampee:
There is no written preference, but I tend to go with matching the filename.  That would probably mean go for Routing instead of routing.

dampee:
As the link is working now, I would like to suggest to close the PR. 

Are you OK with that?

abjerner:
Yup ;)

PR #814, opened 2018-05-27
Another broken link

abjerner:
Don't think this link ever worked :o


dampee:
And it probably won't with the index.md on OUR...  So i've removed it.

Thanks for your contribution!

abjerner:
@dampee If I'm not mistaken, Our removes the filename automatically. Either way, it should work on Our once deployed ;)

PR #816, opened 2018-05-27
Another broken link

abjerner:


dampee:
As it is Poining to an external website, I would suggest to use the full URL.

Issue #763, opened 2018-05-09
There are two File and Folder Permissions pages


marcemarc:
https://our.umbraco.org/documentation/Getting-Started/Setup/Server-Setup/permissions
https://our.umbraco.org/Documentation/Getting-Started/Setup/Install/permissions

probably only need the 'Server Setup' one and then link through to the page from the install folder

FransdeJong:
Maybe also emphasize that the the old folders (usercontrols, xslt, masterpages, macroscripts) can be missing because it's not used by umbraco anymore?

AndyFelton:
Just completed a pull request to resolve the original issue. @FransdeJong can I suggest you add your comment as a new issue.

AndyFelton:
#822

PR #822, opened 2018-05-28
Sort permission pages (#763)

AndyFelton:
Deleted permissions page from the Install directory and changed links to point at the permissions page in the Server Setup directory


AndyFelton:
#763

Jeavon:
I have migrated the intro from the to be deleted file into the one being kept as it has been updated more recently, so this can now be merged!

PR #828, opened 2018-05-28
Fixed broken link

abjerner:
There is no class called `ApplicationTreeService`, as the `ApplicationTreeService` returns an instance of the `IApplicationTreeService` interface. I've updated the link to reflect that.

I'm not totally sure whether the rest of the text might need some adjustment as well. But now at least the link works ;)


dampee:
Great catch!

PR #831, opened 2018-05-28
Change in step order

BatJan:
Switched the order of steps so checking .net version becomes before choosing MVC since that is the order in, which things are presented to you when you create a new project in Visual Studio.


BatJan:
Yay! I must do the gif meme now! ;-)

I hope so!...Wondering if I should make another clarifying that one should use VS2017 since Harriet experienced some errors using VS2015 in https://our.umbraco.org/forum/using-umbraco-and-getting-started/92209-i-am-getting-an-error-trying-to-install-the-client-side-part-of-umbraco-headless

PR #832, opened 2018-05-28
fixing a lot of it's and its errors

prjseal:


prjseal:
Hi Damiaan
Do you prefer individual PRs with single file commits, or ones with lots of files like this?

Paul

PR #839, opened 2018-05-29
Mention that installation of NodeJs is assumed

BatJan:
Some people who might not be used to playing around with NodeJs might forget or be unaware that installing NodeJs first is a requirement for successfully installing the umbraco-headless package even though it has "NodeJs" written all over it :-)

### DIFF
--- a/Umbraco-Cloud/Headless/Headless-Node-Client/index.md
+++ b/Umbraco-Cloud/Headless/Headless-Node-Client/index.md
@@ -1,6 +1,8 @@
 # Umbraco-Headless-Client-NodeJs
 
-Javascript Client Library for working with the Umbraco Headless API
+Javascript Client Library for working with the Umbraco Headless API.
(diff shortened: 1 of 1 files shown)


dampee:
Hi Jan
Good point!  Do you know if there is a minimum version?  It should be good to have this in there too.
Maybe we should ask @madsrasmussen ?

Thanks for your submission to OUR docs :-)  #H5YR

BatJan:
Hi Damiaan

That's indeed a good point too - Let's see what @madsrasmussen says.

#H5YR right back :-)

madsrasmussen:
Hm, that is actually a good question. I am currently on v 8.11.2 which is the LTS version and that works. So I think we will be fine just mentioning that one.

BatJan:
@dampee @madsrasmussen - Updated PR with mention of recomended version

Issue #845, opened 2018-05-31
Make the Umbraco UI documentation easier to find by adding a link in the docs navigation.


readingdancer:
Currently it doesn't seem like the Umbraco UI API docs are linked to from the main documentation section, so unless you have Googled it and come across the link, it's almost impossible to find.

I think there should be a clear link in the main documentation navigation tree on the left across to the API docs:

https://our.umbraco.org/apidocs/ui/#/api

marcemarc:
Would you have it on the top level, or in the reference section? think we just need @nul800sebastiaan to setup a redirect

readingdancer:
My gut feeling is it should be within one of the sections so that we can describe what it is used for. If people see it on the top level they might just click in and be confused as to its usage?

As you said perhaps in the reference section or in a second that talks about back office development. ( I am not at my PC currently so not 100% sure )

marcemarc:
@readingdancer have added a couple of 'landing pages' in reference and extending, that might be a start, should at least make  them appear in the navigation... https://github.com/umbraco/UmbracoDocs/pull/850

what do you think?

dampee:
I think we should at least make them appear in the search  :-)

marcemarc:
@readingdancer ok that's merged in! Have a look see if that's a step in the right direction; thanks for pointing it out... Be interesting to see in the analytics if more people discover them in this way

marcemarc:
@readingdancer forgot to close the issue!

PR #850, opened 2018-06-05
Increase the visibility of the existence of the API documentation

marcemarc:
See issue: https://github.com/umbraco/UmbracoDocs/issues/845 #845 

Essentially if you are within the documentation pages, and using the navigation there isn't any scent that would lead you to the auto generated / API documentation for C# or Backoffice UI

So this PR adds a couple of 'Landing Pages' that sign post the existence of the documentation to the reference and to the extending sections...

### DIFF
--- a/Extending/Backoffice-UI-API-Documentation/index.md
+++ b/Extending/Backoffice-UI-API-Documentation/index.md
@@ -1,0 +1,14 @@
+# Backoffice UI API Documentation
+
+A library of API Reference documentation is auto-generated from the comments within the Umbraco Source Code.
+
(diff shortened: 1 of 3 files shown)


readingdancer:
Hi @marcemarc that looks good to me, I think this will be really useful to people. I am sure a lot of developers don't have a clue that these auto generated documentation sections even exist :)

hemraker:
@marcemarc not empty! It was cholk-full of good intentions 😉

dampee:
I love good intentions.  ;-)

marcemarc:
@hemraker yes looking closer, not empty!, at the bottom you can see good intentions, aspirations and I can make out some hope, which we will hang onto.

PR #843, opened 2018-05-30
Fixed lots more typos.

readingdancer:
I think the title says it all :)

PR #852, opened 2018-06-06
Chris making documentation great again v2

readingdancer:
Hi @dampee this branch should have all my previous changes and exclude the files you mentioned, it's also rebased on top of the current master.


dampee:
So this replaces #843 :-)
Thanks a lot for the update @readingdancer !

PR #856, opened 2018-06-07
Add example of @Umbraco.Field() method in partials

BatJan:
I just discovered that it's not documented anywhere that when you're trying to use @Umbraco.Field() from within a partial you need to pass the context to it otherwise you will not get any value. Therefore I added this information and a basic example.

See my discovery if you have the time to read this reply - At first I actually thought it was not possible so had to correct myself - It's been a while...:-) 

https://our.umbraco.org/forum/using-umbraco-and-getting-started/92301-change-url-on-image#comment-291932

### DIFF
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -21,7 +21,7 @@ This is probably the most used method which simply renders the contents of a fie
 	
 If you're using the Field method from within a partial view then be aware that you will need to pass the context so the Field method knows where to get the desired value from. For instance you can pass "CurrentPage" like this:
 
-	@Umbraco.Field(CurrentPage, "bodyContent")
+	@Umbraco.Field(Model.Content, "eventLink")
 
 You will also need to pass the "Context" to the @Umbraco.Field() method if you're looping over a selection like this where we pass the "item" variable.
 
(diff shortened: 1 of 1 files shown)


dampee:
That is right indeed.  You can do that. :-)  Didn't knew it was not documented. Great catch!
We are trying to get rid of all dynamics in the examples. Could you update the example?

PS I've already fixed a typo

Reference/Templating/Mvc/views.md
```diff
@@ -18,6 +18,21 @@ All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoTemplatePage` which expos
 This is probably the most used method which simply renders the contents of a field for the current content item.
 
 	@Umbraco.Field("bodyContent")
+	
+If you're using the Field method from within a partial view then be aware that you will need to pass the context so the Field method knows where to get the desired value from. For instance you can pass "CurrentPage" like this:
+
+	@Umbraco.Field(CurrentPage, "bodyContent")
```
dampee: CurrentPage is a dynamic in this case.  Could we change that?

BatJan:
Hi Damiaan

This should be fixed now :-)

BatJan:
Indeed does - I've been privileged to only work with strongly typed models upon our own "umbraco baseline" so I'm still confused about the different options when it comes to "vanilla" Umbraco :) But slowly getting my memory back together :D

PR #860, opened 2018-06-10
Removing "CurrentPage" example of rendering

BatJan:
Damiaan Peeters told me in another PR I made that we're trying to get rid of Dynamic code examples so therefore I modified this file to only show the typed example.

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textbox.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textbox.md
@@ -21,3 +21,13 @@ Textbox is a simple HTML input control for text
 	       <p>@(Model.Content.GetPropertyValue<string>("pageTitle"))</p>
 	   } 
 	}
+
+
+### Dynamic (Obsolete)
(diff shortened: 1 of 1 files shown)


dampee:
Thanks.  Ik wilde mergen this immediately, bit i al going tot ask permission with my fellow Documentation curators.

marcemarc:
My vote would be for marking them as obsolete, for a period of time, then removing them, perhaps when we have next version released. It's good for people if they start working on an existing Umbraco site, and it has this CurrentPage dotted everywhere to be able to discover what this is/was and that actually it's not the  recommended approach, removing all reference to it, means people will potentially still think it is ok to use, it's just missing from the docs, if that makes sense...

I updated the rendering a grid in the grid template in a similar way:

https://our.umbraco.org/Documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Grid-Layout/Render-Grid-In-Template

dampee:
That makes sense @marcemarc 

@BatJan can you find some time to update the PR?

Jeavon:
I agree, adding "Obsolete" is the way to go

BatJan:
Thank you for the input folks - I agree that leaving it in there and marking it as obsolete is the better option for the same reasons as Mark mentioned.

I'll update it later this evening.

BatJan:
Updated version has been comitted @dampee :-)

dampee:
Thanks a lot @BatJan !
 We can't merge right now because it will mess up OUR.  But it will be merged in the upcoming days!

PR #861, opened 2018-06-11
mentioned the namespace for the extension methods

dampee:


dampee:
It's a patch from 4 years ago on a file which does not exists anymore

PR #863, opened 2018-06-11
Marking dynamic approach as obsolete

BatJan:
The dynamic method has been marked as being obsolete and I have added a text explaining the approach is not recommended


dampee:
Hi @BatJan 
This was the last merge from your "obsolete dynamics PR series".  
Thanks a lot for your effort and to think about a good way to communicate to our readers!

H5YR 🥇 

Kind regards
Damiaan

BatJan:
Thank you @dampee - And thank you for the small discussion we've had. I'll keep'em coming. And #h5yr for being so quick at reviewing :)

PR #864, opened 2018-06-11
Mark Dynamic as obsolete

BatJan:
Marked the dynamic example as being obsolete and added a text explaining that the dynamic approach is not the recommended one.

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Date.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Date.md
@@ -22,7 +22,7 @@ The only setting that is available for manipulating the Date property is to set 
 
 ### Dynamic (Obsolete):
 
-The below example is using Dynamic access content access, which is considered obsolete and is not recommended to use. However the example is included for historical reasons if for instance a developer has overtaken a project where this approach is being used. This approach will be obsolete when Umbraco 8 is released and therefore is best to use the strongly typed example listed above.
+See [Common pitfalls](https://our.umbraco.org/documentation/reference/Common-Pitfalls/#dynamics) for more information.
 
 	@{
 		@CurrentPage.datePicker.ToString("dd-MM-yyyy")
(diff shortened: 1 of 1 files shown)


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Date.md
```diff
@@ -20,7 +20,9 @@ The only setting that is available for manipulating the Date property is to set
 
 	@(Model.Content.GetPropertyValue<DateTime>("datePicker").ToString("dd MM yyyy"))
 
-### Dynamic: 
+### Dynamic (Obsolete):
+
+The below example is using Dynamic access content access, which is considered obsolete and is not recommended to use. However the example is included for historical reasons if for instance a developer has overtaken a project where this approach is being used. This approach will be obsolete when Umbraco 8 is released and therefore is best to use the strongly typed example listed above.
```
Jeavon: I would prefer to link to the common pitfalls doc rather than repeating this, I would suggest this:

Please see [Common Pitfalls](../../../../Reference/Common-Pitfalls/index.md#dynamics) for information about why the Dynamic approach is obsolete

BatJan:
@Jeavon Aaah! I did not realize that there was such a section. I'll update all my PR's to point to this document instead.

BatJan:
@Jeavon @dampee - Is it better now? If so I'll update the other PR's too. And sorry for the spam I'll clone the repo for the next PR's I'm going to make so stuff like this will be done in 1 single PR.

dampee:
Wouldn't it be better if we mention "dynamics" or "obsolete" somewhere in the sentence below? Certainly because the link has a dynamics tag.

See [Common pitfalls](https://our.umbraco.org/documentation/reference/Common-Pitfalls/#dynamics) for more information.

BatJan:
Hmm...How about "See Common pitfalls for more information about why the dynamic approach is obsolete" ? Will that work? :) @dampee

Jeavon:
Yes @BatJan 

`See [Common Pitfalls](../../../../Reference/Common-Pitfalls/index.md#dynamics) for information about why the Dynamic approach is obsolete`

BatJan:
Cool @Jeavon - Will update all open PR's I have regarding this with this text after some dinner :)

The next time I do a batch like this I make sure to clone the docs instead of using the "Edit this" button - Don't want to overwork you guys :)

BatJan:
Text is updated - Thanks for the discussion mates :)

PR #869, opened 2018-06-11
Mark Dynamic as obsolete

BatJan:
Marked the dynamic example as being obsolete and added a text explaining that the dynamic approach is not the recommended one.


dampee:
@BatJan any time left to look into the conflict?

dampee:
Hi@BatJan  
resolved the conflict myself.  The file was already updated by another PR of the great @BatJan  :-D

PR #880, opened 2018-06-14
Updated the documentation for the localization service

abjerner:
Per a chat I had with Shannon for the documentation of another service, the documentation really shouldn't list all the methods of a given service, as that is just extra stuff to keep updated, and we can refer to the auto-generated API docs instead. So with this PR, the method list for the localization service is now gone.

I also added some samples for referencing languages by using the localization service. The samples show to to get languages - either individually or as a collection. So far I've called the new page *Referencing languages*, but I'm not sure whether *referencing* is the best term here. Perhaps *Getting languages*? Or something else?


dampee:
Hi  @abjerner 

Thanks a lot for your PR!  Really apprecieated!

Moving the file makes sense.  Because the foldername is the same, we won't need to ask a redirect to Sebastiaan.

Concerning the naming, I would like to ask @marcemarc or @Jeavon  to give their thoughts.

marcemarc:
@abjerner great contribution! I'm thinking maybe 'Working with languages' as you go on to explain 'how to work with them' using the service, whereas I tend to agree referencing, sort of infers not working with them!

marcemarc:
@abjerner general curatorial opinion says "Working with languages and the localizationservice" would be a good title for the sample page.

abjerner:
@marcemarc Isn't the part with the localization service implied if we have a breadcrumb and URL saying LocalizationService?

Also, if we look a bit more general, my thought was to have one page showing how to *get* instances of a certain type, and another page for *creating* new instances etc.

Languages are a bit special compared to other objects in Umbraco, as there doesn't appear to be methods in the localization service for creating/adding a new language, but otherwise I would have created a specific page for *creating*

In my pull request for the user service (which I still need to wrap up), I tried something like this:

- [Create-a-new-user.md](https://github.com/abjerner/UmbracoDocs/blob/patch-5/Reference/Management/Services/UserService/Create-a-new-user.md)
- [Get-a-reference-to-a-user.md](https://github.com/abjerner/UmbracoDocs/blob/patch-5/Reference/Management/Services/UserService/Get-a-reference-to-a-user.md)
- [Update-user-information.md](https://github.com/abjerner/UmbracoDocs/blob/patch-5/Reference/Management/Services/UserService/Update-user-information.md)

Generally my thought was to have a page for each CRUD operation - or at least when it makes sense. I'm think *reading* isn't a good name either, but otherwise it could be *creating {type}*, *updating {type}* and *deleting {type}*. There are probably also some cases where more pages would make sense.

[The Wikipedia page for CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) mentions *retrieving* as an alternative word for *reading*, which might be more ideal. So it could be (again generally speaking):

- creating {type}
- retrieving {type}
- updating {type}
- deleting {type}

marcemarc:
@abjerner yes, I think the retrieving is preferable to reading, but also think we should get your update into the docs - would it take you long to rebase? and then we could merge in and reflect?

abjerner:
@marcemarc I think it would be easiest to just create a new PR, so I'll look into that ;)

PR #882, opened 2018-06-17
Update egghead link

BatJan:
I propose updating the Egghead link to point explicitly at the AngularJS fundamentals course since it's currently only pointing to the egghead homepage. Can be confusing for people to find the AngularJS course instead of the Angular course so it's better to be explicit about it.


dampee:
I agree!

PR #883, opened 2018-06-17
Update egghead link

BatJan:
I propose updating the Egghead link to point explicitly at the AngularJS fundamentals course since it's currently only pointing to the egghead homepage. Can be confusing for people to find the AngularJS course instead of the Angular course so it's better to be explicit about it.

Don't know if perhaps we should consider creating a page with angularJs information in the documentation, which can be referenced instead? Like the "common pitfalls" page? Does it make sense what I'm trying to say? :) - Don't know if there are more places where these 3 links are mentioned though...Just spotted these today browsing through the documentation randomly.


dampee:
I checked out the repo and found only two files with 2 links each to egghead.

Making a specialized file about angularjs makes sense to me.  Could you create an issue about this question?

dampee:
Oh, and thanks a lot for updating the link!  H5YR!

PR #884, opened 2018-06-17
Add example of localize attribute

BatJan:
I added an example showing how attributes can be localized, which is also described in the documentation of the <localize> directive here https://our.umbraco.org/apidocs/ui/#/api/umbraco.directives.directive:localize


dampee:
Nice addition.  Not everyone knows where to look in the autogenerated docs.  So well worth merging in!

Thanks a lot Jan!

PR #887, opened 2018-06-19
Reuse a bit of the repo instruction from core

BatJan:
Untill now I have not needed to clone the repository in order to create PR's for the documentation. But now that I need to do a little bigger contribution (adding a new page etc.) I thought it would be better to clone the repo. Then I realized nothing is currrently mentioned about how to keep the fork in sync with the main repository like it is here on the Core repo: https://github.com/umbraco/Umbraco-CMS/edit/dev-v7/docs/CONTRIBUTING_DETAILED.md - So thought that I would add some details about it here too heavily inspired by the core text.


marcemarc:
Thanks @BatJan!

PR #888, opened 2018-06-19
How to query headless using the .NET Client

hartvig:


dampee:
Hi Thanks Niels,

I've fixed a small code issue.  For the rest it seems fine.  

As this is a new file: I wanted to be sure that it's linked from somewhere so it is findable.  Can you make sure you link from related content to this page? 
Or was this supposed to be only available from the menu?

Thanks a lot for your suggestion!  We really appreciate it!  ;-)

PR #891, opened 2018-06-21
Fixing example in Examine Fluent API

steroberts89:
The recently added example for the fluent API had an error with the ordering. This should be done on the search query, not the results - which doesn't have an OrderBy method or extension.

This was introduced in https://github.com/umbraco/UmbracoDocs/commit/13212de060d35ebbbe98edaa417338083436b636#diff-d142ee10fd27747e8fd653df9b3b0ef1

Reference material: https://github.com/Shazwazza/Examine/wiki/Sorting-results


dampee:
So happy to see your first contribution to the Documentation!  #H5YR

steroberts89:
Thanks!

PR #893, opened 2018-06-22
typo correction

liamlaverty:
just corrects a typo


dampee:
Hehe :-) Funny  it slipped through the typo-fix rush after codegarden.

Thanks a lot for fixing!

PR #896, opened 2018-06-23
Fix a wee typo

greystate:
*I know!* They said there were no more typos :-)


marcemarc:
Thanks @greystate ... typos now score double points

PR #897, opened 2018-06-24
Added a page on Security Hardening

JeffreyPerplex:
First write-up of security hardening in Umbraco based on the presentation I gave during CG18


Reference/Security/Security-hardening/index.md
```diff
@@ -0,0 +1,59 @@
...
+By default there are some folders in your Umbraco installation that should be only used by authenticated users. It’s considered a good practice to lock down these folders to specific IP-addresses and/or IP-ranges to make sure not everyone can access these.
+The folders we want to lock down are App_Plugins, Config, Umbraco_Client and Umbraco.
+The prerequisite of this to work is that you’re using [IISRewrite](../../routing/iisrewriterules/index.md) 
+
+If you’ve made sure that you’ve installed this on your server we can start locking down our Umbraco-folders. This can be down by following these three steps.
+
+1. We are going to lock down /Umbraco/, but because API-controllers and Surface-controller will use the path /umbraco/api/ and /umbraco/surface/ these will also be locked down. Our first rule in the IISRewrite.config will be used to make sure that these are not locked by IP-address.
+
+	<rule name="Ignore" stopProcessing="true">
+		<match url="^(?:umbraco/api|umbraco/surface)/" />
```
marcemarc: in the olden days of traditional load balancing, you'd need to also permit umbraco/webservices

Reference/Security/Security-hardening/index.md
```diff
@@ -0,0 +1,59 @@
...
+	</rule>
+
+2. Get the IP-addresses of your client and write these down like a regular expression. If the IP-addresses are for example 213.3.10.8 and 88.4.43.108 the regular expression would be ‘213.3.10.8|88.4.43.108”.
+
+3. Lock down the folders App_Plugins, Config, Umbraco_Client and Umbraco (or the renamed version of this folder) by putting this rule into your IISRewrite-rules
+
+	<rule name="Allowed IPs" stopProcessing="true">
+		<match url="^(?:app_plugins|config|umbraco|umbraco_client)(?:/|$)" />
+		<conditions>
+			<add input="{REMOTE_ADDR}" negate="false" pattern="213.3.10.8|88.4.43.108”>
```
marcemarc: If an organisation can't identify it's editors by IP range - would it be a reasonable alternative to have an internal dns entry, only on the company's network for the backoffice access?, so have a condition like:

 
            <add input="{HTTP_HOST}" pattern="backoffice.internaldnsentry.*" negate="false" />

or does that have other side affects or is it more obscurity than security?

Reference/Security/Security-hardening/index.md
```diff
@@ -0,0 +1,59 @@
+# Umbraco Security Hardening
+
+Here you find some tips and trick for hardening the security of your Umbraco installation.
+
+## Rename your Umbraco-folder
+
+By default the login page of Umbraco is available at the path /umbraco/. This page is the entrance to your installation and it’s considered a good practice to rename your path to a more secure path.
```
marcemarc: Are there any packages that rely on the umbraco folder being called 'umbraco' - I know in the past I've accidentally created some.

Reference/Security/Security-hardening/index.md
```diff
@@ -0,0 +1,59 @@
...
+3. Lock down the folders App_Plugins, Config, Umbraco_Client and Umbraco (or the renamed version of this folder) by putting this rule into your IISRewrite-rules
+
+	<rule name="Allowed IPs" stopProcessing="true">
+		<match url="^(?:app_plugins|config|umbraco|umbraco_client)(?:/|$)" />
+		<conditions>
+			<add input="{REMOTE_ADDR}" negate="false" pattern="213.3.10.8|88.4.43.108”>
+		</conditions>
+		<action type=”AbortRequest” />
+	</rule>
+
```
marcemarc: Is it worth suggesting people run their site through https://asafaweb.com/ ?

Reference/Security/Security-hardening/index.md
JeffreyPerplex: Hi Marc, thanks for the quick replies!

We could of course add /umbraco/webservices/ if we want too, but I never came across this issue. Maybe as a remark under this that for older systems you maybe have to include this one as well?

Reference/Security/Security-hardening/index.md
JeffreyPerplex: Never used this, but if you can confirm you've used this before I will add this

Reference/Security/Security-hardening/index.md
JeffreyPerplex: Not that I'm aware of, but we will found it out pretty quickly indeed. Do you know what the correct way is in Umbraco to determine the 'Umbraco-path'. And maybe we can give that as hint to package developers to use.

Reference/Security/Security-hardening/index.md
JeffreyPerplex: Depends on your vision on the documentation I guess... Do you want to refer to other sources in your documentation, but also for example to specific blogposts of the community, or videos of CodeGarden-sessions?

We also test our websites with:

- https://www.ssllabs.com/ssltest/ (for correct SSL configuration)
- https://www.htbridge.com/ssl/ (also for SSL)
- https://www.htbridge.com/websec (bit simalar to asafaweb)
- https://securityheaders.io/ (for the correct headers)

dampee: If we add these extnernal links, we need to make sure that the docs have "rel=nofollow" attributes added on the A tags.  Otherwise we will receive a bunch of SEO link builders submitting links...  Maybe @Jeavon knows by heart, if this can be done by default with the Markdown lib we are using.

hemraker:
@JeffreyPerplex don't think anyone answered your question about how best to preview markdown.
Two things: 

1. I like to use VS Code. It already has built in markdown preview (ctrl+shift+v) and you can do side by side writing and preview (even auto-scrolls 😃 ) 
2. Clone Our and drop your docs in the documentation folder (think it's in App_Data). This let's you preview your changes as they will actually look on Our.

It's been a while since I did 2, so not sure if all the recent great work has changed this. But I'm sure the curators knows 😃)

Reference/Security/Security-hardening/index.md
Jeavon: Yes, we can add .UseNoFollowLinks() to the pipeline, it will add rel=nofollow to all links, I think this will be ok....?

dampee:
@hemraker I think it is the /documentation folder (not app_data)

JeffreyPerplex:
@hemraker Thanks for the reply. I installed a markdown editor (https://marketplace.visualstudio.com/items?itemName=MadsKristensen.MarkdownEditor) in Visual Studio and that works way easier indeed. It's probably a bit the same as VS Code.

1) Should we update the general docs a bit (https://github.com/umbraco/UmbracoDocs) so it's easier for new contributors to help documenting by pointing them to right tools. Because I never used markdown before contributing to these docs. Good idea?

2) Small sidestep: I still think it's really hard for some persons (webdesigners or projectmanagers that are using Umbraco, but not using these fancy developer-tools) to contribute. I know we cannot fix these easily and soon, but I can't ask my projectmanagers to help documenting some stuff because they get really scary only thinking about GitHub or MarkDown... If I would have to redesign this part of Our I would definitely put it in Umbraco so everyone could edit

JeffreyPerplex:
Hey documentation team, thanks for all the replies! What is the expected next step for me to get this on Our? Do I have to update the docs with your recommendations? Or is it okay for now, as it is?

I have 10 more tips, but haven't got time to document them right now, but I would say to improve the docs step-by-step in the coming period. 

Let me know,

Jeffrey

marcemarc:
@JeffreyPerplex it is certainly wonderful what you have right now, my comments were just my thoughts from reading through, I think it is close to pulling 'as is', and then it can be improved over time, etc but one thing though is we had a wee chat amongst the documentation curators, and between us we haven't got a site where we've successfully renamed the Umbraco folder! (I had an issue on an Umbraco 4 site once :-P)- so in these circumstances, we're duty bound to just run the advice via UmbracoHQ, more of a procedural thing that a downer on the tactic, so I've pinged @nul800sebastiaan  and if he waves the blue flag of 'ok' then we'll merge it asap, as it's super valuable information!!

Looking forward to the next set of tips too, but the hardening is a real FAQ, so we're itching to get this in asap!

JeffreyPerplex:
Hi @marcemarc, thanks for the feedback and the update. I was a bit confused what the next steps were, and feedback of four persons was a bit overwhelming :)

The renaming of the Umbraco-folder works out of the box and tested it quite well, but I must admit we haven't applied it to any of our production websites. We have scheduled it to do this somewhere in August, but we haven't decided on the name of the folder yet :P

Thank you again, sir(s), for the help and the curating!

Jeffrey

nul800sebastiaan:
I haven't read the whole thread in detail (sorry) but just to chime in: 

- It is sound advise - at least to stop evil people like Jeffrey (just kidding!) from easily scanning your site
- It doesn't sound like anyone in this thread has combined all of the advise here in practice (correct me if I'm wrong). I would opt for first finding someone who has experience with this kind of setup and ask for their advise
- Related: remember that we (HQ) will need to support the scenario if we document it - so we'll need a solid setup
- Unfortunately, we do currently rely on the `/umbraco` folder to be there on Umbraco Cloud, the automatic upgrade magic only works with the default setup
- Sidenote: I just realized, don't we have the same problem with the `/Umbraco_client` directory?

Other than that documenting this setup is great and I look forward to reading the results when you've nailed it. 👍

Reference/Security/Security-hardening/index.md
Shazwazza: There's lots of packages with hard coded "/umbraco" paths ;) they shouldn't be hard coded but they are.

JeffreyPerplex: Hi @Shazwazza, and what should we tell these package developers? Which code should they use to get the correct path? I don't think that we should read it from the web.config, or do we? And ../../ is a possibility neither I guess?

If you can tell me how to do it, I can document that as well. Thanks!

marcemarc:
Thanks @JeffreyPerplex for the updates, it's much clearer now

PR #899, opened 2018-06-26
Spelling fix in Load-Balancing/index.md

shahzad-0:


dampee:
Thanks a lot for your catch @shahzad-0 
Do you have a profile on http://our.umbraco.com?  If so, what is the link?

PR #904, opened 2018-06-29
Add a note about @CurrentPage being obsolete

BatJan:
Added a variation of the "standard" about the dynamic approach being obsolete. Have not added "Obsolete" to the headline since it will be awfully long.

Could perhaps just switch "(dynamically)" with "(obsolete)" instead to highlight it already in the headline and then skip the bold "Attention" text?


dampee:
H5YR 🚀

PR #905, opened 2018-06-29
Added a note about UDI not having an official definition

BatJan:
And by official definition I mean...I don't think anyone has ever agreed officially what it's short for, right? We just call it a UDI because it sounds like something that is unique for the way Umbraco identifies nodes...sort of, right? :)


dampee:
haha  :-) So true.  

It's itching to merge this in.  But I am going to ask @Jeavon and @marcemarc  if it ok.

marcemarc:
@BatJan @dampee ... I refer to this tweet: https://twitter.com/zpqrtbnk/status/827204958734974976 as official an announcement as it gets... my vote is it remains uppercase throughout UDI even though it's technically just a 'word' someone has made up and should be Udi... it's more recognisable as to what it is if it remains in uppercase

PR #900, opened 2018-06-26
#170 documentation for working with tags

BatJan:
As instructed by @dampee in this issue https://github.com/umbraco/UmbracoDocs/issues/895#issuecomment-400261030 I hereby open a PR for documenting the "Tags" property editor - This PR relates to issues #170 and #230 

This is the initial commit and it's work in progress and this should not currently be merged.

Issue #170, opened 2015-01-05
Documentation for working with Tags


Lechus:
1. There is:
   if (Model.HasValue("articleTags")){
   Gives me error:
   CS1928: 'Umbraco.Web.Models.RenderModel' does not contain a definition for 'HasValue' and the best extension method overload 'Umbraco.Web.PublishedContentExtensions.HasValue(Umbraco.Core.Models.IPublishedContent, string)' has some invalid arguments
2. @foreach (var tagItem in Model.articleTags.Split(','))
   gives me:
   Parser Error Message: Unexpected "foreach" keyword after "@" character.  Once inside code, you do not need to prefix constructs like "foreach" with "@".

Shazwazza:
The docs required need to deal with how we work with Tags with the UmbracoHelper

dampee:
@Jeavon what do you think of this commit: 24d49bbeee84b1932cc033d42404a80649b96dab

sofietoft:
Closing this issue, as it looks like Tags has been documented.

PR #910, opened 2018-07-02
Feature/#170 documentation for working with tags versioned

BatJan:
Adding a YAML "tagged" version of the current tags documentation + legacy documentation for versions before Umbraco 7.6 using the vpost file prefix.

This PR relates to #170 and #900


dampee:
Hi Jan
Because we merged the other pull request, this one is now having conflicts.
Can you rebase (or cherry pick the right commits)?

Really looking forward on merging this.  But it has to wait for umbraco/OurUmbraco#236 to be merged.  Right?

BatJan:
I'll have a look and make sure to sort out the troubles later today :)

BatJan:
Hi @dampee I think this is good for another review? :)

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags-vpost-7.0.md
```diff
@@ -0,0 +1,77 @@
+---
+keywords: tags property editors v7 version7
+versionFrom: 7.0.0
```
dampee: How sure are you about the versionFrom?

dampee:
Looks ace indeed!

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags-vpost-7.0.md
BatJan: Gut feeling to be honest - I can try and setup some 7 and 7.3 to see if there are any differences between the version though. I think that should be sufficient, no? Or we can wait and see if anyone reports issues in the forum if/when they come across it?

BatJan:
Wee, first mover! 💃

BatJan:
Hi @dampee - Just did some dogfooding and noticed that image refernces were broken and the @ sign was missing in the code examples when using the @foreach loop.

Also learned that the documentation I have done pre 7.6 only work from 7.4 where models builder was introduced so I have renamed the filename and updated the yaml tags.

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags-vpost-7.0.md
BatJan: Yup, agree. Just forgot how much has really happened since v7.0 was fresh and shiny! :) - It's been updated now.

PR #915, opened 2018-07-06
Mention that credentials usage has changed

BatJan:
Since the release of 7.7.0 the e-mail has been used for logging in instead of the username but it does not seem to be mentioned anywhere so thought I would add a section about it and mention how to keep using the username instead if the change is causing too much confusion when people do upgrades.


marcemarc:
Thanks @BatJan yes, and useful to document that setting!

PR #921, opened 2018-07-14
Add missing images + putting the note in italics

BatJan:


marcemarc:
@BatJan thanking you once more...

PR #923, opened 2018-07-14
Adding documentation for the textarea property editor

BatJan:


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textarea.md
```diff
@@ -0,0 +1,27 @@
...
+## Content Example:
+
+![Textarea Content Example](images/textarea/7/textarea-content.png)
+
+
+## MVC View Example:
+
+	@{
+	   if (Model.Content.HasValue("textarea")){
+	       <p>@(Model.Content.Textarea</p>
```
marcemarc: This bit is missing the closing round bracket...

... also got me thinking whether the example should use an alias that is a different name to the DataType, in case that led to some confusion?

Also in the example, if we're using HasValue, should the example use GetPropertyValue... or is there a Modelsbuilder equivalent to HasValue?

BatJan:
Argh, yes I'm afraid that all the PR's I submitted after adding the checkbox list images are the same since I apparently merged the branch into my local master before branching out the following PR branches.

But since the checkbox-list pr. is merged does it then matter or will I need to remove that part? :-)

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textarea.md
BatJan: Good point Mark - I'll change the alias and fix the missing bracket.

I'll also modify the example to using GetPropertyValue like it does in Textbox https://our.umbraco.com/documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textbox

BatJan: Just made a new commit with the suggested fixes

marcemarc:
Yes I think that's all good, as you say it's the same as the earlier PR, thanks for updating, apologies completely missed that you had!

BatJan:
@marcemarc Cool, no worries mate :-) Hope you enjoy the summer a bit too!

PR #927, opened 2018-07-18
Add documentation for the color-picker editor

jmayntzhusen:


marcemarc:
Thanks @jmayntzhusen although of course in black and white it irks me it's not called the Colour Picker ins tead of Color Picker.. .shudder... nice work on the page though!

PR #935, opened 2018-07-26
Link to the "Textarea" documentation

BatJan:
Apparently forgot to reference the .md file when submitting my PR for the textarea datatype. Do'h! :)


marcemarc:
Doh, I think that's the sort of thing I'm meant to notice!

BatJan:
But it's so hot everywhere you go that no one can think straight these days! :)

PR #939, opened 2018-07-30
Added some documentation about using Session-variables

JeffreyPerplex:
Discovered the hard way (and after searching a lot) that you cannot use dots in a sessionname for some sort of reason.


marcemarc:
@JeffreyPerplex Thanks! - and we feel your pain with the dot issue... hopefully updating here will save future devs (probably yourself too)...

PR #942, opened 2018-08-01
Introduce gitignore file

callumbwhyte:
I found a few weird temp files were visible in Git after solving a merge conflict.

I've added a `.gitignore` file to prevent these from showing up in future.

Surprised this wasn't here already!


marcemarc:
Thanks @callumbwhyte !!!!

PR #945, opened 2018-08-07
How to manage environments

jmayntzhusen:


jmayntzhusen:
I added it as a new "Manage Environments" section under Set Up, maybe there would be a better placement?

PR #934, opened 2018-07-26
Add documentation for the "Numeric" datatype

BatJan:

PR #947, opened 2018-08-09
Correcting the code example in step three

BatJan:
The current example results in the following error System.Web.HttpCompileException (0x80004005): c:\Git\Sites\Umbraco-7-11-1\Views\MacroPartials\LatestBlogposts.cshtml(21): error CS0428: Cannot convert method group 'Count' to non-delegate type 'double'. Did you intend to invoke the method?

I'm no expert in LINQ but when adding .ToList() it works like it should. The code that should be replaced with the example from step three also converts the output using .ToList() - .Take(3) should also not be necessary to add since paging is already setup, which listes 3 posts pr. page.

I noticed this issue when I tried to re-create the error Nic got here https://our.umbraco.com/forum/using-umbraco-and-getting-started/93182-error-editing-partial-macro-in-the-umbraco-starters-guide - I have not heard from him yet about whether my change works for him or not but I'm 99.99995% certain that it will since it works for me :)

### DIFF
--- a/Tutorials/Starter-kit/Lessons/2-Add-a-Blog-Post-Publication-Date/part-3.md
+++ b/Tutorials/Starter-kit/Lessons/2-Add-a-Blog-Post-Publication-Date/part-3.md
@@ -11,7 +11,7 @@ Finally in Part Three we shall change the blog listing.
 ```
 3. Scroll further back up the page to where we get the blog posts (as the **Children** of this page).  Change this to:
 ```
-var blogposts = startNode.Children.OrderByDescending(x => x.GetPropertyValue<DateTime>("PublicationDate")).Take(3);
+var blogposts = startNode.Children.OrderByDescending(x => x.GetPropertyValue<DateTime>("PublicationDate")).ToList();
 ```
 Because we are sorting by a custom property we need to use the generic GetPropertyValue method.
 4. Then click *Save*. A confirmation message should appear confirming that the Partial view has saved.  


BatJan:
Nic just confirmed in the referenced post that the fix I suggested works 😄

marcemarc:
@BatJan cool, the reason the error occurs is that in the calculation of pageCount. **blogPosts.Count** is called, where Count is a property of a List<T> object

eg    ` var pageCount = (int)Math.Ceiling((double)blogposts.Count / (double)numberOfPosts);`

.Children, however returns an IEnumerable<IPublishedContent> which does not have a Count property, hence the error message!

calling ToList() forces the enumeration to be evaluated into memory, and causes the Count property to be populated.

but alternatively, you could call the Linq .Count() method group on the IEnumerable instead, eg don't call ToList() and have
```
var blogposts = startNode.Children.OrderByDescending(x => x.GetPropertyValue<DateTime>("PublicationDate"));
var pageCount = (int)Math.Ceiling((double)blogposts.Count() / (double)numberOfPosts);
```
Now I can find stack overflow articles saying you should always always use .Count() because it first tries to cast to an ICollection object before enumerating to achieve the count, so you are always better off, and I can find stack overflow articles saying you should never use .Count() as it has to check first whether the object can be cast to ICollection :-) so if you know you have a List, just use .Count...

I'm pretty sure for simple examples there will be no difference, but it would be good if the tutorial worked! 

will see if @dampee @Jeavon have a preference...

dampee:
@dawoe  is right all the way.  

No preference for me concerning the Count vs Count().  As long as the example is simple, straightforward and works.

BatJan:
@marcemarc Thank you for the explanation Mark - Much appreciated and sorry for me being slow at getting back to you. UmbracoBBQ planning and an upcoming wedding is a bit much at once apparently 😅 

Will have a look at your suggestions in #934 soonish!

@dampee where did @dawoe comment on this? I'm confused 😃

PR #958, opened 2018-08-21
Documenting the "showDeprecatedPropertyEditors" setting

BatJan:
I had a use for this setting today and discovered that it was actually not documented - So here it is :)


dampee:
I didn't knew we missed this one!  Nice catch @BatJan !

PR #968, opened 2018-08-29
Update index.md

ProNotion:
Add additional detail about configuring your site key for the Recaptcha2 field type.

### DIFF
--- a/Add-ons/UmbracoForms/Editor/Creating-a-Form/Fieldtypes/index.md
+++ b/Add-ons/UmbracoForms/Editor/Creating-a-Form/Fieldtypes/index.md
@@ -56,6 +56,12 @@ Displays a recaptcha field (words seen in distorted text images), where the user
 
 ![Recaptcha](recaptcha.png)
 
+For Recaptcha2 you will need to configure your site key by adding the following key/value pair to your appsettings in the web.config as follows:
+
+    <add key="RecaptchaSiteKey" value="000-sHZZenninFziVUV9TN24FqhwZvc2b4e8BLrG"/>
+    
+You can create a key by [logging into your Recaptcha account](https://www.google.com/recaptcha/).
+
 ## Hidden
 Hidden field, won't be visible to the user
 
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hey @ProNotion 

First of all, thanks a lot for submitting this PR #h5yr!

I was under the impression the reCaptcha key was set here:

 /App_Plugins/UmbracoForms/UmbracoForms.config in that file there are two fields that look like this:

<setting key="RecaptchaPublicKey" value="xxxxxx" />
<setting key="RecaptchaPrivateKey" value="xxxxxx" />

Maybe they both work and one of them is preferred over the other? @warrenbuckley maybe you can weigh in on this? 😃

ProNotion:
@jmayntzhusen In the view file in my project there is the following:

    if (!string.IsNullOrEmpty(siteKey))
    {
        <script src="https://www.google.com/recaptcha/api.js" async defer></script>
        <div class="g-recaptcha" data-sitekey="@siteKey" data-theme="@theme" data-size="@size"></div>
    }
    else
    {
        <p class="error">ERROR: ReCaptcha v.2 is missing the Site Key - Please update the web.config to include 'key="RecaptchaSiteKey"'</p>
    }

I actually wasn't aware of the config settings you've noted so will check that out. In either case there should be details on this page about how to configure it so if I am wrong I am happy to update and submit a new PR but the view should also be updated to reflect the correct location.

jmayntzhusen:
@ProNotion ah yes I see, I haven't tested either of them to be honest, have just been told about the UmbracoForms.config one. If they both work we should figure out which one is considered best practise and include that - otherwise just the one, but I absolutely agree that it should be in the docs 😉

ProNotion:
@jmayntzhusen I can confirm that your solution works and mine does not so the information provided in the view is incorrect. I will amend my PR.

jmayntzhusen:
@ProNotion thanks a bunch for testing it all out and for submitting this PR! #h5yr!
I'm sure this will be helpful to many others.

ProNotion:
@jmayntzhusen No problem. What needs to happen in order for the view to updated in the core for UmbracoForms?

jmayntzhusen:
@ProNotion I am not sure what view you are referring to 😁

I am also not sure that this is the right place to discuss it, you can send me an email with a screenshot or something and I will have a look 😉 jmh@umbraco.com

To ensure a faster response time you should reach out to contact@umbraco.com

If this is something you feel there is missing documentation for could you create an issue instead? 😉

PR #983, opened 2018-09-19
Updating "Aditional Notes" and "Getting env in.."

dzordzs:
In some places, we said that you should run `echo> deploy` in `/data/revisions` folder which is not true.


jmayntzhusen:
Hey @dzordzs 

You are absolutely right, good spot! I will merge this in for you, thanks for submitting this PR. #h5yr!

Issue #980, opened 2018-09-17
Extend the markdown generation classes


dampee:
## What type of issue is it?
Discussion

## What article/section is this about? 
All :-)

## Describe the issue:
Sometimes you want to warn the user, or make a side node. We currently have no possibility to include this kind of extra information in the documentation. And because Microsoft Docs has been doing awesome things, I've been looking at them.  :-)

While I know the solution will need to be a PR on the https://github.com/umbraco/OurUmbraco, my best take is to create the issue in the docs issues because it only affects the documentation.  :-)

Microsoft docs has a neat way of rendering this with simple markdown:
```
> [!NOTE]
> In dit artikel gaat u een app implementeren in App Service onder Windows. Zie [Een PHP- en MySQL-web-app bouwen in Azure App Service onder Linux ](./containers/tutorial-php-mysql-app.md) om een app te implementeren in App Service onder _Linux_.
>
```
Generates this: 
![afbeelding](https://user-images.githubusercontent.com/1857460/45644306-bdc21a80-babd-11e8-9efe-94e1cadb22e1.png)

_Oh, have you seen that NOTE is translated into the uses language. _

## And there is more!
I think we don't only need a "note", but also other things

* Note
* Tip
* warning
* ...  (please comment below)

### a checklist
```
> [!div class="checklist"]
> * Een MySQL-database in Azure maken
> * Een PHP-app verbinden met MySQL
> * De app implementeren in Azure
```
becomes:
![afbeelding](https://user-images.githubusercontent.com/1857460/45645177-fa8f1100-babf-11e8-8508-2d3833b6e3e8.png)

### embedding other md files
While I don't agree wheter we need this, it's an easy way to insert a commercial plug to e.g. umbraco cloud trial.... in a controllable way
```
[!INCLUDE [quickstarts-free-trial-note](../../includes/quickstarts-free-trial-note.md)]
```

### Code examles
```
` ` `sql 
CREATE DATABASE sampledb;
` ` ` 
```

makes me just jealous if you see this:
![afbeelding](https://user-images.githubusercontent.com/1857460/45645363-75f0c280-bac0-11e8-9263-4fc4463232f9.png)

### next section button
Look at this code
```
> [!div class="nextstepaction"]
> [Een bestaande aangepaste DNS-naam toewijzen aan Azure Web Apps](app-service-web-tutorial-custom-domain.md)
```
[...]

PR #990, opened 2018-09-20
Adding a link to the "new" Related Links data type

readingdancer:
I am not sure who named the new one, but as this is marked as obsolete it would make sense that we highlight clearly what people should now be using instead!


sofietoft:
Hi Chris,

It makes a lot of sense to link to the new type here 👌 
I do think it would do better to put it right above the _Data Type Definition Example_ heading, instead of under the main title.
What do you think about that?

Have a good one!
-Sofie

dampee:
@sofietoft I think it's just fine for now.  Would be better if we can replace this later with our classes introduced by issue #980 
But we could revisit this later.  Merging now.

PR #992, opened 2018-09-24
Update index.md

StefanoChiodino:


sofietoft:
Hi @StefanoChiodino 

I'm not really sure I understand the change you've made here.
When you are using Forms v6+ and add a Form, there is an option in the backoffice to _Exclude scripts_.

The code snippet you've added does exactly that, correct?

StefanoChiodino:
Almost. The doc currently references to a "property" Exclude Scripts, the terminology is confusing because 1) is not a property but a parameter and 2) it's excludeScripts, not Exclude Scripts 3) this applies to the macro only, but the code in the next lines is using action instead.

So, with my proposed changes you get actual code snippets that work.

sofietoft:
Thanks for explaning, @StefanoChiodino 

I'll do some testing, just to make sure I understand exactly how this work! 😁

sofietoft:
Hi @StefanoChiodino 

I apologize for the delay! 😁 
I've had some time to test this script rendering when you are using Forms.

I'm still not 100% sure I understand the example you've suggested through this PR.
Sounds like you might still be using Forms 4? 

Anywho. After doing some extensive testing today, I've taken the liberty of updating the documentation to match my findings.

Have a look here: https://our.umbraco.com/documentation/Add-ons/UmbracoForms/Developer/Rendering-Scripts/ 

-Sofie

StefanoChiodino:
Hi,

I'm using the latest Umbraco Forms. I'm not sure what I need to explain, so could you please ask for more specific questions? I've taken the snippets of code directly from my working codebase.

sofietoft:
Sure, no problem 😄 

Where exactly are you using this code snippet:
```
@Html.Action("RenderForm", "UmbracoForms", new { formid = Model.Form, theme = "yourThemeName", includeScripts = false })
```

I tried using it in my templates, but it gave me a YSOD.
Perhaps I'm doing something wrong? 🤔

StefanoChiodino:
That line renders the form, without the scripts. Not sure what YSOD you have got, it won't work out of the box ofc, you need to have that theme with name `yourThemeName` and a `Form` in your `Model`.

PR #998, opened 2018-09-25
Updated name of my markdown editor

prjseal:
To save confusion when going along with this tutorial, now that there is a markdown editor data type we need to be specific about the name of the data type we are choosing from the dropdown. So the correct name is 'My markdown editor'. This will help the user to not pick the wrong one like I did the first time.


dampee:
Thanks @prjseal 
We'll merge this soon!

prjseal:
Resolved conflict

PR #1003, opened 2018-09-26
Add docs for Umbraco.Decimal

bjarnef:
Added `Umbraco.Decimal` to the list of core property editors.


sofietoft:
So, @bjarnef , I just realized that I merged this without reading it properly 🙈 

Looks like you've used the alias as the heading.
Any chance I could get you to update that?
The heading should read "Decimal" instead of "[Umbraco.Decimal]" - right?

bjarnef:
Ahh, yes .. it should be "Decimal".
I was a bit too fast to submit this PR 🙈

PR #1010, opened 2018-09-30
Added sending user model description

dawoe:
In v7.7.10 this event got added to the EditorModelEventManager class


dampee:
Thank a lot @dawoe

PR #1017, opened 2018-10-01
Added version number since when custom routing is supported

dawoe:
This is supported since v7.3. So it should be clear from documentatoin


jmayntzhusen:
Hey @dawoe !

As I wrote on your other PR - #1016, thanks for helping out with this, we will merge it in once version specific documentation is ready 😄

PR #1018, opened 2018-10-01
Added version number of umbraco that supports the tour functionality

dawoe:
This is supported since 7.8 so that should be clear


jmayntzhusen:
Hey @dawoe !

As I wrote on your other PR - #1016, thanks for helping out with this, we will merge it in once version specific documentation is ready 😄

PR #1021, opened 2018-10-01
Updated and renamed ContentService.cshtml

abjerner:
In line with some of my other PRs for the services, I've removed the list of methods as we already have them in the auto-generated documentation.

As there may be more pages about the content service in the future, I've renamed and moved the `ContentService.md` file to a sub folder so future files can be added to the same folder.

The code for the documentation at our.umbraco.com automatically detects the file in the sub folder instead of the old location, so no changes are necessary for this file:

https://github.com/umbraco/UmbracoDocs/blob/master/Reference/Management/Services/index.md


abjerner:
Btw - my plan is to do the same for the other services as well. Do you prefer a big PR for all the remaining services? Or one PR for each service? I'd imagine the latter would be easier to handle in regards to avoiding merge conflicts.

dampee:
Hi @abjerner 

Thanks a lot.

Concerning your question:
if it can be reviewed : one PR is ok.  If it becomes a mess, please make different PR's.  :-)  What you could do is reference each time an issue where you described (once) what you are doing.

PR #1022, opened 2018-10-01
Rewrite section about config transforms for package developers

### LINKED ISSUE #977: Hard-to-follow example in config-transforms documentation
***This is the page (and section) in question: https://our.umbraco.com/documentation/Umbraco-Cloud/Set-Up/Config-Transforms/#including-transforms-in-umbraco-packages***

Though there's some very clever stuff going on in this section, I find it a little confusing and hard to follow - I've turned to this section more than a few times, and I always have to really slow down when trying to decipher it, mainly because of two things:

## 1. Naming

It's does not explain how I should name the file(s) other than _"it can be any word, for example `RollingStones` or `uPaintItBlack`"_ — so which is it? Where do they come from? If you can literally  name them anything, how do you prevent a name clash with another package? Probably much better to state that you should try to use your package's name or alias to prevent clashing with any other transform file(s)? (Only when reading it the second or third time and then picturing to myself what I'd do, did it occur to me that they _can_ be named literally anything because they're just executed.)

## 2. Sample values

While describing the three different values for the example, it's kinda confusing to say that in one place it should be _a red **door**_ so the setting should be `red`, and then suddenly talk about _a green **sea**_ for the second value? Then immediately state that this could just as easily be _a deeper `blue`_ instead? Only if you _really_ know the lyrics to that song will this make an inch of meaning, and I'm guessing that's not the majority of the developers looking this up 😬 

I would prefer a simpler example that just sets a different value (e.g. "Setting the `environmentColor`") for each environment - will probably cover what more than 80% of people looking this up is looking for, right?

I know I could be the only one having trouble with this, but I will (of course) take a stab at rephrasing this section myself, and propose a PR for further discussion.

greystate:
Closes #977

### DIFF
--- a/Umbraco-Cloud/Set-Up/Config-Transforms/index.md
+++ b/Umbraco-Cloud/Set-Up/Config-Transforms/index.md
@@ -10,11 +10,11 @@ Config Transforms are a way to transform your config files without changing the 
 
 To transform a config file, you need to create a new file in the same folder with the following naming convention: `{config-file name}.{environment}.xdt.config`.
 
-If you want to do a transform on your `web.config` file for the Live environment on your project, the config transform you need to create will look like this:
+If you want to do a transform on your `Web.config` file for the Live environment of your project, the config transform you need to create will look like this:
 
-`web.live.xdt.config`
+`Web.live.xdt.config`
 
-The `{environment}` part needs to be replaced with the target environment, for which there's currently 3 possibilities for each project:
+The `{environment}` part needs to be replaced with the target environment, for which there are currently 3 possibilities for each project:
 
 1. `development`
 2. `staging`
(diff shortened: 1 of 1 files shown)


clausjensen:
Thanks @greystate 😃

nul800sebastiaan:
Hi there @greystate,

First of all: **A big #H5YR** for contributing to Umbraco during Hacktoberfest! We are very thankful for the huge amount of PRs submitted, and all the amazing work you've been doing :pray:

If you haven't heard already we have had a record breaking amount of PR's in both the CMS and the Documentation repository. You can read all about the numbers in the [Umbraco Hacktoberfest Aftermath blog post](https://umbraco.com/blog/the-2018-hacktoberfest-aftermath/).

In the spirit of Hacktoberfest we've prepared some **exclusive Umbraco swag** for all our contributors - _including you_! 

[Receive your swag!](https://goo.gl/forms/smjQwEMDM5cUnz6x1) :point_left: Please follow this link to fill out and submit the form, **before November 21st 2018**.

Following this date we'll be sending out all the swag, which also means that it might not reach your doorstep before December, so please bear with us and be patient :pray:

The only thing left to say is **thank you so much for participating in Hacktoberfest**! We really appreciate the help!

Kind regards,
The Umbraco PR Team & The Documentation Curators

PR #995, opened 2018-09-24
- consistent naming for backoffice.

prjseal:
- decided to go with the naming format already used in the navigation menu 'backoffice'

PR #1026, opened 2018-10-02
Renamed back-office to backoffice.

ed-parry:
Picked up the few remaining areas (from PR #995) where "back-office" is used, it was only those that appear with the hyphen that were left, but now should all be in-line with the CMS, using "backoffice".


dampee:
Thanks a lot for your effort @ed-parry !

PR #1035, opened 2018-10-03
Updating Path too long docs

dzordzs:
Added steps on how to fix the "path too long" issue.


jmayntzhusen:
Hey @dzordzs 

Thanks a lot for helping out with the documentation! #h5yr!

PR #1044, opened 2018-10-03
Updating to include more information about Migrations

mikkelhm:
This will update the docs to include a bit more information about the recomendation to use Migrations when using custom tables in the database.

Im unsure wather it is ok to link to sebastiaans blogpost on migrations, but I dont think we have any other documentation on how to do custom migrations, that could be linked to


jmayntzhusen:
Hey @mikkelhm 

Linking to Sebastiaans blogpost is fine, I fixed a few typos but other than that it's all good - I will merge this now 😉

PR #1053, opened 2018-10-04
Updated tags default storage type

prjseal:
Added information about the change to the default storage type for tags since 7.12 and fixed a spelling mistake.

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags.md
@@ -15,7 +15,7 @@ The **Tag group** setting provides a way to categorize your tags in groups. So f
 
 ### Storage type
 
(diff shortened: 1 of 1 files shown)


dampee:
Hi @prjseal 
We have a new system in place to make sure that the current documentation is always for the current version.

What I suggest in this case is to create a new file `tags-vpre-7.12.md` describing the old way.  Then add YAML tags on the top of the new file. Use in the new file (describing old stuff) the versionTo marker.

```
---
versionTo: 7.12.0
---
```

  If you need any help just give a shout!

prjseal:
Thanks @dampee 
I will do this later.

jmayntzhusen:
We have documentation on the versioning now if you need an overview 😉

https://our.umbraco.com/documentation/Contribute/#multi-version-documentation

prjseal:
Thanks for your guidance with this. I have submitted a new PR with those changes. #1078

PR #1054, opened 2018-10-04
Fixed file name error for umbracoSettings.config

prjseal:


dampee:
Haha!  Good catch @prjseal

PR #1075, opened 2018-10-09
Fix invalid URL's for Courier Nightly URL's

### LINKED ISSUE #912: Invalid URL's for Courier Nightly URL's
I think this one needs someone from HQ, so maybe @sofietoft can help?

On this page:

https://github.com/umbraco/UmbracoDocs/blob/master/Add-ons/UmbracoCourier/CommonIssues.md

All the links are invalid, they look like they should be re-directs that then point to the current location of these files, but we don't have access to fix these re-directs, I assume this is something the HQ can do. I could update these to point directly to the files, but I assume the re-directs are used to create statistics / make it possible to change the location of the stored files.

Also, I am pretty sure the links are pointing to quite old versions and should be updated to point to more recent nightly versions of these files.

All the nightly builds can be found here:

http://nightly.umbraco.org/?container=umbraco-courier

danimart1991:
Fixes #912


jmayntzhusen:
Thanks for updating the links @danimart1991 , I think this whole page needs an overhaul, the current Courier version is 3.1.7 and here it mentions Umbraco v4 and 6 which are also heavily outdated.

danimart1991:
I include changes that mentions Umbraco v4 and v6, and fix spacings

dampee:
I do also think we should not link to the "latest" version on this page.  Or even mention the latest version number on this page.

jmayntzhusen:
Hey @danimart1991 

As this has been a bit inactive and another person submitted a more complete PR #1121 that fixes this issue, this will now be closed. Thanks for taking the time to submit this PR 🙂

PR #1088, opened 2018-10-11
Add links to the Umbraco models in reference

pxgamer:
I'm not sure if these were left out intentionally due to there not being much content in the model docs for them. But I was trying to find out more about these and saw they weren't links.


jmayntzhusen:
Hey @pxgamer 

I think these were all automatically generated and aren't linked to because noone has added the information yet.. If you compare these two you will see what I mean:

https://our.umbraco.com/documentation/Reference/Management/Models/language
https://our.umbraco.com/documentation/Reference/Management/Models/Media

Maybe we should make an issue instead so maybe someone can add these 🤔

PR #1095, opened 2018-10-12
Update index.md

dharmendra-lingaiah:
Minor grammar change

### DIFF
--- a/Cheatsheets/index.md
+++ b/Cheatsheets/index.md
@@ -11,6 +11,6 @@ Covers:
 
 ## [Converting Masterpages to Views](masterpagestoviews.md)
 
-Covers how you convert common syntax in Umbraco Masterpages to Umbraco Views.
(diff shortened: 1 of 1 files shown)


Cheatsheets/index.md
```diff
@@ -11,6 +11,6 @@ Covers:
 
 ## [Converting Masterpages to Views](masterpagestoviews.md)
 
-Covers how you convert common syntax in Umbraco Masterpages to Umbraco Views.
+Covers how you can convert common syntax in Umbraco Masterpages to Umbraco Views.
```
dampee: This implies that there are other ways to convert.  Could you add in the pull request which other ways are available?

dharmendra-lingaiah: I agree there could be other ways. We can retain original style if you feel it makes sense

jmayntzhusen:
Hey @dharmendra-lingaiah 

As this seems like an unnecessary change I will go ahead and close this, thanks for the effort 🎉

PR #1099, opened 2018-10-14
Moved files to their own sub folders

abjerner:
As a follow up on https://github.com/umbraco/UmbracoDocs/pull/1021 and some of my other PRs, I've moved the Markdown files for the remaining services so they rather than being placed in the `Services` folder are placed as index files of their own sub folders. This makes sense if we later are creating additional (sub) pages for each service - which I plan to ;)


dampee:
Seems correct.  I don't think we need 301 redirects as the link stays the same anyway.

PR #1102, opened 2018-10-15
Update website-managed.md

stevetemple:
This won't run without additional app settings


jmayntzhusen:
Hey @stevetemple 

This is a fairly new method, and you are right, it takes the Config as a parameter 😁 Thanks for helping out with the documentation #h5yr!

PR #1105, opened 2018-10-15
Merge gone wrong?

abjerner:
This may be a merge gone bad, or just the side effect of two people working on PRs for the same file simultaneously.

I made a PR for renaming `ContentService.md` to `ContentService/Index.md`, in which I also removed some of the redundant content: https://github.com/umbraco/UmbracoDocs/pull/1021 

While waiting for this to be merged, another user worked in this PR for changing the same file: https://github.com/umbraco/UmbracoDocs/pull/1062

As a result, the file was re-added at it's old location. #1062 fixed a few spelling mistakes, but as far as I can tell, this was for the part of the text that I removed in my PR.

So my suggestion would simply be to delete the re-added file - hence this PR.


jmayntzhusen:
Hey @abjerner 

I think you are right, good catch! I'll merge this now 😄

PR #1110, opened 2018-10-16
Updated health check folder name and word usage

ed-parry:
Quick update for the folder name so that the section appears as "Health Check" in navigation, to match the name used in Umbraco. Also added a missing code snippet around an attribute name and updated the section name as well for consistency.

### DIFF
--- a/Extending/Health-Check/index.md
+++ b/Extending/Health-Check/index.md
@@ -1,4 +1,4 @@
-# Health check
+# Health Check
 
 The developer section of the Umbraco backoffice holds a dashboard named "Health Check". It is a handy list of checks to see if your Umbraco installation is configured according to best practices. It's possible to add your custom built health checks.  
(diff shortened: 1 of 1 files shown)


dampee:
@ed-parry We can't just rename files as that would invalidate old links .  Therefor a new 301 redirect will need to be created. 
I need to ask @jmayntzhusen or @sofietoft how they do that.

ed-parry:
Ah yes @dampee - a quick look suggests there's two or three links to that page in the documentation, but agree a redirect would be preferred for external links too.

jmayntzhusen:
Hey @ed-parry 

I agree, this is a better naming so I will add a PR to add a rewrite for this on Our, then we can hold off on merging this until the redirect is in place 😄

ed-parry:
Great, thanks @jmayntzhusen! Will keep that in mind for the future :)

PR #1113, opened 2018-10-16
Updated the ContentTypeService/Index.md file a bit

abjerner:
As a follow up on https://github.com/umbraco/UmbracoDocs/pull/880 and some of my other PRs, I've removed the list of methods from the Markdown file for the content type service.

I've also extended the line about how to obtain a reference to the content type service.


sofietoft:
Thanks @abjerner 😄 

@marcemarc , could you follow up on this one?

marcemarc:
Thanks @abjerner yes, if the autogenerated docs will be the point of truth, then agree makes no sense to list methods out here, as per previous discussions, thanks for spotting and hoovering this up!

PR #1114, opened 2018-10-16
#H5IS

abjerner:
`contentService` should obviously be `ContentService`. My bad 😊


marcemarc:
@abjerner doh! or should that be Doh!

PR #1127, opened 2018-10-18
Fixed images paths

pbres:
Images paths were not updated and were pointing to the folder one level higher than they should do


jmayntzhusen:
Hey @pbres 

Thanks for the catch, this is not a huge deal as the YAML parser will strip the `..` anyways, but it's better to keep it consistent! 😉

Will merge now!

PR #1134, opened 2018-10-19
Adds the V8 Cake Content App as a C# example

warrenbuckley:
Adds an example of how to do a Content App as a C#

### DIFF
--- a/Extending/Content-Apps/index-v8.md
+++ b/Extending/Content-Apps/index-v8.md
@@ -114,4 +114,67 @@ In a similar way you can limit your content app according to user roles (groups)
 
 ## Creating a Content App in C#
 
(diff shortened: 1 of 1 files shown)


prjseal:
Hi Warren, 
I am guessing we will still need the files:

package.manifest
mycontentapp.html
mycontentapp.controller.js

I am aslo guessing that with this C# code, the manifest file doesn't need the contentApps property on it, so could the manifest file be simplified like this when using the C# approach?

```
{
    // array of files we want to inject into the application on app_start
    javascript: [
        '~/App_Plugins/MyContentApp/mycontentapp.controller.js'
    ]
}
```

If that is correct, I wonder if it is worth saying that in the file?

PR #1139, opened 2018-10-20
Added more specific info about placing 'show' directive

pbres:
Current info was a little bit confusing for me, I didn't know whether should I add "show" directive under 'contentApps' item or on the same level as Content Apps definition.


jmayntzhusen:
Hey @pbres 

I'm all for clarifying if it's hard to follow so let's get this merged in 🙂

PR #1145, opened 2018-10-22
Small changes to word order and grammar

emmaburstow:
One word needed moving and while I was in there, I made some of the grammar more consistent so that numbered bullet points all begin with capitals rather than a mix across the page and also so that bullets where some ended with a full stop and some didn't now all end with full stops. :)


jmayntzhusen:
Hey @emmaburstow ,

Thanks for the contribution! #h5yr!

PR #1155, opened 2018-10-23
Relation Type model definition

Matthew-Wise:
Added the model definitions of the RelationType model. Most of the text taken from the Summary comments in the code :)

### DIFF
--- a/Reference/Management/Models/RelationType.md
+++ b/Reference/Management/Models/RelationType.md
@@ -1,8 +1,13 @@
+---
+keywords: relation type RelationType relations related
+versionFrom: 6.0.0
+---
 # RelationType
+
 
 **Applies to Umbraco 6.x and newer**
 
-Represents a relation type between two object types of either media or content.
+The `RelationType` class represents a relation definition between two node types (content or media). For example keeping track of node usage across the site, in order to avoid deleting content that is used else where. When querying a relation this is done using the parent node key unless the `RelationType` is set to bidirectional in which case you can also query using the child node key.
 
  * **Namespace:** `Umbraco.Core.Models` 
  * **Assembly:** `Umbraco.Core.dll`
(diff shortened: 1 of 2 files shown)


Reference/Management/Models/RelationType.md
```diff
@@ -17,6 +17,47 @@ All samples in this document will require the following using statement:
...
+	return relationType.ParentObjectType;
+    
+ ### .ChildObjectType
+  Gets or sets the Childs object type id as `Guid`
+  
+	// Given a `RelationService` object get RelationType by its Id and return IsBidirectional
+	var relationType = relationService.GetRelationTypeById(1234);
+	return relationType.ChildObjectType;
+    
+## Methods
```
jmayntzhusen: Is there a reason you left a `Methods` headline with no content under it? 🙂

Matthew-Wise: No I should of just removed it

dampee:
The idea is that we are not starting to copy paste all properties in the docs. The auto-generated content should be auto-generated.  Otherwise, we have a massive copy-paste job.
Having a description at the top is good, as well as examples, guides or tips on how and when to (not) use it.

While we like to thank you for the effort @Matthew-Wise, I think we should partially reject your PR.  I hope you understand the "we do not like copy paste"-vision. Do you see a chance to update your PR accordingly?

Matthew-Wise:
@dampee  I agree on the copy paste content, I think some of it is unavoidable as adding further explanation to some properties would be unnecessary. I will however update some that need a bit more explaining.

I'll take a look at the meta/yaml as well.

Matthew-Wise:
I also noticed on the Content one it includes properties from "Entity" so I will add those. 

Or should an Entity page be created and that inheritance made clear? So we avoid maintaining Entity in so many places.

dampee:
@Matthew-Wise 
Sorry for the waiting, there were a few issues with the markdown, but i've fixed them for you!
Thanks a lot for the contribution

PR #1170, opened 2018-10-24
Reordered some paragraphs, clean up + LightInject example

mzajkowski:
I've reordered some sections to be more logically separated and cleaned up IApplicationEventHandler example for Unity. I also added LightInject example with helper method to register services from other assemblies in the domain. Hopefully it'll be useful to merge!

### DIFF
--- a/Reference/using-ioc.md
+++ b/Reference/using-ioc.md
@@ -11,6 +11,20 @@ We don't use IoC in the Umbraco source code. This isn't because we don't like it
 In most IoC frameworks you would setup your container in your global.asax class. To do that in Umbraco, you will need to inherit from our global.asax class called: `Umbraco.Web.UmbracoApplication`. You should then override the `OnApplicationStarted` method to build your container and initialize any of the IoC stuff that you require.
 Alternatively you can implement the `Umbraco.Web.IApplicationEventHandler` interface.
 
+## What assemblies and controllers do I need to register?
+
+You need to register all assemblies that may contain MVC or WebApi controllers. In Umbraco this is the `umbraco` assembly which you can get a direct assembly reference to using the example syntax used above:
+
+```
+typeof(UmbracoApplication).Assembly
+```
+
+If you don't register assemblies that contain controllers you may end up with YSOD errors. If you do not register a controller then ASP.NET will try to create the controller but if it doesn't have an empty constructor you'll get a YSOD.
+
+## Things to note
+
+We use a custom MVC controller builder in our code called `Umbraco.Web.Mvc.MasterControllerFactory`, which needs to always be the default controller factory, if you change this Umbraco will probably not work anymore. The good news is that you can specify 'slave' factories so you can specify custom controller factories for different purposes. You would just need to create a new class that inherits from `Umbraco.Web.Mvc.IFilteredControllerFactory` and ensure that the class is public (so it can be found). If your IoC implementation affects the default controller factory, you may have to modify it in order to support this implementation. For the most part, most IoC frameworks will just target setting a custom DependencyResolver which is 100% ok.
+
 ## Autofac Example
 
 This example will setup Autofac to work with Umbraco (see [their documentation](https://autofac.readthedocs.org/en/latest/) for full details). Our examples make use of the following NuGet packages: `Autofac`, `Autofac.Mvc5`, `Autofac.WebApi2`.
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
I don't know how IoC works, hopefully @Shazwazza can give it a quick look 😄

mzajkowski:
Speaking about it.. I'll add something else to this description. I think it's worth to mention Shannon's package: https://github.com/Shazwazza/Our.Umbraco.IoC as an option as well. Wait for it!

Reference/using-ioc.md
```diff
@@ -11,6 +11,20 @@ We don't use IoC in the Umbraco source code. This isn't because we don't like it
 In most IoC frameworks you would setup your container in your global.asax class. To do that in Umbraco, you will need to inherit from our global.asax class called: `Umbraco.Web.UmbracoApplication`. You should then override the `OnApplicationStarted` method to build your container and initialize any of the IoC stuff that you require.
 Alternatively you can implement the `Umbraco.Web.IApplicationEventHandler` interface.
 
+## What assemblies and controllers do I need to register?
+
+You need to register all assemblies that may contain MVC or WebApi controllers. In Umbraco this is the `umbraco` assembly which you can get a direct assembly reference to using the example syntax used above:
```
jmayntzhusen:
```suggestion
You need to register all assemblies that may contain MVC or WebApi controllers. In Umbraco this is the `umbraco` assembly which you can get a direct assembly reference to using the example syntax used below:
```
It's below, right? 🙂

Reference/using-ioc.md
```diff
@@ -227,3 +220,57 @@ And here is the merged UnityActivator
...
+            container.EnableWebApi(GlobalConfiguration.Configuration);
+
+            // Setting up the resolver
+            var resolver = new LightInjectWebApiDependencyResolver(container);
+            GlobalConfiguration.Configuration.DependencyResolver = resolver;
+            DependencyResolver.SetResolver(new LightInjectMvcDependencyResolver(container));
+		}
+	}
+    
+If we're going to call/implement this class outside of the Web application or inside of the package for example, it might be useful (and sometimes required) to also register all the other existing Controllers and API Controllers inside of the whole solution. Especially if we'll be willing to use them or inject them in our own registrations. Below code is scanning through the assemblies within the current domain and registering existing controllers which are implementing IHttpController interface.
```
jmayntzhusen:
```suggestion
If we're going to call/implement this class outside of the Web application or inside of the package for example, it might be useful (and sometimes required) to also register all the other existing Controllers and API Controllers inside of the whole solution. Especially if we'll be willing to use them or inject them in our own registrations. The code below scans through the assemblies within the current domain and registering existing controllers which are implementing the IHttpController interface.
```

Reference/using-ioc.md
```diff
@@ -227,3 +220,57 @@ And here is the merged UnityActivator
...
+                    container.Register(controllerType, new PerRequestLifeTime());
+                }
+            }
+        }
+        
+We can call the helper method after our own services registrations.
+
+## Our.Umbraco.IoC
+
+There is also another way of dealing with IoC containers in Umbraco solution and it's with using the package created by one of the Core developers - Shannon Deminick - called: Our.Umbraco.IoC (https://github.com/Shazwazza/Our.Umbraco.IoC).
```
jmayntzhusen:
```suggestion
There is also another way of dealing with IoC containers in Umbraco solutions and it's with using the package created by one of the Core developers - Shannon Deminick - called: [Our.Umbraco.IoC](https://github.com/Shazwazza/Our.Umbraco.IoC)
```

nul800sebastiaan:
Hi there @mzajkowski,

First of all: **A big #H5YR** for contributing to Umbraco during Hacktoberfest! We are very thankful for the huge amount of PRs submitted, and all the amazing work you've been doing :pray:

If you haven't heard already we have had a record breaking amount of PR's in both the CMS and the Documentation repository. You can read all about the numbers in the [Umbraco Hacktoberfest Aftermath blog post](https://umbraco.com/blog/the-2018-hacktoberfest-aftermath/).

In the spirit of Hacktoberfest we've prepared some **exclusive Umbraco swag** for all our contributors - _including you_! 

[Receive your swag!](https://goo.gl/forms/smjQwEMDM5cUnz6x1) :point_left: Please follow this link to fill out and submit the form, **before November 21st 2018**.

Following this date we'll be sending out all the swag, which also means that it might not reach your doorstep before December, so please bear with us and be patient :pray:

The only thing left to say is **thank you so much for participating in Hacktoberfest**! We really appreciate the help!

Kind regards,
The Umbraco PR Team & The Documentation Curators

dampee:
Hi @mzajkowski 
Could you review the changes by @jmayntzhusen and rebase your branch to solve the merge conflict?

Reference/using-ioc.md
Shazwazza: @jmayntzhusen yes i think so :) // @mzajkowski

Reference/using-ioc.md
```diff
@@ -11,6 +11,20 @@ We don't use IoC in the Umbraco source code. This isn't because we don't like it
...
+
+```
+typeof(UmbracoApplication).Assembly
+```
+
+If you don't register assemblies that contain controllers you may end up with YSOD errors. If you do not register a controller then ASP.NET will try to create the controller but if it doesn't have an empty constructor you'll get a YSOD.
+
+## Things to note
+
+We use a custom MVC controller builder in our code called `Umbraco.Web.Mvc.MasterControllerFactory`, which needs to always be the default controller factory, if you change this Umbraco will probably not work anymore. The good news is that you can specify 'slave' factories so you can specify custom controller factories for different purposes. You would just need to create a new class that inherits from `Umbraco.Web.Mvc.IFilteredControllerFactory` and ensure that the class is public (so it can be found). If your IoC implementation affects the default controller factory, you may have to modify it in order to support this implementation. For the most part, most IoC frameworks will just target setting a custom DependencyResolver which is 100% ok.
```
Shazwazza: We'll need to change the term 'slave' to 'sub' (or similar)

Reference/using-ioc.md
```diff
@@ -156,12 +156,9 @@ Leave the UnityConfig file as is.
 
 Now create the following files to configure Unity correctly for an Umbraco site.
 
-	class UnityEvents : IApplicationEventHandler
+	public class UnityEvents : ApplicationEventHandler
```
Shazwazza: This is incorrect, there is a reason why it is using the interface! If you use the base class, then you need to override both: `ExecuteWhenApplicationNotConfigured` and `ExecuteWhenDatabaseNotConfigured` and return true for both of these, by default this is false because this base class is generally just used for plugins/extensions but your container must always be populated.

Reference/using-ioc.md
Shazwazza: @mzajkowski agreed with the above request

Reference/using-ioc.md
Shazwazza: I think this should go at the very top, people don't need to manually implement all of this and it would save a lot of headaches to just use this as the recommended approach.

jmayntzhusen:
As we haven't seen any replies from @mzajkowski  for a long time I will close this.

PR #1180, opened 2018-10-26
- removed duplicate word 'the'

prjseal:


jmayntzhusen:
Thanks for the fix! 😄

PR #1182, opened 2018-10-26
- removed duplicate word 'to'

prjseal:


jmayntzhusen:
Thanks for the fix! 😄

PR #1187, opened 2018-10-26
- changed word 'the' to 'to'

prjseal:


jmayntzhusen:
Thanks for the fix! 😄

PR #1189, opened 2018-10-26
Update index.md

czakarian:
Fixed the file directory  for Creating-a-Property-Editor


jmayntzhusen:
Hey @czakarian 

Thanks for taking the time to help out with our documentation, I will have to reject this PR though as you are "fixing" a link that already works - the way our routing is set up you don't actually need to link to the full file path, we have a PR on it's way that should clarify that in our contribution guidelines 🙂

PR #1191, opened 2018-10-27
Update index.md - grammar fixes

abbas10m:
Fixed some incorrect grammar

### DIFF
--- a/Implementation/Default-Routing/Execute-Request/index.md
+++ b/Implementation/Default-Routing/Execute-Request/index.md
@@ -30,7 +30,7 @@ Working with dynamics is simpler in some cases especially with regards to refere
 
 	@CurrentPage.markDown
 
-However, dynamics does not provide any intellisense and is compiled at runtime.
+However, a dynamic object does not provide any intellisense and is compiled at runtime.
 
 ## [UmbracoHelper](../../../Reference/Querying/UmbracoHelper/index.md)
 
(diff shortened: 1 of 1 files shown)


Implementation/Default-Routing/Execute-Request/index.md
```diff
@@ -30,12 +30,12 @@ Working with dynamics is simpler in some cases especially with regards to refere
 
 	@CurrentPage.markDown
 
-However, dynamics do not provider any intellisense and are compiled at runtime
+However, dynamics does not provide any intellisense and is compiled at runtime.
```
jmayntzhusen: This sounds a bit wrong to me, is the word `dynamics` not plural? So `do` and `are` are the correct words?

marcemarc: @jmayntzhusen it's confusing because 'dynamics' can be considered to be the 'subject' of working with dynamic objects, but I think in this context it would be clearer still to have:

However, a _dynamic_ object does not provide any intellisense and is compiled at runtime.

jmayntzhusen: Yep @marcemarc 's suggestion should clear it up, could you make that change @abbas10m ? 🙂

abbas10m: Updated. Thanks for the feedback @jmayntzhusen and @marcemarc  :)

jmayntzhusen:
Awesome, thanks for your contribution! 🎉

PR #1192, opened 2018-10-27
Removed method lists from the remaining services

abjerner:
Another follow-up from https://github.com/umbraco/UmbracoDocs/pull/880

With this PR, I've removed the method lists for the the remaining services.

While most of the methods described are similar to the XML documentation, a few of them also has an example or an improved description. I'll try keeping those in mind when I create examples like I have done for a few of the services.


jmayntzhusen:
Hey @abjerner 

Amazing job you have been doing cleaning up the services 👍🎉

Due to the examples in these ones I think it would be a waste to have them all deleted. 

I know you may have plans to add the examples again at a later time, but maybe this PR should be amended to only delete the regular methods and then leave the examples - then you can remove them when you add them elsewhere 🙂 

What do you think?

abjerner:
@abjerner I'll have a look at what makes the most sense. It's only two services that had examples as part of the method list, so perhaps I can just add new pages those through new PRs. I'll get back to you on that ;)

sofietoft:
Hi @abjerner !

We're going through the PRs here - this one has been here for a while 🙈 , and also seems like quite a task to merge!

Should we still keep it open, or would it be okay to close it for now?
What do you think? 😄  You're of course welcome to re-open and continue working on it at a later point.

abjerner:
Hi @sofietoft,

I still think the changes make sense, but as some of the files has changed along the way, wouldn't it make sense that I create new PR(s)?

And if so, should I create a single PR similar to this one, or should I split it up in individual PRs? I'd imagine the latter would give the least headaches in terms of potential merge conflicts.

dampee:
I am pretty sure I can easily resolve the conflicts.  I'll try to look at it later.

dampee:
Ha!  I finally nailed it (after 3 retries).  It would have been faster to redo this manually.  But it worked out in the end.  :-)

abjerner:
Famous last words: 

> I am pretty sure I can easily ....

😂 

Joke aside, great work with the merge @dampee 👍

dampee:
I tried to solve it with a merge, and that gave never the result I really wanted. So ended up with a rebase and a force push.

PR #1196, opened 2018-10-27
Update index-v8.md

crgrieve:
Copying some info from content apps page as an intro for v8 specific landing page and fixing link.

### DIFF
--- a/Extending/index-v8.md
+++ b/Extending/index-v8.md
@@ -8,6 +8,9 @@ versionFrom: 8.0.0
 These features are only available in v8
 :::
 
-## [Content Apps](Content-Apps/index-v8)
+## [Content Apps](Content-Apps/index-v8.md)
 
-New feature in v8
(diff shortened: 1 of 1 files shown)


Extending/index-v8.md
```diff
@@ -8,6 +8,9 @@ versionFrom: 8.0.0
 These features are only available in v8
 :::
 
-## [Content Apps](Content-Apps/index-v8)
+## [Content Apps](Content-Apps/index-v8.md)
```
jmayntzhusen: Adding .md here is not necessary as the link manager looks for both folders and files with the name, if you test it out on the current page the link works just fine 😉

dampee: I am merging this, as it's a file anyway.  The .md extension makes it work on github too...

dampee:
Thanks a lot @crgrieve

PR #1197, opened 2018-10-27
 Updated the documentation for the content type service

abjerner:
Added some information about retrieving content types and content type containers.

I have a general question about headlines. Should it be something like **Getting a single content type** or **Getting a Single Content Type**? Or something else?

Anyways - I've used fenced code blocks as these is now supported by Our (YAY :tada:). Also, with https://github.com/umbraco/OurUmbraco/pull/332 the documentation should look something like below:

image

### DIFF
--- a/Reference/Management/Services/ContentTypeService/Index.md
+++ b/Reference/Management/Services/ContentTypeService/Index.md
@@ -1,6 +1,8 @@
-# ContentTypeService
+---
+versionFrom: 6.0.0
+---
 
-**Applies to Umbraco 6.x and newer**
+# ContentTypeService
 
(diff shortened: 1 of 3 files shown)


marcemarc:
@abjerner excited to merge this in because of the formatting will raise on docs meeting tomorrow about headings etc

marcemarc:
@abjerner Merged!!! - when there is text saying applies to version 6 and above, we're removing and adding YAML to describe the version it applies to - @dampee has updated the PR for this with that info in.

With regard to the headings, we all prefer:

 Getting a single content type

(but imagine it's Crazily all over the Place throughout the Docs)

looks good, thanks!

PR #1198, opened 2018-10-27
Should be ".NET" and "C#" (AKA upper case)

abjerner:
They should both use capital letters to match the official spelling.


jmayntzhusen:
Nice catch, thanks!

PR #1201, opened 2018-10-29
LICENSE: Fix indentation

ayan-b:
Previously scrolling was required. The reason being the previous file (taken from https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/LICENSE.md) is of .md format but the added one in this repo is of .txt format. The indentation is fixed so that scrolling is not required. Another fix can be changing the file format.

Current file: https://github.com/umbraco/UmbracoDocs/blob/e1efd035191627eb83fdf4358797e324e452a68e/LICENSE
Previous file: https://github.com/prog1dev/UmbracoDocs/blob/f07975c9bbd83e9db871898837617f4d10e68d2c/LICENSE

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,9 +1,19 @@
-# The MIT License (MIT) #
-
 Copyright (c) 2013-present Umbraco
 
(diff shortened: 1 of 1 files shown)


ayan-b:
Please review  @jmayntzhusen

jmayntzhusen:
Hi @ayan-b 

I updated the file extension, I don't think manually making the lines shorter is a good fix. Thanks for bringing it to my attention though 🙂 Will close this PR now.

PR #1205, opened 2018-10-29
Updated code examples

darren-welch:
I believe the examples are wrong, unless I misunderstand something (which is highly likely):  Whenever I use nested content, I need to use "Model.**Content**.GetPropertyValue".  

Model.GetPropertyValue always raises an error.

So I believe this will fix the examples.  If I am wrong then please discard this.

Thanks, Darren

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Nested-Content.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Nested-Content.md
@@ -79,7 +79,7 @@ Example:
 
     @inherits Umbraco.Web.Mvc.UmbracoViewPage
     @{
-        var items = Model.GetPropertyValue<IEnumerable<IPublishedContent>>("myPropertyAlias");
+        var items = Model.Content.GetPropertyValue<IEnumerable<IPublishedContent>>("myPropertyAlias");
     
         foreach(var item in items)
         {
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hi @darren-welch 

I think you are right, but we have a few other issues with the page as well, it uses @Umbraco.Field and it inherits from @inherits Umbraco.Web.Mvc.UmbracoViewPage

Maybe what we need here is for someone to set this up using the best practices and then remake the examples 🤔

You are very welcome to set this up and test and do a complete update 😉

dampee:
Hi @darren-welch @jmayntzhusen 

Thanks a lot for your effort, but - as you thought - the you are incorrect. :-)

If you use `UmbracoTemplatePage` at the `@inherits`  you will need `@Model.Content.GetPropertyValue(...)` otherwise it is `@Model.GetPropertyValue(...)`

Concerning the best practices:

1. Umbraco serves UmbracoTemplatePage by default.
2. UmbracoTemplatePage will disappear in v8; being replaced with UmbracoViewPage
3. UmbracoTemplatePage exposes the dynamic "CurrentPage" which is deprecated for v8 AND is discourages from being used.

SO... i think having UmbracoTemplatePage is just fine.  Therefor I'm closing this PR.

@darren-welch thanks a lot for trying to improve the docs! It is appreciated a lot. We are looking forward to your next contribution (and hope it will be merged in).

PR #1214, opened 2018-10-30
Remove the children part as well here

Aaen:
I removed the children part for the non-strongly typed version so the example works too


marcemarc:
And thanks again @Aaen here...

PR #1229, opened 2018-11-04
Simplify security section

dampee:
I was trying to read the Security section, and it is impressively LONG...

So enormous that I could not find what I needed and kept scrolling.  That was until i noticed that the different sections had titles.  So I moved all the things into different files.  Made a small introduction where necessary.  

And I also added 

* some version information.
* information from https://issues.umbraco.org/issue/U4-6753 (see the file auto-linking.md)


dampee:
For anyone who wishes to review:

* verify I have moved information, not removed something by accident
* verify that all new pages have valid links and can be reached

dampee:
Thanks a lot @marcemarc

PR #1230, opened 2018-11-04
add version info for Content Templates

dampee:


jmayntzhusen:
This seems straight forward - let's just merge it 👍😁

PR #1241, opened 2018-11-11
UmbracoSettings version cleanup

dampee:
Introduced some version information in different files + code cleanup of the index file

### DIFF
--- a/Reference/Config/umbracoSettings/index-vpost7.5.0.md
+++ b/Reference/Config/umbracoSettings/index-vpost7.5.0.md
@@ -1,0 +1,11 @@
+---
+versionFrom: 7.5.0
+---
+
+# New settings in 7.5.0
+
+## `<allowPasswordReset>`
+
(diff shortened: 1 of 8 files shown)


jmayntzhusen:
This is great @dampee !

Only thing I noticed is that you named your files something like index-vpost7.5.0.md in our guidelines we write it should be something like index-vpost-v7.5.0.md

I don't remember the code we have for this, but either the guidelines need an update or your filenames do 😁

dampee:
We need to have the guidelines updated, because we search for "-v", it is our marker.  We should not have 2 markers...

dampee:
@jmayntzhusen 
Just reviewed the guidelines.  The text is wrong, the examples are correct.

https://github.com/umbraco/UmbracoDocs/blob/master/Contribute/file-naming-conventions.md#examples

dampee:
And updated the guidelines too...  :-)

PR #1250, opened 2018-11-20
Content Apps: added fenced code blocks for syntax highlighting

abjerner:
There is probably a lot of work in updating old documentation pages with fenced code blocks, but there aren't really that many v8 pages yet - so I've now fixed the first one.

With fenced code blocks also comes syntax highlighting, which comes with a number of different issues for the JSON examples in this article. Per the JSON specification:

1. property names should be enclosed with double quotes
2. string values shopuld be enclosed with double quotes
3. comments are not part of the JSON specification

for both 1. and 2., no quotes or single quotes make the JSON invalid. As such, GitHub's syntax highlighting looks like as shown here:

image

So the big question is now: how much should we actually care about this? While some of the examples are invalid according to the JSON specification, JSON.net is more forgiving, so the examples will work in Umbraco. But editors like Visual Studio and VS Code will complain.

Single quoted strings and comments are also currently not supported by my syntax highlighter package, which is now used on Our - but both will be supported in the next release (the code is already on GitHub, I just haven't pushed a new release yet).

My proposal is that property names and string values always should be enclosed in double quote per the JSON specification, but comments will still be allowed, as they may have a great explanatory effect. I'd be happy to help with that for existing and as well as new pages.

Any thoughts?

### DIFF
--- a/Extending/Content-Apps/index-v8.md
+++ b/Extending/Content-Apps/index-v8.md
@@ -49,23 +49,25 @@ Next we need to create a manifest file to describe what this content app does. T
 
 Add the file `/App_Plugins/CakeContentApp/package.manifest` and inside add the JSON to describe the content app. Have a look at the inline comments in the JSON below for details on each bit:
 
-    {
-        // define the content apps you want to create
-        contentApps: [
-            {
-                "name": "Cake", // required - the name that appears under the icon, everyone loves cake, right?
-                "alias": "appCake", // required - unique alias for your app
-                "weight": 0, // optional, default is 0, use values between -99 and +99 to appear between the existing Content (-100) and Info (100) apps
-                "icon": "icon-cupcake", // required - the icon to use
-                "view": "~/App_Plugins/CakeContentApp/cakecontentapp.html" // required - the location of the view file
-            }
-        ]
-        ,
-        // array of files we want to inject into the application on app_start
-        javascript: [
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hey @abjerner 

You are right, we should try to keep at least new things up to date with the awesome new syntax highlighting you made 🎉

I agree that we should just enforce double quotes so we don't have any issues - with regards to the comments there is apparently a workaround of highlighting it as json5 on Github, should allow comments I tested it here: https://github.com/jmayntzhusen/UmbracoDocs/pull/3/files 

Don't know if your package supports this though?

dampee:
I think we should make sure examples can be "copy-pasted".
As it's a config file in this case, we should make sure people do not have errors when they copy the code.

My best take on this: 
1. move the comments to the explanation below.  Or
2. make sure we support JSON5 :-)

abjerner:
@jmayntzhusen @dampee I've now updated the examples so they're more valid, which mostly goes for the JSON examples. I've also switched `json` out with `json5`.

Regarding syntax highlighting for `json5`, it should be fine merging this PR for now, as the syntax highlighting will kick in when my [other PR](https://github.com/umbraco/OurUmbraco/pull/360) for Our Umbraco is merged.

dampee:
Great!  Lets merge this!

PR #1254, opened 2018-11-26
Updated Debug and CustomErrors section.

dzordzs:
We set `debug` to `true` and `CustomErrors` to `Off` only on Cloud environments. 
If you clone it down locally, you'll have Debug set to RemoteOnly and Compilation to False.


jmayntzhusen:
Hey @dzordzs 

You are correct, but maybe since you already wrote it here, you could update this PR to include the expected behavior on local? 🙂

PR #1255, opened 2018-11-30
Update code example

Aaen:
I have updated the code example so it works and renders items out


dampee:
Hi Dennis
Was the sample not working?  What is wrong with the current code?
We try to use ModelsBuilder syntax as much as possible.

PR #1263, opened 2018-12-08
Remove apostrophes

nathanwoulfe:
Remove unneccesary apostrophes - SurfaceControllers and UmbracoApiControllers are plural, not possessive, so no need for apostrophe.


jmayntzhusen:
Hey @nathanwoulfe 

Thanks for the update, I will take your word for it since you are a native speaker 😄

PR #1265, opened 2018-12-08
Errant apostrophe

nathanwoulfe:
Found another one - `a content editors'` should be `a content editor's`, as the `a` makes it singular but `editors'` is the possessive plural. Since it's possisive singular, we want `editor's` instead


dampee:
Thanks @nathanwoulfe 
Pinging @marcemarc to review  :-)

PR #1267, opened 2018-12-12
Update routing.md

StefanoChiodino:
`urlHelper.GetUmbracoApiService("GetAllProducts")` doesn't use a valid signature

### DIFF
--- a/Reference/Routing/WebApi/routing.md
+++ b/Reference/Routing/WebApi/routing.md
@@ -33,5 +33,4 @@ Generally a UrlHelper instance will be available on most base classes like Contr
 
 	var requestContext = HttpContext.Current.Request.RequestContext;
     var urlHelper = new System.Web.Mvc.UrlHelper(requestContext);
-    var apiControllerUrl = urlHelper.GetUmbracoApiService("GetAllProducts");
-
+    var url = urlHelper.GetUmbracoApiService("GetAllProducts", "Controller");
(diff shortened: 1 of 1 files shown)


Reference/Routing/WebApi/routing.md
```diff
@@ -33,5 +33,4 @@ Generally a UrlHelper instance will be available on most base classes like Contr
 
 	var requestContext = HttpContext.Current.Request.RequestContext;
     var urlHelper = new System.Web.Mvc.UrlHelper(requestContext);
-    var apiControllerUrl = urlHelper.GetUmbracoApiService("GetAllProducts");
-
+    var url = urlHelper.GetUmbracoApiService("GetAllProducts", "Controller");
```
jmayntzhusen:
```suggestion
    var url = urlHelper.GetUmbracoApiService("GetAllProducts", "ProductsApiController");
```
Hey @StefanoChiodino 

You are right, the method takes a second controllername parameter, "Controller" is maybe not the best naming convention - and ProductsApiController was used in the example just above, so would you mind if we changed it to be consistent? 🙂

Thanks for taking the time to update the documentation!

StefanoChiodino: Sure, I didn't see that.

jmayntzhusen:
Thanks again for the update! 🎉😃

Issue #359, opened 2016-07-11
DatabaseServerMessengerOptions needs docs to show how to configure options on startup


Shazwazza:
During startup the `DatabaseServerMessengerOptions` can be tweaked to control how often the load balancing instructions from the database are processed and pruned.

Jeavon:
I can write this, but do you know a use case?

As far as I can tell sample code would be like this:

	namespace BearTheLoad.Logic.ServerRegistrars
	{
		using Umbraco.Core.Sync;

		public class BearTheLoadDatabaseServerMessengerOptions : DatabaseServerMessengerOptions
		{
			public BearTheLoadDatabaseServerMessengerOptions()
			{
				DaysToRetainInstructions = 2; // 2 days
				ThrottleSeconds = 5; // 5 second
				MaxProcessingInstructionCount = 1000;
				PruneThrottleSeconds = 60; // 1 minute
			}
		}
	}

And then in ApplicationStarting:

	ServerMessengerResolver.Current.SetServerMessenger(
		new DatabaseServerMessenger(applicationContext, true, new BearTheLoadDatabaseServerMessengerOptions())
	);

Shazwazza:
You don't need to subclass the options, just set the properties when you create the options. 

If people don't want it checking instructions so often they could adjust here, i guess it depends on how much the site is serving and how often things are being published.

Shazwazza:
Or maybe you know you take servers offline for a few days, I'm which case you'd want to retain instructions longer. 

Lastly, maybe you want to cold boot with less pending Max instructions

Really depends on what your site is doing i guess.

Jeavon:
Cool, I'll write it up.

Code sample something like this then:

	ServerMessengerResolver.Current.SetServerMessenger(
		new DatabaseServerMessenger(
			applicationContext, 
			true, 
			new DatabaseServerMessengerOptions() {
				DaysToRetainInstructions = 2, // 2 days
				ThrottleSeconds = 5, // 5 second
				MaxProcessingInstructionCount = 1000,
				PruneThrottleSeconds = 60 // 1 minute
			}
		)
	);

sofietoft:
Hi @Jeavon,

I'm currently working my way through the issue tracker here on the Docs.
Any chance you can give me a status on this? 😁 
I haven't been able to locate anything about this in the docs, but I might very well be looking the wrong places.

-Sofie

Jeavon:
@sofietoft I will create the docs for this, assigned to myself :)

PR #1270, opened 2018-12-14
Documented Database Messenger Options for issue #359

Jeavon:


Getting-Started/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
```diff
@@ -48,9 +49,10 @@ The first thing to do is create a couple classes for your front-end servers and
 			return null;
 		}
 	}
+```
 
 then you'll need to swap the default `DatabaseServerRegistrar` for the your custom registrars during application startup.
-You'll need to create an [ApplicationEventHandler](/Documentation/Reference/Events/Application-Startup) and override `ApplicationStarting`. 
+You'll need to create an [ApplicationEventHandler](../../../../Reference/Events/Application-Startup.md) and override `ApplicationStarting`.
```
jmayntzhusen:
```suggestion
You'll need to create an [ApplicationEventHandler](../../../../Reference/Events/Application-Startup) and override `ApplicationStarting`. 
```

So when the .md is added to a URL it will find the file, but it will also find it without the .md extension - the bonus is if we some day extend upon that article and move it to a folder with an image folder and call it index.md, then it will find that folder name instead and then automatically look for the index file 😉

Getting-Started/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
```diff
@@ -86,3 +88,31 @@ the [Explicit master scheduling server](#explicit-master-scheduling-server) conf
 Now that your front-end servers are using your custom `FrontEndReadOnlyServerRegistrar` class, they will always be deemed 'Slave' servers and will not 
 attempt any master election or task scheduling and because you are no longer using the default `DatabaseServerRegistrar` they will not try to ping
 the umbracoServer table.
+
+## Controlling how often the load balancing instructions from the database are processed and pruned
+
+During start up the `DatabaseServerMessengerOptions` can be adjusted to control how often the load balancing instructions from the database are processed and pruned.
+
+e.g. This example should be added within a [`ApplicationStarting`](../../../../Reference/Events/Application-Startup.md#startup-methods) event
```
jmayntzhusen: Is this not normally put in ApplicationStarted instead of ApplicationStarting? 🤔

Getting-Started/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
Jeavon: hmm, if we don't have the .md the docs are not navigable on GitHub though which I believe we are trying to maintain as well as via Our?

Getting-Started/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
Jeavon: Not this one, it won't work if added to ApplicationStarted

Jeavon:
@hartviglarsen was taught by the best so I'm sure he's correct

Getting-Started/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
jmayntzhusen: I would argue having it maintainable on Our would be more important than linking between pages on Github, but we can talk about it on our Curators meeting tomorrow 🙂

PR #1271, opened 2018-12-16
Update Member helper methods on UmbracoHelper

marcemarc:
The 'obsolete' versions of MemberHasAccess and IsProtected were displayed, there are now versions when you only have to pass in the path of the node (as that includes the id anyway!)


dampee:
Good one @marcemarc 
I would like to know when it was changed so we can keep it in a "previous version" file...

dampee:
Big hugs @marcemarc   :-)

PR #1272, opened 2018-12-18
Healthcheck info - Missing end of coding block indication '''

marcemarc:
The second code block isn't closed and so includes the next few headings of text, making it hard to read


jmayntzhusen:
Good catch 👍

PR #1275, opened 2018-12-20
added "C#" to the "best results" line

liamlaverty:


jmayntzhusen:
Thanks @liamlaverty 

This should make it a bit clearer! 😁

PR #1278, opened 2018-12-23
Converted code example to use fenced code block and C# syntax highlighting

abjerner:


jmayntzhusen:
Looks good @abjerner 👍

PR #1283, opened 2019-01-03
specifies which button causes the rebuild action

liamlaverty:
Changed a couple of instances where the button to click was ambiguous.


jmayntzhusen:
Hey @liamlaverty 

Thanks for making the documentation more clear, #h5yr! 😄

PR #1312, opened 2019-01-18
Note regarding smtp settings

skttl:
Since 7.13, Umbraco won't send mails regarding user invitations or forgotten passwords if the from attribute of the smtp settings is set to noreply@example.com.

https://github.com/umbraco/Umbraco-CMS/pull/3308


jmayntzhusen:
Hey @skttl 

Thanks for adding this, I'm sure you will prevent many headaches in the future 😄

PR #1313, opened 2019-01-18
Typo ("databse").

Nicholas-Westby:


jmayntzhusen:
Hey @Nicholas-Westby 

Good catch, I was a bit too quick to merge that last one. Thanks for correcting it! 😉

Issue #1266, opened 2018-12-11
Missing syntax highlighting in the documentation


jmayntzhusen:
Our has recently been updated to support syntax highlighting https://github.com/umbraco/OurUmbraco/pull/332 

This means we are missing it everywhere in the documentation, the way to add it is to go in the markdown files and add something like this:
````
```html

<h1>headline</h1>

```
````

The currently supported highlighted languages are csharp, html, json5, xml

To make it easier we can split this task up in the subfolders of the documentation repo:

![2018-12-11_0854](https://user-images.githubusercontent.com/36075913/49786626-c0cb5080-fd24-11e8-9113-b6b5ae3f4dab.png)

- [x] Add-ons
- [x] Cheatsheets
- [x] Contribute
- [x] Development-Guidelines
- [x] Extending
- [ ] Getting-Started
- [ ] Implementation
- [ ] Reference
- [ ] Tutorials
- [ ] Umbraco-Cloud

dampee:
👍

jmayntzhusen:
As pretty much everything has been updated - big thanks to @BatJan! - I will close this issue 🙂🎉

PR #1314, opened 2019-01-18
Align the layout of xml examples

BatJan:
Just discovered that not all XML snippet where tagged with ```xml making the documentation look inconsistent - It should be fixed on all instances now 😃

### DIFF
--- a/Reference/Config/umbracoSettings/index.md
+++ b/Reference/Config/umbracoSettings/index.md
@@ -14,10 +14,11 @@ Below you can see the settings that affect the Umbraco backoffice (v7.8+)
 ### Tours
 
 The section is used for controlling the backoffice tours functionality.
-
-    <backOffice>
-        <tours enable="true"></tours>
-    </backOffice>
+```xml
+<backOffice>
+    <tours enable="true"></tours>
+</backOffice>
+```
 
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hey @BatJan 

Thanks for helping out with this, we are actually missing it for csharp, javascript and html some places too: https://github.com/umbraco/UmbracoDocs/issues/1266 😉

BatJan:
@jmayntzhusen Uh :) I'll be happy to help out with that too - Still many pages to update. Should I do one big PR instead? Or is it better to keep doing it the way I've been doing it so far? I mean some pages have also had some warnings or notes added and if it's all in one big PR it might be harder to approve and merge? My OCD can't handle the syntax highlighting being out of sync 😄  So will be more than happy to go through as many pages as I can 👍

jmayntzhusen:
That would be amazing! I just wen't through section by section, which is also why some sections are ticked on that issue, you can create 1 PR per section so it's a bit easier to manage 😃 

Adding warnings and notes are fine, if I or one of the other curators disagree with something we can just leave a review and take it from there 😉

BatJan:
Ok, I'll do a batch now then 👍

PR #1315, opened 2019-01-18
Use the XML tag for XML snippets

BatJan:
I noticed the XML snippet where still using the old code markup instead of the new and shiny XML formatting so I updated the code examples to make use of it 😃


jmayntzhusen:
Thanks #h5yr!

PR #1316, opened 2019-01-18
Format the code snippet using the XML tag

BatJan:
Updated the code example to make use of the XML tag / Snippet 😃


jmayntzhusen:
Thanks #h5yr! 😄

PR #1322, opened 2019-01-19
Update code formatting and add version and a warning

BatJan:
I have added some nicer code formatting for this piece of documentation and I have also added a warning about this piece of documentation only being relevant for Umbraco 6. And I have also added the YAML meta data about the versions.

Even thought the /Base concept has been around since... v3 I think it seems that the way to use /base changed in v6 where the restExtensions.config was renamed to baseRestExtensions.config also containing another syntax, which is the one referenced in this piece of documentation.

### DIFF
--- a/Reference/Config/BaseRestExtensions/index.md
+++ b/Reference/Config/BaseRestExtensions/index.md
@@ -1,15 +1,24 @@
+---
+keywords: base v6 version6
+versionFrom: 6.0.0
+---
+
+:::warning
+If you're using Umbraco 7 you should consider using Umbraco Web API instead of /Base since it's deprecated and only kept for backwards compatibility. Please see the documentation on Umbraco Web API here: https://our.umbraco.com/documentation/Reference/Routing/WebApi/
+:::
+
 # BaseRestExtension.config
 
 BaseRestExtension.config contains the data necessary for the /Base system when exposing the methods in your class library.
 
-	<?xml version="1.0" encoding="utf-8"?>
-	<BaseRestExtensions>
-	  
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hmm so adding a versionFrom YAML tag here without having a second file with a different version would end up a bit weird on Our I think. Maybe we should add a second index file that has versionFrom 7.0.0 that has that warning and then remove the warning from this one, what do you think? 🤔

BatJan:
Yes that is probably better - Then the only thing that will be displayed for v7+ is the warning stating that Web API is the way to go these days. But I guess that means I'll need to do this PR another way since I don't know if it's possible to actually add new files using the Web UI? :)

dampee:
@BatJan You actually can. Go to [your branch on your own repo](https://github.com/BatJan/UmbracoDocs/tree/patch-20/Reference/Searching/Examine) an click the "create new file" on top.  If you want some explanation on how to work with a git client and vscode, i'll glad to help you out!

@jmayntzhusen  having only a `versionFrom` is no issue.  Because it includes the current version.

Therefore I'll merge this (and fix if needed...)

BatJan:
@dampee Ah I did not know that - Thanks! But I know my way the other way around, no problem. It's just that using the Web UI is sometimes faster when doing documentation. But thank you anyway 😃 

But judging from your comment to @jmayntzhusen it seems like there is no reason to create a specific v7+ version? 💭

jmayntzhusen:
@dampee I just seemed to remember it only showing the versioning box if there is more than 1 version? It also seems to not be present on the page: https://our.umbraco.com/documentation/Reference/Config/BaseRestExtensions/

PR #1325, opened 2019-01-19
XML formatting code and fixing minor glitch

BatJan:
Apart from formatting the XML code snippets I also fixed "area" and wrapped it in <> so it reads <area> - It was already marked as code but was not properly formatted.


jmayntzhusen:
Thanks #h5yr! 😄

PR #1337, opened 2019-01-19
Markup higlighted text as a note

BatJan:
I considered using a tip instead but I think this is more like a note about the models builder project being an open source project on it's own more than a tip about something that can improve your code or what not.


jmayntzhusen:
I think often tip and note could be used interchangeably. Note is definitely fine in this regard 😉

PR #1361, opened 2019-01-21
Markup code samples as XML

BatJan:
I'm wondering if perhaps this needs to be versioned? I don't know if the XSLT example works in Umbraco 7.13 these days? But anyhow we might as well prepare for the release of v8 where this way will no longer be supported...right? 😄


jmayntzhusen:
Yea if it is removed in v8 we should at least make a v8 version saying it won't work 😄 is that something you would like to do?
If in doubt how the versioning is set up you can see an example here: https://github.com/umbraco/UmbracoDocs/commit/ca1b4294aa1bb217f42aa5d1bf3fe1e045279471#diff-f1a1f94d33ffff4a22ae3c44674ff157

dampee:
@BatJan how sure are you that this is not working in v7 anymore?

BatJan:
@dampee 97.32% 😁 - I checked the /config folder in a v7.0.0 build and was not able to find the xsltExtensions.config file where the changes need to be made.

Even though XSLT is still supported in v7 the extensions stuff changed so I don't think the documentation is relevant for v7 unsless I'm missing something obvious? 😃

greystate:
Got tagged. Will have a look :)

greystate:
So - Umbraco 7 did change the way XSLT Extensions were registered. Instead of using the `xsltExtensions.config` file you add an `[XsltExtension]` attribute in the code for the extension.

A blank "Clean" XSLT file in Umbraco 7 already includes the Examine namespace, so it's actually a bit simpler in v7 because you don't need to perform the initial setup steps 1 - 3.

Reference/Searching/Examine/xslt-example.md
```diff
@@ -1,31 +1,45 @@
...
+keywords: examine indexing v4.5 v6
+versionFrom: 4.5.0
+versionTo: 6.2.6
+---
+
 # Using Examine with XSLT
 
 The XSLT implementation of Examine requires a little bit of setup. The steps are outlined below.
 
 1. Add the Examine xslt extensions to the xslt config in file ~/Config/XsltExtensions.config
```
greystate: If I may, can I suggest fixing the casing while we're at it? 😁 
E.g.:

```
1. Add the Examine XSLT extensions to the config in `~/Config/XsltExtensions.config`
```

Reference/Searching/Examine/xslt-example.md
```diff
@@ -1,31 +1,45 @@
...
+```
 
 3. And then ensure that you add "Examine" to the exclude-result-prefixes value, so the attribute might look something like:
 
-    `exclude-result-prefixes="Examine umbraco.library Exslt.ExsltCommon Exslt.ExsltDatesAndTimes Exslt.ExsltMath Exslt.ExsltRegularExpressions Exslt.ExsltStrings Exslt.ExsltSets ">`
+```xml
+exclude-result-prefixes="Examine umbraco.library Exslt.ExsltCommon Exslt.ExsltDatesAndTimes Exslt.ExsltMath Exslt.ExsltRegularExpressions Exslt.ExsltStrings Exslt.ExsltSets ">
+```
 
 Now we can use the Examine xslt extensions in our xslt. This will render the exact same results as the above 3 examples:
```
greystate: Here as well:

```
Now we can use the Examine XSLT extensions in our XSLT. This will render the exact same results as the above 3 examples:
```

Reference/Searching/Examine/xslt-example.md
```diff
@@ -1,31 +1,45 @@
...
 The XSLT implementation of Examine requires a little bit of setup. The steps are outlined below.
 
 1. Add the Examine xslt extensions to the xslt config in file ~/Config/XsltExtensions.config
 
-    `<ext assembly="UmbracoExamine" type="UmbracoExamine.XsltExtensions" alias="Examine" />`
+```xml
+<ext assembly="UmbracoExamine" type="UmbracoExamine.XsltExtensions" alias="Examine" />
+```
 
 2. In your Xslt for your macro ensure to add the namespace:
```
greystate: Also:

```
2. In the XSLT for your macro ensure to add the namespace:
```

dampee:
Thanks @greystate 

@BatJan I suggest we add a new file for v7 on how this works.  What do you think?

BatJan:
@dampee Yep that's probably a good idea.

@greystate Did I fix the casing correctly? I got confused by the "Resolved dialog" 😀

dampee:
The only thing to add to merge this pull request is a page that we should use [XsltExtention] in v7. 

Pretty sure it won't work anymore in v8. But we would need to test.

BatJan:
@dampee In V8 there is no XSLT support at all if I'm not mistaken I believe it should be fine. I'll look into adding the v7 stuff now

BatJan:
Ok, so I have had some trouble with the commits since my user was not setup correctly - So now the history seems a bit broken...Not sure how to properly fix that... and I'm not sure what the issue is with the change request? :)

However I have added a version for pre v 7 and then modified the original file to be about v7 - However I'm not sure where to include the part with the `[XsltExtension]` that you mentioned @greystate ? Also I feel like the XSLT example should perhaps be displayed in full since the namespaces for examine are added by default?

greystate:
> However I'm not sure where to include the part with the `[XsltExtension]` that you mentioned @greystate ? Also I feel like the XSLT example should perhaps be displayed in full since the namespaces for examine are added by default?

I don't think we should mention the `[XsltExtension]` attribute - it's only something you need to know when building your own custom-extensions.

dampee:
@greystate True.  We could add it, but we don't nee to leave this PR open for it.  Thanks for your input.

And to  @BatJan thanks a lot for your effort(s) and patience!

PR #1419, opened 2019-01-23
Adding relevant code syntaxes and notes throughout the page

BatJan:
Phew, this was a big one! I think I caught it all so things should be marked up correctly using css, js, json and html markup. Also converted all things marked at "note" in bold and converted it to :::note


sofietoft:
Quick comment on this one:
There are few places, where there are a couple of notes right after each other.
Wouldn't it be nicer to just merge those together?

So, in line 90 to 100 there are three seperate notes.
Could we make that just one note? 🤔

And the same in line 367 to 373?
Just a thought!

/Sofie

BatJan:
Yes, you're absolutely right! I'll go and fix it now 👍

PR #1451, opened 2019-01-23
Add C# syntax to code examples - And wondering if...

BatJan:
we also have a syntax notation for xpath? 🤔 😃

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Multinode-Treepicker.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Multinode-Treepicker.md
@@ -47,38 +47,44 @@ Enter `typeAlias,altTypeAlias` to only allow selecting nodes with those alias'. 
 
 ### Typed:
 
-    @{
-        var typedMultiNodeTreePicker = Model.Content.GetPropertyValue<IEnumerable<IPublishedContent>>("banner");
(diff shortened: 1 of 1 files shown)


sofietoft:
Good question 🤔 

Think we need @dampee or @jmayntzhusen on this one?

sofietoft:
Good question 🤔 

Think we need @dampee or @jmayntzhusen on this one?

BatJan:
So it's actually something we need to ask @abjerner about maybe? 😃

abjerner:
@BatJan nope - no support for XPath at the moment. Got an example you wish to get highlighted?

BatJan:
@abjerner Hmm, just checked this file and I must have been tired when I wrote it since this does not contain any actual xpath.... 🙄 ... I'll let you know when if/when I find a relevant example 👼

abjerner:
@BatJan no problem. I haven't found any good examples yet - probably because XPath isn't a file format on it's own. GitHub doesn't seem to support this either.

But if there are XPath examples in the documentation (or it will make sense to have them in the future), I can have a look at adding support for it. Although I'm not an XPath expert, I think it will be somewhat easy to add highlighting support for it 😉

BatJan:
@abjerner Yeah, that's true of course... So perhaps it's actually ok to use either the csharp or xml notation for those instead - Nevermind me! 😉

PR #1465, opened 2019-01-23
Add note syntax to note and wondering...

BatJan:
if we should remove the "Coming soon" text? I doubt people will want to spend time documenting this with v8 coming up? 😃


sofietoft:
Hi Jan,
Amazing job with all these PRs 😄 🎉 

And good point about the 'Coming soon' text.
I see no problem in removing that!

/Sofie

BatJan:
Thank you - My OCD can't handle the differences in the syntax highlightning 😁 😉 

But wow there are many pages to go through! 😄 

"Coming soon" has been removed.

PR #1492, opened 2019-01-24
Add C# syntax to code examples

BatJan:


Reference/using-ioc.md
```diff
@@ -17,128 +17,140 @@ This example will setup Autofac to work with Umbraco (see [their documentation](
...
+    {
+    }
+}
+```
 
-		public HomeController(MyAwesomeContext myAwesome)
-		{
-			_myAwesome = myAwesome;
-		}
+In this example, we will assume that we have a Document Type called 'Home'. Now we're going to create a custom controller to hijack a route for all content pages of type Home *(NOTE: we can target custom template names too - see the [Hijacking routes](routing/custom-controllers) documentation for full details).* Notice that the constructor accepts a parameter the custom class; this will be injected via IoC.
```
sofietoft: You missed a note 😱 😉

BatJan: argh! I thought I was flawless... down to 99.991% 😢 

😁  - It's been fixed now 👍

PR #1519, opened 2019-01-26
Update Tags-vpost-7.4.md

marcemarc:
Not sure if this is the correct place, but we don't have anything on Setting Tags via ContentService and there is a nice extension method, particularly with the options to store as json/csv

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags-vpost-7.4.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags-vpost-7.4.md
@@ -5,6 +5,7 @@ versionTo: 7.5.14
 ---
 
 # Tags
+
 `Alias: Umbraco.Tags`
 
 `Returns: CSV` or `JSON`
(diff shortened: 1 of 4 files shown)


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags-vpost-7.4.md
```diff
@@ -80,3 +80,38 @@ Notice that Newtonsoft.Json is referenced in the below example. It's already a p
     </ul>
 }
 ```
+### Setting Tags Programatically:
+You can use the ContentService to create and update Umbraco content from c# code, when setting tags there is an extension method (SetTags) on IContentBase that helps you set the value for a Tags properties. Remember to add the using statement for `Umbraco.Core.Models` to take advantage of it.
```
jmayntzhusen:
```suggestion
You can use the ContentService to create and update Umbraco content from c# code, when setting tags there is an extension method (SetTags) on IContentBase that helps you set the value for a Tags property. Remember to add the using statement for `Umbraco.Core.Models` to take advantage of it.
```

marcemarc: Yes, it should be in the other two, I was editing online so I'm not sure how it editing this version - the extension method has been there for always - it's been renamed in V8 as AssignTags, if we can add that to a list somewhere!

PR #1521, opened 2019-01-26
Adding Methods to Content Service.

crgrieve:
Adding methods to make it more consistent with other documentation such as Media Service.
Info taken from API docs.


dampee:
Hi @crgrieve 
The reason why your PR is not merged, is that we would like to get rid of all the methods.  Auto generated code should not be a manual process.
If there is an addition, like examples or an explanation we like to reconsider.  But for now I think we'll decided on closing this without merging.  

what do you think?

jmayntzhusen:
As there has been no reply for a while I will close this 🙂

PR #1527, opened 2019-01-29
First property editor versioned for v8

jmayntzhusen:
With v8 coming out "soon", we should update the for example the built-in property editors to work with v8, this is the first one and will create an issue for the rest 🙂


jmayntzhusen:
The full list is here: #1528

PR #1530, opened 2019-01-30
Changed the number of custom hostnames to 15

FransdeJong:
I changed the number of custom hostnames to 15 instead of 20 since we are not able to add more than 15 in the cloud interface.

Taken from cloud:
Manage hostnames
You can add up to 15 custom hostnames for each environment.


jmayntzhusen:
Good catch, you are absolutely right! 😃

I see this is your first contribution to Umbraco, do you have a profile on our.umbraco.com? We give all contributors a contribution badge there 😉

FransdeJong:
Hi, 

That was fast! My profile on Our is @FranscdeJong

[](https://our.umbraco.com/members/id:119542/[](url))

PR #1532, opened 2019-01-31
fixed contributing link

prjseal:


jmayntzhusen:
Another good catch! 😄

PR #1536, opened 2019-02-06
ContentApps: fixes C# code and updates for 8.0.0-alpha.58.2091

PerplexDaniel:
Example code did not build as it contained a nested switch statement among other things, appears to be caused by unintentional duplicated code.
It also referenced UmbracoComponentBase and IUmbracoUserComponent which are no longer present in Umbraco v8 as it seems.
Updated code was tested in 8.0.0-alpha.58.2091.


dampee:
Hi @PerplexDaniel 
That was indeed a bad merge in the past!  Thanks for the catch.
This was your first contribution to the docs, but you have already a c-trib badge.  Thanks a lot for your effort!
Kind regards

PR #1555, opened 2019-02-17
Update errorpage info in umbracoSettings.config

greystate:
Adding docs on using GUIDs and XPaths, which are both better ways if using multiple environments on Umbraco Cloud.


sofietoft:
This looks really great @greystate !

I'll ping or developers and have them give it a quick review 💪

marcemarc:
this looks good to me and is good advice for when the ids change between environments, so have merged for now!

PR #1571, opened 2019-02-20
Forms V8

bergmania:
Added V8 docs for Forms. 
Some of the images are added in a V8 form -but not all. The overall meaning should still make sense.

Some documentation is removed from v8, because is not available anymore. 
- Auto updates from UI is removed, because it should not be available for editors to do the update
- XPath stuff removed.


Add-ons/UmbracoForms/Installation/index-v8.md
```diff
@@ -0,0 +1,13 @@
+# Installation and upgrading
+
+## [Extending Umbraco with the Forms section](Install)
+Extend Umbraco with the form builder section in a few steps  
+
+## [Configuring a license](../../The-Licensing-model)
+Getting Forms out of trial mode
+
```
sofietoft: @bergmania 
I noticed the **Keeping Forms up to date** is not on this page.
So, it's correct that you will not be able to upgrade Forms directly from the backoffice in Umbraco 8?

bergmania: @sofietoft, Yes. Thats also mentioned in the PR message :

> * Auto updates from UI is removed, because it should not be available for editors to do the update

PR #1583, opened 2019-02-24
Rejigging Compositions

marcemarc:
Hopefully keeping the same information, but in a different order and different words

This is really complicated, but only because it's new and the words for the different elements, are so similar and are abstract concepts. So I've tried early on to give a gist of the concept and the terms, so when they are explained later they are more familiar, rather than just another word beginning with C that I don't quite get! I expect it could be reworked further!


warrenbuckley:
Thanks @marcemarc reads much better than what I had done. Kudos !

Issue #1528, opened 2019-01-29
Update Property Editors for v8


jmayntzhusen:
## What type of issue is it? (Choose one - delete the others)
Missing documentation

## What article/section is this about? 
https://our.umbraco.com/documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/ 

## Describe the issue:
With v8 coming out later this year we should update Property editors to be have up to date examples for v8 🙂

- [x] Checkbox (Umbraco.TrueFalse)
- [x] Checkbox list (Umbraco.CheckBoxList)
- [x] Color Picker (Umbraco.ColorPicker)
- [x] Content Picker (Umbraco.ContentPicker)
- [x] Date (Umbraco.Date)
- [x] Date/Time (Umbraco.DateTime)
- [x] Decimal (Umbraco.Decimal)
- [x] Dropdown (Umbraco.DropDown.Flexible)
- [x] Email address (Umbraco.EmailAddress)
- [x] File upload (Umbraco.UploadField)
- [x] Grid layout (Umbraco.Grid)
- [x] Image Cropper (Umbraco.ImageCropper)
- [x] Label (Umbraco.NoEdit)
- [x] List view (Umbraco.ListView)
- [x] Markdown editor (Umbraco.MarkdownEditor)
- [x] Media Picker (Umbraco.MediaPicker)
- [x] Member Group Picker (Umbraco.MemberGroupPicker)
- [x] Member Picker (Umbraco.MemberPicker)
- [x] Multi Url Picker (Umbraco.MultiUrlPicker)
- [x] Multinode Treepicker (Umbraco.MultiNodeTreePicker)
- [x] Nested Content (Umbraco.NestedContent)
- [x] Numeric (Umbraco.Integer)
- [x] Radio button list (Umbraco.RadioButtonList)
- [x] Repeatable textstrings (Umbraco.MultipleTextstring)
- [x] Rich Text Editor (Umbraco.TinyMCEv3)
- [x] Slider (Umbraco.Slider)
- [x] Tags (Umbraco.Tags)
- [x] Textarea (Umbraco.TextArea)
- [x] Textbox (Umbraco.TextBox)
- [x] User picker (Umbraco.UserPicker)

Adolfi:
PR for the label property:  https://github.com/umbraco/UmbracoDocs/pull/2113

bgunnarsson:
PR for EmailAddress: #2762

renaa:
[...]

PR #1584, opened 2019-02-25
Second property editor versioned for v8

elit0451:
Updating the built-in property editors to work with v8, part 2.

child of: #1528


jmayntzhusen:
This looks great 🙂 Thanks for helping out @elit0451 ! 🎉

PR #1594, opened 2019-02-28
Update Application Event Handler info for v8

marcemarc:
this is a weird one as we have a page called ApplicationEventHandler, but that class now that no longer exists, so weird to keep the url of that page and not mention ApplicationEventHandler - makes me think this should perhaps say this approach is retired in V8 and redirect to the Composing page, but I've updated the text to reflect the new way in this PR - so we can discuss


marcemarc:
We should just have a page that says application event handler has been removed in V8 - and refer to composing page

PR #1597, opened 2019-03-01
Patch/v8 sectiontrees

marcemarc:
Was going to update info on ISearchableTree, but in order to get there tried to create a Custom Tree in V8 - so instead I've updated the info on creating a custom tree!! and sorted out the section a little bit, as we had a landing page to divert people to v6 & v7 tree info pre 'versionable' documentations - so now v6 and v7 pages are just the tree information pages for those versions, and index.md is now just the info for v8.

Have referenced the core source code, to fathom how I think it now works, hopefully it makes sense (different types of adding custom ActionMenuItem's was a bit vague on) , I've gone through the examples, and gotten the custom favourite things tree working, and wired up to the events, and some custom controllers for the menu items, all seems to work.


Extending/Section-Trees/trees.md
```diff
@@ -1,31 +1,303 @@
...
+and
+* GetMenuForNode (returns a *MenuItemCollection*) - Responsible for returning the menu structure to use for a particular node within a tree.
+
+Decorate your '*TreeController*' with the *Tree* Attribute, which is used to define the name of the section the Tree should be loaded in, which 'Tree Group' it should belong to and also define an alias and title for your custom tree.
+
+
+**For example**
+
+```csharp
+[Tree("settings", "favouriteThingsAlias", TreeTitle = "Favourite Things Name", TreeGroup="favouritesGroup", SortOrder=5]
```
LottePitcher: Was giving this a try, and it's missing the closing ")"

marcemarc: Thanks @LottePitcher ! - have added it in... 

Did everything else work ok, based on the info above?

LottePitcher: Yes, got your “favourite things” tree appearing. However:

1) I wanted a single node in the tree and couldn’t work out how to hook up that to a view (in your example only child items hook up to views) 

2) It seemed that after I built again the favourite things tree disappeared (as did my one) when I reloaded back office

But it was quite late at night. So was going to wait until I’d tried again before I reported anything!

marcemarc: cool, did you have: IsSingleNodeTree = true in your Tree attribute settings?

also Warren's LogViewerTreeController only has a single node, maybe that will give a hint:

https://github.com/umbraco/Umbraco-CMS/blob/91c52cffc8b7c70dc956f6c6610460be2d1adc51/src/Umbraco.Web/Trees/LogViewerTreeController.cs

he has this line in the CreateRootNode that sets the custom path

 root.RoutePath = string.Format("{0}/{1}/{2}", Constants.Applications.Settings, Constants.Trees.LogViewer, "overview");

LottePitcher: Got that all working, thank you for pointing me in the right direction! My observations:

I don't think the IsSingleNodeTree attribute was needed here. It's not specified on the LogViewer or Languages core single node tree items. It is however specified on the Packages and Users tree.  Both those sections do not display a side tree pane. So, assuming that this flag is used to hide the tree pane, I removed the attribute in my fork of umbraco and indeed, the Packages section did then show the left tree pane. I think the key thing is that root.HasChildren = false, and your example CreateRootNode already has a comment about setting that to false if a single node.

So a suggestion for that new section that you added:

Change heading to be 'Single Node Trees / Customising the Root Node Action' with something along the lines of:

By default clicking the root node will display the dashboard of the current section.  

If you want a 'tree' consisting of only a single node - perhaps just to provide an area to control some settings or a placeholder for a single page backoffice app (see the LogViewer in the settings section for a good example) - then you will want to override this behaviour. Or perhaps you want to display an 'introduction' page to your tree (as happens with Content Templates in the settings section)

Overriding the CreateRootNode method allows you to set the 'RoutePath' to where your single page application / help text lives. This should be in the format of section/treeAlias/method. For example:

	root.RoutePath = $"{Constants.Applications.Settings}/favouriteThingsAlias/overview";

As our example controller uses the PluginController attribute, clicking the root node would now request /App_Plugins/favouriteThings/backoffice/favouriteThingsAlias/overview.html. If you are not using the PluginController attribute, then the request would be to /umbraco/views/favouriteThingsAlias/overview.html

marcemarc: @LottePitcher cool, think we got there in the end, IsSingleNodeTree is a very misleading attribute :-) anyway - have updated to reflect the info, I think, I'm trying not to use 'You' and 'Your' in the text for documentation pages that aren't tutorials... I'm finding it really difficult.

LottePitcher: It was very timely of you to have been working on this, so thank you very much.  
Noted re avoiding 'you', writing words is tricky!

PR #1599, opened 2019-03-02
Update Manual Installation - MySql V8

marcemarc:
Support for MySql - the popular open source Sql database management system has been dropped in V8 - https://github.com/umbraco/Umbraco-CMS/pull/4143


jmayntzhusen:
Hey @marcemarc 

As this guide shows you how to install Umbraco using Webmatrix which is deprecated I'm thinking we should instead remove it entirely as we already have other guides showing how to install Umbraco via NuGet and as a zip using VSCode and IIS Express. Do you think anyone would miss this Webmatrix guide?

marcemarc:
@jmayntzhusen yep go for it, it was a random update, after a forum post about MySql support.

PR #1610, opened 2019-03-10
Updating Route Hijacking Custom Controller information

marcemarc:
Need to double check the examples are correct


marcemarc:
Have checked the examples, and added a bit on controller injection

PR #1611, opened 2019-03-11
Update ActiveDirectory users

drpeck:
I worked this change out with a bit of trial and error. Works on my computer, and I can't see this info anywhere else.


sofietoft:
Hi @drpeck 

Thanks for this, it looks good!
We have versioning on the documentation, so instead of simply overwriting the existing article, it would be great to have 2 separate articles for each version.

Would you be up for making this change? (have a look at this for reference: https://our.umbraco.com/documentation/Contribute/file-naming-conventions) 
Otherwise, I'd be happy to do it.

/Sofie

dampee:
Hi @drpeck 

I've fixed it, so this is all fine to merge in now.
Thanks a lot for your effort, it'll certainly help other people!

PR #1614, opened 2019-03-12
Fix name of Models Builder dashboard name

bjarnef:


marcemarc:
Thanks @bjarnef !! good spot

PR #1619, opened 2019-03-16
V8 - Update EditorModelEvents

marcemarc:
need to check examples work

(when we switch the automated generated docs to be for V8 we should update the overview of model sending classes here to instead point to autogenerated definitions


marcemarc:
Updated examples

PR #1638, opened 2019-03-26
Fix heading

StefanoChiodino:


jmayntzhusen:
Nice catch, thanks for taking the time to update it! 🙂

PR #1642, opened 2019-03-27
Dashboard tutorial v8

jmayntzhusen:
This is a WIP to convert the old tutorial to v8.

Currently hit a stopping point in the getUserLog method, seems like the audit table isn't logging that much in v8? I get nothing at least.


jmayntzhusen:
Closing as Github is for some reason including a bunch of PRs that are merged already in this.

PR #1643, opened 2019-03-27
Update index.md

ProNotion:
Grammar correction


jmayntzhusen:
Good catch, thanks! 🙂

PR #1645, opened 2019-03-28
Extract additional properties

bjarnef:
Just adding some additional documentation on how to extract addtional properties as show in the screenshots.
https://our.umbraco.com/documentation/getting-started/backoffice/Property-Editors/Built-in-Property-Editors/Nested-Content/

PR #1648, opened 2019-03-29
Tiny update

bjarnef:
... to previous PR https://github.com/umbraco/UmbracoDocs/pull/1645

We are already inside the code block 😁


jmayntzhusen:
Ah good catch 😁

PR #1649, opened 2019-03-29
Update IPublishedContentDocs 

rhyshamilton:
I split the IPublishedContent documentation into the appropriate V7 and V8 index files (minor change). This just helps to prevent old V7 docs being overwritten, whilst setting the V8 doc as the main page for this section. The V8 doc still requires more updates, but it splits things up a bit.


Reference/Querying/IPublishedContent/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-versionFrom: 7.0.0
+versionFrom: 8.0.0
 needsV8Update: "true"
```
jmayntzhusen:
```suggestion
```
This yaml tag is for docs pages that needs to be verified they work for v8. If we have a v8 version then it will no longer be needed 🙂

Reference/Querying/IPublishedContent/index -v7.md
```diff
@@ -0,0 +1,32 @@
+---
+versionFrom: 7.0.0
+needsV8Update: "true"
```
jmayntzhusen:
```suggestion
```
This yaml tag is for docs pages that needs to be verified they work for v8. If we have a v8 version then it will no longer be needed 🙂

rhyshamilton:
Thanks for the feedback on this! :) I've resolved the typo issue, as there shouldn't have been a space in the filename to begin with.

jmayntzhusen:
Awesome, thanks for helping out with the documentation! #h5yr! 🎉😄

PR #1651, opened 2019-04-01
Update install guide for v8

### LINKED ISSUE #1647: Update install guide with important project naming caveat.
The documentation on the install guide should be updated to say that it is important not to name the VS solution as "Umbraco". This causes a critical failure due to namespace conflict with the CMS itself.

People shouldn't have to go to stackoverflow to resolve a critical failure despite following the install guide correctly. A one liner in the install guide would suffice.

https://stackoverflow.com/questions/25461798/umbraco-7-could-not-load-type-umbraco-web-umbracoapplication

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/master/Getting-Started/Setup/Install/install-umbraco-with-nuget.md***

rhyshamilton:
Resolves #1647

### DIFF
--- a/Getting-Started/Setup/Install/install-umbraco-with-nuget-v7.md
+++ b/Getting-Started/Setup/Install/install-umbraco-with-nuget-v7.md
@@ -51,6 +51,8 @@ It's important to pick only one of those as other templates cause errors beyond 
 ### Visual Studio 2013/2015/2017
 Go to **File > New Project** and pick an ASP.NET Web Application.    
 
+**Note:** Refrain from naming your solution `Umbraco`, as this will cause a namespace conflict with the CMS itself.
+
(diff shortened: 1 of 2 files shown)


jmayntzhusen:
Hey @rhyshamilton 

Thanks for the fix! I think this same issue applies on v7 though, so to completely resolve the issue we will need to have it changed in the [v7 version of the article](https://github.com/umbraco/UmbracoDocs/blob/master/Getting-Started/Setup/Install/install-umbraco-with-nuget-v7.md) as well 🙂

rhyshamilton:
Thanks for the feedback :) I've updated the guide on v7 as well now, so hopefully that should resolve things.

jmayntzhusen:
Very nice, thanks for the contribution! 👍🙂

PR #1659, opened 2019-04-07
Update example for setting a custom DefaultRenderMvcController in V8

marcemarc:


Implementation/Default-Routing/Controller-Selection/index.md
```diff
@@ -15,20 +14,28 @@ The MVC Action that executes by default for every request is the `Index` action
...
+        {
+            composition.SetDefaultRenderMvcController<CustomMvcController>();
+        }
     }
 }
 ```
 
 It is a requirement that your custom controller inherit from `Umbraco.Web.MVC.RenderMvcController`.
-You can override the `Index` method to perform any customizations that you require.
+You can override the `Index` method to perform any customisations that you require.
```
Jeavon: customizations?

marcemarc: I don't know why, but I can't bear the use of the z - US English is probably the standard I know, but there is a bloody z in there, and it's just not right :-P

PR #1661, opened 2019-04-08
Incorrect reference in example

darren-welch:
I followed this guide and saw it had the wrong name for the Custom Component.


marcemarc:
good spot! thanks for updating!

PR #1666, opened 2019-04-10
Change "user group" to "user"

BatJan:
A colleague of mine pointed out that the text should probably read "...administered for the specific user, but rather for the user group(s)..." rather than "...administered for the specific user group, but rather for the user group(s)..." since permissions are now administered on the group level and not the user level, right? :-)


jmayntzhusen:
Yep that makes total sense! You should invite that colleague to contribute as well 😉

Thanks for the update to both of you though! 😄

BatJan:
It's not like I'm not trying, believe me 😄 Hopefully they'll start doing it automatically themselves one fine day 🙏

jmayntzhusen:
Thanks for trying then 😁👍 #h5yr!

PR #1686, opened 2019-04-30
readme.md grammar mistake

kcharecki:
Indefinite pronouns that end in -one are always singular


jmayntzhusen:
Good catch, thanks for updating! 🙂

PR #1691, opened 2019-05-01
v8 additions, clarified v7 specifics

dangrussell:


sofietoft:
Hi @dangrussell 

This is good stuff! 💪 
Would you mind taking a crack at splitting the article up in two? One for Umbraco 7 and one for Umbraco 8?
We're using YAML tags and file names for the versioning. You can read more about the details in the [Contributing](https://our.umbraco.com/documentation/Contribute/file-naming-conventions) section.

I'd also be happy to do it as well. Just let me know 😁 
/Sofie

PR #1698, opened 2019-05-03
Updated the docs about IIS logging on UC

Aaen:
I added a bit more about where to find it. I had a client that thinks it was in Kudu. 
Also added the part to make people aware that the environment will restart


jmayntzhusen:
Thanks @Aaen 👍

PR #1702, opened 2019-05-09
Add link to reference (more on tags)

TheKevin-J:
I found the page lacking a link to the UmbracoHelper reference where you can learn more on querying and working with the tags. This makes the current page frustrating because it shows up in search more prominently but misses this information.

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Tags/index.md
@@ -98,3 +98,7 @@ namespace Our.Documentation.Examples.Controllers
     }
 }
 ```
+
+### More on working with Tags
+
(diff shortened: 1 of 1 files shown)


dampee:
Hi @TheKevin-J 

Thanks a lot for your contribution.  You are right.  Finding information that if it is not linked can be very hard.  

I replaced the link with a relative path as we do not want to hardcode the hostname in the documentation.

What is your profile on http://our.umbraco.org so I can make sure we add a contributor tag?  

Kind regards
Damiaan

esn303:
Hey @dampee and @TheKevin-J, would an absolute link be ok (omitting the host) like so: /Documentation/Reference/Querying/UmbracoHelper/#working-with-tags

Because right now the new link in the updated docs leads to a 404 ... 😞 
[https://our.umbraco.com/Reference/Querying/UmbracoHelper/#working-with-tags]

PR #1703, opened 2019-05-09
Update GetLogger example so that it works

Jeavon:
The existing example only worked with some pre-release versions of v8. GetLogger was moved into CoreRuntime before v8.0 was released


jmayntzhusen:
@warrenbuckley agreed that the namespace changed, and the example looks fine, haven't tested but I guess you did? If so we can just merge 🙂

Issue #1685, opened 2019-04-30
Missing Examine docs for v8


jmayntzhusen:
## What type of issue is it? (Choose one - delete the others)
Missing documentation

## What article/section is this about? 
https://our.umbraco.com/documentation/Reference/Searching/Examine/

## Describe the issue:
The way Examine works has drastically changed for v8. This issue is going to be a sort of overview of what has and will be documented and will be a place for anyone using v8 to report anything they feel is missing from the Examine documentation so we can try to include it all.

## Finished documentation for Examine in v8
- [Examine quickstart](https://our.umbraco.com/documentation/Reference/Searching/Examine/quick-start/) - _shows how to set up a search on a template and perform different kinds of searches on an index_

## Upcoming documentation for Examine in v8
- [x] IndexProviderTransformingIndexValues - _the v8 equivalent to GatheringNodeData, a way to combine fields into one to search through all fields on all nodes at once_
- [x] Creating a new index 
- [ ] Setting up an IndexPopulator - _used to be ISimpleDataService_
- [ ]  add headers (see comment below)

Please feel free to post anything else here you feel should be documented for Examine and we will do what we can to provide documentation for it. As always this is open source, so anyone comfortable pitching in would be greatly appreciated!

References:
- There is a nice [Web Demo](https://github.com/Shazwazza/Examine/tree/v1.0.0/src/Examine.Web.Demo) of Examine on the Github repo with examples to follow.

dawoe:
We use examine a lot to also store data from external systems. : http://sleslie.me/2016/indexing-external-data-using-examineumbraco/

I think the ISimpleDataService does not exist anymore. Would love to see a example of this, because this is used quite often.

Dave

dampee:
The page should be cleaned up to like @Shazwazza mentioned here: https://github.com/umbraco/UmbracoDocs/pull/1994#issuecomment-549590894

sofietoft:
Hi @jmayntzhusen !
Do you know the status of the Examine Documentation?
Is still issue still valid, or can it be closed? 😁

jmayntzhusen:
Ah this completely slipped my mind. Last I checked there was a bug prohibiting me from completing the IndexPopulator documentation, but it was fixed in 8.4 (https://github.com/umbraco/Umbraco-CMS/pull/6824).

[...]

PR #1704, opened 2019-05-09
Add info on TransformingIndexValues

jmayntzhusen:
Continued work on https://github.com/umbraco/UmbracoDocs/issues/1685 🙂


Reference/Searching/Examine/index-v7.md
```diff
@@ -0,0 +1,37 @@
...
+versionFrom: 7.0.0
+---
+
+# Examine
+
+_Examine uses Lucene as its search and index engine. Searching using Examine with Lucene can be very powerful and fast and it can also be really simple to setup and use._
+
+## What is Examine?
+
+Examine allows you to index and search data quickly and easily. It is a library that sits on top of Lucene, a very high performance search engine library written in Java. Examine is built on top of a .NET-implementation of Lucene to provide very efficient APIs to make searching and indexing as straight forward as possible. Umbraco provides a further layer on top, UmbracoExamine, that opens up the Umbraco-specific APIs for indexing and searching content and media out of the box.
```
Shazwazza: This block should be:

> Examine allows you to index and search data quickly and easily. Examine is a library that sits on top of Lucene.Net, a high performance search engine library. Examine provides easy to use APIs to make searching and indexing as straight forward as possible. Umbraco provides a further layer on top, UmbracoExamine, that opens up the Umbraco-specific APIs for indexing and searching content and media out of the box.

probably with a link to the lucene.net website: https://lucenenet.apache.org/

it's not written in Java! ... it's a c# port of the original Java source so best not to confuse people.

Reference/Searching/Examine/index-v7.md
```diff
@@ -0,0 +1,37 @@
...
+
+Examine is provider based so is very extensible and allows you to configure as many indexes as you like and each may be configured individually. The backoffice search in Umbraco also uses this same search engine, so you can trust that you're in good hands.
```
Shazwazza: Should be:

> Examine is provider based so it is extensible and allows you to configure your own custom indexes if required. The backoffice search in Umbraco also uses this same search engine, so you can trust that you're in good hands.

No need to use words like 'very' all of the time, we don't need to try to impress people reading the docs :) ... but i suspect this was copy/pasted from before. We should also not encourage people to create 'as many indexes as they want' because we don't want people to go creating a ton of indexes since each index added adds overhead equivalent to another index.

Reference/Searching/Examine/index.md
```diff
@@ -13,25 +12,21 @@ Examine allows you to index and search data quickly and easily. It is a library
 
 Examine is provider based so is very extensible and allows you to configure as many indexes as you like and each may be configured individually. The backoffice search in Umbraco also uses this same search engine, so you can trust that you're in good hands.
```
Shazwazza: Like the above comment, this should be changed:

> Examine is provider based so it is extensible and allows you to configure your own custom indexes if required. The backoffice search in Umbraco also uses this same search engine, so you can trust that you're in good hands.

Reference/Searching/Examine/index.md
```diff
@@ -13,25 +12,21 @@ Examine allows you to index and search data quickly and easily. It is a library
...
 
-## [Overview & Explanation - "Examining Examine by Peter Gregory"](overview-explanation.md)
+## [TransformingIndexValues](transforming-index-values.md)
 
-A detailed overview from top to bottom of how to use Examine
-
-## [Full configuration markup example](../../Config/ExamineSettings/index.md)
-
-Shows all configuration options with an explanation for each
+Provides a way to create custom fields of all your data so you can search across many fields in one query.
```
Shazwazza: It does allows you to do more than just this, I'm also not convinced that `TransformingIndexValues` is a great headline? Maybe this headlines should be "Examine Events" and say something like

> Details about subscribing to Examine events which can provide a way to modify the data being indexed. 

Then change the page `transforming-index-values` to `examine-events.md` since this page can contain a lot more information about different events in the future

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
+---
+versionFrom: 8.0.0
+---
+
+# TransformingIndexValues
```
Shazwazza: The file name and main headline of this page should be changed to examine-events (see above comment)

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+        //todo
+    }
+
+    public void Terminate()
+    {
+    }
+}
+```
+
+If this syntax is confusing you can read more about [Components](../../../Implementation/Composing/index.md). We can now add the logic to combine fields in the `IndexProviderTransformingIndexValues` method:
```
Shazwazza: Again, refrain from assuming what you audience thinks. Saying things like this seems to indicate that we know our APIs are confusing but i believe this is only true if you've never seen them before. I would just say:

> You can read more about this [syntax and Components here](../../../Implementation/Composing/index.md). We can now add the logic to combine fields in the `IndexProviderTransformingIndexValues` method:

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
+---
+versionFrom: 8.0.0
+---
+
+# TransformingIndexValues
+
+In the [Quick Start](Quick-Start/index.md) documentation you can see how to perform a simple search with Examine. That is great if you want to search between node names or you know that you always want to search for a specific field - e.g. `bodyText`.
```
Shazwazza: This whole section on TransformingIndexValues seems to indicate that this event is for one single purpose only which it isn't. This is just one example of what you can do with this event. 

Before going into this example we need to explain what this event is actually for.

> The TransformingIndexValues event allows you to manipulate the data that will be indexed during an index operation. With this event you can add, remove or modify the data that is going into the index which can be helpful in many scenarios. 

Then perhaps another sub headline: `Example`

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
+---
+versionFrom: 8.0.0
+---
+
+# TransformingIndexValues
+
+In the [Quick Start](Quick-Start/index.md) documentation you can see how to perform a simple search with Examine. That is great if you want to search between node names or you know that you always want to search for a specific field - e.g. `bodyText`. 
+
+However, what if you want to search through several different node types and search across many different fields, you likely will not want to have a query that looks like this:
```
Shazwazza: this is assuming what the audience thinks. Instead it can just be written:

> However, what if you want to search through several different node types and search across many different fields, you will typically need to have a query that looks like this:

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+```cs
+var results = searcher.CreateQuery("content").Field("nodeName", searchTerm)
+                    Or().Field("bodyText", searchTerm)
+                    Or().Field("description", searchTerm)
+                    Or().Field("about", searchTerm)
+                    Or().Field("otherText", searchTerm)
+                    .Execute();
+```
+
+This quickly becomes unmanageable, would have to add each new field to the query as you make them.
```
Shazwazza: This should be

> This can be simplified

Otherwise it sounds like the APIs we are providing are not good enough out of the box and we are assuming that everyone must do it this way which isn't true.

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+                    Or().Field("bodyText", searchTerm)
+                    Or().Field("description", searchTerm)
+                    Or().Field("about", searchTerm)
+                    Or().Field("otherText", searchTerm)
+                    .Execute();
+```
+
+This quickly becomes unmanageable, would have to add each new field to the query as you make them. 
+
+Instead you can use **TransformingIndexValues** (used to be called GatheringNodeData in Umbraco 7). TransformingIndexValues is an Examine event where a new field is created in the index that contains the data from several other fields, you can then search on just that field. This is done via a [composer](../../../Implementation/Composing/index.md).
```
Shazwazza: should be 

> You can use the **TransformingIndexValues** event (used to be called GatheringNodeData in Umbraco 7) to add a custom field to the indexed data to combine the data from several fields and then you can search on that one field. This is done via a [composer](../../../Implementation/Composing/index.md).

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+        //we need to cast because BaseIndexProvider contains the TransformingIndexValues event
+        if (!(index is BaseIndexProvider indexProvider))
+            throw new InvalidOperationException("Could not cast");
+
+        indexProvider.TransformingIndexValues += IndexProviderTransformingIndexValues;
+    }
+
+    private void IndexProviderTransformingIndexValues(object sender, IndexingItemEventArgs e)
+    {
+        //todo
```
Shazwazza: This will need to be filled in

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+```
+
+If this syntax is confusing you can read more about [Components](../../../Implementation/Composing/index.md). We can now add the logic to combine fields in the `IndexProviderTransformingIndexValues` method:
+
+```cs
+if (e.ValueSet.Category == IndexTypes.Content)
+{
+    try
```
Shazwazza: This is a huge try/catch block... but why? Large try/catch blocks are bad practice and lead to developers thinking that this is a fail safe to proper error checking, this also leads to errors being swallowed and problems will go on silently in the background with developers not knowing what is happening. It's better to let unhandled errors be thrown so people actually know what the problems are.

To do proper error checking you need to determine what can go wrong in this block of code:

* Creating an object `var combinedFields = new StringBuilder();` ... cannot throw, so no try/catch needed
* Iterating over this object `foreach (var fieldValues in e.ValueSet.Values)` ... potentially there could be a null check in there, so need to figure out if `e` or `e.ValueSet` or `e.ValueSet.Values` could ever be null ... I don't think they can so no try/catch needed
* Iterating over this object `foreach (var value in fieldValues.Value)`, just like the above, only a null reference exception could occur if `fieldValues.Value` could ever be null... I don't think it can so no need for try/catch
* `combinedFields.AppendLine(value.ToString());` ... you are already checking for null above, so no exception can occur, no need for a try/catch
* `e.ValueSet.Add("combinedField", combinedFields.ToString());` , we already decided that `e.ValueSet` cannot be null and we definitely know that `combinedFields` cannot be null because we created it. However, an exception can be thrown here if the "combinedField" key already exists in this dictionary, this could happen if someone created a property type called "combinedField", so this should be properly error checked which is simple by just changing the code to `e.ValueSet.TryAdd("combinedField", combinedFields.ToString());` which will do it's own error checking.

We should then remove this try catch so we aren't promote the use of large try/catch/swallow code blocks

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+```
+
+So at this point we have done something along the lines of:
+- Get the ExternalIndex
+- Get the valueset for the ExternalIndex
+- For each field, add the content to a new field called `combinedField`
+
+Before this works the component will have to be registered in a composer. If you already have a composer you can add it to that one, for this example we will make a new composer:
```
Shazwazza: missing the word `but`, so should be `but for this example we will make a new composer:`

Reference/Searching/Examine/transforming-index-values.md
```diff
@@ -0,0 +1,120 @@
...
+
+```cs
+public class CustomComposer : IUserComposer
```
Shazwazza: There's another option to this - but will leave it up to you to decide what to document. You can create a composer for a single component in a strongly typed way so for this example, you could do:

```cs
//This is a composer which automatically appends the ExamineEvents component
public class ExamineComposer : ComponentComposer<ExamineEvents>, IUserComposer 
{ 
   // you could override `Compose` if you wanted to do more things, but if it's just registering a component there's nothing else that needs to be done.
}
```

Reference/Searching/Examine/transforming-index-values.md
jmayntzhusen: Will be filled in in the next step 🙂

Reference/Searching/Examine/transforming-index-values.md
jmayntzhusen: Great point, fixed it now 👍

Reference/Searching/Examine/transforming-index-values.md
jmayntzhusen: Think we have plenty of examples in the docs of the "standard" way that I did it, will use your example here 🙂

PR #1708, opened 2019-05-19
Add lots of examples around using Services with DI

marcemarc:
This is very much WIP, and may suit rather a tutorial format, but my gut feeling is having one place to show off by example the different nuances of injecting services that people will run into when implementing a typical site would be useful.

Particular keen to suggest people follow the pattern of creating their own custom services for implementation logic, instead of writing code in Views or Controllers...

but not sure how much of that is my own preference, and whether this should be re-centred around using ModelsBuilder or if that is a separate 'good practice' outline...

Some of the examples could be improved upon, and need to do a sense check of wording...

but wanted to know if this seemed like a good idea etc - was in the right place - should be 3 different pages etc

### DIFF
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -136,7 +136,7 @@ namespace Umbraco8.Components
 
         public void Terminate()
         {
-            throw new NotImplementedException();
+           
         }
     }
 }
@@ -168,7 +168,7 @@ namespace Umbraco8.Components
     {
         private readonly IUmbracoContextFactory _umbracoContextFactory;
 
-        public PublishEventsComponent(IUserService userService, IUmbracoContextFactory umbracoContextFactory)
+        public PublishEventsComponent(IUmbracoContextFactory umbracoContextFactory)
         {
             _umbracoContextFactory = umbracoContextFactory;
         }
@@ -194,7 +194,7 @@ namespace Umbraco8.Components
 
         public void Terminate()
         {
-            throw new NotImplementedException();
+         
         }
     }
 }
@@ -241,7 +241,7 @@ namespace Umbraco8.Extensions
 }
 ```
 
-anywhere there is reference to the UmbracoHelper, and a reference is added to the namespace the extension belongs to, it's possible to write `Umbraco.GetNewsSection()` - for a large site though, this can get a little bit out of control!
+anywhere there is reference to the UmbracoHelper, and a reference is added to the namespace the extension belongs to, it is possible to call the method by writing `Umbraco.GetNewsSection()`
 
 ### Custom Services and Helpers
 
@@ -266,9 +266,9 @@ namespace Umbraco8.Services
         IPublishedContent GetContactUsPage();
     }
 }
-
+```
 Create the concrete service class that implements the interface:
-
+```csharp
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -13,7 +13,7 @@ The general rule of thumb is that Management Services provide access to allow th
 Although there is a Management Service compelling called the `ContentService` - only use this to modify content - do not use the `ContentService` in a View/Template to pull back data to display, this will make database requests and be slow - here instead use the generically named `UmbracoHelper` to access the `PublishedContentQuery` methods that operate against a cache of published content items, and are significantly quicker.
 :::
 
-The Management Services and Helpers are all registered with Umbraco's underlying DI framework - this article aims to show examples of gaining access to utilise them in multiple different scenarios (there are subtle differences to be aware of depending on what part of Umbraco is being extended) - and also suggest how to follow a similar pattern to encapsulate custom 'site specific' logic, registered with the DI framework, to avoid repetition and promote consistency and readability in a solution.
+The Management Services and Helpers are all registered with Umbraco's underlying DI framework - this article aims to show examples of gaining access to utilise these resources in multiple different scenarios (there are subtle differences to be aware of depending on what part of Umbraco is being extended) - and also suggest how to follow a similar pattern to encapsulate custom 'site specific' implementation logic, in similar services and helpers, registered with the underlying DI contain, to avoid repetition and promote consistency and readability within an Umbraco site solution.
 
 ## Accessing Management Services and Helpers in a Template/View
 
@@ -394,7 +394,7 @@ namespace Umbraco8.Controllers
 
 #### Using the SiteHelperService inside a View
 
-If strictly following the paragdigm of MVC, calling custom Services from Views might feel like an anti-pattern. However there isn't one single best practice approach to working with Umbraco. Alot depends on circumstance and expertise. Allowing Umbraco to handle the flow of the requests to the particular template, and writing implementation logic in Views/Templates, is still a very common approach. There are circumstances too where the custom implementation logic shared is very 'View' specific - custom logic for constructing 'Alternative Text' for images or different crop urls for img srcsets can be neatly handled in a Service without having to 'route hijack' the request and build a complex ViewModel.
+If strictly following the paragdigm of MVC, calling custom Services from Views might feel like an anti-pattern. However there isn't necessarily one single 'best practice' approach to working with Umbraco. A lot depends on circumstance, expertise and pragmatism. Allowing Umbraco to handle the flow of incoming requests to a particular page + template, and writing implementation logic in Views/Templates, is still a very common approach. There are circumstances too, where the custom implementation logic shared is very 'View' specific - custom logic for constructing 'Alternative Text' for images or different crop urls for img srcsets can be neatly handled in a custom Helper/Service without having to create a hijacked MVC route for the request and build a complex ViewModel. Custom Services called from Views, can help separate the concerns, even if the 'plumbing' isn't pure MVC.
 
 You 'could' create an instance of your custom service inside the view
 
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -175,12 +175,6 @@ namespace Umbraco8.Components
 }
 ```
 
-#### Accesing ContentCache from a SurfaceController, RenderMvcController UmbracoApiController
-
-If your working inside a class/controller that inherits from one of the special base classes, you have access to the cache from the UmbracoHelper
-
-UmbracoHelper.
-
 #### Accesing ContentCache from a Content Finder / UrlProvider
 
 Inside a ContentFinder you have access to the content cache via the PublishedRequest
@@ -205,6 +199,7 @@ Let's create a custom service, that's responsible for finding key pages within a
 
 Create an interface to define the service:
 
+```csharp
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -242,16 +237,18 @@ namespace Umbraco8.Services
         }
     }
 }
+```
 
 Register the service with an IUserComposer
 
+```csharp
 using Umbraco.Core;
 using Umbraco.Core.Composing;
 using Umbraco8.Services;
 
 namespace Umbraco8.Composers
 {
-    public class RegisterSuperTestServiceComposer : IUserComposer
+    public class RegisterSiteServiceComposer : IUserComposer
     {
         public void Compose(Composition composition)
         {         
@@ -259,9 +256,11 @@ namespace Umbraco8.Composers
         }
     }
 }
+```
 
 Implementing the service, we can inject Umbraco's IPublishedContentQuery to find the pages we're after
 
+```csharp
 using System.Linq;
 using Umbraco.Core.Models.PublishedContent;
 using Umbraco.Web;
@@ -271,7 +270,7 @@ namespace Umbraco8.Services
     public class SiteService : ISiteService
     {
         private readonly IPublishedContentQuery _contentQuery;
-        public AnotherSiteService(IPublishedContentQuery contentQuery)
+        public SiteService(IPublishedContentQuery contentQuery)
         {
             _contentQuery = contentQuery;
         }
@@ -289,11 +288,13 @@ namespace Umbraco8.Services
         }
     }
 }
+```
 
 (if we planned to use our SiteService outside of a request where UmbracoContext is not available we'd use the UmbracoContextFactory + EnsureUmbracoContext() methods described above to access the content cache)
 
 #### Using the SiteService inside a Controller
 
+```csharp
 using System.Web.Mvc;
 using Umbraco.Web.Models;
 using Umbraco8.Services;
@@ -325,27 +326,31 @@ namespace Umbraco8.Controllers
         }
     }
 }
+```
 
 #### Using the SiteService inside a View
 
 Sometimes you'll want to encapsulate some kind of View Logic into a service, or perhaps you prefer not to RouteHijack every request to build a custom ViewModel for a page.
 In these circumstances you 'could' instanstiate your service inside your view:
 
+```csharp
 @using Umbraco8.Services
 @inherits UmbracoViewPage
 @{
 
     Layout = "master.cshtml";
-    IAnotherSiteService siteService = new AnotherSiteService(Umbraco.ContentQuery);
+    ISiteService siteService = new SiteService(Umbraco.ContentQuery);
     IPublishedContent newsSection = siteService.GetNewsSection();
 }
 <section class="section">
     <div class="container">
         <article>
+```
 
 but this isn't making use of the DI container to control which concrete service implements IAnotherService
 so you could get the instance from the Umbraco.Web.Composing.Current DI container:
 
+```csharp
 @using Umbraco8.Services
 @using Current = Umbraco.Web.Composing.Current;
 @inherits UmbracoViewPage
@@ -358,9 +363,10 @@ so you could get the instance from the Umbraco.Web.Composing.Current DI containe
 <section class="section">
     <div class="container">
         <article>
-
+```
 or another strategy is to create your own 'CustomViewPage' to use as the basis for your views and to wire up any of your custom services
 
+```csharp
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -377,7 +383,7 @@ namespace Umbraco8.ViewPages
 {
     public abstract class CustomViewPage<T> : UmbracoViewPage<T>
     {
-        public readonly ISiteService _siteService;
+        public readonly ISiteService SiteService;
         public CustomViewPage() : this(
             Current.Factory.GetInstance<ISiteService>(), 
             Current.Factory.GetInstance<ServiceContext>(),
@@ -387,7 +393,7 @@ namespace Umbraco8.ViewPages
         }
         public CustomViewPage(ISiteService siteService, ServiceContext services, AppCaches appCaches)
         {
-            _siteService = siteService;
+            SiteService = siteService;
             Services = services;
             AppCaches = appCaches;
         }
@@ -420,20 +426,21 @@ namespace Umbraco8.ViewPages
         }
     }
 }
+```
 
 with this in place your view would look like this:
-
+```csharp
 @using Umbraco8.ViewPages
 @inherits CustomViewPage
 @{
 
     Layout = "master.cshtml";
-    IPublishedContent newsSection = _siteService.GetNewsSection();
+    IPublishedContent newsSection = SiteService.GetNewsSection();
 }
 <section class="section">
     <div class="container">
         <article>
-
+```
 
 
 
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -380,7 +380,7 @@ namespace Umbraco8.Controllers
 
         public override ActionResult Index(ContentModel model)
         {
-            var newsSection = _siteService.GetNewsSection();
+            var newsSection = _siteHelperService.GetNewsSection();
             var blogPostViewModel = new BlogPostViewModel(model);
             blogPostViewModel.NewsSection = newsSection;
             //etc          
@@ -405,7 +405,7 @@ You 'could' create an instance of your custom service inside the view
 
     Layout = "master.cshtml";
     ISiteHelperService siteHelperService = new SiteHelperService(Umbraco.ContentQuery);
-    IPublishedContent newsSection = siteService.GetNewsSection();
+    IPublishedContent newsSection = siteHelperService.GetNewsSection();
 }
 <section class="section">
     <div class="container">
@@ -423,7 +423,7 @@ to 'get around this' you could get the current concrete instance from the Umbrac
 
     Layout = "master.cshtml";
     ISiteHelperService siteHelperService = Current.Factory.GetInstance<ISiteHelperService>();
-    IPublishedContent newsSection = siteService.GetNewsSection();
+    IPublishedContent newsSection = siteHelperService.GetNewsSection();
 }
 <section class="section">
     <div class="container">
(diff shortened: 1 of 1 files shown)


Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+                       IMedia newFolder = _mediaService.CreateMedia(contentItem.Name, -1, "Folder");
+                        _mediaService.Save(newFolder);
+                    }
+                }
+            }
+        }       
+
+        public void Terminate()
+        {
+            throw new NotImplementedException();
```
warrenbuckley:
```suggestion
```

Recommend you remove this @marcemarc as when Umbraco tries to gracefully shutdown its gonna YSOD

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+                //query the published content cache
+                IEnumerable<IPublishedContent> rootNodes = contentCache.GetAtRoot();
+                //retrieve a particular item with id 1234 from the published content cache
+                IPublishedContent particularItem = contentCache.GetById(1234);
+            }
+        }
+
+        public void Terminate()
+        {
+            throw new NotImplementedException();
```
warrenbuckley: Same as above ⬆️

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+    public class PublishEventsComposer : ComponentComposer<PublishEventsComponent>
+    {
+
+    }
+    //the component safely accessing the published content cache
+    public class PublishEventsComponent : IComponent
+    {
+        private readonly IUmbracoContextFactory _umbracoContextFactory;
+
+        public PublishEventsComponent(IUserService userService, IUmbracoContextFactory umbracoContextFactory)
```
warrenbuckley: Typo? As you don't use `userService` in the example

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+            //NewsCategories //NewsSection
+            public static IPublishedContent GetNewsSection(this UmbracoHelper umbracoHelper)
+            {
+            return umbracoHelper.ContentSingleAtXPath("root/homePage/newsSection");
+            }
+     }
+}
+```
+
+anywhere there is reference to the UmbracoHelper, and a reference is added to the namespace the extension belongs to, it's possible to write `Umbraco.GetNewsSection()` - for a large site though, this can get a little bit out of control!
```
warrenbuckley: I would give context as to why it could get out of control

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+namespace Umbraco8.Services
+{
+        public interface ISiteHelperService
+    {
+        IPublishedContent GetNewsSection();
+        IPublishedContent GetContactUsPage();
+    }
+}
+
+Create the concrete service class that implements the interface:
```
warrenbuckley: Forgot to close the code example above & thus you have a single text line in the code example, that people are bound to copy & paste

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+            }
+     }
+}
+```
+
+anywhere there is reference to the UmbracoHelper, and a reference is added to the namespace the extension belongs to, it is possible to call the method by writing `Umbraco.GetNewsSection()`
+
+### Custom Services and Helpers
+
+Another option, is to make use of the underlying DI framework, and create custom Services and Helpers, that in turn can have the 'core' Management Services and Umbraco Helpers injected into them - (the same rules about transient services apply!). This approach enables the grouping together of similar methods within a suitably named service, and promotes the possibility of testing this custom logic outside of controllers and views.
```
Shazwazza: Not sure if this 'transient' rule makes sense ... see my notes above.

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+In this example, we create a custom helper service, that's responsible for finding key pages within a site, eg the News Section or the Contact Us page - these methods will commonly be called in different places throughout the site, and it's great to encapsulate the logic to retrieve them in a single place - we'll call this helper service: SiteHelperService
```
Shazwazza: In core, we are trying to remove anything called `Helper`, i'm not sure if we want to promote that to developers too. `Helper` is very vague, it doesn't really have a meaning. In this case it could be called `ISiteService` or since it's their site `IMySiteService` (naming is hard)

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+    {
+        public void Compose(Composition composition)
+        {         
+            composition.Register<ISiteHelperService, SiteHelperService>(Lifetime.Request);          
+        }
+    }
+}
+```
+
+Note we register the service in 'Lifetime.Request' scope - as we'll be making use of the transient UmbracoContext
```
Shazwazza: It is **not** Transient, it is also Request lifetime. 

I have another note below about how to structure these types of things, registering things as Lifetime.Request should be avoided wherever possible and perhaps we make a note of that.

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+            var siteRoot = _contentQuery.ContentAtRoot().FirstOrDefault();
+            var contactUs = siteRoot.Children(f => f.ContentType.Alias == "contactUs").FirstOrDefault();
+            return contactUs;
+        }
+    }
+}
+```
+
+:::Note
+If we planned to use our SiteHelperService outside of a request where UmbracoContext is not available we'd use the UmbracoContextFactory + EnsureUmbracoContext() methods described above to access the content cache.
```
Shazwazza: Based on my previous note, it is preferable to not use Request based services. What happens now if this current service is needed on a background thread or you want to be able to inject this into a singleton - well you can't!

Instead, registering things as singletons where possible makes the most sense. Then if you want your object to have access to the UmbracoContext, you will inject the `IUmbracoContextAccessor` ... in fact, this is best practices. We don't want people to inject UmbracoContext or IPublishedContentQuery directly - the only places this is 'ok' is in Controllers - but as noted already, these things are already injected into controllers.

For custom services, best practice is register as a singleton, if you need an UmbracoContext, inject `IUmbracoContextAccessor`. Then it's also important to check if there is one available. For example, `if (umbCtxAccessor.UmbracoContext == null) { //there is no current context, so now you should use the UmbracoContextFactory }`. However with that said, you could simply always just use the `UmbracoContextFactory` instead since this already checks for the existence of a current umb context on the `IUmbracoContextAccessor`. The benefit however of using `IUmbracoContextAccessor` is that you know if you are within an http request or not.

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+:::
+
+:::Note
+Check if _contentQuery.SingleAtXPath("root/homePage/contactUs") is faster... (it would be in V7)
```
Shazwazza: As noted above - xpath is slower in v8.

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+
+namespace Umbraco8.Controllers
+{
+    public class BlogPostController : Umbraco.Web.Mvc.RenderMvcController
+    {         
+        private readonly ISiteHelperService _siteHelperService;
+        private readonly ILogger _logger;
+
+        public BlogPostController(ILogger logger, ISiteHelperService siteHelperService)
```
Shazwazza: Just like the first note about base constructors we probably want to change this. Also with the use of a custom _logger doesn't actually make much sense.

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+        {
+            _logger = logger;
+            _siteHelperService = siteHelperService;
```
Shazwazza: Should our examples add a null check? 

```cs
_siteHelperService = siteHelperService ?? throw new ArgumentNullException(nameof(siteHelperService));
```

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+        public BlogPostController(ILogger logger, ISiteHelperService siteHelperService)
+        {
+            _logger = logger;
+            _siteHelperService = siteHelperService; 
+        }
+
+        public override ActionResult Index(ContentModel model)
+        {
+            var newsSection = _siteHelperService.GetNewsSection();
```
Shazwazza: I know we are trying to be concise but there should be a null check on `newsSection`

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+---
+
+# Services and Helpers
+
+Umbraco has a range of 'Core' Services and Helpers that act as a 'gateway' to Umbraco data and functionality to use when extending or implementing an Umbraco site.
+
+The general rule of thumb is that Management Services provide access to allow the modification of Umbraco data, (and therefore aren't optimised for displaying data), whereas Helpers provide access to readonly data with performance of displaying data taken into consideration.
+
+:::warning
+Although there is a Management Service compelling called the `ContentService` - only use this to modify content - do not use the `ContentService` in a View/Template to pull back data to display, this will make database requests and be slow - here instead use the generically named `UmbracoHelper` to access the `PublishedContentQuery` methods that operate against a cache of published content items, and are significantly quicker.
```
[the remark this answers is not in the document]
marcemarc: capitalisation gone!

and 'compelling' removed - the sense it is trying to convey, is that although there is a service called 'ContentService' don't use this to readonly access Service.. so it is compellingly named.... I am compelled to use it for Content - but actually don't! Naming is hard! - I have made it clearer in the 'warning box' what the warning is! without 'embellishing' with fancy words... there I go again :-P

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+:::
+
+The Management Services and Helpers are all registered with Umbraco's underlying DI framework - this article aims to show examples of gaining access to utilise these resources in multiple different scenarios (there are subtle differences to be aware of depending on what part of Umbraco is being extended) - and also suggest how to follow a similar pattern to encapsulate custom 'site specific' implementation logic, in similar services and helpers, registered with the underlying DI contain, to avoid repetition and promote consistency and readability within an Umbraco site solution.
+
+## Accessing Management Services and Helpers in a Template/View
+
+Inside a view/template or partial view that inherits from UmbracoViewPage, access is provided to Umbraco's services via the `Services` [service context](../../Reference/Management/Services/) and the `Umbraco` - [UmbracoHelper](../..//Reference/Querying/UmbracoHelper) - is a gateway to an array of useful functionlity.
```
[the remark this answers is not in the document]
marcemarc: agreed and fixed

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+@inherits UmbracoViewPage<Blogpost>
+@using ContentModels = Umbraco.Web.PublishedModels;
+@{
+    Layout = "master.cshtml";
+
+    // retrieve an item from Umbraco's published cache with id 123
+    IPublishedContent publishedContentItem = Umbraco.Content(123);
+
+    // it is 'unlikely' to need to use a Management Service in a view:
+    var relationService = Services.RelationService;
```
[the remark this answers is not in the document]
marcemarc: commenting out the code, and keeping the message in, want it to be visible that this isn't generally a done thing - if we remove completely then people won't implicitly know this isn't a good thing

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+namespace Umbraco8.Controllers
+{
+    public class BlogPostController : Umbraco.Web.Mvc.RenderMvcController
+    {        // GET: BlogPost   
+        private readonly ILogger _logger;
+        private readonly UmbracoHelper _umbracoHelper;
+        private readonly RelationService _relationService;
+
+        public BlogPostController(ILogger logger, UmbracoHelper umbracoHelper, RelationService relationService)
```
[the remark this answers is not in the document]
marcemarc: Yes, I've reworked the article in several places to reflect this - we need to document for those that choose to use DI for the full DI experience leading through to unit testability and beyond, and also for those who perhaps choose not to use it to that extent, but will gain benefit from the simplicity of the Service Locator fallback 'just working' for missing constructors, and hopefully can still be encouraged to utilise grouping common implementation code into a service...

... I've tried to think about how best to try and convey this to different audiences...

what I have is two examples, the first showing the fallback 'just works', followed by a warning regarding DI followed by the proper best practice for DI approach - and added the tip and the gif to generate them!

Hopefully this doesn't scare people off with a ton of constructors they've never seen before, but lures them into the 'why' and where 'it could go' with DI - but easy to delete the first bit if it's felt the documentation shouldn't try to engage in this way.

I guess in some implementations the length of the constructors could get quite long... so yes +1 for, is it a Facade Service?, or SuperControllerContextObject

Implementation/Services/index.md
marcemarc: thought so... any circumstances where it would be better to steer people to use XPath, for these methods?

Implementation/Services/index.md
marcemarc: I think this is resolved too by the refactor, eg logger not injected, and showing both the Service locator fallback + full DI ways

Implementation/Services/index.md
marcemarc: indeed, when we can show sound implementation examples, habits will be picked up

Implementation/Services/index.md
marcemarc: and null checking here too

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+        {
+            _umbracoHelper = umbracoHelper;
+            _logger = logger;
+            _relationService = relationService;
+        }
+        public override ActionResult Index(ContentModel model)
+        {
+            // write helpful messages to the Umbraco Trace logs to aid with debugging
+            _logger.Info<BlogPostController>("Using core logger implementation");
```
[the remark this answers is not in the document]
marcemarc: Yep, reworked to highlight you do not need to inject these helpers or services into the constructors of a controller to use them

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+            // retrieve an item from Umbraco's published cache with id 123
+            IPublishedContent publishedContentItem = _umbracoHelper.Content(123);
+            // it is unlikely to use a Management service when rendering content from a custom controller
+            //(when using relationService like this you would want to provide a layer of caching)
+            var allRelatedUmbracoItems = _relationService.GetByParentId(model.Id);_
+        }
+```
+
+:::Note
+You still have access to Umbraco. and Services. as as away to access UmbracoHelper and any Management Services.
```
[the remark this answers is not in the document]
marcemarc: yep,. removed

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+So the above could be Umbraco.Content(123) and Services.RelationService.GetByParentId(model.Id)
+:::
+
+## Accessing Core Services and Helpers when there is no UmbracoContext eg in a Component or C# Class
+
+Controllers and Views have an UmbracoContext, however this is not always the case 'everywhere in Umbraco', for example common extension points: Components,ContentFinders or Custom C# Classes.
+
+:::Warning
+If the UmbracoContext is not guaranteed to exist you cannot inject any helpers or services that rely on the UmbracoContext, eg UmbracoHelper, PublishedContentQuery and if you try to inject these helpers into the constructors of these context less classes...  Umbraco will report a 'boot' error on startup.
```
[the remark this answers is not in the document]
marcemarc: agreed and updated

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+        public void Terminate()
+        {
+           
+        }
+    }
+}
+```
+
+Management Services that are not transient (eg do not rely on UmbracoContext), eg `ContentService`, `MediaService` can be injected anywhere.
```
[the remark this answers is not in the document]
marcemarc: a-ha! thanks, right, I've added a section on Lifespans with the above information, which is really useful and indeed means the explanation is now less woolly over why and what you can use where!

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+### Accessing Published Content Cache from a Component or C# Class
```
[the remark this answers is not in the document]
marcemarc: updated

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+Injecting `UmbracoHelper` or `IPublishedContentQuery` into classes that aren't guaranteed to have an UmbracoContext will trigger a boot error, however there is a technique that allows the querying of the Umbraco Published Content Cache, using the `UmbracoContextFactory` and calling `EnsureUmbracoContext()`.
```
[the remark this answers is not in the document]
marcemarc: also updated

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+        private void ContentService_Published(IContentService sender, Umbraco.Core.Events.ContentPublishedEventArgs e)
+        {
+            //first call EnsureUmbracoContext
+            using (var umbracoContextReference = _umbracoContextFactory.EnsureUmbracoContext())
+            {
+                //the UmbracoContextReference provides access to the ContentCache
+                var contentCache = umbracoContextReference.UmbracoContext.ContentCache;
+                //query the published content cache
+                IEnumerable<IPublishedContent> rootNodes = contentCache.GetAtRoot();
+                //retrieve a particular item with id 1234 from the published content cache
```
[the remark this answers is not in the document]
marcemarc: It isn't trying to suggest that, it is just a shit example! - just showing the technique without any real thought for the context of it's usage, other than somehere that is outside of a request - you would never really 'do that' kind of thing  in 'that event' - had a bit of a think - keen not to guide people to be sending emails 'on publish' and my CacheUpdated examples seemed too convoluted ... examples are hard too...

... what I came up with, which may not be desirable either ...

is when unpublishing event fires - you might want to retrieve what the published Url once was to the unpublished item to store in a db table to be able to serve proper 410 'page gone' status codes via an IContentFinder... for future requests instead of a 404. It's easily understandable as a concept but is getting the IPublishedContent object from content cache, in order to read the Url convoluted? or make sense in the more complex world of variants... 

just needs to be something not awful, but allows the reader to focus on the technique not what is trying to be done by the example...

I've put that into the PR anyway, it is better than what was there.

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+        public void Terminate()
+        {
+         
+        }
+    }
+}
+```
+
+#### Accesing Published Content Cache from a Content Finder / UrlProvider
```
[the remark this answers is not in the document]
marcemarc: Added a note, and again like the controllers, people don't need to inject in order to access content / cache, so hopefully made that clear too

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+using Umbraco.Web;
+
+namespace Umbraco8.Extensions
+{
+    public static class UmbracoHelperExtensions
+    {
+            //NewsCategories //NewsSection
+            public static IPublishedContent GetNewsSection(this UmbracoHelper umbracoHelper)
+            {
+            return umbracoHelper.ContentSingleAtXPath("root/homePage/newsSection");
```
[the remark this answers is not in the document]
marcemarc: updated to linq

Implementation/Services/index.md
marcemarc: No it doesn't, I have had another attempt to make it make clearer in sense! having the Lifespan info on the same page, makes it clearer too

Implementation/Services/index.md
marcemarc: Ha, all my examples use ISiteService, and SiteService, I added Helper in the example to try and follow what I thought was some sort of Umbraco driven convention where broadly Helper's are performant, safe ways to access things - and Services are slow database and CRUD... but Helper is a pet dislike of mine... agree it's vague... maybe for helping manipulate a string, but not for serving resources from elsewhere, curious as to why you persisted with UmbracoHelper in V8 - updated in the PR

Implementation/Services/index.md
marcemarc: Updated, and the Lifetime info added

Implementation/Services/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+Controllers and Views have an UmbracoContext, however this is not always the case 'everywhere in Umbraco', for example common extension points: Components,ContentFinders or Custom C# Classes.
+
+:::Warning
+If the UmbracoContext is not guaranteed to exist you cannot inject any helpers or services that rely on the UmbracoContext, eg UmbracoHelper, PublishedContentQuery and if you try to inject these helpers into the constructors of these context less classes...  Umbraco will report a 'boot' error on startup.
+:::
+
+### Injecting Services into a Component
+
+It's possible to inject Management Services that do not rely on the `UmbracoContext` into the constructor of a component. This example shows injecting the MediaService in a Component to create a corresponding Media Folder for every 'landing page' that is saved in the Content Section, by subscribing to the 'Content Saved' event.
```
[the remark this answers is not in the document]
marcemarc: added

Implementation/Services/index.md
marcemarc: Ok on the first point, the subtle thing is we 'do' want people to create custom services that are only used in Views/Controllers! - that is the inspiration for the article - Yes people have access to UmbracoHelper in a controller or view - but it leads to implementation of logic in the view or controller - eg Umbraco.ContentAtRoot().FirstOrDefault().Children(f=>f.IsVisible() && f.ContentType.Alias == "landingPage" && f.Children.Any());  which is sort of fine when the implementation logic is written once, in one place, but you tend to find duplication and subtle variations on implementation building up overtime, bring a lack of consistency, where somebody forgot 'where else' they pulled back news articles in date order - but what can be a neater alternative, is if this logic is put into a Service to be called in multiple places - it's the first step towards some degree of separation. WIth V8 the wonderful thing is the complicated bit of wiring up DI is done for peoples - so I'm trying to highlight in a Murky rather than super Clean way, how easy it is to get those benefits of separation, without it getting super complicated. The familiarity of the IPublishedContentQuery methods is helpful!

But agree the warnings against using the custom service outside of a http request aren't clear...
... and agree it's just as easy to use the IPublishedContentCache methods ... it's just the unfamiliarity - I wonder if over time we could sync the method naming a little?

... anyway my suggested solution for now is to include two examples, one showing injecting the familiar interface methods and then add the warning, and then show the 'better practice' way, again my hope is here, people are lured in by the easier example first and then conclude themselves that the second example is much better, and not so hard to use to avoid getting caught out.

I feel if we only show the second example, then the first example will be used by people anyway and they'll run into trouble, but this might not be the approach the official documentation should take to engaging people on this subject.

With IUmbracoContextAccessor, I haven't added anything - feel like we should, but wasn't clear of the benefit/how to explain best, eg if there is no UmbracoContext you'd have to use the UmbracoContextFactory approach, so you'd have to inject both into your custom service, unless you were sure you were only going to use your service where there is an UmbracoContext... then you could just use UmbracoContextAccessor - but if the UmbracoContextFactory always checks via the UmbracoContextAccessor anyway - it might be simplest to say just use that always? but I sense I'm failing to grasp the significance of knowing whether you are in a http request or not inside the context of your custom service :-(

 (perhaps there could be a flag on the UmbracoContextReference object that indicates whether you are or not?)

marcemarc:
@Shazwazza thanks for the review! I've been through all the points/comments and created a new PR here: https://github.com/umbraco/UmbracoDocs/pull/1762

Where I think I've resolved the comment in the PR, I've resolved the conversation here... and on other items where I wasn't sure or had a question, I've left the conversation open... but feel free to discuss on the other PR if that is easier!

PR #1713, opened 2019-05-26
More consisting markup + readability improvements

Migaroez:
- Use **_Bold Italic_** for specific terms (could be linked later to a lexicon)
- Use [Exact Text] to indicate UI items
- Use "_Input value_"
- Shortened some sentences
- Reworked awkward wording
- Added a comment to indicate a content improvement

I looked at the markdown specifications but couldn't find any style guide for terms/ui/input so I tried to use what was already present. 

If accepted I will continue this style of work on the rest of the tutorial.

### DIFF
--- a/Tutorials/Creating-Basic-Site/Document-Types/index.md
+++ b/Tutorials/Creating-Basic-Site/Document-Types/index.md
@@ -6,9 +6,9 @@ versionFrom: 8.0.0
 ## Data first 
 ### nothing in = nothing out!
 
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hey @Migaroez 

You are correct, we do not currently have a style guide, We have an open RFC for creating one, so if you feel these rules would be good to have in it please swing by and leave a comment at https://github.com/umbraco/rfcs/pull/9

The work you did in this PR is great though, thanks for contributing!

Issue #1579, opened 2019-02-22
V8: Update all Application Starting and Application Started event examples to use Composer approach


marcemarc:
Throughout the documentation when showing how to extend different parts of Umbraco, there is usually included a snippet to show how to wire up an event or register a collection etc via creating a class that implements ApplicationEventHandler, for V8 a different approach is used, creation a IUserComposer to create a composition to customise the functionality of Umbraco.

Therefore all these pages need their examples updated for V8 version to use the new approach:

A list of pages here:

Application Starting

- [x] Extending\Section-Trees\Searchable-Trees\index.md
- [x] Getting-Started\Setup\Server-Setup\Load-Balancing\flexible-advanced.md
- [x] Getting-Started\Setup\Server-Setup\Load-Balancing\flexible.md
- [x] Implementation\Default-Routing\Controller-Selection\index.md
- [x] Reference\using-ioc.md
- [x] Reference\Events\Application-Startup.md
- [ ] Reference\Events\Understanding-Bootmanagers.md
- [x] Reference\Routing\custom-controllers.md
- [x] Reference\Routing\Request-Pipeline\IContentFinder.md
- [x] Reference\Routing\Request-Pipeline\outbound-pipeline.md
- [ ] Reference\Templating\Modelsbuilder\IPublishedContentModelFactory.md

Application Started

- [x] Getting-Started\Code\Subscribing-To-Events\index.md
- [x] Getting-Started\Code\Umbraco-Services\index.md
- [x] Reference\Events\ContentService-Events.md
- [ ] Reference\Events\Extending-UmbracoApplication.md
- [x] Reference\Events\MediaService-Events.md
- [x] Reference\Events\EditorModel-Events\index.md
- [x] Reference\Routing\custom-routes.md
- [x] Reference\Routing\Authorized\index.md

Some of these will be quick wins, eg just updated the example to use the IUserComposer approach will make the documentation V8 ready... whereas on other pages, it will cause confusion, seeing the V8 method for wiring things up will imply the whole page of documentation applies to V8 when it might not

nul800sebastiaan:
Hi @marcemarc - just to let you know, I've added the "idea" label on this since it will be a longer-term project and it's great that you keep working on it, but it doesn't need to be in the overview for HQ for open issue that require immediate attention.

Keep up the great work!

marcemarc:
[...]

PR #1714, opened 2019-05-27
Updating outbound pipeline info for v8

marcemarc:
Updated references to application event handler
Updated  UrlProvider and SegmentProvider examples / explanations
Not sure about how best to explain SiteDomainHelper... so have left the same...


marcemarc:
Related issue: https://github.com/umbraco/UmbracoDocs/issues/1579

PR #1719, opened 2019-05-29
Syntax error fix on example code block.

neurox:


jmayntzhusen:
Hey @neurox 

Thanks for both reporting and fixing this error! 👍🙂

PR #1730, opened 2019-06-11
Add TIP that this step is not required for v8.x +

paulsterling:


sofietoft:
Hi @paulsterling 

I do believe you still need to add these dependencies, in order for things like validation to work 🤔
At least, the validation doesn't work for me, if I do not add these scripts to my template - in Form 8.1.1.

Which version of Umbraco Forms have you been testing with?

/Sofie

paulsterling:
Hi @sofietoft 

I added this after guidance from HQ Support who advised that I do not need the scripts with Umbraco v8+ and Forms v8+. When I follow the (current) steps to add the scripts, Forms displays an error message where the Insert Forms Marco is rendered in a View that informs that the scripts are missing? But without the scripts, there is no error message. Sounds like HQ Support (and me) did not fully consider the implications of leaving the scripts off.

If you'll leave this PR open, I'll revisit the correct approach for v8 and update accordingly. Please advise, if you know of the correct approach already.

-Paul

sofietoft:
Hi @paulsterling 

Ah, an error? Well, that doesn't sound right 🤔
I think what you should do here, is create a bug report on the [Umbraco Forms Issue Tracker](https://github.com/umbraco/Umbraco.Forms.Issues/issues) instead.
If you could provide some steps on how you reproduce this error, I'll do some more testing as well.

/Sofie

paulsterling:
Closing as I can't consistently repo ;-) 

There is a difference in the jQuery scripts version for Forms 7 and Forms 8, so I'll send a new PR for that, but happy to say this seems to be a local issue only.

-Paul

PR #1731, opened 2019-06-12
The do not upgrade list does not exsist anymore

Aaen:
Jim and Jacob have decided that we should not have the do not upgrade list anymore. So there will not be new on this list. 
Jim and I will work with the list over the next couple of weeks.


sofietoft:
Thanks @Aaen 🎉

Issue #1735, opened 2019-06-14
Searching for documentation in Bing or Duck Duck Go sometimes extra empty parts in the path


liamlaverty:
## What type of issue is it?
Other - SEO I guess?

## What article/section is this about? 
`ContentServices Events`

## Describe the issue:
When searching in Bing (I know, I know) or Duck Duck Go for `contentservice.trashed umbraco` the path for the document is `https://our.umbraco.com/documentation/Reference/Events/_/_/ContentService-Events` with an extra couple of `_/_/` in the middle, which leaves the page title with `-_-`'s in it, and the breadcrumb with blank links inside of it 

### Example
![image](https://user-images.githubusercontent.com/7046713/59510865-b8f39100-8eac-11e9-8728-47020b1eaa0b.png)

### Bing Example (broken)
https://www.bing.com/search?q=contentservice.trashed+umbraco

### Google example (working as expected)
https://www.google.com/search?q=contentservice.trashed+umbraco

### Duck Duck Go (broken) 
https://duckduckgo.com/?q=contentservice.trashed+umbraco

### Ask Jeves (working)
https://uk.ask.com/web?q=contentservice.trashed+umbraco

liamlaverty:
Just as a sidenote, this widget doesn't appear when the link has these extra path sections in it:

![image](https://user-images.githubusercontent.com/7046713/59511613-a5e1c080-8eae-11e9-996c-e37cdab02203.png)

nul800sebastiaan:
That's super weird, not sure how it's being indexed like that. Will have to look into this. 🤔

liamlaverty:
I found another one of these, so went to look into it a bit more. It looks like it's happening because of relative path `./` links in github, which are being imported as non-relative HTML `/_/` links in our.umbraco.com. Some pages have many relative links between eachother, so the paths can get enormous if `page A` references `page B` which references `page A`. Here's an example: 

https://our.umbraco.com/documentation/Tutorials/Starter-kit/Lessons/2-Add-a-Blog-Post-Publication-Date/part-2

Watch as the url grows as you navigate back and forward between pages. 
![url-bug-example](https://user-images.githubusercontent.com/7046713/59755153-5705a380-927f-11e9-82b6-b19b1369c05c.gif)

Not sure if this should be changed by updating the links in the github repo (I'm happy to do the grunt work on each link I find), or by updating the way the .md files are imported (I'm not sure I know enough about the importer to do that task).

[...]

PR #1743, opened 2019-06-19
fixes some of #1735

liamlaverty:
Fixes some of #1735


jmayntzhusen:
Sorry for the slow response, thanks a lot for fixing these though! 😊

PR #1744, opened 2019-06-19
multi url picker for umbraco v7.14+

beumsk:


sofietoft:
This is awesome @beumsk Thanks! 😁 

I have only one thing:
Shouldn't we add a link to this article on https://our.umbraco.com/Documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/index-v7 ? 🤔
Just an idea!!

/Sofie

beumsk:
Thanks @sofietoft !

I would like it, but it could confuse pre 7.14 versions users.
If I recall correctly the multi url picker is native starting that version only.

sofietoft:
Hi @beumsk 
Yeah, you're right 🤔 

Hm! Let me see if I can come up with a good plan here.
For now, I'll just merge this in 😁 

Thanks again!

Issue #1741, opened 2019-06-19
Example view of content app  should use umb-box directives


dawoe:
I see the view is using the markup generated by the umb-box directives.

Would it not be better to use these directives in examples. 

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/master/Extending/Content-Apps/index.md***

jmayntzhusen:
Hey @dawoe 

I made the example and am not that familiar with both AngularJS and the Umbraco backoffice directives, which is why it was made this way 😁

If there are changes that can improve the example and make it closer to a best practise then you are very welcome to provide a PR! 👍

dawoe:
As you requested. PR provided

jmayntzhusen:
Thanks for the PR 👌

PR #1746, opened 2019-06-21
Use angular directives for umb-box

dawoe:
Fixes issue #1741


jmayntzhusen:
Wow that was quick, thanks!

PR #1751, opened 2019-06-27
Add documentation about UmbracoMapper

jmayntzhusen:
Initial documentation for UmbracoMapper, a new mapping tool replacing AutoMapper in Umbraco 8.1.

https://github.com/umbraco/Umbraco-CMS/pull/5087

### DIFF
--- a/Reference/Mapping/index-vpre8.1.md
+++ b/Reference/Mapping/index-vpre8.1.md
@@ -1,0 +1,7 @@
+---
+versionTo: 8.1.0
+---
+
+# Mapping in Umbraco
+
+Prior to Umbraco 8.1 mapping was handled by an external dependency AutoMapper, to improve startup times and reduce external dependencies it has been replaced by UmbracoMapper in 8.1.0+.
--- a/Reference/Mapping/index-vpre8.1.md
+++ b/Reference/Mapping/index-vpre8.1.md
@@ -2,6 +2,7 @@
 versionTo: 8.1.0
 ---
 
-# Mapping in Umbraco
+# Object Mapping in Umbraco
 
-Prior to Umbraco 8.1 mapping was handled by an external dependency AutoMapper, to improve startup times and reduce external dependencies it has been replaced by UmbracoMapper in 8.1.0+.
+Often in code there is a need to 'map' one object's properties to another type of object, and the 'type of objects' are not related by inheritance or interface. (Think database layer object, passing information to a presentation layer ViewModel etc). In these circumstances, it can save time and provide consistency to consolidate the logic to map between the options into one set of 'Mapping' rules.
+Prior to Umbraco 8.1 mapping was handled by an external dependency AutoMapper, to improve startup times and reduce external dependencies it has been replaced in 8.1.0+ by a simpler Umbraco specific implementation called UmbracoMapper.
--- a/Reference/Mapping/index.md
+++ b/Reference/Mapping/index.md
@@ -12,9 +12,13 @@ This is not to be confused with the [package by Andy Butland](https://our.umbrac
 
 ## Accessing the UmbracoMapper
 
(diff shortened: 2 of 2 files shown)


Reference/Mapping/index-vpre8.1.md
```diff
@@ -0,0 +1,7 @@
+---
+versionTo: 8.1.0
+---
+
+# Mapping in Umbraco
```
marcemarc: Mapping objects from one type to another in Umbraco

or

Object Mapping in Umbraco

or

Model Mapping in Umbraco

Reference/Mapping/index-vpre8.1.md
```diff
@@ -0,0 +1,7 @@
...
+
+Prior to Umbraco 8.1 mapping was handled by an external dependency AutoMapper, to improve startup times and reduce external dependencies it has been replaced by UmbracoMapper in 8.1.0+.
```
marcemarc: I would add an opening paragraph:

Often in code there is a need to 'map' one object's properties to another type of object, and the 'type of objects' are not related by inheritance or interface. (Think database layer object, passing information to a presentation layer ViewModel etc). In these circumstances, it can save time and provide consistency to consolidate the logic to map between the options into one set of 'Mapping' rules.

marcemarc: replaced in 8.1.0+ by a simpler Umbraco specific implementation called UmbracoMapper.

Reference/Mapping/index.md
```diff
@@ -0,0 +1,190 @@
+---
+versionFrom: 8.1.0
+---
+
+# UmbracoMapper
+
+:::tip
+UmbracoMapper replaced AutoMapper which was an external dependency. AutoMapper builds the mapping code dynamically, based upon mapping profiles, which are defined as C# expressions. UmbracoMapper relies on static code, i.e. mappings need to be hand-written.
+
+This is not to be confused with the [package by Andy Butland](https://our.umbraco.com/packages/developer-tools/umbraco-mapper)
```
marcemarc: confused with the [V7 UmbracoMapper package by Andy Butland] of the same name.

Reference/Mapping/index.md
```diff
@@ -0,0 +1,190 @@
...
+:::
+
+## Accessing the UmbracoMapper
+
+The UmbracoMapper is registered with Dependency Injection (DI). It can therefore be injected in constructors, wherever DI is used. Alternatively, it is also exposed via `Current.Mapper`.
```
marcemarc: injected into constructors, of controllers, custom classes etc, wherever DI is used. Alternatively, it is also exposed via the Current service locator: 'Current.Mapper'

Reference/Mapping/index.md
leekelleher: Andy's "Umbraco Mapper" [also supports v8 too](https://web-matters.blogspot.com/2019/03/umbraco-mapper-new-releases-supporting-v8.html), _just to add to the confusion._

jmayntzhusen:
@marcemarc based on Stephans reply here I would say we just should specify it is free to use: https://github.com/umbraco/Umbraco-CMS/pull/5087#issuecomment-492118578

Will work on your other suggestions! 👍

Reference/Mapping/index.md
```diff
@@ -0,0 +1,194 @@
...
+:::
+
+Mapping with the UmbracoMapper works in ways very similar to AutoMapper:
+
+```cs
+// assuming source is ISource, create a new target instance
+var target = umbracoMapper.Map<ITarget>(source);
+
+// assuming both source and target already exists
+target = umbracoMapper.Map<source, target);
```
bergmania: There is a typo from the document where you copied this from 
```suggestion
target = umbracoMapper.Map(source, target);
```

PR #1754, opened 2019-06-28
Examine indexing

hartviglarsen:
This will add a very basic example of how to create a new index that only includes nodes of a certain document type. While it is not much, I believe it is better than having nothing. :) 

I am quite busy these days.. but let me know if there is something you would like added and I will see if I can manage. :)


jmayntzhusen:
Thanks for submitting this @hartviglarsen ! 

I think @Shazwazza recommends NOT to use custom indexes for things like that though, searching through a single doc type should be done on the query level, not the index, so may need to make a warning on that in there. Hopefully he can weigh in 😄

hartviglarsen:
@jmayntzhusen We recommended the same in V7 but people did it anyway, hence my PR. 😄 

I'll add a note about it.

jmayntzhusen:
Thanks @hartviglarsen 

This completes one of the tasks in https://github.com/umbraco/UmbracoDocs/issues/1685 as well, great job 🎉

PR #1756, opened 2019-07-02
Updated label names as per Umbraco 8.0.2

spudstuff:
Some of the labels in the documentation were using old names.


sofietoft:
Awesome @spudstuff  🎉 

Thanks for correcting these!
I've given you a Contrib badge on our 😎

/Sofie

PR #1757, opened 2019-07-02
Corrected import for Umbraco.Core.Composing

spudstuff:
Dashboard removal import for `Umbraco.Core.Components` was incorrect and should be `Umbraco.Core.Composing`.


sofietoft:
Thanks for catching this @spudstuff 🎉 😄

PR #1762, opened 2019-07-05
Fixed up examples and advice on Services Pattern

marcemarc:
Still a bit to do


marcemarc:
resolved merge conflict

Implementation/Services/index.md
```diff
@@ -232,10 +240,11 @@ namespace Umbraco8.Extensions
...
     public static class UmbracoHelperExtensions
     {
-            //NewsCategories //NewsSection
-            public static IPublishedContent GetNewsSection(this UmbracoHelper umbracoHelper)
+        public static IPublishedContent GetNewsSection(this UmbracoHelper umbracoHelper)
             {
-            return umbracoHelper.ContentSingleAtXPath("root/homePage/newsSection");
+                //assuming a single site with a single News Section at the top level
+                IPublishedContent siteRoot = umbracoHelper.ContentAtRoot().FirstOrDefault();
+                return siteRoot != null ? siteRoot.FirstChild(f => f.ContentType.Alias == "newsSection") : default(IPublishedContent);
```
Shazwazza: I don't think we should use `default(IPublishedContent)`, it should just be `null`. `IPublishedContent` isn't a struct so this equals null anyways.

Shazwazza: This can be rewritten like, but maybe you are keeping the syntax explicit for a reason?

```cs
var newsSection = siteRoot.?siteRoot.FirstChild(f => f.ContentType.Alias == "newsSection") ?? null;
```

Implementation/Services/index.md
```diff
@@ -302,19 +317,36 @@ using Umbraco8.Services;
...
+**"Singleton"** services can be injected into "Singletons" and below ⤵.
+- "Singleton" means that only a single instance of this type will ever be created for the lifetime of the application.
+
+**"Request"** services can be injected into "Request" based lifespans only
+- "Request" based lifetime means anytime this type is needed one new instance of this type will be created for the duration of the current HttpRequest. The object will be disposed of at the end of the current HttpRequest.
+
+#### Implementing the service
 
-Note we register the service in 'Lifetime.Request' scope - as we'll be making use of the transient UmbracoContext
+##### 1 - The service will ONLY be used in a Controller or View
```
Shazwazza: Maybe say:

> 1 - The service will ONLY be used during a request like in a Controller or View

Implementation/Services/index.md
```diff
@@ -323,40 +355,76 @@ using Umbraco.Web;
...
             return contactUs;
         }
     }
 }
 ```
 
-:::Note
-If we planned to use our SiteHelperService outside of a request where UmbracoContext is not available we'd use the UmbracoContextFactory + EnsureUmbracoContext() methods described above to access the content cache.
-:::
+##### 2 - Recommended approach: the custom service can be used anywhere in Umbraco!
```
Shazwazza: I'd prefer if we don't say 'recommended' here. This will mean that people will do this for everything which isn't really what we want. Maybe we just say 

> 2 - The service can be used within or outside of a web request

Side note - What we really would prefer people to do is never inject `UmbracoContext` at all and instead inject `IUmbracoContextAccessor` which is a singleton so it can be injected into Singletons. This follows the same principals as in asp.net core and we have accessors for many things. But there's a caveat to that, since `IUmbracoContextAccessor.UmbracoContext` will still be null in a non-request thread so you'd still need to do the `IUmbracoContextFactory` if it was in a non-request thread. I dunno, maybe we shouldn't confuse this so much. I just want to mention somewhere that `IUmbracoContextAccessor` should be the thing people inject if they need an UmbracoContext and they are not using one of our base classes. It also means they will know if they are running in a web request or not.

Implementation/Services/index.md
```diff
@@ -369,51 +437,80 @@ namespace Umbraco8.Controllers
...
             //etc          
             // Do other stuff here!, then return the custom viewmodel to the template view.
             return CurrentTemplate(blogPostViewModel);
-
         }
     }
 }
 ```
+:::warning
+This isn't truly 'best practice' when using DI. This 'only works' in Umbraco because when a dependency for the base RenderMvcController isn't supplied via a constructor, Umbraco 'falls back' and uses the Service Locator pattern to inject the missing elements. This enables developers to choose to ignore DI, but if trying to following DI best practice, and to make the controller 'unit testable' - use the following example instead.
```
Shazwazza: should say "use the following example instead which supplies all constructor parameters for the base class."

Implementation/Services/index.md
marcemarc: updated, you 'cool kids' with your fancy c# 6 syntax... it will never catch on :-P have added a comment above to be clear, and updated other instances of same.

Implementation/Services/index.md
marcemarc: Text updated...but 

In terms of IUmbracoContextAccessor, I get that this much more preferable than injecting UmbracoContext, but looking at IUmbracoContextFactory, my understanding is this uses IUmbracoContextAccessor to determine if an Umbraco Context exists first before falling back to create one. (I may be wrong)

https://github.com/umbraco/Umbraco-CMS/blob/fca74ad8d7e72e4a5981cd90e4725ad0f7cd4a82/src/Umbraco.Web/UmbracoContextFactory.cs#L74

So is there any downside when trying to get hold of an UmbracoContext to just using  UmbracoContextFactory? it seems to always do what you'd want, eg it always checks to use the existing UmbracoContext, but gets a new one if it isn't there... (and as a lazy dev, I don't have to do anything or don't care, I got my context... and I've just one thing to inject...)

I can see that using IUmbracoContextAccessor would be useful in the scenario of finding out whether or not you were in a non-request thread or not...  but I can't think of a useful example of this scenario (where you don't just always want an UmbracoContext anyway...). Maybe there could be a flag on the UmbracoContext saying IsRequestThread? if this were useful?

Anyway I think people will run into IUmbracoContextAccessor in IntelliSense, and from its name deduce what it could be used for - so I think it's good to acknowledge its existence and how it's better than 'UmbracoContext injecting' and the caveat of the context of a non-request thread... if people are puzzled why sometimes it's null... 

... so I've added an Aside paragraph 

###### Aside: What is the IUmbracoContextAccessor then?

hope that makes sense and warns people off injecting UmbracoContext...

Shazwazza: @marcemarc thanks! and yes that is fine ... possibly in project UniCore ... we are trying to refrain from having request based only dependencies, so in UniCore, you won't be able to just inject the UmbracoContext, you will need to access it by the IUmbracoContextAccessor, just like in .Net Core you must inject the IHttpContextAccessor if you want to use it in services. BUT ... this isn't set in stone, so we'll see what happens. In the meantime, in v8 there's various things people can use and I think you've highlighted them nicely.

PR #1767, opened 2019-07-09
Updated the update instructions for 8.1

FransdeJong:
I updated the update instructions for 8.1 because they were unclear. They stated that we need to update `PropertyValueConverterBase` to `IPublishedPropertyType`. That's not true. Within this class the reference to `PublishedPropertyType` needs to be updated to `IPublishedPropertyType`.


nul800sebastiaan:
Thanks, well spotted! 👍

PR #1775, opened 2019-07-15
index-v7.md Examine Fluent Api example doesn't work

dchallener:
Amended v7 Fluent API example to convert ISearchResults to IEnumerable<IPublishedContent> so Url and Name are exposed.
Raised on forum https://our.umbraco.com/forum/using-umbraco-and-getting-started/98150-fluent-api-following-documentation-errors

### DIFF
--- a/Reference/Searching/Examine/Quick-Start/index-v7.md
+++ b/Reference/Searching/Examine/Quick-Start/index-v7.md
@@ -87,18 +87,19 @@ var query = Request.QueryString["query"];
 var searcher = Examine.ExamineManager.Instance.SearchProviderCollection["ExternalSearcher"];
 
 var searchCriteria = searcher.CreateSearchCriteria(Examine.SearchCriteria.BooleanOperation.Or);
-var searchQuery = searchCriteria.Field("nodeName", query.Boost(5)).Or().Field("nodeName", query.Fuzzy()).And().OrderByDescending("createDate");
+var searchQuery = searchCriteria.Field("nodeName", query.Boost(5)).Or().Field("nodeName", query.Fuzzy()).And().Field("__IndexType", "content").And().OrderByDescending("createDate");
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Hey @dchallener 

Thanks a lot for taking the time to update this example, there are a few flaws in it like you pointed out.

Another flaw which is also in your example is if the results return an index type which is not content, so media for example it will YSOD when trying to find the typed content for it.

I tested it out a bit and found this to work, what do you think about replacing the current example with this?

```cs
@{
    var query = Request.QueryString["query"];
    var searcher = Examine.ExamineManager.Instance.SearchProviderCollection["ExternalSearcher"];

    var searchCriteria = searcher.CreateSearchCriteria(Examine.SearchCriteria.BooleanOperation.Or);
    var searchQuery = searchCriteria.Field("nodeName", query.Boost(5)).Or().Field("nodeName", query.Fuzzy()).And().Field("__IndexType", "content").And().OrderByDescending("createDate");
    var searchResults = searcher.Search(searchQuery.Compile());
    if(searchResults.Any())
    {
    <ul>
        @foreach (var result in searchResults)
        {
        var node = Umbraco.TypedContent(result.Id);
        <li>
            <a href="@node.Url">@node.Name</a>               
        </li>
        }
    </ul>
    }
}
```

The only differences to your example is that I added a content filter on the results: `And().Field("__IndexType", "content")` - and that I ran the Umbraco helper inside the loop on the already exposed id, not sure of any performance implications this may have.

dchallener:
Hi @jmayntzhusen 

Thanks for reviewing this - agreed with the TypedContent and limiting the search index to just content - in reality, I think there are quite a few things we'd do differently but this is at least now a working snippet.

I'm not sure of the performance either but have updated the commit to your way as it reads a little easier!

jmayntzhusen:
Great, thanks again for helping out with the documentation! 🙂👍

PR #1780, opened 2019-07-23
Update index.md

AxendoEelco:
On inspecting the RedirectTrackingComponent component, which is the v8 equivalent of RedirectTrackingEventHandler I've noticed there no longer is a check for the existence of a 3rd party DLL with redirect functionality in the solution. Changed the documentation accordingly. 

I don't know of this is by design or that the check is simply missing in the v8 codebase, so maybe this should be checked first.


jmayntzhusen:
Hey @AxendoEelco 

I talked to one of the Umbraco Core devs, and it is by design currently, mainly because no packages exist for v8 that supports this. It may change in the future, but we also have heard from a package developer that it may work side by side with our implementation without needing to disable the core way.

Will merge this in for now, thanks for helping to keep the docs up to date! 🎉👍

PR #1782, opened 2019-07-24
Update outbound-pipeline.md

benoitcornette:
Change of var name "orginalUrl" to "originalUrl" (typo in the current documentation)


jmayntzhusen:
Nice catch, thanks! 🙂

PR #1785, opened 2019-07-24
fixed typo breaking the controller

beumsk:


jmayntzhusen:
Yep that looks right, thanks for making the update 🙂

PR #1798, opened 2019-08-08
amended example to correctly set the minimum level and take the setti…

tom-ingeniuus:
…ng from an AppSetting in web.config

The original example didn't work as the ReadFromConfigFile overwrote the call to set a level.
I also felt it should be simple to set different levels for different environments
t

### DIFF
--- a/Getting-Started/Code/Debugging/Logging/index.md
+++ b/Getting-Started/Code/Debugging/Logging/index.md
@@ -218,7 +218,7 @@ In the `/config/serilog.user.config` file you can add the following lines, which
 
 ### Full C# control over Serilog configuration
 
-If you like using Serilog but prefer to use C# to configure the logging pipeline then you can do so with the following example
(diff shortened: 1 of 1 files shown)


tom-ingeniuus:
totally agree Marc, namespaces should always be included in coed samples, added them now,
best
t

marcemarc:
@tom-ingeniuus thanks - I've merged!, I'll ask Jesper to trigger the build to update Our Umbraco

gavinfaux:
Can anyone assist with an updated solution for 8.6 as both the WebRuntime constructor and GetLogger override are marked as [Obsolete]?

The new constructor requires a mainDom argument but can't work out how to resolve/inject as the implementation in base class uses private types (to support new SQL dom locking features?)

sofietoft:
@gavinfaux - apologies for the delay!

I've had a chat with the rest of the docs team about this, and it seems that what you're trying to do here is basically impossible in 8.6.0 and 8.6.1.
There has, however, been made a PR: https://github.com/umbraco/Umbraco-CMS/pull/7998
When this is merged, it should be possible - we'll see what we can do about having updating docs ready for the release as well! 💪 
I hope that to some degree answers your questions.

gavinfaux:
@sofietoft thanks for swift reply will keep on eye on PR

PR #1799, opened 2019-08-08
Add automatic style guide testing and description

jmayntzhusen:
This PR will add the Vale files required to do automatic style rules testing as well as an article explaining what the rules are and how it works.

NOTE: When we added the .github folder in the past it was added to the sitemap and thus the menu of the documentation. Because the sitemaps top level is only based on folders the `_vale.ini` file will not cause any issues.

As for the reason the Vale config file is in the root, it is because the Vale tool for VS Code doesn't know how to find it if it's somewhere else, even though Vale itself has a --config="path" parameter in the cmd line.


Contribute/Style-Guide/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Using Vale locally
+
+One of the big strengths of Vale is that it is possible as a contributor to simply run the tests locally before you create a PR, there is also a VS Code extension for it.
+
+First step to running Vale locally is to install it following [their documentation](https://errata-ai.github.io/vale/#installation).
+
+Next you can open a command line tool in the documentation repo and run the following command:
+
+```vale --glob='*.md' .```
```
sofietoft: Not sure what's wrong here, but it looks like the formatting of the rest of the article is a bit funky after this line 🤔

jmayntzhusen: It looked fine in a Markdown previwer locally, and also looks fine if you preview the document in Github, but yea the syntax coloring is weird in Github after this..

sofietoft: Probably just a GitHub thing then 😁

dampee: Maybe try replacing it with 4 spaces in front.
Or put in on three lines.

PR #1807, opened 2019-08-16
V8 Create DB table example

warrenbuckley:
Adds in some example code to show how to create a custom DB table by using Migrations - which is a big topic in itself to try & cover


jmayntzhusen:
Yo @warrenbuckley 

This is great! I added a bit of text and an image at the top. Will merge it in now, thanks a lot for the addition! 😁

Issue #1810, opened 2019-08-20
applications.config no longer present in v8 - document needs updating.


John-Blair:
applications.config no longer present in v8 - document needs updating.

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/master/Extending/Section-Trees/sections.md***

nul800sebastiaan:
Yep, that should probably have a "versionTo" field (I can't remember the exact notation). This file doesn't exist on v8 and never will. :)

crgrieve:
I have set his using versionRemoved rather than versionTo... hope this works?

https://github.com/umbraco/UmbracoDocs/pull/1812

crgrieve:
@nul800sebastiaan I submitted a PR for the applications.config adding versionRemoved. Hopefully this is right way?

If so, could also do the dashboard.config? Is it completely removed in v8 too?

John-Blair:
I'd add 
404handlers.config
baserestextensions.config
embeddedmedia.config
examineindex.config
examinesettings.config
feedproxy.config
filesystemproviders.config
log4net.config
scripting.config
trees.config

Thanks

jmayntzhusen:
While none of these exist in v8 the question is how to present it in the docs.. 

I'd say ideally we would have an 8+ version of these articles simply saying the config file is not there in v8 and then linking to an article describing how to do it instead..

Of course this would require that we had articles describing the v8 equivalent of doing stuff in the config files which is currently not the case 😬

Think we will discuss the best way to handle these kinds of things at our next [docs curator](https://our.umbraco.com/community/the-documentation-curators/) meeting 🙂

crgrieve:
ok, thanks @jmayntzhusen :) I'll wait to see the approach before updating any others. Happy to update and contribute further based on what you decide.

jmayntzhusen:
Hey everybody

Will close this issue for now, Marc has made a new more general issue for what to do with the missing config options in v8 here: #1817 

We have decided that the most user friendly thing to do here is to have v8 versions of all of the config articles that then explain that the config file is not there in v8, and optimally it would also link to or include some examples of how things would then be configged through code 🙂

PR #1812, opened 2019-08-21
Adding the version removed

crgrieve:
Adding the version removed as per issue https://github.com/umbraco/UmbracoDocs/issues/1810

### DIFF
--- a/Extending/Section-Trees/sections.md
+++ b/Extending/Section-Trees/sections.md
@@ -1,5 +1,6 @@
 ---
 versionFrom: 7.0.0
+versionRemoved: 8.0.0
 ---
 
 # Sections
--- a/Extending/Section-Trees/sections.md
+++ b/Extending/Section-Trees/sections.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
-versionRemoved: 8.0.0
 ---
 
 # Sections
--- a/Reference/Config/applications/index.md
+++ b/Reference/Config/applications/index.md
@@ -1,6 +1,6 @@
 ---
 versionFrom: 7.0.0
-needsV8Update: "true"
+versionRemoved: 8.0.0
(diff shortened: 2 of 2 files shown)


jmayntzhusen:
Thanks a lot for updating this article @crgrieve 

Will leave it pending for now until our next curators meeting, read more here: https://github.com/umbraco/UmbracoDocs/issues/1810#issuecomment-523403540

jmayntzhusen:
Hey @crgrieve 

So sorry I forgot to update this PR after our meeting 😬

We ended up agreeing on the following:
For each config file that no longer exists in v8 we would like a v8 versioned article for them. Those articles can initially say something along the lines of:

> This config file no longer exists in Umbraco 8. 
>The configurations that this file used to handle are very likely done through code in Umbraco 8, but for now we have no docmentation on doing that.
> If you hit this page and needed to know how to do this, please create an issue on the [documentation issue tracker](https://github.com/umbraco/UmbracoDocs/issues) and we will try to bump it up on our list of things missing documentation!

Then with time of course - and feel free to help out with this if you can - we would like to edit that original message to actually tell people how to do that config in v8 (or link to it if examples already exists). 

We welcome any suggestions you may have to this approach, we feel it will atleast prevent people from trying things that we are sure aren't working, and long term we will provide the equivalent value, that just isn't set through config files but rather through code (likely composers if I understand it correctly).

crgrieve:
Thanks for the reply @jmayntzhusen :)

I can add in they v8 placeholder versions if that's useful? Then can think of the v8 specific instructions for each config as you have mentioned?

dampee:
Hi @crgrieve 
Sorry for the waiting.  But we had to discuss the meta tags as we where not sure anymore ourselves. :-D
Thanks for the effort!
Kind regards
Damiaan

PR #1813, opened 2019-08-21
Adding content app tutorial link

crgrieve:
I was going to create a content app tutorial but there already is one :) So just linking to it from tutorials section incase people are expecting it to be there with other tutorials.


jmayntzhusen:
Thanks @crgrieve !

We unfortunately have a few tutorial style articles out there not in the tutorials section. Will require a larger restructure for now to get it all moved around, so will leave for when we do that in the future, a link is great though!

PR #1819, opened 2019-08-27
updates a link with `//` to `/`

liamlaverty:
A tiny typo fix on the more `info -> querying` link


sofietoft:
Good catch @liamlaverty 👏 
Thanks!!

PR #1823, opened 2019-08-29
Update code snippets language

rasmusjp:
Right now the C# code snippets is broken on Our, I've changed the language from `cs` to `csharp` which should fix it


sofietoft:
Awesome, @rasmusjp 

Let's try it out! 😁

PR #1824, opened 2019-09-01
Fix typos in code examples

ProNotion:
Fixes missing quote and typo in variable name


sofietoft:
Good catch @ProNotion ! 💪 

Thanks for fixing this!

PR #1831, opened 2019-09-04
Adds missing parameter to *ModZero IsHelpers

### LINKED ISSUE #1830: Missing information about some IsHelpers
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/master/Reference/Querying/IPublishedContent/IsHelpers.md***

The `.IsModZero()` and `.IsNotModZero()` helpers are missing the important first `modulus`  parameter.

Also, in the general explanation about the usage, it's not clear what happens if you only supply a single string argument and the condition evaluates to `false` - does it return `null` or maybe an empty string? (_Spoiler alert:_ It's an empty string).

<details>
<summary>Why was I looking for this?</summary>
There's a subtle difference, since they're all primarily used for adding classes or other attributes, and a `null` tells Razor not to render the attribute (with a couple of exceptions) whereas an empty string will always be rendered.
</details>

I'll provide a PR with the missing docs shortly...

greystate:
This adds the missing `modulus` parameter to the `.IsModZero()` and `.IsNotModZero()` helpers.

Also cleaned up the descriptions a bit and added information about the scenario where you've supplied a single string argument to an IsHelper and the condition evaluates to `false` - where an empty string gets returned.

Closes #1830

### DIFF
--- a/Reference/Querying/IPublishedContent/IsHelpers.md
+++ b/Reference/Querying/IPublishedContent/IsHelpers.md
@@ -65,11 +65,11 @@ Test if the current node is at the specified index in the collection
 
 Test if the current node is not at the specified index in the collection
 
-### .IsModZero([string valueIfTrue][,string valueIfFalse])
+### .IsModZero(int modulus[,string valueIfTrue][,string valueIfFalse])
 
 Test if the current node's position is evenly divisible by a given number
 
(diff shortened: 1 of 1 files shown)


nul800sebastiaan:
Awesome, thanks very much on the clarification! 👍

PR #1832, opened 2019-09-05
Fix syntax bug in code example

abhamla:
Removed second bracket ")" under **Without ModelsBuilder**


sofietoft:
Good catch, @abhamla ! 😄 
Thanks for fixing the syntax bug.

Can see it's your first time contributing to the Umbraco Docs? That's awesome!! #h5yr 😎 
Do you have an account on [Our Umbraco](https://our.umbraco.com)?
If you do, give me your username / user ID, and I'll grant you to "C-Trib" badge. You've sure earned it 😁

PR #1834, opened 2019-09-09
Added missing property to ExamineIndex instructions

f33n3y:
The instructions for using Examine pre v0.1.80 with Azure were missing the second property which is required to use the 'Sync' behaviour.
I have added this missing property in.


jmayntzhusen:
Hey @f33n3y 

Thanks again for updating the documentation! I have given you a contributor badge on your Our profile now 🎉

PR #1849, opened 2019-09-15
Added link for the Visual Studio extension

abjerner:
There are currently two Models Builder extensions in the Visual Studio marketplace. One is deprecated, and there shouldn't be used.

This change makes sure to link directly to the new and correct Visual Studio extension.


dampee:
Thanks a lot for noticing @abjerner !

PR #1850, opened 2019-09-15
Models Builder - Updated the section about API mode

abjerner:
The section explained how to run the *Custom tool*, which was for the old Visual Studio extension. The new extension works slightly different, so I've updated the docs accordingly, as well as included a few more details.


sofietoft:
This is great, @abjerner 👏 

As soon as I've had some one here at Umbraco HQ give it a quick look, we'll merge it in!

jmayntzhusen:
Thanks @abjerner 🙂

PR #1856, opened 2019-09-16
Clarifying Umbraco version settings

PaulAik:
Config settings for Umbraco XML cache file and other TEMP files differ depending on Umbraco version used. Proposed change clarifies only one entry is required if on version 7.7.3+


sofietoft:
Thanks for this @PaulAik !!

I don't know much about these things, so I'll need to run it by an expert before we can merge! 😁

Shazwazza:
yep looks good

PR #1858, opened 2019-09-17
New dedicated workers public IP address

pijemcolu:
**What**
Adding a new outgoing IP address to the the rewrite rules section.

**Why**
This is our only documentation where the end users can potentially find the outgoing IP addresses that websites are reaching out from.

**Suggestion**
We could think about this change as beloning to a a new networking or a security section where it would be much easier to find networking rules, AP addresses, firewall rules for sql, basic authentication and such.


jmayntzhusen:
Thanks!

PR #1861, opened 2019-09-17
Fix typo

nathanwoulfe:
'useful to flog' becomes 'useful to log'


jmayntzhusen:
Haha nice catch! Don't want to give people the wrong idea 😄

PR #1866, opened 2019-09-20
removed v7

skartknet:
removed v7


jmayntzhusen:
Thanks a lot, not sure how we missed that one 😄

PR #1868, opened 2019-09-23
Update index-v7.md

craigs100:
Just worked through this. The string.split method gives these errors:-
System.Linq.Enumerable+WhereSelectArrayIterator`2[System.String,System.String]


sofietoft:
Thanks for this @craigs100 

Can I ask which version of Umbraco you used when you ran into the errors here? 🤔

craigs100:
Hi @sofietoft, it was 7.14.0.

sofietoft:
I tested this  just now in a much earlier version - Umbraco 7.6.0.
Seems that the snippet in our Docs doesn't work for that version either.

I'll need to track down, when this was changed!
Thanks for catching this @craigs100 👏 🎉

PR #1889, opened 2019-10-01
Update formatting in ReleationService

hartviglarsen:
This triggered me a bit.

I can see that the article needs to be updated/verified for V8. I'll do that one of the following days.


jmayntzhusen:
Thanks for the PR @hartviglarsen 🙂

PR #1905, opened 2019-10-02
updated doc to v8

beumsk:


sofietoft:
Thanks! 😄 🎉 
This is really great. 
We look forward to even more contributions from you!

Issue #1853, opened 2019-09-15
Verify articles for Umbraco 8


sofietoft:
# Help wanted
Many of our articles needs to be verified against Umbraco 8, and for that we need your help!! 🙏

## What can you do to help?
We've put together a list of articles that needs to be verified and updated for Umbraco 8. 
We need you're help to do the following:
* Pick an article on the list
* Read it!
* Test out any code snippets, tutorials and step-by-step's in the latest version of Umbraco 8

Does the article still apply for Umbraco 8?
* **Yes**: Tell us in a here, and create a PR where you remove the `needsV8Update` tag at the top of the article.
* **No**: Tell us in an issue, OR create a new Umbraco 8 compatible article about the same topic!

## The List

- [x] [Backoffice Tours](https://our.umbraco.com/documentation/Extending/Backoffice-Tours/)
- [x] [File System Providers](https://our.umbraco.com/documentation/Extending/FileSystemProviders/)
- [x] [Macro Parameter Editor](https://our.umbraco.com/documentation/Extending/Macro-Parameter-Editors/)
- [x] [Value Converters](https://our.umbraco.com/documentation/Extending/Property-Editors/value-converters)
- [x] [Value Converters - Interface Example](https://our.umbraco.com/documentation/Extending/Property-Editors/value-converters-full-example-interface)
- [x] [Value Converters - Attributes Example](https://our.umbraco.com/documentation/Extending/Property-Editors/value-converters-full-example-attributes)
- [x] [Creating Forms (not Umbraco Forms)](https://our.umbraco.com/documentation/Getting-Started/Code/Creating-Forms/)
- [x] [Debugging](https://our.umbraco.com/documentation/Getting-Started/Code/Debugging/)
- [x] [Source Control](https://our.umbraco.com/documentation/Getting-Started/Code/Source-Control/)
- [x] [Templates](https://our.umbraco.com/documentation/Getting-Started/Design/Templates/)
- [x] [Templating (Reference Section)](https://our.umbraco.com/documentation/Reference/Templating/)
- [x] [Tutorial: Add Google Authentication](https://our.umbraco.com/documentation/Tutorials/Add-Google-Authentication/)
- [x] [Tutorial: Creating a Property Editor](https://our.umbraco.com/documentation/Tutorials/Creating-a-Property-Editor/)
- [x] [Tutorial: Creating an XML Sitemap](https://our.umbraco.com/documentation/Tutorials/Creating-an-XML-Site-Map/)
[...]

PR #1907, opened 2019-10-02
Verifying File System Providers for v8

rhtSingh:
I've tested File System Providers in v8, and there are no changes that need to be made.
I have removed the "needsV8Update" flag.

### DIFF
--- a/Extending/FileSystemProviders/index.md
+++ b/Extending/FileSystemProviders/index.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
-needsV8Update: "true"
 ---
 
 # Custom file systems (IFileSystem)


rasmusjp:
I think the article needs updating for v8, because both `/config/FileSystemProviders.config` and `FileSystemProviderManager` doesn't exist anymore.

sofietoft:
This is great @rhtSingh , thanks for running the tests! 😄 

Could you share the steps you followed to test this?

sofietoft:
Have a look at this comment here: https://github.com/umbraco/UmbracoDocs/issues/1853#issuecomment-539999779 

Seems this is documented in the source code, and there are quite a few changes.
I'll close this one for now.

PR #1911, opened 2019-10-02
Create .nanorc for nano users

Yash-Garg:


jmayntzhusen:
Hey @Yash-Garg 

We are not really interested in cluttering the repo with a bunch of editor specific files. We already have the editorconfig that spans multiple editors. Thanks for taking the time but this PR will be declined. Looking forward to future contributions! 🙂

PR #1913, opened 2019-10-02
Add charset and whitespace trimming in editorconfig

Yash-Garg:


jmayntzhusen:
Hey @Yash-Garg 

My understanding is that it will default to UTF-8 anyways, so no reason to specify it.

And the trim whitespace change is already in the document you just added a duplicate? Will close this PR based on the above 🙂

Yash-Garg:
Ah, didn't see that this was a duplicate.

PR #1918, opened 2019-10-02
v8 folder structure updates

crgrieve:
*Adding new screenshots of folder structure in v8
*Removing references to umbraco_client folder as no longer exists.

Thought I would make this a separate PR as there may be feedback on more items needing updating depending on this.

Thanks,
Carole


kows:
Shouldn't original version be copied to a "index-v7.md" ?
Version info can be provided by https://our.umbraco.com/documentation/Contribute/adding-metadata .

sofietoft:
Thanks @crgrieve ! 😄 

A few notes for you here:
As @kows also mentions, I think we need to version this article.
`Umbraco_client` is still i Umbraco 7, right?
So, I think it would be great if you would make another article, "index-v7.md". It should contain the same content, except leave the mentions of `Umbraco_Client` in the one for Umbraco 7 💪

crgrieve:
Sure, I will make a v7 version :)

Getting-Started/Code/Source-Control/index.md
```diff
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
```
kows: update to 8 ? :)

crgrieve:
@sofietoft , v7 versions added :)

sofietoft:
Brilliant! 😄 Thanks @crgrieve 🎉

PR #1925, opened 2019-10-03
Serilog refactor

crgrieve:
Refactor to move the logging config information to config section where other config files are defined.

Copied this content from Serilog page (and linked to from old location). I thought it was more suited to config section?

I have linked from the config index to this and therefore have created a v7 version with the log4net reference. Have left as 'needsv8update' as there are more considerations for config that are changable so I want to just limit this to logging for this PR.

Hopefully this is ok :)


sofietoft:
This is awesome @crgrieve ! 😄 

It makes so much sense to have all this information in the Config section instead.
I do have one little thing though:
You've made a `-v7` version of the Config/index file, which is awesome!
Both the index files have `verionFrom: 7.0.0` though. Could you update it to `8.0.0` on `index.md`? Leave the `needV8update`, as you also mention, it could use some more work.

Tbh I'm not sure how it will look once we have both tags on one article, but it makes sense logically to have both of them on this article 😅 Let's see what happens!
What'd you think?

crgrieve:
I was wondering about that to be honest, wasn't sure if the *-v7 naming convention would be enough to differentiate the 2. Didn't know if adding versionFrom:8.0.0 would imply this only got added in 8?

But happy to change it :)

crgrieve:
updated :)

jmayntzhusen:
Thanks a bunch yet again @crgrieve 🎉😁

PR #1926, opened 2019-10-03
v8 modifications and v7 versions

crgrieve:
v8 modifications to some config and adding v7 version when needed.

Adding content to the general-v7.md that was accidentally missing from PR #1918 earlier today!

Thanks,
Carole :)

### DIFF
--- a/Getting-Started/Setup/Server-Setup/permissions-v7.md
+++ b/Getting-Started/Setup/Server-Setup/permissions-v7.md
@@ -87,6 +87,14 @@ are used for css files</p>
 </td>
 </tr>
 <tr>
+<th>/Masterpages</th>
+<th>Modify / Full control</th>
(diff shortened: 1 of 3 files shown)


sofietoft:
Thanks again @crgrieve 😄 🎉 
You're really on fire this week - it's amazing!!

Regarding the change you've made to the Web.config file here - have a look at this issue I just created this morning: #1928 

We need to have the entire article verified and updated against Umbraco 8.
My suggestion would be, that you remove it from this PR. Then, if you have time, take a crack at updated the entire article?
Or, you could also add a mention of the `umbracoUseSSL` key, in a comment on the issue 🤔
Completely up to you!

crgrieve:
Thanks @sofietoft , web config change reverted now.

PR #1938, opened 2019-10-06
V8 property type tutorial

crgrieve:
Making a v7 page and small change to v8.

### DIFF
--- a/Tutorials/Creating-a-Property-Editor/index-v7.md
+++ b/Tutorials/Creating-a-Property-Editor/index-v7.md
@@ -101,7 +101,7 @@ Now our basic parts of the editor is done, namely:
 - The controller for wiring up the editor with angular.
 
 ## Register the data type in Umbraco
-After the above edits are done, restart your application. Go to the Settings section, click the 3 dots next to the data types folder and create a new data type called "markdown". In the editor you can now select a property editor, where your newly added "My markdown editor" will appear.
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks @crgrieve ! 😄 

Quick question: Have you tested the whole Property Editor Tutorial against Umbraco 8?
Just wondering, as you haven't made Umbraco 8 articles for those.

crgrieve:
Hi @sofietoft , no not yet.

I think step 4 will require a bit of rework for the DB side of things. I will need to read up on that a bit more before I can do a PR for that.

Thanks,
Carole

sofietoft:
Ah, makes sense 😄 

I'm not sure it makes sense to have only the index of the tutorial versioned, as that would indicate that the entire article has been verified against Umbraco 8 🤔 
I realize that updating the entire tutorial is quite a big task.
For this PR I have 2 suggestions:

1. We close it for now or add a "WIP" mention.
2. We can add a mention on the Umbraco 8 index page, saying that the Umbraco 8 version of the tutorial is yet to be completed.

I'm open for other suggestions as well of course 😄 
What'd you think?

crgrieve:
Hi @sofietoft , I am happy to update to take whichever is the preferred approach :)

sofietoft:
I've done some thinking on this .. 
I think we should close it for now.

When you or me or something else, starts working on getting the entire tutorial here updated, it'll be easier to merge everything at once, without having to worry about other PRs on the same section 😅

PR #1945, opened 2019-10-07
Update index.md

sakshamtaneja21:


sofietoft:
Thanks, but I'm afraid we cannot accept this PR.

The YAML tags, e.g. `versionFrom: 6.0.0` and `versionTo: 7.0.0` needs to be at the top of the article always.

PR #1946, opened 2019-10-07
Update index.md

sakshamtaneja21:


sofietoft:
Thanks, but I'm afraid we cannot accept this PR.

The YAML tags, e.g. `meta.Title`, `meta.Description` and `versionFrom` needs to be at the top of the article always. Otherwise they'll be rendered in the article, which is not the intention.

You can read more about the YAML tags here: https://our.umbraco.com/documentation/Contribute/adding-metadata

PR #1954, opened 2019-10-07
V8 sitemap tutorial

crgrieve:
As per #1853 , verifying v8 article for sitemap.

Thanks,
Carole


Tutorials/Creating-an-XML-Site-Map/index.md
```diff
@@ -91,8 +90,13 @@ Now editors have the ability to set these values for each page of the site. Rath
...
+@inherits Umbraco.Web.Mvc.UmbracoViewPage
+@{
+    Layout = null; 
+    Response.ContentType = "text/xml";
+ }
+ 
+ <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemalocation="http://www.google.com/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">[INSERT SITE MAP CONTENT HERE]</urlset>
 ```
 
 Notice how we're not adding any spaces or carriage returns before the `<urlset>` opening declaration, even though it would be easier to read if we did, we want to avoid making the XML invalid.
```
marcemarc: Hi @crgrieve did changing the view so that it has spaces and carriage returns before the starting <urlset> make the generated sitemap invalid? if it's still valid then we can remove the paragraph about 'Notice how we're not adding any spaces...

crgrieve: Hi @marcemarc , It rendered in browser and I ran the output XML through an XML validation and it passed. I can't run it against the checker on the article as the sitemap isn't public.

Is this ok? If not I can revert back to previous formatting.

marcemarc: cool, I did the original tutorial based on an Umbraco4 snippet, and that if I remember rightly fell foul of the spaces back in the day... so if it all works fine it's much more readable like you have it!, I was just interested :-) not challenging you to a duel! but means we can also take out the paragraph that explains that people should get rid of whitespace too!

crgrieve: Is that ok testing I have done? Was the XML invalid but still loading in browser when it was breaking previously or will it being loading in browser mean all is ok now?

If so I will remove that paragraph :)

marcemarc: ultimate, are we sure, super sure, thing would be to put the output in a .xml file and put it somewhere on a server and point: https://www.xml-sitemaps.com/validate-xml-sitemap.html at it, it's soooo long ago I can't remember but the Xml would be valid with the spaces, it's just something superstitiously I carry around in my head, it may no longer be based on fact :-)

crgrieve: updated to remove that paragraph as discussed :)

marcemarc: woot

sofietoft:
All looks great @crgrieve 😄 🎉 
Thanks!

I'll merge this - @marcemarc has approved of course 😉

PR #1964, opened 2019-10-10
WIP - v8 load balancing documentation

Jeavon:


Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Lucene/Examine configuration
+
+The single instance backoffice Web App should be set to use [SyncTempEnvDirectoryFactory](file-system-replication.md#examine-directory-factory-options).
+
+The multi instance front end Web App should be set to use [TempEnvDirectoryFactory](file-system-replication.md#examine-directory-factory-options).
+
+### Umbraco TEMP files
+
+Store the Umbraco temporary files in the local server's 'temp' folder. Achieve this by changing this configuration setting to 'true' in the web.config. The downside is that if you need to view this configuration file you'll have to find it in the temp files. Locating the file this way isn't always clear.
```
marcemarc: I'm not sure everybody knows which Temp files you are referring to here, but that maybe obvious in the context of where this page sits... but possibly could we say something like

When an instance of Umbraco runs it generates some 'temporary' files on disk... in a non-cloud environment these would be created within the folders of the Web Application...however in an Azure Web App we want these to be created in the local storage of the actual server that Azure happens to be using for the Web App and so  we set this configuration setting to 'true', and the temporary files will be located in the .net temporary folder. This is great for 'speed' of access for Umbraco operation, however, the downside is it's very difficult on Azure to manually browse to this temporary location if you are troubleshooting for any reason.

(Also 'if you need to view this configuration file' ??? with nucache we no longer have an umbraco.config? published cache so is it wrong to refer to it as a configuration file??)

Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md
```diff
@@ -0,0 +1,58 @@
...
+{
+    IgnoreLocalDb = true
+});
+```
+
+### Steps to set-up a environment
+
+1. Create an Azure SQL database
+2. Install Umbraco on your master environment and ensure to use your Azure SQL Database
+3. Install Umbraco on your front-end environment and ensure to use your Azure SQL Database
```
marcemarc: Should we mention the word 'Replica' here? for the front end environment...

front end implies to me a little headless situation, where the actual Umbraco application isn't deployed... but that isn't the case (yet) where in future we might have a decoupled runtime... The ServerRole 'enum' has Master, Replica, Single and Unknown... so I reckon if we start sewing the word Replica in here it will help understanding and debugging for people in the future, can still call it the Replica Front End if that makes sense?

Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md
```diff
@@ -0,0 +1,58 @@
...
+4. Test: Perform some content updates on the master/administration environment, ensure they work successfully on that environment, then verify that those changes appear on the front-end environment
+
+### Scaling
+
+**Do not scale your master/administration environment** this is not supported and can cause issues.
+
+Azure Web Apps can be manually or automatically scaled up or down and is supported by Umbraco's load balancing.
```
marcemarc: Worth saying that the 'Front end replica' Azure Web App can be manually or automatically scaled... in case this is read without the warning... and the need to have a separate master/admin is not understood...

... be great to add a diagram here? Admin Server, Shared Database, and then scalable replicas...

also you may mention this later... but pretty worthwhile making sure the elements are in the same Azure region to avoid lag

Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Deployment considerations
+
+Since you have 2 x web apps, when you deploy you will need to deploy to both places - There is probably various automation techniques you can use to make this more basic. That is outside the scope of this article.
+
+**Important note:** This also means that you should not be editing templates or views on a live server as master and front-end environments do not share the same file server. Changes should be made in a staging environment and then pushed to live environments.
```
marcemarc: Can this be a recommendation that changes are made in a development environment? and then pushed to live environments?

I'm old enough to remember Staging as a place to stage changes before final release, rather than a place you would 'make changes'!

Also development environment can hint at working locally with visual studio.. rather than people thinking they 'must' setup a staging environment...

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
...
+
+If the file system on your servers isn't performing any file replication then no _Umbraco_ configuration file changes are necessary.
+
+Depending on the configuration and performance of the local storage you might need to consider Examine Directory Factory Options and the Umbraco temporary storage location. 
+
+## Synchronised/replicated or shared file system
+
+If the file system on your servers is performing file replication then the Umbraco temporary folder must be excluded from replication.
+
+If the file system on your servers is located on shared storage you will need to configure Umbraco to locate the Umbraco temporary folder outside of the shared storage.
```
marcemarc: is it worth pointing to the advice above and mentioning the config setting to do so?

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Non-synchronised/replicated file system
+
+If the file system on your servers isn't performing any file replication then no _Umbraco_ configuration file changes are necessary.
+
+Depending on the configuration and performance of the local storage you might need to consider Examine Directory Factory Options and the Umbraco temporary storage location. 
+
+## Synchronised/replicated or shared file system
+
+If the file system on your servers is performing file replication then the Umbraco temporary folder must be excluded from replication.
```
marcemarc: What is the 'Umbraco temporary folder' that you would like them to exclude?

app_data?

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
+---
+versionFrom: 8.0.0
+---
+
+## Non-synchronised/replicated file system
```
marcemarc: I might be misreading this but is this

Non-synchronised/non-replicated file system?

I'm not sure what is meant by the slash

particulary with the next heading being

syncronised/replicated or shared file system

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
...
+
+If the file system on your servers isn't performing any file replication then no _Umbraco_ configuration file changes are necessary.
```
marcemarc: I'm not sure people would know, or whether this describes the scenario well enough?

eg if Media is configured to be in blob storage, and changes aren't made to templates on teh production environment, then no form of file replication is required, and no specific umbraco configuration changes are required either?

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
...
+
+Depending on the configuration and performance of the local storage you might need to consider Examine Directory Factory Options and the Umbraco temporary storage location.
```
marcemarc: environment's local storage, you might need to consider...

then I would link to the docs about 'Examine's Directory Factory Options' and also to the bit above where the Temporary Storage Location is set?

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
marcemarc: think the folders to exclude are listed further down. so can link to that section?

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
...
+* [Watch an intro to installing and working with DFS](https://www.youtube.com/watch?v=DYfBoUt2RVE)
+
+There are other alternatives for file replication out there, some free and some licensed. You'll need to decide which solution is best for your environment.
+
+### Non-replicated files
+
+When deploying Umbraco in a load balanced scenario using file replication, it is important to ensure that not all files are replicated - otherwise you will experience file locking issues. Here are the folders and files that should not be replicated:
+
+* ~/App_Data/TEMP/*
+	* Alternatively store the Umbraco temporary files in the local server's 'temp' folder and set Examine to use a [Directory Factory](#examine-directory-factory-options). Achieve this by changing this configuration setting to 'true' in the web.config. The downside is that if you need to view this configuration file you'll have to find it in the temp files. Locating the file this way isn't always clear.
```
marcemarc: we mention this above, probably ought to be in Tip formatting... but again we are talking about umbraco.config file?

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
```diff
@@ -0,0 +1,72 @@
...
+			
+    ```xml
+    <add key="Umbraco.Core.LocalTempStorage" value="EnvironmentTemp" />
+    ```
+            
+* ~/App_Data/Logs/*
+	* This is **optional** and depends on how you want your logs configured (see below) 
+
+If for some reason your file replication solution doesn't allow you to not replicate specific files folders (which it should!!) then you can use an alternative approach by using virtual directories. *This is not the recommended setup but it is a viable alternative:*
```
marcemarc: exclude instead of 'not replicate'?

marcemarc: Think this should have a heading to split it up from the good practice above, it's more an aside for a niche situation...

Getting-Started/Setup/Server-Setup/Load-Balancing/index-v7.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Legacy load balancing
+
+Umbraco versions before 7.3.0 must use a legacy/deprecated load balancing technique.
+
+[Full documentation is available here](traditional.md)  
+
+## Common load balancing setup information
+
+_The below section is common for all load balancing configurations, ensure that the following is instructions are followed regardless of what load balancing setup you choose._
```
marcemarc: the following is instructions...

the following instructions

Getting-Started/Setup/Server-Setup/Load-Balancing/index-v7.md
```diff
@@ -0,0 +1,103 @@
...
+
+There are some logging configurations to take into account no matter what type of load balancing environment you are using.
+
+[Full documentation is available here](logging.md)
+
+### Testing
+
+Your staging environment should also be load balanced so that you can see any issues relating to load balancing in that environment before going to production.
+
+You'll need to test this solution **a lot** before going to production. You need to ensure there are no windows security issues, etc... The best way to determine issues is have a lot of people testing this setup and ensuring all errors and warnings in your application/system logs in Windows are fixed.
```
marcemarc: This doesn't seem to come across as good advice, although the sentiment is correct... it just says 'this is really flakey and hard to set up' but that shouldn't be the message if you follow the instructions.

Could be replaced with a link to the community Load Balancing healthcheck for V7, and V8 version (umm when that's released)... but also perhaps suggest a way of testing...

1) Update content on Master
2) Check update appears on Master Server
3) check update appears on Replica servers
If using Cloud scale up Replica environments and try again
Try Scheduled publishing of content
Check and review application/system logs for errors

emmaburstow:
👋 Hi docs people! Just dropping in to say hey and that we see you. Thanks for the work!

Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md
Jeavon: Frond servers can be Master rather than Replcia but perhaps I'll update these steps to include fixing Master and Replic....

Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md
Jeavon: After many late nights debate, we have settled on "Standalone File System"

Jeavon:
I have updated a load of stuff after many late-night discussions between myself and Marc ;)
I think this is now at the point it needs to be merged and then improved

Shazwazza:
Hi all, it seems with these changes we've completely removed the docs for how to setup a LB scenario in v8 with a shared file system. There's no link to the shared file system doc and no mention of shared file systems anymore. Is there a reason for that?

Shazwazza:
At some point in the file-system-replication.md it mentioned 'shared', see https://github.com/umbraco/UmbracoDocs/blob/cc5926c2866132e5f6dfe01e0d77c3d2ff705fba/Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication.md but then that was removed.

Shazwazza:
Perhaps 'shared' simply means no replication like here https://our.umbraco.com/Documentation/Getting-Started/Setup/Server-Setup/Load-Balancing/file-system-replication#standalone-file-system

The problem is this new doc is a bit confusing because it doesn't line up with the 3 hosting design options we list in load balancing: https://our.umbraco.com/documentation/Getting-Started/Setup/Server-Setup/Load-Balancing/#design

We need to be explicit on this page as to what each one is. IMO it's confusing to say `Standalone File System` and not mentioned 'shared' or NAS/SAN, etc... I wouldn't know if that section pertains to my setup or not.

I'm submitting an urgent PR to fix some of the links, etc... and update our description of SqlMainDom until 8.6.x is out with a fix for that in LB scenarios but i think we should update this page too.

PR #1966, opened 2019-10-10
Update RelatedLinks article

sofietoft:
Remove needsV8update label from article and updated some stuff on index pages


jmayntzhusen:
🎉👏

PR #1968, opened 2019-10-11
Umbraco.tv Episode links update

larsskjoldiversen:
Removed link to an episode that does not exist on TV.


sofietoft:
Thanks @larsskjoldiversen 😄 🎉

PR #1997, opened 2019-10-15
Updating v8 link for tv

crgrieve:
Creating a new page for v7 and updating index for v8


sofietoft:
Good catch @crgrieve !

Could you add a link to the Umbraco 7 article here: https://our.umbraco.com/Documentation/Tutorials/Creating-Basic-Site/Conclusions-Where-Next/

I think both versions of the tutorial is currently linking to the same article, as it hasn't been versioned 😅

sofietoft:
Thanks @crgrieve 😄

PR #2004, opened 2019-10-17
Make C# app factory sample non breaking 

dawoe:
There is a PR open at core for introducing content apps on members as well. There is discussion going on how big this breaking change is. 

The discussion is mainly sparked because this will break the C# Content app factory code example. Because this throws a exception when the item is not a media or content item. It would have been better to return null. 
But that would make the entire switch statement look strange. So I changed the example a bit to only show the app when on content items.

I also made clear from the description that content apps can be used on media as well


sofietoft:
Thanks for the update @dawoe 😄 
Exciting that Content Apps might be coming to Members as well!!

PR #2012, opened 2019-10-19
fix(umbraco-services): fix broken paths in Reference links

daniloff200:
These links work well on https://our.umbraco.com/documentation/Getting-Started/Code/Umbraco-Services/ , but,  broken in .md files

This PR fixed it

P.S. Feel free to decline this PR, if it's not a meaningful/important change  : )


sofietoft:
Thanks @daniloff200 😄 

Yeah, it's correct that these links doesn't work directly in Github. I'm afraid changing them, will break how they work on the frontend on Our 🙈 so, I'll have to reject this PR.

I think what we need to do instead, is rename the actual files, so they don't have the capital `I`.
There's a very specific way of doing that: https://stackoverflow.com/a/20907647 

Would you be up for making that change instead? 😁

PR #2020, opened 2019-10-21
Using External services with cloud

sofietoft:


Umbraco-Cloud/Set-Up/External-Services/index.md
```diff
@@ -0,0 +1,25 @@
+# Using external services with Umbraco Cloud
+
+In some cases Cloud might not be the only service you are working with. You might need to work with other services as well - this could be either internal or third party services. In either case, it will be services external to Umbraco Cloud.
```
jmayntzhusen: Should we be consistent with calling it `Cloud` or `Umbraco Cloud` all the time? 🤔It is sort of changing between the two a few times in this article.

Other than that consideration this is all good 👍

Issue #1875, opened 2019-09-25
Help wanted: Docs about Querying


sofietoft:
# Help wanted: Docs about Querying
Many of our articles needs to be verified against Umbraco 8, and for that we need your help!! 🙏

Our article series about **Querying** needs to be verified and updated against Umbraco 8. 

We need you're help to do the following:
* Pick an article on the list
* Read it!
* Test out any code snippets, tutorials and step-by-step's in the latest version of Umbraco 8

Does the article still apply for Umbraco 8?
* **Yes**: Tell us in a here, and create a PR where you remove the `needsV8Update` tag at the top of the article.
* **No**: Tell us in an issue, OR create a new Umbraco 8 compatible article about the same topic!

## The List

- [x] [Querying: UDI Identifiers](https://our.umbraco.com/documentation/Reference/Querying/Udi)
- [x] [IPublishedContent: IsHelpers](https://our.umbraco.com/documentation/Reference/Querying/IPublishedContent/IsHelpers)
- [x] [Dynamic Published Content](https://our.umbraco.com/documentation/Reference/Querying/DynamicPublishedContent/)
- [x] [Dynamic Published Content: Properties](https://our.umbraco.com/documentation/Reference/Querying/DynamicPublishedContent/Properties)
- [x] [Dynamic Published Content: Collections](https://our.umbraco.com/documentation/Reference/Querying/DynamicPublishedContent/Collections)
- [x] [Umbraco Helper](https://our.umbraco.com/Documentation/Reference/Querying/UmbracoHelper/)
- [x] [Membership Helper](https://our.umbraco.com/documentation/Reference/Querying/MemberShipHelper/)

Thanks for helping us out!
We appreciate all your help 😄

poornimanayar:
@sofietoft Do you still need help with this? Happy to help!

sofietoft:
@poornimanayar We could definitely still use some help getting the articles in this list verified for Umbraco 8! 😄 

Would be awesome, if you could pitch in! 💪

poornimanayar:
Assign it to me!!! Will have a look later today :-)

poornimanayar:
@sofietoft , I have been looking at the IsHelpers() for IPublishedContent. Some of the Is Helper is available on an `IndexedArray` which you can loop through while some others are not. How do you want me to approach it?

poornimanayar:
[...]

PR #2029, opened 2019-10-22
Removed needsv8Update DynamicPublishedContent

poornimanayar:
Helping out with the issue #1875 
DynamicPublishedContent is not supported in v8, so removed the tag

### DIFF
--- a/Reference/Querying/DynamicPublishedContent/index.md
+++ b/Reference/Querying/DynamicPublishedContent/index.md
@@ -1,5 +1,6 @@
 ---
 versionFrom: 7.0.0
+versionRemoved: 8.0.0
 ---
 
 # DynamicPublishedContent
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks @poornimanayar 😄 

We actually have a tag called `versionRemoved` as well. You think it would make sense to use that here?
It will add a "warning" like the one you see here: https://our.umbraco.com/documentation/Add-ons/UmbracoCourier/

sofietoft:
Whoops, forgot to add the full YAML tag 😅 
If we go with the `versionRemoved` tag, it'll look like this:
```
---
versionFrom: 7.0.0
versionRemoved: 8.0.0
---
```

poornimanayar:
Absolutely @sofietoft ! Apologies, I was not aware of the tags. So please bear with me :-)

sofietoft:
No worries at all @poornimanayar 😄 
I'm not sure we even mention this tag in our contribution guidelines 🤔 I'll double-check and make sure it's added 💪 

Thanks again for verifying the articles 🎉 Much appreciated!

PR #2037, opened 2019-10-24
Update badge default color

nathanwoulfe:
Default notification badge (when 'type' not set) uses Umbraco-blue, it's no longer turquoise.


sofietoft:
Good catch @nathanwoulfe 😄 🎉 Thanks!

nathanwoulfe:
No worries @sofietoft - every little bit helps!

PR #2040, opened 2019-10-24
Grammar changes in the readme

emmaburstow:
Hello lovely docs people. While looking for a big and impressive addition I could make to the docs repo, I found two stray commas and had to take them out. Then a few other bits and bobs. Not the most exciting change but as you say, every little counts.


sofietoft:
Thanks @emmaburstow 😄 🎉 

Yes, every little bit counts! 💪

Issue #1877, opened 2019-09-25
Help wanted: Docs about MVC Templating


sofietoft:
# Help wanted: Docs about MVC Templates
Many of our articles needs to be verified against Umbraco 8, and for that we need your help!! 🙏

Our article series about **MVC Templating** needs to be verified and updated against Umbraco 8. 

We need you're help to do the following:
* Pick an article on the list
* Read it!
* Test out any code snippets, tutorials and step-by-step's in the latest version of Umbraco 8

Does the article still apply for Umbraco 8?
* **Yes**: Tell us in a here, and create a PR where you remove the `needsV8Update` tag at the top of the article.
* **No**: Tell us in an issue, OR create a new Umbraco 8 compatible article about the same topic!

## The List

- [x] [MVC Views](https://our.umbraco.com/documentation/Reference/Templating/Mvc/views)
- [x] [MVC Views: Examples](https://our.umbraco.com/documentation/Reference/Templating/Mvc/examples)
- [x] [MVC Partial views](https://our.umbraco.com/documentation/Reference/Templating/Mvc/partial-views)
- [x] [MVC Child actions](https://our.umbraco.com/documentation/Reference/Templating/Mvc/child-actions)
- [x] [MVC Querying](https://our.umbraco.com/documentation/Reference/Templating/Mvc/querying)
- [x] [MVC Forms](https://our.umbraco.com/documentation/Reference/Templating/Mvc/forms/)
- [x] [MVC Forms: Partial Views](https://our.umbraco.com/documentation/Reference/Templating/Mvc/forms/tutorial-partial-views)
- [x] [MVC Forms: Child Actions](https://our.umbraco.com/documentation/Reference/Templating/Mvc/forms/tutorial-child-action)
- [x] [MVC Forms: Custom markup](https://our.umbraco.com/documentation/Reference/Templating/Mvc/forms/tutorial-custom-markup)

Thanks for helping us out!
We appreciate all your help 😄

greystate:
I think I've just found my **Hacktoberfest** assignment 😆 — happy to tackle this.

greystate:
Sidenote: The first two does not show the warning about not having been verified with V8 on Our, even though they have the `needsV8Update` tag. Is that a "the docs have not been built" issue, or something else that should ideally be reported as an issue itself?

sofietoft:
That's awesome to hear @greystate 😎 🎉

If you check the articles directly in Github, you'll see that they actually have the "needsV8update".
[...]

PR #2041, opened 2019-10-24
Updating MVC views file for v8 to get rid of the old Umbraco.Field 

emmaburstow:
Lots of changes here. I've tested the snippets and replaced (or straight up removed) the ones that were no longer working in v8. While I've tried to explain in full how we access property data in v8, I'm worried I may have wandered into modelsbuilder territory. I did try doing all this on a clean install where I hadn't set it up though so hopefully I've avoided that. I hope this helps some!

### DIFF
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -52,11 +52,11 @@ Looping over a selection works in a similar way. If you have a property that con
 If you want to convert a type and it's possible, you can do that by typing a variable and assigning the value from your property to it. This could look like the example below.
 ```csharp
 @foreach (TeamMember person in Model.TeamMembers)
-                        {
-                           <a href="@person.Url">                                                              
-                             <p>@person.Name</p>
-                           </a>
-                         }
+   {
+     <a href="@person.Url">                                                              
+        <p>@person.Name</p>
+     </a>
+   }
  ```
 
 In this example, we are looping through a list of items with the custom made type TeamMember assigned. This means we are able to access the strongly typed properties on the TeamMember item.
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -29,7 +29,7 @@ This is probably the most used method which renders the contents of a field usin
 If you're using the method from within a partial view then be aware that you will need to inherit the context so the method knows which type to get the desired value from. You'd do this at the top of partial view and so strongly typed properties can then be accessed in the partial view. For instance you can pass "HomePage" like this:
 
 ```csharp
-@inherits UmbracoViewPage<SiteSettings>
+@inherits UmbracoViewPage<HomePage>
 ...
 @Model.Value("title")
 ```
@@ -50,14 +50,14 @@ Lopping over a selection works in a similar way. If you have a property that con
 </ul>
 ```
 If you want to convert a type and it's possible, you can do that by typing a variable and assigning the value from your property to it. This could look like the example below.
-`
+```csharp
 @foreach (TeamMember person in Model.TeamMembers)
                         {
                            <a href="@person.Url">                                                              
                              <p>@person.Name</p>
                            </a>
                          }
- `
+ ```
 
 In this example, we are looping through a list of items with the custom made type TeamMember assigned. This means we are able to access the strongly typed properties on the TeamMember item.
 
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -61,22 +61,6 @@ If you want to convert a type and it's possible, you can do that by typing a var
 
 In this example, we are looping through a list of items with the custom made type TeamMember assigned. This means we are able to access the strongly typed properties on the TeamMember item.
 
-## Rendering a field using @CurrentPage (dynamically)
-
-**Attention! This approach is considered obsolete** - See [Common pitfalls]
-
-:::warning
-Attention! This approach is considered obsolete - See [Common pitfalls](../../Common-Pitfalls/index.md#dynamics) for more information about why the dynamic approach is obsolete. for more information about why the dynamic approach is obsolete.
-:::
-
-The UmbracoHelper method provides many useful parameters to change how the value is rendered. If you however want to render value "as-is" you can use the @CurrentPage property of the view. The difference between @CurrentPage and @Model.Content is that @CurrentPage is the dynamic representation of the model which exposes many dynamic features for querying. For example, to render a field you use this syntax:
-
-```csharp
-@CurrentPage.bodyContent
-```
-
-*NOTE: When accessing content dynamically you will not get intellisense if you are using Visual Studio to edit your templates.*
-
 ## <a name="renderingMacros"></a>Rendering Macros
 
 Rendering a macro is easy using UmbracoHelper. There are 3 overloads, we'll start with the most basic:
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -20,7 +20,7 @@ All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.Na
 
 ## Rendering a field in a strongly typed view
 
-This is probably the most used method which renders the contents of a field for the using the alias of the content item.
+This is probably the most used method which renders the contents of a field using the alias of the content item.
 
 ```csharp
 @Model.Value("bodyContent")
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -1,6 +1,6 @@
 ---
-versionFrom: 7.0.0
-needsV8Update: "true"
+versionFrom: 8.0.0
+needsV8Update: "false"
 ---
 
 
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -12,7 +12,7 @@ _Working with MVC Views and Razor syntax in Umbraco_
 
 All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.NameOfYourDocType>` along with the using statement `@using ContentModels = Umbraco.Web.PublishedModels;`. This exposes many properties that are available in razor. The properties on the Document Type can be accessed in a number of ways:
 
-* @Model (of type `Umbraco.Web.Mvc.RenderModel`) -> the model for the view which contains the standard list of IPublishedContent properties but also gives you access to the typed current page (of type whatever type you have added in the angled brackets).
+* @Model (of type `Umbraco.Web.Mvc.ContentModel`) -> the model for the view which contains the standard list of IPublishedContent properties but also gives you access to the typed current page (of type whatever type you have added in the angled brackets).
 * @Umbraco (of type `Umbraco.Web.UmbracoHelper`) -> contains many helpful methods, from rendering macros and fields to retrieving content based on an Id and tons of other helpful methods. [See UmbracoHelper Documentation](../../Querying/UmbracoHelper/index.md)
 * @Html (of type `HtmlHelper`) -> the same HtmlHelper you know and love from Microsoft but we've added a bunch of handy extension methods like @Html.BeginUmbracoForm
 * @UmbracoContext (of type `Umbraco.Web.UmbracoContext`)
--- a/Reference/Templating/Mvc/views.md
+++ b/Reference/Templating/Mvc/views.md
@@ -34,9 +34,9 @@ If you're using the method from within a partial view then be aware that you wil
 @Model.Value("title")
 ```
 
-You will also need to pass the "Context" to the @Umbraco.Field() method if you're looping over a selection like this where we pass the "item" variable.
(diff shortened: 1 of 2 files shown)


emmaburstow:
Oh and it's linked to this issue #1877

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
 
 ## Properties available in Views
 
-All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoTemplatePage` which exposes many properties that are available in razor:
+All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.NameOfYourDocType>` along with the using statement `@using ContentModels = Umbraco.Web.PublishedModels;`. This exposes many properties that are available in razor,. This means that properties on the Document Type can be accessed in a number of ways:
```
sofietoft:
```suggestion
All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.NameOfYourDocType>` along with the using statement `@using ContentModels = Umbraco.Web.PublishedModels;`. This exposes many properties that are available in razor. The properties on the Document Type can be accessed in a number of ways:
```

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
-@Umbraco.Field("bodyContent")
+@Model.Value("bodyContent")
 ```
 
-If you're using the Field method from within a partial view then be aware that you will need to pass the context so the Field method knows where to get the desired value from. For instance you can pass "CurrentPage" like this:
+If you're using the method from within a partial view then be aware that you will need to inherit the context so the method knows which type to get the desired value from. You'd do this at the top of partial view and so strongly typed properties can then be accessed in the partial view. For instance you can pass "HomePage" like this:
 
 ```csharp
-@Umbraco.Field(Model.Content, "eventLink")
+@inherits UmbracoViewPage<SiteSettings>
```
sofietoft: `SiteSettings` .. shouldn't that be `HomePage` according to the text above? 🤔

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
-@Model.Content.GetPropertyValue("bodyContent")
-```
-
-You can also specify the output type that you want from the property. If the property editor or value does not support the conversion then an exception will be thrown. Some examples:
-
-```csharp
-@Model.Content.GetPropertyValue<double>("amount")
-@Model.Content.GetPropertyValue<RawXElement>("xmlContents")
-```
+If you want to convert a type and it's possible, you can do that by typing a variable and assigning the value from your property to it. This might could look like the example below.
```
sofietoft:
```suggestion
If you want to convert a type and it's possible, you can do that by typing a variable and assigning the value from your property to it. This could look like the example below.
```

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
 
 ## Properties available in Views
 
-All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoTemplatePage` which exposes many properties that are available in razor:
```
marcemarc: I think we need to show the Modelsbuilder approach 'first' as out of the box Modelsbuilder is turned on - but I think we also need to show what happens when Modelsbuilder is turned off - eg the view will inherit from UmbracoViewPage... also you can have partials etc inherit from UmbracoViewPage and still work if Modelsbuilder is turned on...

emmaburstow:
Amazing. I'll look at this on the train today. Thanks for the feedback!

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
+All Umbraco views inherit from `Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.NameOfYourDocType>` along with the using statement `@using ContentModels = Umbraco.Web.PublishedModels;`. This exposes many properties that are available in razor,. This means that properties on the Document Type can be accessed in a number of ways:
 
+* @Model (of type `Umbraco.Web.Mvc.RenderModel`) -> the model for the view which contains the standard list of IPublishedContent properties but also gives you access to the typed current page (of type whatever type you have added in the angled brackets).
```
marcemarc: I think RenderModel has become ContentModel... but also I think in using UmbracoViewPage the Model is IPublishedContent?

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
 ```csharp
-@Umbraco.Field(Model.Content, "eventLink")
+@inherits UmbracoViewPage<SiteSettings>
+...
+@Model.Value("title")
 ```
 
 You will also need to pass the "Context" to the @Umbraco.Field() method if you're looping over a selection like this where we pass the "item" variable.
 
+Lopping over a selection works in a similar way. If you have a property that contains, for instance, an IEnumberable collection, you can access the individual items using a foreach loop. Below illustrates how you might do that using "item" as a variable.
```
marcemarc: Lopping?

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
+If you're using the method from within a partial view then be aware that you will need to inherit the context so the method knows which type to get the desired value from. You'd do this at the top of partial view and so strongly typed properties can then be accessed in the partial view. For instance you can pass "HomePage" like this:
 
 ```csharp
-@Umbraco.Field(Model.Content, "eventLink")
+@inherits UmbracoViewPage<SiteSettings>
+...
+@Model.Value("title")
 ```
 
 You will also need to pass the "Context" to the @Umbraco.Field() method if you're looping over a selection like this where we pass the "item" variable.
```
marcemarc: There is a lingering reference to Umbraco.Field here

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
 ```
 
-If you're using the Field method from within a partial view then be aware that you will need to pass the context so the Field method knows where to get the desired value from. For instance you can pass "CurrentPage" like this:
+If you're using the method from within a partial view then be aware that you will need to inherit the context so the method knows which type to get the desired value from. You'd do this at the top of partial view and so strongly typed properties can then be accessed in the partial view. For instance you can pass "HomePage" like this:
 
 ```csharp
-@Umbraco.Field(Model.Content, "eventLink")
+@inherits UmbracoViewPage<SiteSettings>
+...
+@Model.Value("title")
```
marcemarc: If we're showing people the wiring up of the view using the Modelsbuilder approach... eg
@inherits UmbracoViewPage<BlogPost>

should we show the first rendering out of a field (and are we still using the term field to refer to properties?) using a more Modelsbuildery technique eg

@Model.BodyContent

instead of 

@Model.Value("bodyContent")

Don't get me wrong I much prefer the simpler non-modelsbuilder approach... but as it's 'on' by default - would it make sense to people new to Umbraco to show that methodology first...

... but also showing the Model.Value("") way too?

It depends a little on how further the codebase for Modelsbuilder becomes part of the core and not a plugin

There is a section on Modelsbuilder at the end of the article
but that's probably a bigger question than just updating the info to see if it works with V8 :-)

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
+                        {
+                           <a href="@person.Url">                                                              
+                             <p>@person.Name</p>
+                           </a>
+                         }
+ `
+
+In this example, we are looping through a list of items with the custom made type TeamMember assigned. This means we are able to access the strongly typed properties on the TeamMember item.
 
 ## Rendering a field using @CurrentPage (dynamically)
```
marcemarc: I would update the warning to mention that dynamic has been removed (and remove the info on how to use it)- but leave this info in a V7 version of this documentation page

Reference/Templating/Mvc/views.md
emmaburstow: Yes! My mistake

Reference/Templating/Mvc/views.md
```diff
@@ -10,82 +10,56 @@ _Working with MVC Views and Razor syntax in Umbraco_
...
+If you want to convert a type and it's possible, you can do that by typing a variable and assigning the value from your property to it. This might could look like the example below.
+`
+@foreach (TeamMember person in Model.TeamMembers)
+                        {
+                           <a href="@person.Url">                                                              
+                             <p>@person.Name</p>
+                           </a>
+                         }
+ `
+
```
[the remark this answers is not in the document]
emmaburstow: Thanks Jesper! I'll do that now

Reference/Templating/Mvc/views.md
emmaburstow: 👍

Reference/Templating/Mvc/views.md
emmaburstow: 👍

Reference/Templating/Mvc/views.md
emmaburstow: I agree. I felt as though I'd keep the structure the same so begin with Model.Value although it's something I wouldn't really use... I'm not sure int he value (no pun intended) of doing things this way. I'm going to get Modelsbuilder at the forefront - it's my preferred way but now I was reluctant at first to do it. As it's out of the box now, I think there's no danger in doing it. It's actually simpler than the 'simple' way.

Reference/Templating/Mvc/views.md
emmaburstow: Versioning... I want to have a v7 version and a v8. I am kinda writing a v8 one. Should I change the version in the meta data to 8 and then create a new file for 7? Or should I keep this as 7 and then create a new v8?

Reference/Templating/Mvc/views.md
emmaburstow: If you type the model in the partial with angle brackets like this

@inherits UmbracoViewPage<SiteSettings>

You get the model returned as it is in the brackets

emmaburstow: I've flipped these now and added a little more info - do we think it's resolved? It might be that there's still more to do.

Reference/Templating/Mvc/views.md
marcemarc: The syntax for writing out a property is simpler with Modelsbuilder... however the concept of generated models, and extending them along with the inherits statement being different at the top of each view can be a little more confusing when you are getting started... (depending on what you already know of course in terms of C# and MVC!) - you don't get caught out with the Models being out of sync between environments if you use: Model.Value("bodyContent")... (stringly typed models :-)) and if you use UmbracoViewPage everywhere, then you don't run into problems with the wrong model being passed to a shared partial view etc... also the 'Insert' menu  and Query builder use the @Model.Value approach 
image and generate 'queries' without the Modelsbuilder syntax - so if you utilise these which is more common when starting out, it's weird if this isn't mentioned in the documentation!

The conclusion we came to in docs discussions was that both methods should probably be presented equally, but perhaps we always mention 'Modelsbuilder' first since that is the way we think HQ wants people to go... (eg pure live being on by default, and the work to put modelsbuilder 'lite' into the core) but the other approach is 'simpler' conceptually... (it depends I think on the context of the project - Umbraco has such a broad spectrum of usage there isn't one best way! - but at least now we are down to two good ways instead of 5 ways!)

Reference/Templating/Mvc/views.md
marcemarc: create a views-v7.md file and have this containing V7 specific info with the YAML saying From 7.0.0 then in then in the file you are editing view.md file it can be totally V8 focussed with the YAML saying from 8.0.0   ... that way this info is presented first without the noise of legacy information, but the dropdown allows people to see the V7 version... if they wish... it is a little complicated to understand and hastily cobbled together... but so far hasn't completely broken down!

Reference/Templating/Mvc/views.md
emmaburstow: Fab - that clears it all up a little for me. I agree that 2 ways is better than 5 too!

emmaburstow:
Ok - I'm not certain as there were quite a few changes, but are we there now? If there's more to do, point it out for me and I'll get on it.

sofietoft:
Thanks a bunch @emmaburstow 😄 🎉 

We have in the Docs team talked about how it at some point would make sense to have this entire article restructured. Like, making it clearer which approach to use, whether you're using the ModelsBuilder or not.

I think for now, we'll make an issue for it here on the repo, and take it from there 💪 

I can see @jmayntzhusen has requested some changes - I'll just send this past him, and then I believe we're good to merge here 😎

sofietoft:
Thanks again @emmaburstow ! 😁 

I've created an issue, where we can continue the discussion on how this article should be structured - #2054

PR #2042, opened 2019-10-24
Updated docs for hosting on Azure

Shazwazza:
These updates remove the FCN reference since that is not really relevant anymore, moves up the recommended configuration settings to make it more clear and adds the disabling overlapping appdomains setting. This is all relevant for both v7 and v8


JoseMarcenaro:
@Shazwazza - I was asking myself exactly the same as @jmayntzhusen: What about the load balanced sites?

Shazwazza:
@jmayntzhusen yes, it needs to be done for __ALL__ sites on Azure websites

PR #2046, opened 2019-10-25
Added two additional examples for the Media Service

abjerner:
Ideally there should be examples for all the CRUD operations, but here are at least some more for creating media items.

### DIFF
--- a/Reference/Management/Services/MediaService/Index.md
+++ b/Reference/Management/Services/MediaService/Index.md
@@ -84,7 +84,7 @@ In addition to the three mandatory parameters as shown above, you may also speci
 
 ### Creating a new media from an uploaded file
 
-The example below shows how to create a new file (on in this case, an image) from a HTTP upload. For illustrative purposes the example is a Razor view.
+The example below shows how to create a new file (in this case, an image) from a HTTP upload. For illustrative purposes the example is a Razor view.
 
 
 
--- a/Reference/Management/Services/MediaService/Index.md
+++ b/Reference/Management/Services/MediaService/Index.md
@@ -111,7 +111,7 @@ The example below shows how to create a new file (in this case, an image) from a
         // Initialize a new image at the root of the media archive
         IMedia media = Services.MediaService.CreateMedia("Hello", Constants.System.Root, Constants.Conventions.MediaTypes.Image);
         
-        // Set the property value (Umbraco will handle the underlying magic)
+        // Set the property value (Umbraco will automatically update related properties)
         media.SetValue(Services.ContentTypeBaseServices, Constants.Conventions.Media.File, "hello.jpg", file);
         
         // Save the media
--- a/Reference/Management/Services/MediaService/Index.md
+++ b/Reference/Management/Services/MediaService/Index.md
@@ -77,7 +77,7 @@ For the `CreateMedia` method, the first parameter is the name of the folder to b
 
 The second parameter is the ID of the parent media item. `Constants.System.Root` is a constant defined in Umbraco with the value of `-1`, which is used for indicating the root of the media archive. Instead of specifying the numeric ID of the parent, you may instead specify either a `Guid` ID or an `IMedia` instance representing the parent media.
 
-The third parameter is the alias of the media type. As Umbraco comes with a folder type by default, we can use the `Constants.Conventions.MediaTypes.Folder` constant to specify that the alias of the media type is `Folder`.
+The third parameter is the alias of the Media Type. As Umbraco comes with a Folder Type by default, we can use the `Constants.Conventions.MediaTypes.Folder` constant to specify that the alias of the Media Type is `Folder`.
 
 In addition to the three mandatory parameters as shown above, you may also specify a numeric ID for a user to which the creation of the media should be attributed. If not specified, the media will be attributed to the user with ID `-1`.
 
--- a/Reference/Management/Services/MediaService/Index.md
+++ b/Reference/Management/Services/MediaService/Index.md
@@ -152,4 +152,4 @@ using (Stream stream = File.OpenRead("C:/path/to/my-image.jpg"))
 }
 ```
 
-Again Umbraco will make sure the necessary properties are updated
+Again Umbraco will make sure the necessary properties are updated.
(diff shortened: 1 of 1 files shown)


Reference/Management/Services/MediaService/Index.md
```diff
@@ -61,9 +61,30 @@ IMediaService mediaService = Umbraco.Core.Composing.Current.Services.MediaServic
 
 ## Samples
 
+### Creating a new folder
+
+To create a new folder at the root of the media archive, your code could look as:
```
sofietoft:
```suggestion
To create a new folder at the root of the media archive, your code could like the following:
```

Reference/Management/Services/MediaService/Index.md
```diff
@@ -61,9 +61,30 @@ IMediaService mediaService = Umbraco.Core.Composing.Current.Services.MediaServic
...
+
+// Save the folder
+Services.MediaService.Save(folder);
+```
+
+For the `CreateMedia` method, the first parameter is the name of the folder to be created.
+
+The second parameter is the ID of the parent media item. `Constants.System.Root` is a constant defined in Umbraco with the value of `-1`, which is used for indicating the root of the media archive. Instead of specifying the numeric ID of the parent, you may instead specify either a `Guid` ID or an `IMedia` instance representing the parent media.
+
+The third parameter is the alias of the media type. As Umbraco comes with a folder type by default, we can use the `Constants.Conventions.MediaTypes.Folder` constant to specify that the alias of the media type is `Folder`.
```
sofietoft:
```suggestion
The third parameter is the alias of the Media Type. As Umbraco comes with a Folder Type by default, we can use the `Constants.Conventions.MediaTypes.Folder` constant to specify that the alias of the Media Type is `Folder`.
```

Reference/Management/Services/MediaService/Index.md
```diff
@@ -61,9 +61,30 @@ IMediaService mediaService = Umbraco.Core.Composing.Current.Services.MediaServic
...
+
+In addition to the three mandatory parameters as shown above, you may also specify a numeric ID for a user to which the creation of the media should be attributed. If not specified, the media will be attributed to the user with ID `-1`.
+
+
 ### Creating a new media from an uploaded file
 
-The example below shows how to create a new file from a HTTP upload. For illustrative purposes the example is a Razor view.
+The example below shows how to create a new file (on in this case, an image) from a HTTP upload. For illustrative purposes the example is a Razor view.
```
sofietoft:
```suggestion
The example below shows how to create a new file (in this case, an image) from a HTTP upload. For illustrative purposes the example is a Razor view.
```

Reference/Management/Services/MediaService/Index.md
```diff
@@ -106,3 +127,29 @@ When creating a new media from a file (eg. of the types **Image** or **File**),
 
 Umbraco uses this instance to determine the type of the media you're creating, as well as handling a few things "under the hood" so you don't have to. For instance Umbraco will automatically set other properties relatd to the file - such as file size and image dimensions.
 :::
+
+
+### Creating a new media from a stream
```
sofietoft: Sounds a bit like you're missing an `item` here? 🤔 

```suggestion
### Creating a new media item from a stream
```

Reference/Management/Services/MediaService/Index.md
```diff
@@ -106,3 +127,29 @@ When creating a new media from a file (eg. of the types **Image** or **File**),
...
+
+Similar to specifying a `HttpPostedFileBase`as shown in the example above, you can instead specify a `Stream` for the contents of the file that should be created.
```
sofietoft:
```suggestion
Similar to specifying a `HttpPostedFileBase` as shown in the example above, you can instead specify a `Stream` for the contents of the file that should be created.
```

Reference/Management/Services/MediaService/Index.md
```diff
@@ -106,3 +127,29 @@ When creating a new media from a file (eg. of the types **Image** or **File**),
...
+
+```csharp
+// Open a new stream to the file
+using (Stream stream = File.OpenRead("C:/path/to/my-image.jpg"))
+{
+
+    // Initialize a new image at the root of the media archive
+    IMedia media = Services.MediaService.CreateMedia("My image", Constants.System.Root, Constants.Conventions.MediaTypes.Image);
+
+    // Set the property value (Umbraco will handle the underlying magic)
```
sofietoft: Is it really magic though? 😅 
```suggestion
    // Set the property value (Umbraco will handle the underlying magic)
```

Reference/Management/Services/MediaService/Index.md
```diff
@@ -106,3 +127,29 @@ When creating a new media from a file (eg. of the types **Image** or **File**),
...
+    media.SetValue(Services.ContentTypeBaseServices, Constants.Conventions.Media.File, "my-image.jpg", stream);
+
+    // Save the media
+    Services.MediaService.Save(media);
+
+}
+```
+
+Again Umbraco will make sure the necessary properties are updated
```
sofietoft:
```suggestion
Again Umbraco will make sure the necessary properties are updated.
```

Reference/Management/Services/MediaService/Index.md
abjerner: I've now changed the text to "Umbraco will automatically update related properties" 😉

abjerner:
@sofietoft thanks - I've now updated the PR to reflect your suggestions 😉

For future PRs, can you elaborate on why it should be *Media Type* instead of *media type*? For the most part, I were referring to *a* media type, not *the* media type.

sofietoft:
Thanks @abjerner 😄 

Yeah, that's a great question!
We're trying to be a bit more consistent in the Docs - eg writing `Document Types` instead of `document types`, `Media Type` instead of `media type`, and so on.

These are concepts in Umbraco, and are also written with capital letters in the backoffice and most code, so writing it in the same way in the Docs, gives it a more consistent look..

I do see your point on reference to _a_ Media Type .. but I still think, as it *is* a concept, we should use the capital letters!

I'd be happy to hear your thoughts on this! 😁 
In the mean time, I'll make sure this is merged in.

PR #2056, opened 2019-10-31
Fix two typos

greystate:
Not much, I know - but GitHub makes it sooo easy to submit a fix right when you see a typo :)


greystate:
Guessing the warnings are because the document describes all the rules, thus violating them 😄

jmayntzhusen:
Yea don't worry about the style guide errors, I fix them in https://github.com/umbraco/UmbracoDocs/pull/2061.

Thanks for the corrections!

PR #2062, opened 2019-11-02
Link to the more in depth examples of DI

marcemarc:
The Services and Helpers page contains more indepth examples of using DI in different contexts and creating your own services to inject, so I think it's useful to link to that information from here...

### DIFF
--- a/Reference/Using-Ioc/index.md
+++ b/Reference/Using-Ioc/index.md
@@ -210,3 +210,7 @@ namespace Doccers.Core
     }
 }
 ```
+
+### Using DI in Services and Helpers
+
+[Services and Helpers] (../../Implementation/Services/index.md) - For more examples of using DI and gaining access to Services and Helpers, and creating your own custom Services and Helpers to inject.
--- a/Reference/Using-Ioc/index.md
+++ b/Reference/Using-Ioc/index.md
@@ -213,4 +213,4 @@ namespace Doccers.Core
 
 ### Using DI in Services and Helpers
 
-[Services and Helpers] (../../Implementation/Services/index.md) - For more examples of using DI and gaining access to Services and Helpers, and creating your own custom Services and Helpers to inject.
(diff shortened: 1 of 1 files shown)


marcemarc:
@hartviglarsen I'm wondering in the SiteService example you have in this page, whether we should be warning that the SiteService would only be reusable when there is an UmbracoContext? - I've linked to the Services and Helpers page that covers that in more detail - is that enough?

sofietoft:
I think it's great to have a link added in this article!! 💪 
Was thinking it could be moved up a bit in the article - but can see how it makes sense to add it in the end .. Perhaps we could use `##` instead of `###` ? Just to make a bit more visible 🤔 

Anywho.
I've pinged @hartviglarsen to hear his thoughts as well.

hartviglarsen:
> @hartviglarsen I'm wondering in the SiteService example you have in this page, whether we should be warning that the SiteService would only be reusable when there is an UmbracoContext? - I've linked to the Services and Helpers page that covers that in more detail - is that enough?

Perhaps it should be written as a note to make it a bit more noticeable? E.g.

```
### UmbracoHelper

:::note
The use of the UmbracoHelper is only possible when there's an instance of the UmbracoContext. >You can read more here<.
:::

```

marcemarc:
@sofietoft shall we add @hartviglarsen 's note (which has a link) and then leave it at the bottom too as a bit of a 'read more'... ?

Also I think we should link to the lightinject site, and include the sneaky trick of accessing the lightinject container via casting the concrete property:

this is also a really common question about registering multiple services at once for those coming from autofac that would be good to answer: https://our.umbraco.com/forum/using-umbraco-and-getting-started/99810-lightinject#comment-313880 as an example...

sofietoft:
@marcemarc - Yeah, I think adding it as a note is a good idea! 💪 

And also a good idea adding a link to the lightinject site.
Perhaps use the `Tip` formatting for that one? 🤔

marcemarc:
@sofietoft updated I thinK!

sofietoft:
Awesome! 😄 And don't even have any changes to request 💪 Thanks @marcemarc 

Merging!

PR #2069, opened 2019-11-06
Fixed a typo

ctf-umbraco:
I fixed a typo in the security section there was a "t" missing in "the" after 7.1.5


sofietoft:
Good catch @ctf-umbraco 😄 Thanks!

PR #2072, opened 2019-11-06
Fix quotes in Security Hardening v7 code example

matthew-darby:


sofietoft:
Thanks for catching these @matthew-darby 😄 🎉

PR #2076, opened 2019-11-08
Change hardcoded index name to Constant.

Adolfi:
Instead of specifying the "ExternalIndex" string in the TryGetIndex method, use the Constants.UmbracoIndexes.ExternalIndexName.


sofietoft:
This is great @Adolfi ! 😄 

Let me just run it by someone who knows a bit more on examine then me! 😁

sofietoft:
Looks like we're all good here @Adolfi 😄 

Thanks @Jeavon for helping out.
I'll merge this now! 💪

PR #2077, opened 2019-11-08
Code snippets does not get syntax highlighting.

Adolfi:
This page code does not get syntax highlighting with the ```cs, which also create uncompiling code in the ExamineComposer section, since the type paramter is not rendered on Our.Umbraco.

The code in the docs on Our does not combile, since ComponentComposer requires 1 type argument. 
Currently if you code and paste the ExamineComposer on this page (https://our.umbraco.com/documentation/Reference/Searching/Examine/examine-events) into your code, you will get these two error: 
- Using the generic type 'ComponentComposer<TComponent>' requires 1 type arguments
- 'ExamineComposer' does not implement interface member 'IComposer.Compose(Composition)'

However by using ```c# like all the other pages, we should both get syntax highlighting and hopefully also the missing type paramter.

### DIFF
--- a/Reference/Searching/Examine/examine-events.md
+++ b/Reference/Searching/Examine/examine-events.md
@@ -16,7 +16,7 @@ In the [Quick Start](Quick-Start/index.md) documentation you can see how to perf
 
 However, what if you want to search through several different node types and search across many different fields, you will typically need to have a query that looks like this:
 
-```cs
+```c#
 var results = searcher.CreateQuery("content").Field("nodeName", searchTerm)
                     Or().Field("bodyText", searchTerm)
                     Or().Field("description", searchTerm)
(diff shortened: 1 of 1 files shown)


sofietoft:
This is very interesting @Adolfi !

Did not know that the syntax highligting would affect the code when you copy it!
I'd say we should definitely try the `c#`!

Adolfi:
Wow that was fast, already live! :)
Looks a lot better now. 
Yes we should use c# everywhere and not cs.

Have a great weekend.

sofietoft:
Yes! That helped quite a lot 😄 

Sure makes a lot of sense to use `c#` !
Thanks for catching this @Adolfi !

Have a fab' weekend 🦄

PR #2080, opened 2019-11-08
Logviewer apps on newline.

Adolfi:
The last word in the sentence “dashboard” does not end with a dot also there is no new line before the badges, which makes it look wonky on mobile device. See screenshot: https://paste.pics/e0c1c15790f75eaacdfe87392dd94b2d


sofietoft:
Good catch @Adolfi 😄

PR #2083, opened 2019-11-13
Add custom package actions article

jmayntzhusen:
This should provide a good understanding of how to create package actions. I have added a list of current v8 open source packages that use package actions for examples.

Very curious to hear feedback on the "When to use a Package Action" section 🙂

### DIFF
--- a/Extending/Packages/Package-Actions/custom-package-actions.md
+++ b/Extending/Packages/Package-Actions/custom-package-actions.md
@@ -1,0 +1,120 @@
+---
+versionFrom: 8.0.0
+---
+
+# Creating custom package actions
+
+In addition to utilizing the [built-in package actions](index.md), you can also create your own package actions. Package actions are custom code that runs on install and uninstall of a package. You can do whatever you want in a package action - however some things are more common than others.
+
+## Basic package action implementation
+
+To create a custom package action you need to create a new class and implement the `IPackageAction` interface. The empty package action would then look like this:
+
+```cs
+using System.Xml.Linq;
+using Umbraco.Core.PackageActions;
+
+namespace CustomPackageAction.PackageActions
+{
+    public class CustomPackageAction : IPackageAction
+    {
+        public string Alias() => "CustomPackageAction";
+
+        public bool Execute(string packageName, XElement xmlData)
+        {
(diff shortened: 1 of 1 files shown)


Rockerby:
Looking good!

My understanding of when to use a package action is when you need to perform some initial setup for the package to work correctly/easier - such as a creating a new Data Type for a property editor in the package. I think your section covers this nicely.

It may be worth mentioning that because the action is passed through as an `XElement` this gives full flexibility (like child nodes).  The Starter-Kit project you linked to is a good example of this in action.

Also, I'm thinking it would be useful to add some examples in your first para `some things are more common than others such as creating Document Types, media items and similar`, gives people a bit more of a steer straight away.

Extending/Packages/Package-Actions/custom-package-actions.md
```diff
@@ -0,0 +1,120 @@
...
+            return true;
+        }
+    }
+}
+```
+
+A package action consists of 3 mandatory methods:
+
+**Alias()** 
+In this method you specify the package action alias. This is used when you select a package action in the backoffice. It can look like this:
```
LottePitcher: "This is used when you select a package action in the backoffice."

_how about changing this to something like_

"This is used in the XML definition of the package actions that you specify when creating a package in the backoffice."

Extending/Packages/Package-Actions/custom-package-actions.md
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+}
+```
+
+Here you can see we added an attribute called `rootNodeName` and then retrieved it within the `Execute()` method from the `xmlData` object. This may not be super useful if you are only creating one package, but if you want reusable package actions across packages you can add configuration like this.
+
+## Examples of custom Package Actions
+
+Even though you can do whatever you want within a package action, most packages use package actions for either installing content / media or adding configuration. To help you get started here are some examples of popular packages using these package actions:
```
sofietoft:
```suggestion
Even though you can do whatever you want within a package action, most packages use package actions for either installing content / media or adding configuration. To help you get started you can find some examples below, of popular packages using these package actions.
```

PR #2085, opened 2019-11-14
Fix links and structure

jmayntzhusen:
Found a few broken links - and added cross links where there should have been some 🙂


sofietoft:
Looks good 💪 😁

PR #2088, opened 2019-11-15
Flips the order of versions for Azure Webapps...

ctolkien:
To be in order version descending.

Previously it went "middle", "oldest", "newest" versions. Now it goes newest to oldest.  

Kept having to scratch my head to figure out which version I should be looking at.


sofietoft:
Great observation @ctolkien 😄 🎉

PR #2093, opened 2019-11-17
Heartcore: Adding a contents list to content delivery API

crgrieve:
As there are many examples and this page is quite long, I thought I'd add a contents section with anchor links to help people find they are looking for.

If this is ok, I will also add it to the content management API documentation pages.

### DIFF
--- a/Umbraco-Heartcore/API-Documentation/Content-Delivery/content/index.md
+++ b/Umbraco-Heartcore/API-Documentation/Content-Delivery/content/index.md
@@ -2,6 +2,8 @@
 
 **BASE URL**: `https://cdn.umbraco.io`
 
+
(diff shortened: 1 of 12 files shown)


sofietoft:
This is awesome @crgrieve 😄 
Makes a lot of sense to have this sort of "table of contents" at the top.

Perhaps we should give the list a heading as well?
"Table of Contents" or something similar. What do you think?

crgrieve:
Hi @sofietoft ,

I have added the "table of contents" and also added it to the content management APIs too :)

Thanks,
Carole

sofietoft:
Yay, that's awesome! 😄 
I'll merge it right away.
Thanks @crgrieve 🎉

PR #2099, opened 2019-11-21
The ID is required for conditional fields

richarth:
Added a few notes to make it clear that conditional fields won't be shown when the criteria are met unless they have an ID on them.


sofietoft:
Thanks @richarth ! Great work 😄 

I did not know this was a requirement! Super to have it claryfied 💪

PR #2103, opened 2019-11-22
Syntax highlighting of health check example

ThomasArdal:


sofietoft:
Wonderful to have some syntax highligting 👏 
Thanks for adding this, @ThomasArdal ! 😎

PR #2104, opened 2019-11-25
Added info about Label (Long string).

Adolfi:
If you input a long string value into a "regular" Label (string) then you will get an error when you try to save & publish saying: "Unable to publish - String or binary data would be truncated.". This was easily fixed by changing the value type from Label (String) to Label (Long string). 
screen

Would have been nice to find information about in the docs, hence this PR. :)

Cheers!

### DIFF
--- a/Getting-Started/Data/Data-Types/default-data-types.md
+++ b/Getting-Started/Data/Data-Types/default-data-types.md
@@ -40,7 +40,9 @@ Allows for the upload and cropping of images by using a focal point. Specific cr
 Is a non-editable control, can only be used to display a present text (string). It can also be used in the
 media section to load in values related to the node, such as width, height and file size.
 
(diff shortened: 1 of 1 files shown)


sofietoft:
This is great @Adolfi !
I agree that it makes sense to have docs on this.

Would it make sense to add some similar information in the section on Property Editors? 🤔
Here, I mean: https://our.umbraco.com/documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/#label 
The "Label" property type doesn't have it's own article yet, and that would perhaps make sense, since there are these different value types to use! 
Just a thought! Let me know what you think 😁

Adolfi:
Thank you @sofietoft!
Great Idea. I will fix a page for the Label property type and create a new PR.
Cheers!

PR #2110, opened 2019-11-26
Heartcore: Adding meta data

crgrieve:
Was looking at docs and realised there wasn't SEO meta data on the Heartcore pages, so I added some basic info :)

Thanks,
Carole


sofietoft:
Great work @crgrieve ! 😄 
Thanks again - we really appreciate all the work you've been doing on the docs lately!

PR #2116, opened 2019-11-29
Fixing example Resolver

steroberts89:


sofietoft:
Thanks @steroberts89 ! 😄 

I'll make sure we have this reviewed as soon as possible!

dampee:
Thanks a lot @steroberts89 
Good catch.  Thanks a lot for taking the effort to fix this!

PR #2122, opened 2019-12-02
Security config prefix change

jynzix:
Without the slash being added before 'media', I get the following error in the logs when image are trying to be served,
"ERROR ImageProcessor.Web.HttpModules.ImageProcessingModule - 
ImageProcessor.Common.Exceptions.ImageProcessingException: ProcessImageAsync 610 : No image exists at ..."

The images are still served but adding the slash at the beginning removes the error.

### DIFF
--- a/Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
+++ b/Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
@@ -111,7 +111,7 @@ You have to manually add `prefix="media/"` to the service element, otherwise Ima
 <security>
   <services>
     <!--<service name="LocalFileImageService" type="ImageProcessor.Web.Services.LocalFileImageService, ImageProcessor.Web" />-->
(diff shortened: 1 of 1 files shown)


sofietoft:
Hi 😄 
You closed this so fast, that I almost didn't get to look at it!
I honestly don't know anough about this piece of docs, to help verify it.

You closing it, does that mean it shouldn't be changed after all?
Let me know if you need help looking more into this!

jynzix:
Hello! I was sure his was an issue but it honestly depends on the configuration.
The way I have my config set up is that I have the container at the end of my host address in the security.config. This actually originally didn't include a trailing slash at the end of the address. If I added one at the end, it also fixed the issue I had above which is why I closed it. 🙂

sofietoft:
Aha, that's interesting! 🤔 I'm glad to hear that you manage to solve your issue at least!

If I understand you correctly, the way your config differs from this article, is that you're adding your container at the end of the "Host" value!
As the example in this article doesn't have the container at the end of the host address, it sounds like we should keep the trailing slash in this example - right?
Just want to make sure I understand 🙈

jynzix:
Yes definitely!

PR #2124, opened 2019-12-02
Adding heartcore link

crgrieve:
I noticed there was no Heartcore link in the main landing page content, so thought I would add it :)


sofietoft:
Ah, that's great @crgrieve !

It's so obvious that it's so easy to miss 😅 Thanks for adding the link! 🎉

PR #2132, opened 2019-12-05
Add docs on IVariationContextAccessor

jmayntzhusen:
Had a question about this today and couldn't find anything about it. Here is a small explanation and example of using it. 

Have created a new section under reference called `Language-Variation`, feel free to move it somewhere else.


sofietoft:
Thanks @jmayntzhusen ! 😄

PR #2134, opened 2019-12-06
Proofing edit

lcichanowicz:


sofietoft:
Great catch @lcichanowicz ! Thanks 😄

PR #2136, opened 2019-12-06
Proofing edits

lcichanowicz:


sofietoft:
Whoops! Looks like a few too many words found their way in to that section 🙈 
Thanks for help getting it sorted! 💪

PR #2138, opened 2019-12-06
Proofing edit

lcichanowicz:


sofietoft:
Good catch @lcichanowicz ! 😄 Amazing how changing just one word in a sentence can make it that much easier to read! 🎉

PR #2140, opened 2019-12-10
Update docs with important info.

MahmoudUmbraco:
Update docs with important info about what happens if you run the uaas.cmd tool as an administrator.


sofietoft:
Looks great @MahmoudUmbraco !

Shouldn't we add it as a warning instead? Sounds like something people wouldn't want to happen this "adding files into Windows/System folder"? 🤔

MahmoudUmbraco:
Yeah, a warning would be great :D 
Thanks for your suggestion @sofietoft 👑

sofietoft:
Would you be up for adding it?
We have some docs on how here: https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/#adding-notes-warnings-tips

I can of course also do it, if you like! Let me know 💪

sofietoft:
Perfect! Thanks 😄

PR #2145, opened 2019-12-12
Update XML sitemap tutorial

mjbarlow:
Just ran through the tutorial, few  changes I could suggest, I added a trim to the blacklisted doc types, removed an unused variable in the full example, changed the variables to implicitly typed (var), and updated the method calls for hasValue and Value to use FallBack.ToAncestors.


marcemarc:
@mjbarlow and hey! checkout that shiny new C-Trib contributor badge you now have on Our Umbraco...

image

PR #2146, opened 2019-12-12
Temp/apicontroller-test

Adolfi:
As promised, here are the Unit Test example for the UmbracoApiController, now possible due to 8.4 release. I added a warning message about the version specific test, hope it follows a suitable convention otherwise I'll update it.
ping @Shazwazza. :)


sofietoft:
This is great @Adolfi ! 😄 

I'll make sure we have this reviewed as soon as possible 💪

Shazwazza:
@Adolfi looks great! Not sure what the convention is for warnings regarding version numbers, seems ok to me but I'll let @sofietoft decide :) 

@Adolfi are you able to resolve the merge conficts? Once that is done i think @sofietoft can merge this.

Thanks!!

sofietoft:
Thanks for the review @Shazwazza ! 😄 

The warning is formatted correctly.
However - it seems like the version info is something you'd wanna know before implementing the code? Perhaps we should move it up a bit - before the code snippet even? 🤔
What do you think @Adolfi ?

Adolfi:
@Shazwazza Thanks, glad you liked it. I have fixed the merge conflict. :) Cheers!

Adolfi:
@sofietoft Yeah that makes sense. This was why wanted to double check if there was a convention of some sort for these kind of warnings. For instance this page: https://our.umbraco.com/Documentation/Reference/Scheduling/ I saw that the 'warning' comes after the code snippet.

Maybe if we would decide on preferred way (for example warnings "always before" and tips "always after") I could create a separate PR fixing this everywhere?

Or if you'd like I can just swap it in this one and we leave the others, works either way for me. :)

Adolfi:
@sofietoft also might be worth updating this page: https://our.umbraco.com/Documentation/Contribute/Markdown-Conventions/ with some info about where to place warnings/tips before/after etc. Thoughts?

sofietoft:
That's a very good point, @Adolfi ! 🤔 

I would say that where to put the warning box very much depends on the scenario.
For this one though, and the other example you found in the Scheduling article, it definitely makes sense to have the warning **before** the snippets.

You're more than welcome to go through all warnings and make a PR! When we have that, we can always take the different scenarios into account!
How does that sound? 😁

Adolfi:
@sofietoft I updated the placement for the warning so it is **before** the code snippet for the ApiController Test, so that this PR can be merged.

I will create another PR in the future, where I will go through all the Warnings and Tips to see if they should be before or after each code snippets. To be continued.. :)

sofietoft:
Awesome @Adolfi 😄 👏 

Thanks again for putting in all the work to do this!

Adolfi:
@sofietoft Thank you!!!

PR #2148, opened 2019-12-16
Changed but to by :)

MahmoudUmbraco:


sofietoft:
Good catch @MahmoudUmbraco ! 😄 Thanks!

PR #2154, opened 2019-12-19
Heartcore versioning and updates

sofietoft:
Added details about upgrade on and related to Umbraco Heartcore projects.

### DIFF
--- a/Umbraco-Heartcore/Versions-and-updates/index.md
+++ b/Umbraco-Heartcore/Versions-and-updates/index.md
@@ -2,19 +2,9 @@
 versionFrom: 8.0.0
 ---
 
-# Versions and upgrades
+# Versions and updates
 
-Umbraco Heartcore is a headless SaaS (software as a service) product, which means that one of the benefits is that you do not have to worry about upgrades. These will all be handled by Umbraco HQ.
-
-In this article you can learn more about how we handle upgrades for both the Umbraco Backoffice and the Umbraco Heartcore Rest API.
-
-## Umbraco versions
-
-The Umbraco backoffice, which on a Heartcore project serves as the UI for content creation, will automatically be upgraded to a newer version whenever new and relevant features are released for the Umbraco CMS.
-
-Upgrades like these will have no effect on your content or your live application.
-
-## API versions
+Umbraco Heartcore is a headless SaaS (software as a service) product, which means that one of the benefits is that you do not have to worry about upgrades. These will all be handled by Umbraco HQ. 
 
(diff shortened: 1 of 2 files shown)


Umbraco-Heartcore/Versions-and-updates/index.md
```diff
@@ -0,0 +1,17 @@
+---
+versionFrom: 8.0.0
+---
+
+# Versions and updates
+
+Umbraco Heartcore is a headless SaaS (software as a service) product, which means that one of the benefits is that you do not have to worry about upgrades. These will all be handled by Umbraco HQ.
```
sitereactor: Umbraco Heartcore is a software as a service (SaaS) offering, which means that one of hte benefits is that you do not need to worry about upgrades. All maintenance related to the cms, its features and APIs are handled by Umbraco HQ.

Umbraco-Heartcore/Versions-and-updates/index.md
```diff
@@ -0,0 +1,17 @@
+---
+versionFrom: 8.0.0
+---
+
+# Versions and updates
+
+Umbraco Heartcore is a headless SaaS (software as a service) product, which means that one of the benefits is that you do not have to worry about upgrades. These will all be handled by Umbraco HQ. 
+
+Umbraco HQ will regularly be adding new features and endpoints to the REST APIs, as well as maintaining the existing structure and endpoints. Each time this is done, a new version of the API will be released. API releases and versioning is handled following the [ASP.NET API versioning guidelines](https://github.com/microsoft/aspnet-api-versioning).
```
sitereactor: Umbraco HQ will regularly be adding new features to the cms and new capabilities to the REST APIs. 
Each time new endpoints are added to the REST API the `api-version` will be updated to reflect the changes. API releases and versioning is handled following the [ASP.NET API versioning guidelines](https://github.com/microsoft/aspnet-api-versioning). We follow semver in terms of versioning, so you can be sure that existing APIs will not break and continue to work as expected. If a breaking change is required then the major version will be incremented, but the previous major version will continue to work.

Umbraco-Heartcore/Versions-and-updates/index.md
```diff
@@ -0,0 +1,17 @@
...
+
+:::note
+It is important that you **use the `API-version` header to specify which version of the API you will be using** when making calls to the API. If no version is specified the latest vesrion of the API will be used, which could potentially break your client if it is setup for a lower version.
```
sitereactor: When making calls to the REST API we recommend that you **use the `api-version` header to specify which version of the API you will be using**. If no `api-version` is specified the latest version of the API will be used.

When using the client libraries the `api-version` is handled as part of the library and generally not something you need to worry about. Just be aware that the client library releases will have certain features available - as an example, if you are using the very first release (1.0) of the .net core client library it will not have the Forms API available. That was made available in the 1.1 version release of the client library.

PR #2160, opened 2019-12-20
Remove unused _siteDomainHelper field.

Adolfi:
In the ProductPageUrlProvider docs the _siteDomainHelper is declared as a local field, for no reason at all (its unused). Maybe its been used in the past?

It should still be injected in the constructor since it is passed down to the DefaultUrlProvider, but it doesnt need to be locally declared in the ProductPageUrlProvider.cs.


sofietoft:
Thanks for the suggested changes here @Adolfi ! 😄 

It's a "bit" outside my scope, so I'll run it by our experts 💪

marcemarc:
@Adolfi you are correct! it's not used, I think the example might have evolved and then simplified, but tidying up of the constructor was missed!

PR #2161, opened 2019-12-20
404 link on Implementation index page.

Adolfi:
Not sure if the discontinued Rest-Api link on the https://our.umbraco.com/documentation/Implementation/ page should be removed or linked to the https://our.umbraco.com/documentation/Implementation/Rest-Api/ page, promoting Umbraco Heartcore. I went for the later option..

Anyway, It shouldn't link to a dead link (404).


sofietoft:
Good point @Adolfi !

I think what you've done here, is the best solution 💪 Thanks a bunch!

PR #2162, opened 2019-12-24
V8 PropertyActions : Update documentation to use the simpler approach

warrenbuckley:
Uses the umbProperty parent reference component/directive stored in $scope in order to us fire the method `setPropertyActions()`

@nielslyngsoe can you double-check this please in the new year and @sofietoft can you do you usual tidying up of my work, please :)

### DIFF
--- a/Extending/Property-Editors/Property-Actions/index.md
+++ b/Extending/Property-Editors/Property-Actions/index.md
@@ -30,44 +30,40 @@ We use `labelKey` and `labelTokens` to retrieve a localized string that is displ
 
 `isDisabled` is used to disable an Action, which change the visual appearance and prevents interaction. Use this option when an action wouldn't provide any change. In the example above, the action `remove all entries` would not have any impact if there is no entries.
 
-## Getting ready for Property Actions
-A Property Editor needs to be implemented as a Component for it to perform the call to expose its Property Actions.
-The Component must be configured to retrieve an optional reference to `umbProperty`. The requirement must be optional because property-editors are implemented in scenarios where it's not presented.
-
-See the following example:
-
-```js
-angular.module('umbraco').component('myPropertyEditor', {
-    controller: MyController,
-    controllerAs: 'vm',
-    require: {
-        umbProperty: '?^umbProperty'
-    }
-    …
-});
-```
-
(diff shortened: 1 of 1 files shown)


warrenbuckley:
Note this has been merged in V8 core this change but has not yet been released @sofietoft

Extending/Property-Editors/Property-Actions/index.md
```diff
@@ -30,44 +30,40 @@ We use `labelKey` and `labelTokens` to retrieve a localized string that is displ
...
 `isDisabled` is used to disable an Action, which change the visual appearance and prevents interaction. Use this option when an action wouldn't provide any change. In the example above, the action `remove all entries` would not have any impact if there is no entries.
 
-## Getting ready for Property Actions
-A Property Editor needs to be implemented as a Component for it to perform the call to expose its Property Actions.
-The Component must be configured to retrieve an optional reference to `umbProperty`. The requirement must be optional because property-editors are implemented in scenarios where it's not presented.
+## Implementing a Property Action
+The implementation of Property Actions consists of two parts. First a definition of the actions and secondly parsing the lists of actions to `$scope.umbProperty.setPropertyActions().
 
-See the following example:
+Note the check if `umbProperty` exists and is not null or undefined
```
sofietoft: Not sure I understand this sentence, @warrenbuckley 🤔 
Does it mean that if `umbProperty` does not exist or is `null` the example will not work?

warrenbuckley: Its just a bit of a safety net - we could probably remove this comment, if you prefer @sofietoft

sofietoft: Sure! But, is my understanding of it correct?

nielslyngsoe:
Hi @warrenbuckley 

Thanks for your contribution to this documentation, I do though think we should keep the approach for Components, and then specify that there also is an option for Controllers. The current documentation is still valid, and the correct way for Components. So we should make it clear that there are two approaches.
I have added it to my tasks and will hopefully get to it this week :-)

sofietoft:
Another approach including this, has been documented in #2170 .
That's why I'm closing this one.

PR #2163, opened 2019-12-27
v8 : Update config files

poornimanayar:
Updated the documentation for config files for v8

### DIFF
--- a/Reference/Config/index.md
+++ b/Reference/Config/index.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 8.0.0
-needsV8Update: "true"
 ---
 
 # Configuration Files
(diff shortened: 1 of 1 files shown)


sofietoft:
Hi @poornimanayar !

Thanks for putting in the work to update these articles on the config files 👏 
As I can see, both Damiaan and Marc have already given you some pointers on what we've been talking about regarding how to handle these "missing" configs.

We have a meeting tomorrow afternoon, and I'll make sure to bring this up, so we can agree on what would be the best approach 😄 
I agree though, that we should still have articles for Umbraco 8, that directs people to docs on how to add the configuration through code instead 💪 

How does that all sound? 
/Sofie

poornimanayar:
Hi @sofietoft,

Thanks for clarifying... I shall start looking at these towards latter part of this week. :-) Looking fwd!

Poornima

sofietoft:
Hi @poornimanayar !

Following our Curator meeting this afternoon, we've updated and cleaned out the Config landing page for Umbraco 8.
Unfortunately that renders this particular PR out-dated.

Hope you understand!

poornimanayar:
Sure thing @sofietoft . Do you still need help on the other PRs?

sofietoft:
We do indeed, so if you're up for it? 😄 

And I don't know if you saw, but Marc added a comment to #2164 .
Our idea with this whole section, is that for the config files that's been removed in Umbraco 8, we instead link to the place in the Documentation that explains how the configuration can be achieved through code. 

So, instead of simply removing the references, we provide them with the new way of doing things 😁 
Does that make sense?

poornimanayar:
Hey @sofietoft ,

Of course I can help! I ll review Marc's comment and then start working on them :-)

Poornima

PR #2164, opened 2019-12-27
Removed needv8update tag from 404handler.config

poornimanayar:
The `404handlers.config` is obsolete. So it does not need v8 update and hence the tag can be removed

### DIFF
--- a/Reference/Config/404handlers/index-v7.md
+++ b/Reference/Config/404handlers/index-v7.md
@@ -1,0 +1,28 @@
+---
+versionFrom: 7.0.0
+---
+
+# 404handlers.config
+
+**Obsolete.  The concept of *NotFoundHandlers* has been replaced by [Content Finders](../../routing/request-pipeline/IContentFinder)**
+
+_Configuration file for legacy *NotFoundHandlers*. These are used to register custom code within the Umbraco incoming request pipeline, including the legacy way to handle a custom 404 error._
+
+The Default Handlers are listed in this configuration file for 'backwards compatibility':
+
+```xml
+<NotFoundHandlers>
+  <notFound assembly="umbraco" type="SearchForAlias" />
+  <notFound assembly="umbraco" type="SearchForTemplate"/>
+  <notFound assembly="umbraco" type="SearchForProfile"/>
+  <notFound assembly="umbraco" type="handle404"/>
+</NotFoundHandlers>
+```
+
+## Backwards compatibility
+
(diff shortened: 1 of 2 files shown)


poornimanayar:
Hi @dampee ,

Just so that I get this right, I am planning to put a holding page in place to say that the config is not relevant for v8 and is replaced by IContentFinder.And then we can link to this document may be? https://our.umbraco.com/Documentation/Reference/Routing/Request-Pipeline/IContentFinder#notfoundhandlers

sofietoft:
Sounds like you're on the right track here, @poornimanayar 💪 

Check out my comment here: https://github.com/umbraco/UmbracoDocs/pull/2163#issuecomment-571050722 

You're of course welcome to add the Umbraco 8 version of the article - I mean, it's always easier to review something that's there, as opposed to an idea of something 😄

poornimanayar:
Will work on these one by one @sofietoft 

Poornima

marcemarc:
Hi Poornima

Yes, it's a question of linking off to the place in the documentation that adequately explains how to achieve the same customisation as the configuration files used to allow people to achieve.

So in your example above, you'd think it would be enough to link here:

https://our.umbraco.com/Documentation/Reference/Routing/Request-Pipeline/IContentFinder#notfoundhandlers

But actually - does that adequately explain how to set a page to be a 404 error page for a site?

Having the heading as Not Found Handlers is really old terminology... so as part of this change what we're aiming to do, is update where these links go to, so they make sense for users who 'just want to set a page as the 404 page'.

eg having the title of 'How to set a 404 Page' is the sort of thing we are hoping for...

regards

Marc

poornimanayar:
Hi @marcemarc , @sofietoft ,

I have made an attempt at the 404 config updates. Let me know if I am on the right track :-)

Poornima

sofietoft:
Looks great @poornimanayar ! 💪 

I think part of what Marc was saying, is that the heading on `routing/request-pipeline/IContentFinder#notfoundhandlers` still uses the old terminology, and we should update that as well?
Is that right, @marcemarc ?

sofietoft:
Apologies for the delay here @poornimanayar !

Just back from our weekly meeting, and we've decided to merge 😄 🎉 
Thanks again for putting in the work to get this updated! Much appreciated.

poornimanayar:
Great @sofietoft , you will have more incoming from me now that I have a direction on this :-)

PR #2170, opened 2020-01-06
Documentation with approaches for both Controller and Component

nielslyngsoe:


Extending/Property-Editors/Property-Actions/index.md
```diff
@@ -30,8 +30,51 @@ We use `labelKey` and `labelTokens` to retrieve a localized string that is displ
...
-## Getting ready for Property Actions
-A Property Editor needs to be implemented as a Component for it to perform the call to expose its Property Actions.
+## Implement Property Actions
+The implementation of Property Actions vary depending on wether your Property Editor is impemented with a Controller or as a Component.
+
+### Controller Implementation
+
+If your Property Editor is implemented with a Controller, you can use the following approach.
+
+See the following example for implementation of Property Actions in a Controller.
```
sofietoft: These two sentences says the same thing, right?
> If your Property Editor is implemented with a Controller, you can use the following approach.

> See the following example for implementation of Property Actions in a Controller.

Can I suggest that we only use one of them?
It could be:
`When your Property Editor is implemented with a Controller, use the following approach for the Property Action:`

Extending/Property-Editors/Property-Actions/index.md
```diff
@@ -47,8 +90,7 @@ angular.module('umbraco').component('myPropertyEditor', {
 });
 ```
 
-## Implementing a Property Action
-The implementation of Property Actions consists of two parts. First a definition of the actions and secondly parsing the lists of actions to `umbProperty.setPropertyActions`.
+See the following example for implementation of Property Actions in a Controller, parsing actions to `this.umbProperty.setPropertyActions`.
```
sofietoft: I'm a little confused about this one.
Shouldn't this be part of the `### Controller Implementation` section?

Extending/Property-Editors/Property-Actions/index.md
nielslyngsoe: Yes, I guess you are 100% right!

nielslyngsoe:
@sofietoft Yes, this is another approach than Warren's suggestion.

I've made the corrections and you can review it again.

Extending/Property-Editors/Property-Actions/index.md
nielslyngsoe: Oh, was a mistake, should say Component...

sofietoft:
Great @nielslyngsoe , thanks! 😄 

Still dependent on a release? And in that case, do you know which release?

nielslyngsoe:
@sofietoft yes.. and for the component part, we require 8.6. So let's just say 8.6 for the whole think to make it simple, even though component-implementation can be done in 8.4.

So controller-people have to wait a little.

PR #2173, opened 2020-01-06
Add documentation on creating a Deploy Valueconnector

jmayntzhusen:


Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+    return value.ToString();
+}
+```
+
+But at this point you can't reattach the process as the code has changed. So we build, go to site1/bin and copy projectalias.Core.dll and projectalias.Core.pdb into site2/bin and then attach to process and try another transfer.
+
+The workflow here is not optimal, but a lot quicker than trying to deploy to Cloud everytime, and with this you can attach the debugger as well to help you out.
+
+Let's sync up, attach the debugger and attempt another transfer. Now you will see that `value` is null a few times, then your hardcoded id a few times, but nothing breaks here. Suddenly you will hit the `FromArtifact` method instead:
```
sofietoft: What exactly does "Let's sync up" mean here? 
Does it mean that everything needs to be pushed to the Cloud environment, and then pulled down again, so the sites are in sync? 🤔 

```suggestion
Let's sync up, attach the debugger and attempt another transfer. Now you will see that `value` is null a few times, then your hardcoded ID a few times, but nothing breaks here. Suddenly you will hit the `FromArtifact` method instead:
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+![Hitting FromArtifact](images/fromArtifact.png)
+
+Here you will notice that the value is what you had returned in `ToArtifact`.
+
+## Creating our ValueConnector
+
+Well enough about testing. You may have realised at this point that the flow is something like this:
```
sofietoft:
```suggestion
You may have realised at this point that the flow is something like this:
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+Site1 content transfer initated -> Hit the `ToArtifact` method on the environment -> Send to site2 -> Hit the `FromArtifact` method on site 2 -> Property data on site2
```
sofietoft:
```suggestion
Site 1 content transfer initated -> Hit the `ToArtifact` method on the environment -> Send to Site 2 -> Hit the `FromArtifact` method on Site 2 -> Property data on Site 2
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+So in our case, what we want to do is to ensure the id is changed in a transfer. We do this by converting the id to a GUID in the `ToArtifact` method, then it gets transfered to site2. On site 2 we will convert it back to an id in the `FromArtifact` method. This way the user will still see an id on the content node, the id they see will be updated to the correct one though.
```
sofietoft:
```suggestion
So in our case, what we want to do is to ensure the ID is changed in a transfer. We do this by converting the ID to a GUID in the `ToArtifact` method on Site 1, which will then get transfered to Site 2. On site 2 we will convert it back to an ID in the `FromArtifact` method. This way the user will still see an ID on the content node, but the ID they see will be updated to the correct one.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+:::warning
+In this example there would be no way for Deploy to know to also transfer the image, so we assume that you would transfer all content and images to ensure it is on the target environment under a different id.
```
sofietoft:
```suggestion
In this example there would be no way for Deploy to know to also transfer the image. We assume that you would transfer all content and images to ensure it is on the target environment under a different ID.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+That is not a good assumption, and you may have noticed that there is a parameter on the ToArtifact method that you could update by finding the image and adding it to `ICollection<ArtifactDependency> dependencies`.
```
sofietoft:
```suggestion
That is not a good assumption, and you may have noticed that there is a parameter on the `ToArtifact` method that you could update by finding the image and adding it to `ICollection<ArtifactDependency> dependencies`. 
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+So in our case, what we want to do is to ensure the id is changed in a transfer. We do this by converting the id to a GUID in the `ToArtifact` method, then it gets transfered to site2. On site 2 we will convert it back to an id in the `FromArtifact` method. This way the user will still see an id on the content node, the id they see will be updated to the correct one though.
+
+:::warning
+In this example there would be no way for Deploy to know to also transfer the image, so we assume that you would transfer all content and images to ensure it is on the target environment under a different id.
+
+That is not a good assumption, and you may have noticed that there is a parameter on the ToArtifact method that you could update by finding the image and adding it to `ICollection<ArtifactDependency> dependencies`. 
+:::
+
+So in order to convert to a GUID in the `ToArtifact` method, we will update the code:
```
sofietoft:
```suggestion
In order to convert to a GUID in the `ToArtifact` method, we will update the code:
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+- [EntityService.GetKey](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.Services.Implement.EntityService.html#Umbraco_Core_Services_Implement_EntityService_GetKey_System_Int32_Umbraco_Core_Models_UmbracoObjectTypes_)
+- [new GuidUdi](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.GuidUdi.html#Umbraco_Core_GuidUdi__ctor_System_String_System_Guid_)
+- [new ArtifactDependency](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.Deploy.ArtifactDependency.html#Umbraco_Core_Deploy_ArtifactDependency__ctor_Umbraco_Core_Udi_System_Boolean_Umbraco_Core_Deploy_ArtifactDependencyMode_)
+
+When stepping through the code we can see that everything seems to work fine:
+
+![Stepped through code](images/steppingThroughCode.png)
+
+Note: Showing the variable values is a feature of [ReSharper](https://www.jetbrains.com/resharper/)
```
sofietoft:
```suggestion
:::note
Note: Showing the variable values is a feature of [ReSharper](https://www.jetbrains.com/resharper/) .
:::
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+- [new GuidUdi](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.GuidUdi.html#Umbraco_Core_GuidUdi__ctor_System_String_System_Guid_)
+- [new ArtifactDependency](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.Deploy.ArtifactDependency.html#Umbraco_Core_Deploy_ArtifactDependency__ctor_Umbraco_Core_Udi_System_Boolean_Umbraco_Core_Deploy_ArtifactDependencyMode_)
+
+When stepping through the code we can see that everything seems to work fine:
+
+![Stepped through code](images/steppingThroughCode.png)
+
+Note: Showing the variable values is a feature of [ReSharper](https://www.jetbrains.com/resharper/) 
+
+By the time we hit `FromArtifact` value of `"umb://media/00c9eff861654f52be7a33367c3561a4"` all that is left to do is convert back to an int.
```
sofietoft:
```suggestion
By the time we hit `FromArtifact` value of `"umb://media/00c9eff861654f52be7a33367c3561a4"` all that is left to do is convert back to an `int`.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+    var getIdAttempt = _entityService.GetId(udi.Guid, UmbracoObjectTypes.Media);
+
+    if (!getIdAttempt.Success) return value;
+
+    return getIdAttempt.Result.ToString();
+}
+```
+
+Here is a gif showing the valueconnectoer in action in the end, but uploading a new image, updating the id on the node, transfering then seeing the image on the new environment and the id updated:
```
sofietoft:
```suggestion
Here is a gif showing the ValueConnector in action. A new image is uploaded, the ID on the node is updated and transferred. Finally the image is on the new environment and the ID is updated:
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+![Full workflow gif](images/valueconnector.gif)
+
+The final ValueConnector code is this:
```
sofietoft:
```suggestion
The final ValueConnector code will look like this:
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+1. You got super lucky and the ID you had on local happened to be the same as what your media item was assigned on your Cloud environment.
+1. Your page will now show a different image as the id you had corresponds to something else on this environment.
+1. You will get a YSOD as it can't find any media items with that id.
+
+To prevent this from happening we will need to use a ValueConnector.
+
+## Testing a ValueConnector
+
+Before we start working on making a ValueConnector a few notes on how to test and work with them. You probably will need to test the values that are being converted, but you probably also doesn't want to build, git push, content transfer to see that the value may not have changed.
```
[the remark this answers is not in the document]
jmayntzhusen: I'm trying to say that testing it by deploying to Cloud is not good, and then I go on to explain how it can be set up locally for faster testing. Any suggestions on how it can be reworded for clarity? 🙂

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,271 @@
...
+Now we will set up these two identical sites to transfer content to eachother, to do so go to `site1/Config/UmbracoDeploy.config` and edit the live environment url to be Site2's url (http://localhost:17025/ in my case).
+Then do the same for site2 but put in site1's domain as the "live" one.
+
+At this point you should be able to go to the backoffice of either environment and do a Content transfer to live, and it should end up on the other (Assuming no errors from your custom connector).
+
+## Debugging a transfer
+
+At this point we haven't done anything to the ValueConverter yet, other than return the original value. Now we will attach Visual Studio to the IIS processes and try a transfer to see what it sends along.
+
+Go to Visual Studio, hit "Attach to Process" (default ALT + CTRL + P), and choose your two IIS processes. Add breakpoints in the `ToArtifact` and `FromArtifact` methods and go to site2, then try to do a content transfer to live (site2).
```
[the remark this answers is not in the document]
jmayntzhusen: Nope that is correct! 😁

jmayntzhusen:
@sofietoft 

I edited a few of your suggestions a little bit, but otherwise all are merged in 🙂

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+}
+```
+
+In this case I cloned the Cloud project down using the [uaas.cmd](https://umbra.co/uaas-cmd) tool, which means that I have a class library that I can add the ValueConnector to. This will automatically have some references included and will build a DLL, eg. `projectalias.core.dll`, and put it in the websites bin folder when building. 
+
+This has no impact on the way you work, but it may help you understand why some things are named the way they are.
+
+At this point I have one clone of the site locally. However, to test this I will push the changes to the Cloud site and then clone it down again. The second clone doesn't need to be cloned with the `uaas.cmd` tool as we aren't developing on it, just need to run it locally.
```
jmayntzhusen:
```suggestion
At this point I have one clone of the site locally. However, to test this I will push the changes to the Cloud site and then clone it down again. The second clone doesn't need to be cloned with the `uaas.cmd` tool as we aren't developing on it, all we need is to run it locally.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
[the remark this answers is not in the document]
sofietoft: It makes sense now. When I was reading it, I didn't realize the next part was the testing part. No need to re-write 💪

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,287 @@
...
+versionFrom: 8.0.0
+---
+
+# Packages on Umbraco Cloud
+
+If you want to use or develop packages for Umbraco Cloud there are a few things to consider and be aware of. One such thing is that custom property editors may need a **ValueConnector** to transform their content between environments.
+
+# ValueConnectors
+
+A ValueConnector is an extension to Deploy that allows you to transform data when you deploy content of any kind between environments. It is mostly used to transfer ID based content between environments.
```
clausjensen: You may want to add in that `ValueConnectors` are also responsible for resolving dependencies. Resolving dependencies is a just as important role of `ValueConnectors` as transforming the values is.

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,287 @@
...
+
+Here you will notice that the value is what you had returned in `ToArtifact`.
+
+## Creating our ValueConnector
+
+You may have realised at this point that the flow is something like this:
+
+Site 1 content transfer initated -> Hit the `ToArtifact` method on the environment -> Send to Site 2 -> Hit the `FromArtifact` method on Site 2 -> Property data on Site 2
+
+So in our case, what we want to do is to ensure the ID is changed in a transfer. We do this by converting the ID to a GUID in the `ToArtifact` method on Site 1, which will then get transfered to Site 2. On site 2 we will convert it back to an ID in the `FromArtifact` method. This way the user will still see an ID on the content node, but the ID they see will be updated to the correct one.
```
clausjensen: Would suggest something like:
"Site 1 content transfer initiated -> Property data is fetched on Site 1 -> Hit the `ToArtifact` method on the environment -> Send to Site 2 -> Hit the `FromArtifact` method on Site 2 -> Property data is saved on Site 2"

and:
"So in our case, what we want to do is to ensure the ID from Site 1 is changed during a transfer to match what the new ID is in Site 2. We do this by converting the ID to a GUID in the `ToArtifact` method on Site 1, which will then get transfered to Site 2. On site 2 we will convert it back to an ID in the `FromArtifact` method. This way the user will still see an ID on the content node, but the ID they see will be updated to the correct one."

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,287 @@
...
+
+:::warning
+
+In this example there would be no way for Deploy to know to also transfer the image. We assume that you would transfer all content and images to ensure it is on the target environment under a different ID.
+
+That is not a good assumption, and you may have noticed that there is a parameter on the `ToArtifact` method that you could update by finding the image and adding it to `ICollection<ArtifactDependency> dependencies`. 
+:::
+
+In order to convert to a GUID in the `ToArtifact` method, we will update the code:
```
clausjensen: I think this should be: 
"In order to add the image as a dependency for the item being transferred, we will update the code in the `ToArtifact` method:"

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,287 @@
...
+
+    if (int.TryParse(svalue, out var intvalue))
+        return null;
+
+    var getKeyAttempt = _entityService.GetKey(intvalue, UmbracoObjectTypes.Media);
+
+    if (getKeyAttempt.Success)
+    {
+        var udi = new GuidUdi(Constants.UdiEntityType.Media, getKeyAttempt.Result);
+        dependencies.Add(new ArtifactDependency(udi, false, ArtifactDependencyMode.Exist));
```
clausjensen: You could add in some info about the different `ArtifactDependencyMode` options that you can use here.

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -0,0 +1,287 @@
...
+{
+    if (string.IsNullOrWhiteSpace(value))
+        return value;
+
+    if (!GuidUdi.TryParse(value, out var udi) || udi.Guid == Guid.Empty)
+        return value;
+
+    var getIdAttempt = _entityService.GetId(udi.Guid, UmbracoObjectTypes.Media);
+
+    if (!getIdAttempt.Success) return value;
```
clausjensen: It isn't really correct to return the value if the media item couldn't be resolved. This will result in a GUID going into a value of a property editor expecting a integer ID - which will in many cases cause this property editor to break or maybe even break the backoffice in a way where you can't edit the item.

It would be better to return a empty value / null or whatever the property editor has when it doesn't have a value.

Whatever is returned from this method is what goes into the value of the property editor .... keep that in mind :)

jmayntzhusen: Good point, changed these two checks to return null:

```
if (!GuidUdi.TryParse(value, out var udi) || udi.Guid == Guid.Empty)
                return null;

if (!getIdAttempt.Success) return null;
```

PR #2181, opened 2020-01-10
Incorrent 'don't'-logic in comment in Navigation example.

Adolfi:
Uncommenting this line **WILL** actually show the @site.Name in the top navigation, 
so the word '_don't_' needs to be removed to fit the current logic.

Probably a leftover from a reversed logic in the passed 
(maybe in the past it was '**comment out** this line if you **don't** want...').

Cheers!


sofietoft:
Wow, that's a great find, @Adolfi ! 🤯
I think you're right that it's a left over from when the sentence was "Comment out this line ..."

Thanks for removing the confusion in the sentence 💪 🎉

Adolfi:
Thanks @sofietoft! Have a great weekend! 🎉

PR #2182, opened 2020-01-10
Update docs regarding renamed UrlMode from UrlProviderMode.

Adolfi:
In Umbraco v.8.1 the UrlProviderMode was renamed to just simply UrlMode. https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/v81-ipublishedcontent-changes#url

These two pages in the docs does not seem to been updated with this info.

The **outbound-pipeline** page already had specific index page targeting 8.1.+ but the **services** page did not. I didn't create a specific pre-8.1 page for the services page since it was such a small change, but instead I extended the note with the renamed info. 
Is that fine? 
Or would you like me to update this PR with a specific page for pre-8.1?

### DIFF
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -224,7 +224,7 @@ public bool TryFindContent(PublishedRequest frequest)
 And inside a UrlProvider the GetUrl method has the current UmbracoContext injected:
 
 ```csharp
-public override UrlInfo GetUrl(UmbracoContext umbracoContext, IPublishedContent content, UrlProviderMode mode, string culture, Uri current)
+public override UrlInfo GetUrl(UmbracoContext umbracoContext, IPublishedContent content, UrlMode mode, string culture, Uri current)
 {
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks for this @Adolfi !

I kinda agree that it doesn't really make sense to make an entire new version of this article, as we're only talking about one minor, where it's different 🤔 
I've asked the rest of the documentation team, as I'd like their thoughts on this as well! 
Will let you what we decide on! 😁

Adolfi:
@sofietoft awesome, let me know what you come up with or if you like me to update it.
Cheers!

sofietoft:
Good morning 💪 

Had a quick chat with the team about this, and we agreed that there's no need for a full new version of this article as we're only talking about a difference in 1 minor version.
I'll merge this now.

Thanks again for putting in the work to get this updated! 😄 🎉

Adolfi:
Great, thanks @sofietoft! Happy to help! :)
Take care, cheers!

PR #2184, opened 2020-01-13
Corrected links to API Browser and Webhooks

poornimanayar:
In the article for Umbraco Backoffice for Heartcore, the links to the API Browser article and the Webhooks article looks interchanged. I have fixed that.


sofietoft:
Ah, good catch, @poornimanayar !

Thanks for setting it right 💪 😄

PR #2186, opened 2020-01-14
Update references to .NET client library

sofietoft:
Update .NET Core references to .NET Standard.


sitereactor:
:shipit: Sounds good to me 👍

PR #2188, opened 2020-01-16
Correction in the code about accessing member data

poornimanayar:
Hi,

I believe a little correction is needed [here](https://our.umbraco.com/documentation/Reference/Templating/Mvc/views#accessing-member-data). It needs to be `Value()` rather than `GetPropertyValue()`

Poornima


sofietoft:
Great catch @poornimanayar ! 😄 
Thanks for getting this corrected 🎉

PR #2190, opened 2020-01-17
MembershipHelper Tests

Adolfi:
Added some test sample for the MembershipHelper.GetCurrentMember();

A bit more complexity than the previous tests, but I think it can be useful since it involves mocking a lot of different dependencies. 

The idea the the documentation page is to start of low level at the top and the further down you get on the documentation page the more complex test samples. Hope you like this approach.

Cheers! 🎉

### DIFF
--- a/Implementation/Unit-Testing/index.md
+++ b/Implementation/Unit-Testing/index.md
@@ -22,18 +22,26 @@ public abstract class UmbracoBaseTest
     public Mock<ICultureDictionaryFactory> CultureDictionaryFactory;
     public Mock<IPublishedContentQuery> PublishedContentQuery;
 
+    public Mock<HttpContextBase> HttpContext;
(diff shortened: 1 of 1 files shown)


sofietoft:
This is great @Adolfi !

I really like the approach about starting "easy" and they making it more complex step-by-step as one move down the page 💪 Super great!

For reviewing, I'm gonna need some outside help - so I'll pass this one along.

Thanks for putting into work to getting these approaches documented 👏 😄

sofietoft:
Looks like we got a _stamp of aproval_ from Jeavon 🎉 
I'll go ahead and get this merged.

Thanks again for writing up the tests for this article @Adolfi ! 😄

Adolfi:
Thanks @sofietoft!! My pleasure!

Issue #2141, opened 2019-12-10
Multiple documentation for v7.0.0


steroberts89:
## What type of issue is it? (Choose one - delete the others)
Wrong documentation

## What article/section is this about?
https://our.umbraco.com/documentation/extending/property-editors/value-converters

## Describe the issue
There are three articles for v7.0.0

sofietoft:
Hi @steroberts89 
Thanks for reporting this 💪 

There's actually already a PR open regarding the Value Converters: https://github.com/umbraco/UmbracoDocs/pull/1865 😄 
It's currently under review with our developers here at Umbraco HQ. Hope to have it live soon!

sofietoft:
Ah, shoot! Looks like I completely misunderstood what you reported here @steroberts89 
Sorry about that!

So, there actually isn't 3 versions for Umbraco 7 😅 I'm afraid what you're seeing, is a flaw in the way we've setup the version targeting.
![image](https://user-images.githubusercontent.com/24529086/70905239-ab59c800-2003-11ea-8838-8d013883f12f.png)

The link points to these two:
https://our.umbraco.com/Documentation/Extending/Property-Editors/value-converters-full-example-interface
and
https://our.umbraco.com/Documentation/Extending/Property-Editors/value-converters-full-example-attributes

As they both start with `value-converters` which is the name of the article you're currently on, it will show you that there are two more versions of the same - even though there isn't 🙈 

I've added a tasks to somehow get this fixed - renaming the two articles should do the trick!

sofietoft:
This has been fixed with #2193 😄

PR #2193, opened 2020-01-17
Renaming files to avoid multiple documentation for value converters

poornimanayar:
Hi ,

I have tried to address the issue [2141](https://github.com/umbraco/UmbracoDocs/issues/2141)

I have tried to rename the files and also update a link which I managed to find. Let me know if this helps.

Poornima


sofietoft:
This is great @poornimanayar ! 😄 

Now, I'm not 100% sure, but I think we might run into a similar issue on these examples, since the sames are still similar - "full-example" .. 🤔 

Perhaps we could rename them instead to:
"full-examples-value-converters.md"
"attributes-example.md"
"interface-example.md"

This would ensure that we do not create a similar "versioning box" ..

Does that make sense? 😁

sofietoft:
Thanks for trying out the file names I suggested 💪 

I'll go ahead and merge, so we can see how this looks on Our 😁

sofietoft:
Yup, looks like we're all good 😎 

https://our.umbraco.com/Documentation/Extending/Property-Editors/value-converters

Thanks again for helping out here @poornimanayar ! 🎉

PR #2003, opened 2019-10-16
Remove the needsv8Update from the Udi documentation

poornimanayar:
Helping out with the issue #1875 

I have removed the v8 update tag from Udi documentation. Verified and its all good for v8.

PR #2194, opened 2020-01-17
Verified membershiphelper for v8

poornimanayar:
Hi,

I have verified the `MembershipHelper` documentation against v8.

This corresponds to issue #1875 

Poornima

### DIFF
--- a/Reference/Querying/MemberShipHelper/index.md
+++ b/Reference/Querying/MemberShipHelper/index.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
-needsV8Update: "true"
 ---
 
 # Membershiphelper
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for verifying the article @poornimanayar ! 😄 
Awesome to see all the work you're putting in here 😁 

So, I do have a single comment here:
On line 10 there's a link to the API Documentation, which is great!
Perhaps I could get you to also add a link to the Umbraco 8 version of those API docs? 

Don't think we need to create a new version of the article, just have links to both versions of the API Docs..
What do you think?

poornimanayar:
Hi @sofietoft ,

I would suggest we approach it the same way we did for Udi documentation in this PR #2003 . Gives it consistency and its one file less to manage.

https://our.umbraco.com/documentation/Reference/Querying/Udi

Poornima

sofietoft:
Ah, yes! That's great @poornimanayar !

Looks like this one is ready then. I'll merge!

Thanks again for putting in the work to get this updated 💪 Much appreciated!

PR #2192, opened 2020-01-17
Updated Content-Apps/index.md

erikjanwestendorp:
Updated code example below "Creating a Content App in C#"

https://github.com/umbraco/UmbracoDocs/issues/2191

PR #2196, opened 2020-01-17
Variable 'content' does not exist in this context.

Adolfi:
`if(content is IContent)` is not compiling, since variable '_content_' does not exist.
The variable we want to use is the passed in object parameter '**source**'.


sofietoft:
Thanks @Adolfi ! 😄 

Looks like someone beat you to it though: #2192 

Perhaps you could review his PR for me instead? 😇 
Would be a great help!

Adolfi:
@sofietoft ah missed that one. Absolutely, I can review it.

PR #2197, opened 2020-01-19
Updated CustomViewPage code example

prjseal:
Removed redundant call to base.InitializePage
Made siteService variable that is passed in to have lowercase s so it is different from the property which has uppercase S.

### DIFF
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -603,16 +603,11 @@ namespace Umbraco8.ViewPages
             )
         {
         }
-        public CustomViewPage(ISiteService SiteService, ServiceContext services, AppCaches appCaches)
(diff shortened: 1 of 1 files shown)


dampee:
Hi @prjseal 
Why would you remove the example override? It is not obvious that this method exists. Wouldn't it be better to add an example why you could override instead of hiding this code from the docs?

prjseal:
I don’t mind. Leave it in if you want but resharper will tell you to delete it.

dampee:
Resharper is not a tool too improve documentation right? Would you be able to update the PR and help us improve the docs?

prjseal:
@dampee updated as requested

dampee:
Thanks a lot @prjseal

PR #2198, opened 2020-01-20
Remove content app badge

erikjanwestendorp:
In https://github.com/umbraco/UmbracoDocs/pull/2192 I fixed some weird indetation suggested by @Adolfi. Accidentally I added a line of code to the first c# code example but this should be deleted since this is a 8.4.x feature and is explained underneath "Notification badges"


sofietoft:
Ah, good catch! 😅 

Let's get that merged right away.
Thanks! 😁

PR #2199, opened 2020-01-20
Make Excluded DocType List more clear

marcemarc:
Excluded Document Types seems a clearer label for document types that will be excluded...


sofietoft:
Makes sense! 😄 👏

PR #2210, opened 2020-01-21
Content apps: use umbraco constants

erikjanwestendorp:
Use Umbraco Constants so that it is more consistent with other pages such as:

https://our.umbraco.com/documentation/Extending/Dashboards/


sofietoft:
That's a great observation @erikjanwestendorp !

I definitely agree that we should do what we can to maintain some consistency.
The changes look good to me, but C# isn't my strongest suite, so I've passed it along to our experts for a quick review 😁

dampee:
Thanks a lot for the improvement @erikjanwestendorp !

PR #2211, opened 2020-01-21
Temp-1853-Verify-Creating-A-Property-Editor-For-v8

Adolfi:
Issue: https://github.com/umbraco/UmbracoDocs/issues/1853
Task: Tutorial: Creating a Property Editor

* I've ran through the tutorial and can confirm that it still works for v8.
* I've set the needsV8Update = "false".
* I noticed that throughout this page we are using 3 different personas: **I, You** and **We**. ex. _"..in the .html file I'll add", "..how you can test your property editor" and "..we will have a markdown editor running"._ I've replaced all of the different personas with "we" since it makes more sense in a documentation.
* I also rephrased the "Register the data type in Umbraco" section to avoid using the term "Developer" or "Settings" section, so we can avoid having to create a new version of this docs-page because of one word (since Data Types has moved to the Settings section in v8).


Adolfi:
A damn, I just noticed that I branched from my other PR which made this PR look awful.
I close this one and create a new one. Sorry.

PR #2213, opened 2020-01-22
Fixed links for v8 reference

soreng:


sofietoft:
Good catch @soreng ! 😄 Thanks for getting the links up to date 💪

PR #2214, opened 2020-01-22
Fix a borked codesample

greystate:
Stumpled upon this on Our and since the **Edit page** button is _right there,_ I hurried over here and fixed it :)


sofietoft:
Great catch @greystate ! 😄 

Yeah, isn't it just easy helping out with the Docs? 😉 

Thanks for fixing the typo! 😁

Issue #1873, opened 2019-09-25
Help wanted: Docs about Routing


sofietoft:
# Help wanted: Docs about Routing
Many of our articles needs to be verified against Umbraco 8, and for that we need your help!! 🙏

Our article series about **Routing** needs to be verified and updated for Umbraco 8. 

We need you're help to do the following:
* Pick an article on the list
* Read it!
* Test out any code snippets, tutorials and step-by-step's in the latest version of Umbraco 8

Does the article still apply for Umbraco 8?
* **Yes**: Tell us in a here, and create a PR where you remove the `needsV8Update` tag at the top of the article.
* **No**: Tell us in an issue, OR create a new Umbraco 8 compatible article about the same topic!

## The List

- [x] [Default Routing](https://our.umbraco.com/documentation/Implementation/Default-Routing/)
- [x] [Default Routing: Inbound Pipeline](https://our.umbraco.com/documentation/Implementation/Default-Routing/Inbound-Pipeline/)
- [x] [Default Routing: Controller Selection](https://our.umbraco.com/documentation/Implementation/Default-Routing/Controller-Selection/)
- [x] [Default Routing: Execute Request](https://our.umbraco.com/documentation/Implementation/Default-Routing/Execute-Request/)
- [x] [Custom Routing](https://our.umbraco.com/documentation/Implementation/Custom-Routing/)

Thanks for helping us out!
We appreciate all your help 😄

sumitkharche:
Hi @sofietoft , 
I am happy to pick this one.

sumitkharche:
Hi @sofietoft, I have completed the changes and raised PR for this issue. So could you please review it and let me know your feedback.

erikjanwestendorp:
Hi @sofietoft I updated the _Default Routing: Execute Request_ page and created a PR. Can you or someone else give it a review and give some feedback if necessary?

erikjanwestendorp:
Hi @sofietoft I updated the _Custom Routing_ page. Can you or someone else give it a review and give some feedback if necessary?

erikjanwestendorp:
@sofietoft the Default Routing page has no `needsV8Update` tag (anymore) but I think the page should be updated. DynamicPublishedContent can be removed for v8.

sofietoft:
Good morning @erikjanwestendorp !

That's a good point. Would be good to have that article in two versions 🤔 
I think it's just a left over from when we added the label to begin with. We kept it off most of the index pages for some reason 😅 

[...]

PR #2215, opened 2020-01-22
Updated the Default Routing: Execute Request page for v8

erikjanwestendorp:
https://github.com/umbraco/UmbracoDocs/issues/1873

- needsV8Update to false
- ContentModel instead of RenderModel
- Deleted DynamicPublishedContent

### DIFF
--- a/Implementation/Default-Routing/Execute-Request/index-v7.md
+++ b/Implementation/Default-Routing/Execute-Request/index-v7.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
-needsV8Update: "false"
 ---
 
 # Executing an Umbraco request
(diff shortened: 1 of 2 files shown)


sofietoft:
This is great @erikjanwestendorp !! 😄 
Thanks for help getting it updated.

All examples and stuff looks great.
2 small things:
- Instead of changing the value to `false` you can also remove the "needsV8Update" - up to you!
- And then, on the article for Umbraco 7 - the links to the Querying section.. Would be great if you could update those, so they point to the Umbraco 7 version of those articles as well?
I believe they've all been versioned 💪

erikjanwestendorp:
@sofietoft Thanks for your quick reply on this.  In my latest commit I deleted "needsV8Update" and updated the links in the v7 article where possible. 

Updated links:

- IPublishedContent
- UmbracoHelper

No specific v7 links for DynamicPublishedContent and MembershipHelper as far as i know.

sofietoft:
Brilliant! 😄 

Yes, you're absolutely right.
Thanks for updating the links, and for putting in the work 💪😁 
I'll get this merged right away!

PR #2222, opened 2020-01-24
Use constants on dashboard page

erikjanwestendorp:
On the [dashboard ](https://our.umbraco.com/documentation/Extending/Dashboards/)page some constants are used for AccessRules. Why not for Sections?  I don't know if using constants is the way to go (Although I prefer Constants) but at least is good to do things the same way 👍. See also https://github.com/umbraco/UmbracoDocs/pull/2210.


sofietoft:
Thanks for the PR @erikjanwestendorp ! 😄 

As with #2210 , I'm not the right person to review - but will make sure to get some of our experts to give this a look as soon as possible 💪

erikjanwestendorp:
@sofietoft Thanks for your quick response! I will be patient :-)

Issue #2128, opened 2019-12-05
Verify and update articles on Services for Umbraco 8


sofietoft:
Below is a list of articles about services that needs to be verified and updates for Umbraco 8.

- [x] [ConsentService](https://our.umbraco.com/documentation/Reference/Management/Services/ConsentService/)
- [x] [ContentService](https://our.umbraco.com/documentation/Reference/Management/Services/ContentService/)
- [x] [ContentTypeService](https://our.umbraco.com/documentation/Reference/Management/Services/ContentTypeService/)
- [x] [DataTypeService](https://our.umbraco.com/documentation/Reference/Management/Services/DataTypeService/)
- [x] [DomainService](https://our.umbraco.com/documentation/Reference/Management/Services/DomainService/)
- [x] [FileService](https://our.umbraco.com/documentation/Reference/Management/Services/FileService/)
- [x] [LocalizationService](https://our.umbraco.com/documentation/Reference/Management/Services/LocalizationService/)
- [x] [MediaService](https://our.umbraco.com/documentation/Reference/Management/Services/MediaService/)
- [x] [MemberGroupService](https://our.umbraco.com/documentation/Reference/Management/Services/MemberGroupService/)
- [x] [MemberService](https://our.umbraco.com/documentation/Reference/Management/Services/MemberService/)
- [x] [MemberTypeService](https://our.umbraco.com/documentation/Reference/Management/Services/MemberTypeService/)
- [x] [RelationService](https://our.umbraco.com/documentation/Reference/Management/Services/RelationService/)
- [x] [SectionService](https://our.umbraco.com/documentation/Reference/Management/Services/SectionService/)
- [x] [UserService](https://our.umbraco.com/documentation/Reference/Management/Services/UserService/)
- [ ] [ApplicationTreeService](https://our.umbraco.com/documentation/Reference/Management/Services/treeservice/)

Related:
Services that still needs to be documented: https://github.com/umbraco/UmbracoDocs/issues/847
ContentService: https://github.com/umbraco/UmbracoDocs/issues/1712
Removing list of available methods: https://github.com/umbraco/UmbracoDocs/pull/1192

sofietoft:
All services on this list has been updated.
Thanks to all who's helped out 💪 
Much appreciated!

PR #2227, opened 2020-01-27
Verified ContentService documentation for v8

poornimanayar:
This is for issue https://github.com/umbraco/UmbracoDocs/issues/2128. I have verified `ContentService` docs for v8. The `ConsentService` has separate docs for v7 and v8. But I have deviated from it as the `ContentService` only needed updated API docs link to v8. I can always make it consistent with `ConsentService` and have separate v7 and v8 pages.

### DIFF
--- a/Reference/Management/Services/ContentService/Index-v7.md
+++ b/Reference/Management/Services/ContentService/Index-v7.md
@@ -1,6 +1,6 @@
 ---
 keywords: services content service
-versionFrom: 8.0.0
+versionFrom: 7.0.0
 ---
 
 # ContentService
@@ -11,7 +11,7 @@ Applies to Umbraco 6.0.0+
 
 The ContentService acts as a "gateway" to Umbraco data for operations which are related to Content.
 
-[Browse the v8 API documentation for ContentService](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.Services.IContentService.html).
+[Browse the v7 API documentation for ContentService](https://our.umbraco.com/apidocs/v7/csharp/api/Umbraco.Core.Services.ContentService.html).
 
  * **Namespace:** `Umbraco.Core.Services`
  * **Assembly:** `Umbraco.Core.dll`
--- a/Reference/Management/Services/ContentService/Index-v7.md
+++ b/Reference/Management/Services/ContentService/Index-v7.md
@@ -1,6 +1,7 @@
 ---
 keywords: services content service
 versionFrom: 7.0.0
+versionRemoved: 8.0.0
 ---
 
 # ContentService
--- a/Reference/Management/Services/ContentService/Index-v7.md
+++ b/Reference/Management/Services/ContentService/Index-v7.md
@@ -5,10 +5,6 @@ versionFrom: 7.0.0
 
 # ContentService
 
-:::note
-Applies to Umbraco 6.0.0+
-:::
-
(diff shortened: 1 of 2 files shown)


sofietoft:
I think it would be great to keep some consistency here 😄 

Would you mind splitting it up into two articles instead? 
And also, I noticed that the link for Umbraco 8 API isn't correct? 🤔 - or it currently gives a 404..
For `ConsentService`, it looks like `IConsentService` is referenced instead on the Umbraco 8 version?

Now, I'm no expert in this at all 😅 But would it be possible that we should link to `IContentService` on this one as well?

poornimanayar:
@sofietoft , That was me being careless! I have fixed it all now and I have split the v7 and v8 into two files. I wasn't sure about removing the mention of v6 in there, so I have retained it for now.

sofietoft:
Thanks @poornimanayar ! 😄 

I think we should get rid of the mention of Umbraco 6 .. on both articles actually!
Doesn't seem relevant anymore 😁

poornimanayar:
Thats done now! Now I have a direction for other services. I shall try and do more during this week if I can :-)

sofietoft:
Sounds great @poornimanayar ! 😄 

I'll run this past the team as well, as I'm no expert in this field 😅

poornimanayar:
Yes please :-)

sofietoft:
I've had a chat with Marc about this now, and we have a few notes for you.

The whole "Getting the Service" part needs to be updated for Umbraco 8 as well. Seems that there are some difference on how to do that, between Umbraco 7 and 8 🤔

So, the whole idea with documenting the Services here on Github, is to link off to the auto-generated docs on the service, a few notes on what it's used for, a few examples of getting a reference to it, and finally a small and very generic use-sample 😁

Reference/Management/Services/ContentService/Index-v7.md
```diff
@@ -0,0 +1,46 @@
+---
+keywords: services content service
+versionFrom: 7.0.0
+versionRemoved: 8.0.0
```
sofietoft: @dampee - By adding the `versionRemoved`, isn't it going to say that the ContentService has been removed in Umbraco 8? It hasn't been removed, has it? 🤔

dampee: @sofietoft Let me verify that. :-o

dampee: It will mention 7.0.0 - 8.0.0 instead of 7.0.0+
So I think the change is ok.

sofietoft:
@poornimanayar - quick update here!

We (the Documentation Curators) have talked about this whole deal about getting all the articles about services up to date for Umbraco 8.
So, if you have a look at the MediaService and how that's been updated for Umbraco 8: https://our.umbraco.com/Documentation/Reference/Management/Services/MediaService

We're looking for a structure similar to that, containing:
* Paragraph explaining what they are
* Perhaps some tips on when to use them
* Link to API
* If relevant: "Real life" use sample
* Link to Services and Helpers documentation (https://our.umbraco.com/Documentation/Implementation/Services/)

How does that all sound to you? 😄

poornimanayar:
Sounds good @sofietoft an example will be super helpful and we have exactly that!

sofietoft:
Hi @poornimanayar - apologies for the late reply here!

I'll merge this PR now. Thanks for helping getting it updated 😄

PR #2236, opened 2020-02-01
Update link to one relevant for v8

sussexrick:
The blog post previously linked was for PetaPoco. 

v8 uses NPoco and the syntax is completely different. The Umbraco docs page is much more helpful.


sofietoft:
Thanks for adding the changes @sussexrick ! 😄 
I agree that it makes sense to update the link here.

However, instead of simply removing the other link - how about we link to both instead? And say something along the lines of "If you using Umbraco 8, check this, and if you're using Umbraco 7, check this..."
What do you think?

sussexrick:
I agree it makes sense to link both, but having done some more work looking into this topic I think the blog link should be on the page I've linked to here. 

Migrations are potentially relevant to any Umbraco install, not just Cloud, so the Cloud page should just have a pointer to the migrations elsewhere either in the docs. If I weren't using Cloud I wouldn't look here.

The destination page should use the ability to have separate versions of the page for v7 and V8. I'm not sure how to do that. The v7 page would link to the blog article I removed. The V8 version should retain the existing content of that page and link to this blog post by Stephan that I found really useful:

https://www.zpqrtbnk.net/posts/migrations-in-v8

sofietoft:
Great idea, adding the external links to the general articles on Databases instead 💪 I like it!

I completely understand that our versioning is a bit confusing 😅 
We've made some docs about it here:
One about how to name the files: https://our.umbraco.com/documentation/Contribute/file-naming-conventions
And one about which labels to add to the articles: https://our.umbraco.com/documentation/Contribute/adding-metadata

Would you be up for taking this on? 😄 I'd be happy to help out if you have any questions.

sofietoft:
Hi @sussexrick !
It's been a while since I've heard from you, so I've taken the liberty to update this PR a bit.

I'll get it merged now, and then I'll create an Issue on how we need to have these details moved to the /Extending/Databases section instead 💪 

Thanks again for putting in the work to make this PR!

PR #2237, opened 2020-02-03
Update dynamicpublishedcontent page

erikjanwestendorp:
Moved @pagename to code block


sofietoft:
Good catch @erikjanwestendorp ! 😄 

Thanks for helping out, once again! 💪

PR #2247, opened 2020-02-06
Generated models namespace changed

Jeavon:
I can't recall which version this was changed in but certainly 8.5+


sofietoft:
Ah, you're right @Jeavon !

Looks like that one slipped past us, when we updated the Models Builder docs.
Thanks 😁

PR #2250, opened 2020-02-07
Add minor change to excludedDocumentTypes check

rachelwong:
To check each page against the excluded document types, the code needs to reference correctly the document's ContentType alias. The correct version is actually already in the final full code sample, but not in this individual step explaining the reasoning. This change makes everything consistent across the entire tutorial and the code can now run correctly.


sofietoft:
Good catch @rachelwong ! 😄 

Thanks for help keeping the documentation consistent! 💪

marcemarc:
thanks @rachelwong !!

PR #2251, opened 2020-02-07
Update jQuery version

jmayntzhusen:
Some of the frontend functionality for Forms requires jQuery.

I did some quick testing with an up to date version and things that use jQuery still work with a later version. So we should recommend people to use that.

Tested the following:
- Conditional fields
- Validation
- Datepicker field (pickaday has a jQuery dependency)
- Mandatory messages


sofietoft:
Thanks @jmayntzhusen !
Great suggestion getting this updated 💪 And thanks for testing 😬

PR #2262, opened 2020-02-12
Fixed constructor name in the PVC example

poornimanayar:
There was a bug in the code, the constructor name is wrong, I have fixed it.


sofietoft:
Great find @poornimanayar ! 😄 Thanks!

PR #2264, opened 2020-02-12
Serilog config description - Incomplete Copy

poornimanayar:
Noticed that Serilog config is missing some copy. I have added that.


sofietoft:
Brilliant! Thanks @poornimanayar ! 😄

PR #2268, opened 2020-02-12
Fix the url to language files

poornimanayar:
The link to language files is 404-ing. Fixed that.


sofietoft:
Ah! Looks like an extra `/Extending` snuck its way in there 🙈 

Thanks for getting it fixed @poornimanayar !

PR #2269, opened 2020-02-12
Update /Reference/Searching/index.md

erikjanwestendorp:
I think the [Examine introduction page](https://our.umbraco.com/documentation/Reference/Searching/) fits to both v7 and v8. 
- Remove needsV8Update


sofietoft:
Looks like you're right @erikjanwestendorp ! 😄 💪 

Thanks for the update.

PR #2270, opened 2020-02-12
Fixed some typos in property actions

poornimanayar:
Hi,

I have fixed some typos and encouraged some consistency.

- 'build-in', shouldn't it be built-in?
- it was worded 'property-editor' in a place, shouldn't it be 'property editor'?

Poornima


sofietoft:
Thanks for adding the corrections @poornimanayar ! 😄 

I agree with all the changes you've made. Much appreciated!
I'll make sure this gets merged right away.

PR #2271, opened 2020-02-13
Bringing consistency to App_Plugins folder wording

poornimanayar:
Hi,

The folder was worded 'app_plugins' and 'App_Plugins' in the same document. I have made it uniform to 'App_Plugins' 

There are at least 40 places in the whole documentation. I can do a replace wherever sensible but I think it will be worth having the lint check for the word like 'backoffice' long term. What do you think?

Poornima

### DIFF
--- a/Extending/Dashboards/index.md
+++ b/Extending/Dashboards/index.md
@@ -12,7 +12,7 @@ The dashboard area of Umbraco is used to display an 'editor' for the selected it
 There are two approaches to registering a custom dashboard to appear in the Umbraco Backoffice:
 
 ### Registering with package.manifest
-Add a file named 'package.manifest' to the app_plugins folder, containing the following json configuration pointing to your dashboard view:
+Add a file named 'package.manifest' to the 'App_Plugins' folder, containing the following json configuration pointing to your dashboard view:
 
 ```json
 {


sofietoft:
Thanks @poornimanayar !

And I think you're right!
It's a small thing, but it would still make sense to have it consistently written the same way.
You're more than welcome to add a PR to update the style guide, if you want?
So, that would be this document, I believe: https://github.com/umbraco/UmbracoDocs/blob/master/.github/valeStyle/Terms.yml 🤔

poornimanayar:
Not sure whether it would work as there is an `ignorecase:true`. Also there are two places where it wouldnt need update so not sure whether we can target them, they are here. 
https://our.umbraco.com/documentation/Reference/Security/Security-hardening/#lock-down-access-to-your-umbraco-folders
https://our.umbraco.com/Documentation/Reference/Security/Security-hardening/index-v7#lock-down-access-to-your-umbraco-folders

Poornima

sofietoft:
I think it's fine 🤔 
The "error level" of the terms is `warning`, so as long as it can be explained why the term isn't changed, I think it's OK to have it.

poornimanayar:
Pr in its way very soon 😉

PR #2274, opened 2020-02-14
Rename Slave to Replica i v8 Load Balancing docs.

Adolfi:
* In v8, Slave is no longer a thing. It has been renamed to Replica, so we should not be using the term Slave anymore.
* Removed empty list-item in Scheduling and master election list


sofietoft:
Great find @Adolfi !

We definitely shouldn't be using the term "slave" any more.
Thanks for putting in the PR to get it removed 💪

PR #2288, opened 2020-02-17
Add link to packages pages from extending index.

KevinJump:
Hi, the packages link was missing, just think it wasn't added when we created the packages section


sofietoft:
Great catch @KevinJump ! 😄 

Let's get this merged right away.

PR #2291, opened 2020-02-18
Cloud package considerations

jmayntzhusen:
Add information on storing data from a package with regards to Umbraco Cloud


Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+Let's take a look at the most common ways of storing data in packages - and what to watch out for on Cloud:
+
+### Package actions
+
+A [package action](../Package-Actions/custom-package-actions.md) are used to run some commands on package install and uninstall. One thing to watch out for with regards to package actions is that whatever code you run in this will only be executed on the environment you install the package on. So if you for example set up a database table in a package action that db table will not be generated in each environments db.
+If you need a db table in each environments db you will probably need to look at migrations below instead.
+
+### Migrations
+
+A [database migration](../../Database/index.md) is a set of database commands you run as part of a migration plan. That migration plan has an id that is stored in the database (in the KeyValue table). This means that when you add new migrations Umbraco will only execute the ones that came after the one with the stored id.
```
dawoe: Migrations are not limited to database operations. You can run just run code as well : https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Core/Migrations/Upgrade/V_8_0_1/ChangeNuCacheJsonFormat.cs

Especially because you mention doing things that generate a Umbraco Schema. This is normally not something that happens when you purely do database migrations, but when use the services (dictionary service, datatype service, ...)

I get the confusion. Because the migrations documentation is located under database.

jmayntzhusen: Ah yes, good point! I knew that we use migrations on Our to generate doc types, templates, data types, etc but didn't make the connection when writing this 😕

Thanks for pointing it out I will fix it up later today!

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+- [Using custom property editors with Deploy](#valueconnectors)
 
-# ValueConnectors
+## Storing data
+
+When developing a package you will sometimes store data, this can be data in many forms - Umbraco schema / content, package settings, etc.
+
+When you develop a package for Umbraco Cloud there are a few things to be aware of when storing data, mainly whether you want that data to be specific to 1 environment or more.
+
+Let's take a look at the most common ways of storing data in packages - and what to watch out for on Cloud:
```
sofietoft:
```suggestion
Let's take a look at the most common ways of storing data in packages - and what to watch out for on Cloud.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+
+### Package actions
+
+A [package action](../Package-Actions/custom-package-actions.md) are used to run some commands on package install and uninstall. One thing to watch out for with regards to package actions is that whatever code you run in this will only be executed on the environment you install the package on. So if you for example set up a database table in a package action that db table will not be generated in each environments db.
```
sofietoft:
```suggestion
A [package action](../Package-Actions/custom-package-actions.md) is used to run some commands on package install and uninstall. One thing to watch out for with regards to package actions is that whatever code you run will only be executed on the environment you install the package on. So if you for example set up a database table in a package action that table will not be generated in each environment.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+If you need a db table in each environments db you will probably need to look at migrations below instead.
```
sofietoft:
```suggestion
If you need a database table in each environment you will probably need to look at migrations below instead.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
sofietoft:
```suggestion
A [database migration](../../Database/index.md) is a set of database commands you run as part of a migration plan. That migration plan has an ID that is stored in the database (in the KeyValue table). This means that when you add new migrations Umbraco will only execute the ones that came after the one with the stored ID.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+
+### Migrations
+
+A [database migration](../../Database/index.md) is a set of database commands you run as part of a migration plan. That migration plan has an id that is stored in the database (in the KeyValue table). This means that when you add new migrations Umbraco will only execute the ones that came after the one with the stored id.
+
+As database migrations are run in the database of the site it also means that they will run on each environment you trigger them on. The most common way to trigger a db migration is to include them in a [composer](../../../Implementation/Composing/index.md), which will ensure they run on site startup. This means any database commands you have in your migration will automatically run when the site starts up. When your package code is pushed to a new environment it will run them from the beginning on that environment as no id is saved in the database.
```
sofietoft:
```suggestion
As database migrations are run in the database of the site it also means that they will run on each environment you trigger them on. The most common way to trigger a database migration is to include them in a [composer](../../../Implementation/Composing/index.md), which will ensure they run on site startup. This means that any database commands you have in your migration will automatically run when the site starts up. When your package is pushed to a new environment it will run the code from the beginning on that environment as no ID is saved in the database.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+
+This is normally a good thing, however if you generate any Umbraco schema then Deploy will automatically create [UDA files](../../../Umbraco-Cloud/Set-Up/Power-Tools/Generating-UDA-files/index.md#what-are-uda-files) based on that schema, and commit them to source control. Which means if you then deploy all your files to the next environment the migration will run again, create duplicates and generate duplicate UDA files - will end up causing a lot of issues.
```
sofietoft:
```suggestion
This is normally a good thing. However if you generate any Umbraco schema then Umbraco Deploy will automatically create [UDA files](../../../Umbraco-Cloud/Set-Up/Power-Tools/Generating-UDA-files/index.md#what-are-uda-files) based on that schema, and commit them to source control. This means that when you deploy all your files to the next environment the migration will run again, create duplicates and generate duplicate UDA files, which could end up causing a lot of issues.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+
+You could consider creating Umbraco schema only during a package action, and then running things like creating db tables in migrations. Another good workaround could be to not run the migrations in a composer, but rather create a dashboard for the package where the user can choose which migrations to run themselves. The package [Articulate has an example of this](https://github.com/Shazwazza/Articulate/blob/master/build/packageManifest.xml#L613).
```
sofietoft:
```suggestion
You could consider creating Umbraco schema only during a package action, and then running things like creating database tables in migrations. Another good workaround could be to not run the migrations in a composer, but rather create a dashboard for the package where the user can choose which migrations to run themselves. The [Articulate package](https://github.com/Shazwazza/Articulate/blob/master/build/packageManifest.xml#L613) has an example of this.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+
+### Creating files
+
+You may sometimes choose to save data in a file, could be a seperate config file for your package or a [config transform file](../../..//Umbraco-Cloud/Set-Up/Config-Transforms/index.md) to fx. add an app setting to web.config.
```
sofietoft:
```suggestion
You may sometimes choose to save data in a file. Could be a seperate config file for your package or a [config transform file](../../..//Umbraco-Cloud/Set-Up/Config-Transforms/index.md) to add an app setting to the web.config.
```

Extending/Packages/Packages-on-Umbraco-Cloud/index.md
```diff
@@ -4,9 +4,42 @@ versionFrom: 8.0.0
...
+If you do this be aware of two things:
+1.  If these files are generated on a Cloud environment they will not be stored in source control, and will be overwritten on next deployment. They need to be installed locally, committed to source control and then pushed up to the Cloud environments. We have an [existing feature request](https://github.com/umbraco/Umbraco.Cloud.Issues/issues/33) on allowing package creators to commit their files directly on Cloud, and it is possible to do so currently but not in a supported way, and it may change suddenly.
```
sofietoft: I like that you mention the existing feature request here, as it makes a lot of sense..
But I do have one concern: How are you keeping track of this? Like, how do we ensure that this articles is updated, once the feature request is realized / added? 😅

jmayntzhusen: Can't guarantee that we will, up to you if it's not worth it 😁

sofietoft: Let's just put it in for now 💪

PR #2293, opened 2020-02-19
Update  Data Persistence page

erikjanwestendorp:
There is a small issue on the [Data Persistence](https://our.umbraco.com/documentation/Implementation/Data-Persistence/) page. 
To get content from the ContentService GetById() should be used instead of Get() which not exists. :-)

### DIFF
--- a/Implementation/Data-Persistence/index.md
+++ b/Implementation/Data-Persistence/index.md
@@ -14,7 +14,7 @@ exposed as a property on all Umbraco base classes such as `SurfaceController`s, 
 So for the majority of cases you can access the services by using this code (for example):
 
 ```csharp
-Services.ContentService.Get(123);
+Services.ContentService.GetById(123);
 ```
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Good catch @erikjanwestendorp ! 💪 

I can see that the article you're editing hasn't been updated for Umbraco 8.
The corrections you're making here - does that apply to both Umbraco 7 and 8?

erikjanwestendorp:
@sofietoft I have also seen that this article still needs to be updated for v8. I wanted to update the page but was unsure about the second part (How to use the services in a non UmbracoContext). I don't know what's the best way to achieve that in v8.  

But the changes I've made in this PR fits to both v7 and v8.

sofietoft:
Ah, sure! No problem @erikjanwestendorp ! 😄 
Also just wanted to double-check.

Let's get this merged in first then 💪 
Thanks again for the PR!

PR #2298, opened 2020-02-20
Retrieving content types update for v8

erikjanwestendorp:
Update the 'retrieving content types" page for v8

### DIFF
--- a/Reference/Management/Services/ContentTypeService/Retrieving-content-types-v7.md
+++ b/Reference/Management/Services/ContentTypeService/Retrieving-content-types-v7.md
@@ -1,0 +1,69 @@
+---
+versionFrom: 7.0.0
+---
+
+# Retrieving content types
+
+## Getting a single content type
+
+A given content type has a few different unique identifier that we can use to look it up via the content type service. For instance, if we know the numeric ID of the content type, we can look it up like this:
+
+```C#
+// Get a reference to the content type by its numeric ID
+IContentType contentType = contentTypeService.GetContentType(1234);
+```
+
+As Umbraco is shifting to use GUIDs instead of numeric IDs, you can also lookup a content type if you already know its GUID ID:
+
+```C#
+// Declare the GUID ID
+Guid guid = new Guid("796a8d5c-b7bb-46d9-bc57-ab834d0d1248");
+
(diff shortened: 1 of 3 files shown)


sofietoft:
Thanks for getting this updated @erikjanwestendorp ! 😄 

I'll run it by the rest of the team,  as I'm not entirely sure I'm equipped to review this!

Jeavon:
Great work, thanks!
I think we can reword this a bit in terms of "As Umbraco is shifting to use GUIDs instead of numeric IDs". v8 is all GUID first, likewise for the code examples, we can add samples using GUIDs instead of IDs or at least in addition too.
I'm happy to do this post-merge

erikjanwestendorp:
@Jeavon Thanks for your review! 
I get your point about GUIDs but not all methods have a GUID counterpart e.g. 
GetAll()  and GetChildren()

Jeavon:
GetAll does exist but you have discovered something missing with GetChildren() as it exists in the service but not the interface so you cannot use it ;( - I have submitted a PR to fix this https://github.com/umbraco/Umbraco-CMS/pull/7700

`.GetAll(new List<Guid>(){Guid.Parse("ca4249ed-2b23-4337-b522-63cabe5587d1"),Guid.Parse("ca4249ed-2b23-4337-b522-63cabe5587d1")});`

erikjanwestendorp:
@Jeavon Awesome!! 

It seems that an overload method of GetAll() has been added in v8.4 and above. So the GUID variant does not exist in v8.3 and below. So what do we have to do with the docs? Create a specific version of this page for v8.4+?

Jeavon:
@erikjanwestendorp you can create index-vpre.8.4.0.md and ensure the versionFrom and versionTo yaml data is set correctly.

erikjanwestendorp:
@Jeavon Thanks again! I created a vpre.8.4.0.md version and also give the GUID examples some more attention!

Jeavon:
Thanks @erikjanwestendorp 
Once v8.7.0 is released you could update it again with examples of using Guid for GetChildren :wink:

erikjanwestendorp:
@Jeavon Great, I'll keep an eye on it :-)

PR #2305, opened 2020-02-23
Added new page about declaring property editors

abjerner:
Right now we have the [**Package Manifest**](https://github.com/umbraco/UmbracoDocs/blob/master/Extending/Property-Editors/package-manifest.md) page that describes how to declare a property editor using `package.manifest` files.

It's also possible to declare property editors using C#, but I haven't been able to find anywhere it the documentation that mentions this. Based on community feedback, only few people seem to know about this part. So with this PR, it's my proposal that this new **Declaring your property editor** page will replace the **Package Manifest** page.

For now this PR only adds the new page, but doesn't update any existing files, as I'd like to hear first whether the documentation team agrees with my proposal.

### DIFF
--- a/Extending/Property-Editors/Declaring-your-property-editor.md
+++ b/Extending/Property-Editors/Declaring-your-property-editor.md
@@ -180,7 +180,7 @@ Optionally one may also specify a third parameter, which is the priority of the 
 
 ### DataEditor class
 
-In the example above, the `SirTrevorEditor ` class doesn't really much. For simple property editors, the C# approach may require a bit more work compared to that of `package.manifest` files. But as property editors grow in complexity, using C# becomes a bit more useful - and also lets you do things not possible with `package.manifest` files.
+In the example above, the `SirTrevorEditor ` class doesn't really do much. For more basic property editors, the C# approach may require a bit more work compared to that of `package.manifest` files. But as property editors grow in complexity, using C# becomes a bit more useful - and also lets you do things not possible with `package.manifest` files.
 
 The [DataEditor](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Core.PropertyEditors.DataEditor.html) class defines a virtual `CreateConfigurationEditor` method. It returns a model which is used for the Angular view when editing the prevalues of a data type.
 
--- a/Extending/Property-Editors/Declaring-your-property-editor.md
+++ b/Extending/Property-Editors/Declaring-your-property-editor.md
@@ -8,7 +8,7 @@ A property editor consists of a number of mandatory properties, and some optiona
 | ---- | ---- | ---- | ---- |
 | `alias` | string | Yes | A unique alias that identifies the property editor. |
 | `name` | string | Yes | The friendly name of the property editor, shown in the Umbraco backoffice. |
-| `editor` | object | Yes |  |
+| `editor` | object | Yes | This describes details about the editor. See the table below for further information. |
 | `icon` | string | No | A CSS class for the icon to be used in the **Select Editor** dialog - eg: `icon-autofill`. |
 | `group` | string | No | The group to place this editor in within the **Select Editor** dialog. Use a new group name or alternatively use an existing one such as **Pickers**. |
 | `isParameterEditor` | boolean | No | Enables the property editor as a macro parameter editor. Can be either `true` or `false` (default). |
--- a/Extending/Property-Editors/Declaring-your-property-editor.md
+++ b/Extending/Property-Editors/Declaring-your-property-editor.md
@@ -1,0 +1,334 @@
+# Declaring your property editor
+
+Generally Umbraco supports two different ways to declare a property editor. Most commonly one would create a `package.manifest` file, and then use it for declaring one or more property editors. But as an alternative, property editors can also be declared using C#.
+
+A property editor consists of a number of mandatory properties, and some optional ones as well. As such, the outer JSON object for the property editor has the following properties:
+
+| Name | Type | Required | Description |
+| ---- | ---- | ---- | ---- |
+| `alias` | string | Yes | A unique alias that identifies the property editor. |
+| `name` | string | Yes | The friendly name of the property editor, shown in the Umbraco backoffice. |
+| `editor` | object | Yes |  |
+| `icon` | string | No | A CSS class for the icon to be used in the **Select Editor** dialog - eg: `icon-autofill`. |
+| `group` | string | No | The group to place this editor in within the **Select Editor** dialog. Use a new group name or alternatively use an existing one such as **Pickers**. |
+| `isParameterEditor` | boolean | No | Enables the property editor as a macro parameter editor. Can be either `true` or `false` (default). |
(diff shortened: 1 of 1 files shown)


Extending/Property-Editors/Declaring-your-property-editor.md
```diff
@@ -0,0 +1,334 @@
...
+
+Generally Umbraco supports two different ways to declare a property editor. Most commonly one would create a `package.manifest` file, and then use it for declaring one or more property editors. But as an alternative, property editors can also be declared using C#.
+
+A property editor consists of a number of mandatory properties, and some optional ones as well. As such, the outer JSON object for the property editor has the following properties:
+
+| Name | Type | Required | Description |
+| ---- | ---- | ---- | ---- |
+| `alias` | string | Yes | A unique alias that identifies the property editor. |
+| `name` | string | Yes | The friendly name of the property editor, shown in the Umbraco backoffice. |
+| `editor` | object | Yes |  |
```
sofietoft: No description on the `editor` property? 😄

Extending/Property-Editors/Declaring-your-property-editor.md
```diff
@@ -0,0 +1,334 @@
...
+
+As shown in the C# example, the [PropertyEditorAsset](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Web.PropertyEditors.PropertyEditorAssetAttribute.html) attribute was used to make Umbraco load the specified JavaScript file.
+
+The constructor of the attribute takes the type of the assets as the first parameter. Possible values are either `ClientDependencyType.Javascript` or `ClientDependencyType.Css`. The second parameter is the URL of the asset.
+
+Optionally one may also specify a third parameter, which is the priority of the asset. This enables you to include your asset either before or after other assets, all depending on their respective priorities.
+
+### DataEditor class
+
+In the example above, the `SirTrevorEditor ` class doesn't really much. For simple property editors, the C# approach may require a bit more work compared to that of `package.manifest` files. But as property editors grow in complexity, using C# becomes a bit more useful - and also lets you do things not possible with `package.manifest` files.
```
sofietoft:
```suggestion
In the example above, the `SirTrevorEditor ` class doesn't really do much. For more basic property editors, the C# approach may require a bit more work compared to that of `package.manifest` files. But as property editors grow in complexity, using C# becomes a bit more useful - and also lets you do things not possible with `package.manifest` files.
```

Extending/Property-Editors/Declaring-your-property-editor.md
```diff
@@ -0,0 +1,334 @@
...
+
+Like mentioned before, the `ddd` class doesn't really do much in this example with the Sir Trevor property editor. But the **Multi Node Tree Picker** and others of Umbraco's build in property editors also override the `ToValueEditor` method. 
+
+This method is used when the strongly typed configuration value is converted to the model used by the Angular logic in the backoffice. So with the implementation of the [MultiNodePickerConfigurationEditor]( https://github.com/umbraco/Umbraco-CMS/blob/ade9bb73246caf25a7073f2b9e5262641a201863/src/Umbraco.Web/PropertyEditors/MultiNodePickerConfigurationEditor.cs) class, some additional configuration fields are sent along - for instance that it's a multi picker and that the ID type should be URI's. These are configuration values that the user should not be able to edit, but the property editor may still rely on them.
```
sofietoft:
```suggestion
This method is used when the strongly typed configuration value is converted to the model used by the Angular logic in the backoffice. So with the implementation of the [MultiNodePickerConfigurationEditor](https://github.com/umbraco/Umbraco-CMS/blob/ade9bb73246caf25a7073f2b9e5262641a201863/src/Umbraco.Web/PropertyEditors/MultiNodePickerConfigurationEditor.cs) class, some additional configuration fields are sent along - for instance that it's a multi picker and that the ID type should be URI's. These are configuration values that the user should not be able to edit, but the property editor may still rely on them.
```

sofietoft:
Thanks for the pr @abjerner, and for putting in all the work to write this article!! 😄 

I was sure that we already had something in the docs about declaring the package.manifest via code, but looking through, I also couldn't find it..
I think it would be great having this added to the docs! I'll take it up with the rest of the Docs team, to hear if they have any thoughts on it 💪 

You mention making more changes - would you mind giving some more details on your plans? 😄 I'm curious.

sofietoft:
I found that we **do** have an example of this already: https://our.umbraco.com/documentation/Extending/Content-Apps/#creating-a-content-app-in-c
It's most likely very specific to creating the Content App, but at least an example .. right? 😅

Extending/Property-Editors/Declaring-your-property-editor.md
abjerner: Not sure why I missed that one. Added now 😉

Extending/Property-Editors/Declaring-your-property-editor.md
abjerner: Great 👍

abjerner:
There may be bits and pieces on how to use C# (eg. for the content app), but still haven't found anything about the property editors. 

Regarding my plans, I think the new page could largely replace the current **Package Manifest** page. Or since it seems to be describing the package manifest more generally than just the property editors, perhaps it should be moved elsewhere? Or do we already have the same information somewhere else?

I'm not entirely certain what do to about it, which was also why I didn't include changes for this page in my PR.

image

Btw - a little side node: it seems that the `PropertyEditorAssetAttribute` is only detected when used on classes inheriting from `DataEditor`. So in the example with the content app, there is apparently no other way to add JavaScript or CSS assets than via a `package.manifest` file. Can you confirm whether this is actually the case? Because I think it should be possible to this from C# as well.

sofietoft:
@abjerner - apologies for not getting back to you on this one.

I've had another look, and I think this article you've written could be moved to the root of the "Extending" section - as it is something that's used for more than "just" property editors 🤔 

I'll make sure to bring it up with the Curators tomorrow, when we have our bi-weekly meeting!

sofietoft:
Hi @abjerner !
Apologies for the delay - we had to move a meeting, etc etc.
But! Good news is we've finally had a chance to talk about your PR here 😄 

We all agree that it would be great to have the concept of package.manifests explained in a more general manner, than just mentioned on our various tutorials!

I mentioned the "Extending" section before .. but we actually think the "Reference" section would be a better fit. Then our guides and tutorials in the Extending section can link to it.

How does that all sound? 😁

sofietoft:
Oh, and I forgot your question:
> Btw - a little side node: it seems that the PropertyEditorAssetAttribute is only detected when used on classes inheriting from DataEditor. So in the example with the content app, there is apparently no other way to add JavaScript or CSS assets than via a package.manifest file. Can you confirm whether this is actually the case? Because I think it should be possible to this from C# as well.

I don't have an answer for you right now, but I'll make sure to look into it, so we can get it resolved 💪

sofietoft:
Hi @abjerner !

Apologies for the delay here! 😄 
On the Docs team we've just agreed to merge this one - it's a lot of great information just sitting there for no one to see.

Thanks for adding the article 💪 
Much appreciated!

PR #2308, opened 2020-02-24
Updated where the settings should be added for LocalDB

Aaen:
I have updated the docs so we tell in which section of the file the setting should be added.


sofietoft:
This is great @Aaen ! 😄 

Good to have this detail added to the docs.
Moved it into a note, as I thought that would make better sense 💪

PR #2309, opened 2020-02-24
Update routing-properties.md

erikjanwestendorp:
Deleted needsV8Update. It seems all routing properties work in v8


sofietoft:
Good morning @erikjanwestendorp ! 😄 

Great to get this article updated! Thanks for verifying it 💪

PR #2325, opened 2020-02-28
Added paragraph about the depth querystring

rasmusjp:
Some of the content endpoints in Heartcore have an optional `depth` querystring parameter, I've added a paragraph describing what it does.

### DIFF
--- a/Umbraco-Heartcore/API-Documentation/Content-Delivery/content/index.md
+++ b/Umbraco-Heartcore/API-Documentation/Content-Delivery/content/index.md
@@ -23,6 +23,13 @@ Api-Version: 2
 Umb-Project-Alias: {project-alias}
 ```
 
+## Depth
+
+The `depth` querystring parameter controls how many levels of referenced Content or Media items that is included in the result.
+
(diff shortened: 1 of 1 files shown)


Umbraco-Heartcore/API-Documentation/Content-Delivery/content/index.md
```diff
@@ -23,6 +23,13 @@ Api-Version: 2
 Umb-Project-Alias: {project-alias}
 ```
 
+## Depth
+
+The `depth` querystring parameter controls how many levels of referenced Content or Media items that is included in the result.
+
+Lets say a Content item have a `Multi Node Tree Picker` and one of the Content items that can be picked have a `Media Picker`, if level is set to `1` the returned data will contain the referenced Content items, but their Media property will be null.
```
sofietoft:
```suggestion
Lets say a Content item have a `Multi Node Tree Picker` and one of the Content items that can be picked have a `Media Picker`. In this case, if the level is set to `1` the returned data will contain the referenced Content items, but their Media property will be null.
```

Issue #2285, opened 2020-02-14
Missing example from Creating a Property Editor tutorial


kenburcham:
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/master/Tutorials/Creating-a-Property-Editor/index.md***

After following this tutorial all the way through (which was pretty great, btw), I found that there is no information or example for how to show the actual markdown output of the "My Markdown Editor" property that gets created as a value on a template. 

The tutorial takes you through creating the Data Type and adding it to a Document Type. Then I can go into a content node and edit using the new Markdown Editor. But the final step would be to see the Markdown displayed on an actual content node of the website in a template. But I can't figure out how to do that and it is skipped in the tutorial.

Thank you!

sofietoft:
Hi @kenburcham ! Thanks for your feedback, and I'm happy to hear you enjoyed the tutorial.

Good point about there not being an example of how to render the property value on the frontend 🤔 - that IS what you're saying, right?

I'll look into how we can get that added as soon as possible!

umbrabot:
Hi @kenburcham,

We're writing to let you know that we've added the `Up For Grabs` label to your issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly PR team bot :-)

kenburcham:
@sofietoft - Yes, that's exactly it, I'd love the tutorial to have an example for how to render the property value on the front end. It would render the markdown that is entered as the value of the property on the page. It looks like it might have something to do with Property Value Converters? https://our.umbraco.com/documentation/Extending/Property-Editors/value-converters But I'm not sure how it should work and an example that implements it for the tutorial would be so helpful!

sofietoft:
Thanks for helping to clear this up! 😁 
Sure would be great to have an example of how to render the custom property on the frontend.

Not sure about the Value converters though 🤔 
The `community/up-for-grabs` label has been added.
Perhaps @Adolfi , who helped get the Property Editor tutorial up to date, could help with an example here? 😄

Adolfi:
@sofietoft @kenburcham absolutely, good point. I will have a look at it.

Adolfi:
[...]

PR #2326, opened 2020-02-28
Added missing example for rendering markdown value.

Adolfi:
From request in https://github.com/umbraco/UmbracoDocs/issues/2285.

I've added an example of how you would render the markdown value in html.
Cheers!


Adolfi:
Friday afternoon PR. Felt a bit rushed. I can do better. Will do a new PR on Monday. 🙃

PR #2330, opened 2020-03-02
Updates docs for Azure Websites with v8

Shazwazza:
Makes v7/v8 specific docs. 

The v8 docs target 8.6 because we introduced a new setting in 8.6 (even though it's still in RC) that needs to be set based on this issue https://github.com/umbraco/Umbraco-CMS/issues/6546 and this PR https://github.com/umbraco/Umbraco-CMS/pull/7306.

Perhaps we should wait until 8.6 is fully released before merging this in.


sofietoft:
This is great @Shazwazza ! Thanks for helping updating this article 😄 🎉 

So, the new index file is for Umbraco 8.6+, and then there's a version for Umbraco 7.
How about this setting when you're running Umbraco 8 to Umbraco 8.5.x?
Just wondering, if we shouldn't add that back in somewhere?

if it IS different in 8.0-8.5.x we could add a note about it, in the Umbraco 8 article? I think that would make most sense.
What do you think? 🤔

Jeavon:
@Shazwazza to clarify with v8.6 and the Umbraco.Core.MainDom.Lock app setting the WEBSITE_DISABLE_OVERLAPPED_RECYCLING setting isn't required. :sparkles:
I can update the Azure Web App load balancing docs on this branch to reflect this also. - https://github.com/umbraco/UmbracoDocs/blob/master/Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md

sofietoft:
@Shazwazza - did you see mine and Jeavons comments here? 😄

Shazwazza:
Yep sorry just haven't had time, its still starred in my inbox. But yes i just wasn't sure how the docs really work with the version tags. You tell me what you think is best:

* Change the version tag to just be 8.0 and then have a bunch of 8.x specific info in the one doc
* Keep the version tag as 8.6 but also have a bunch of 8.x stuff in the one doc
* Have a bunch of different files for different 8.x versions

sofietoft:
No worries! 💪 And thanks for the suggestions.

I think it would be best if we change the version tag to 8.0, and then for stuff applying only to 8.6+ add that as a note or something 🤔 

Then we can always reconsider this, should we run into more version specific changes that needs to be added to this document. 

I'll also bring it up with the Docs team tomorrow at our meeting - would be great with some specific guidelines on how we handle these kind of things 😅

Shazwazza:
@sofietoft I have pushed an update and just added a "NOTE" about 8.0-8.5, let me know if that works i think it's prob the simplest way to explain it

sofietoft:
It's great @Shazwazza ! Thanks for adding the note.

I'll merge this as soon as Umbraco 8.6 is released 💪

PR #2332, opened 2020-03-03
Update IContentFinder.md

JonathonCove:
Corrected `.Children` to be a property rather than a function.

Added null checks in the case that `_domainService.GetAll(true)` returns false (as happened on my project)

Changes are for 8.4.1, I'm not sure how to change for a specific version.


sofietoft:
Thanks for the PR @JonathonCove ! 😄 

I'll make sure we get this reviewed as soon as possible!

sofietoft:
Apologies for the delay here @JonathonCove !

What you've made here looks great 😄 💪 
I'll make sure to merge it in!

Thanks again for putting in the work to get this updated.

PR #2336, opened 2020-03-04
Added missing example from Property editor tutorial

Adolfi:
Issue: https://github.com/umbraco/UmbracoDocs/issues/2285

I added a _:::note_ on the first step of the tutorial, that Umbraco now has a built in markdown editor. As we've discussed @sofietoft, this tutorial is a nice tutorial for how to create your own property editor plugin from start to finish using AngularJs, html, custom configurations, injecting your own or 3rd party JavaScript libraries and using Umbraco’s native dialogs and services etc. but if all you are looking for is a markdown editor, then you don’t need to re-invent the wheel, you can just use the built-in editor. Hence this _:::note_.

I also added an example in step 3 of how to convert the markdown to html format and render it in a view. Aslo included a _:::note_ regarding the built-in Markdown editor and Models Builder.

### DIFF
--- a/Tutorials/Creating-a-Property-Editor/index.md
+++ b/Tutorials/Creating-a-Property-Editor/index.md
@@ -32,6 +32,10 @@ By the end of this guide, we will have a markdown editor running inside of Umbra
 registered as a Data Type in the backoffice, assigned to a Document Type, and the editor can
 create and modify data.
 
+:::note
+This tutorial is a good example of how to create your own property editor from start to finish, however you should know that Umbraco now has a built-in [Markdown editor](https://our.umbraco.com/Documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/#markdown-editor) which is the recommended option for markdown syntax editing.
+:::
+
 ## Setting up a plugin
 
 The first thing we must do is create a new folder inside `/App_Plugins` folder. We will call it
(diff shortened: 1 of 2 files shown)


sofietoft:
Awesome @Adolfi ! 😄 
Thanks for adding in the sample on how to render on the frontend - and all the other additionas you've added - looks great.

I'll make sure to give it a proper review as soon as possible 💪

sofietoft:
I've been trying to set this up today, and I ran into some issues.
When I use the code snippet you added to these docs, I run into an error on my frontend:

`The type or namespace name 'Markdown' could not be found (are you missing a using directive or an assembly reference?)`

I'm just adding it to a standard Template, so my code looks like:

```
@inherits Umbraco.Web.Mvc.UmbracoViewPage<ContentModels.TextPage>
@using ContentModels = Umbraco.Web.PublishedModels;
@{
	Layout = null;
}

@{
    var value = Model.Value<string>("myMarkdownPropertyAlias");
    var html = new Markdown().Transform(value);
}
@Html.Raw(html)
```

I'm most likely doing something wrong - perhaps you can see what it is? 🤔

Adolfi:
@sofietoft oh, my bad. It seems like there´s a missing using statement. I fix straight away.

Adolfi:
@sofietoft please try again. :)

sofietoft:
Ah, perfekt @Adolfi ! 😄 

Now it works brilliantly. 
Thanks again for putting in the work to get this updated.
Much appreciated 👏

Adolfi:
Thanks @sofietoft !!!!

PR #2344, opened 2020-03-06
Add slightly obscure namespace for the cache helpers

Migaroez:


sofietoft:
Thanks for the review @dampee , and for the PR @Migaroez ! 😄 

Looks indeed like it's true for Umbraco 7 - but I'm a bit unsure about Umbraco 8.
In any case, the article updated here, has not yet been updated for Umbraco 8, so I'll go ahead and merge this one for now 😁

PR #2345, opened 2020-03-06
Add back LightInject section

sofietoft:


Reference/Using-Ioc/index.md
```diff
@@ -189,6 +189,21 @@ namespace Example.Core.Components
...
 
+### Accessing LightInject container
+
+Should you need to carry out more complicated registrations beyond the minimalist Umbraco DI implementation, you can access the underlying DI container via the `Concrete` property of the `composition`.
+
+```csharp
+var container = composition.Concrete as LightInject.ServiceContainer;
+container.Register<IFoo, Foo>();
+
+// It's not currently possible to do the following without some workarounds
```
rustybox: I think maybe this

```csharp
// It's not currently possible to assembly scan without some workarounds
// see https://github.com/umbraco/Umbraco-CMS/issues/7502 for details
// container.RegisterAssembly(/* Any method signature */);
```

sofietoft:
Thanks for the suggestion @rustybox ! 💪 

I've applied. Let me know what you think, and we can get this merged today!

rustybox:
Looks good to me, much less likely to cause confusion!

sofietoft:
Brilliant. Thanks for the report and for helping out @rustybox ! 😄 

PR #2354, opened 2020-03-09
Change example of accessing the logger to use DI instead of Umbraco.Core.Composer.Current

TwoMoreThings:
Umbraco.Core.Composing.Current is not recommended as per this discussion (among others)
https://our.umbraco.com/forum/umbraco-8/96270-using-umbracohelper-in-a-custom-class-in-v8#comment-304427
so this showed a bad practice.

I'm not 100% my link to Composing is correct so this will need to be verified

regards
t

### DIFF
--- a/Getting-Started/Code/Debugging/Logging/index.md
+++ b/Getting-Started/Code/Debugging/Logging/index.md
@@ -91,10 +91,9 @@ If you are writing classes that inherit from one of these special Umbraco base c
 
 Then you can access the logging functionality via a special 'Logger' property included in those base classes and use the friendlier syntax of `Logger.Info<T>` to pass the type, if you add a reference to `Umbraco.Core.Logging` as a Using statement.
 
-Outside of these places, eg a ContentFinder or your own custom code, you can get a reference to the logger from the `Umbraco.Core.Composing` namespace, using the `Current.Logger` instance.
+Outside of these places, eg a ContentFinder or your own custom code, you can get a reference to the logger via Dependency Injection.
 
 ```csharp
-using Umbraco.Core.Composing;
 using Umbraco.Core.Logging;
 using Umbraco.Web.Routing;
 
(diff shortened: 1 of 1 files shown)


Getting-Started/Code/Debugging/Logging/index.md
```diff
@@ -91,27 +91,36 @@ If you are writing classes that inherit from one of these special Umbraco base c
...
-            Current.Logger.Info<MyContentFinder>("Trying to find content for url {RequestUrl}", frequest.Uri);
+            _logger.Info<MyContentFinder>("Trying to find content for url {RequestUrl}", frequest.Uri);
 
             //Do Content Finder Logic...
         }
     }
 }
 ```
 
+You will, of course, need to register your ContentFinder [using a Composer](/documentation/Implementation/Composing/index.md)
```
sofietoft: I think this is the correct link! 😁 

```suggestion
You will, of course, need to register your ContentFinder [using a Composer](../../../Implementation/Composing/index.md)
```

sofietoft:
I've had a chat with one of our experts on this, and we like the chance you've made!

However, we think that we should still mention the option to use `Current.Logger` if for some reason DI is not an option.
Perhaps we could say something like:
> While using Dependency Injection is the recommended way, it is possible to `Current.Logger` instead, if DI is not an option.

What do you think? 🤔

TwoMoreThings:
Yes, that sounds good

Getting-Started/Code/Debugging/Logging/index.md
```diff
@@ -91,27 +91,36 @@ If you are writing classes that inherit from one of these special Umbraco base c
 
 Then you can access the logging functionality via a special 'Logger' property included in those base classes and use the friendlier syntax of `Logger.Info<T>` to pass the type, if you add a reference to `Umbraco.Core.Logging` as a Using statement.
 
-Outside of these places, eg a ContentFinder or your own custom code, you can get a reference to the logger from the `Umbraco.Core.Composing` namespace, using the `Current.Logger` instance.
+Outside of these places, eg a ContentFinder or your own custom code, you can get a reference to the logger via Dependency Injection.
```
sofietoft:
```suggestion
Outside of these places, eg a ContentFinder or your own custom code, you can get a reference to the logger via Dependency Injection. While using Dependency Injection is the recommended way, it is possible to Current.Logger instead, if DI is not an option.

The following example uses DI.
```

TwoMoreThings: Hi Sophie, sorry I missed the notification of your comments. I'm happy with that. Thanks, t

sofietoft:
Thanks @TwoMoreThings ! 😄 

I've added the suggestions - both the link and sentence about it still being possible to use Current.Logger.
Let's get this merged 😁

PR #2357, opened 2020-03-10
fix text showing as code

beumsk:


sofietoft:
Ah, good catch @beumsk ! 😄 
Thanks for correcting it 💪

PR #2362, opened 2020-03-11
Update user service

erikjanwestendorp:
Belongs to: https://github.com/umbraco/UmbracoDocs/issues/2128

- Created a v7 version
- Updated UserService for v8


Reference/Management/Services/UserService/Index.md
```diff
@@ -18,24 +15,50 @@ All samples listed in this document will require references to the following dll
...
+using Umbraco.Core.Models;
+using Umbraco.Core.Services;
+```
 
 ## Getting the service
 
-If you wish to use use the user service in a class that inherits from one of the Umbraco base classes - eg. `SurfaceController`, `UmbracoApiController` or `UmbracoAuthorizedApiController` - you can access the user service through a local `Services` property:
+### Services property
+
+If you wish to use use the user service in a class that inherits from one of the Umbraco base classes (eg. `SurfaceController`, `UmbracoApiController` or `UmbracoAuthorizedApiController`), you can access the user service through a local `Services` property:
```
sofietoft:
```suggestion
If you wish to use the UserService in a class that inherits from one of the Umbraco base classes (eg. `SurfaceController`, `UmbracoApiController` or `UmbracoAuthorizedApiController`), you can access the service through a local `Services` property:
```

erikjanwestendorp: @sofietoft Thanks for reviewing!  Maybe good to implement the UpperCamelCase suggestion with other services as well? See for example: https://our.umbraco.com/documentation/Reference/Management/Services/MemberService/

PR #2383, opened 2020-03-18
Reference config

erikjanwestendorp:
Looks like the page has been updated and the needsV8Update can be removed.


sofietoft:
Good observation @erikjanwestendorp ! 😄 

It's not that long ago we, the Docs team, re-structured this whole article.
We might have had a reason for keeping the `needsV8update` tag - but I can't recall 🤔 

Could be because some of the articles explaining the configs.
Hm, let's merge this for now 😁

PR #2385, opened 2020-03-18
Fixed "remove trailing slash" rule to negate /umbraco path

Jay-umbr:
This should also prevent issues with previews when that rule is added.
Courtesy of Jesper M.


sofietoft:
Thanks for this @Jay-umbr ! Looks great.

Looks like there's a merge conflict though - would you be up for clearing that out? 😁

Jay-umbr:
> Thanks for this @Jay-umbr ! Looks great.
> 
> Looks like there's a merge conflict though - would you be up for clearing that out? 😁

Yeah, already done, not sure why there was a conflict though 🤔

sofietoft:
No, it seems kinda strange 😅 

Anywho! Thanks for fixing it.
Now, let's get this merged 💪 

Thanks again for help keeping this up to date! Highly appreciated.

PR #2387, opened 2020-03-18
Fixed misspelling of Elasticsearch

ThomasArdal:


sofietoft:
Good catch @ThomasArdal ! 😄 Thanks for help getting this updated 💪 

PR #2390, opened 2020-03-20
Update Get a reference to a user

erikjanwestendorp:
- Created a v7 version
- Added code blocks to both versions
- Changed ApplicationContext.Services.UserService to Services.UserService in UmbracoViewPage in the v8 version.


sofietoft:
This is great @erikjanwestendorp ! 😄 

I'll make sure we get it reviewed as soon as possible!

PR #2395, opened 2020-03-24
Added reference to cached media

kjsguld:
Add a reference to the documentation on querying Media from the cache, instead of the short explanation on why using an IMedia object is wrong.


sofietoft:
Thanks for adding the reference @kjsguld ! 😄 

PR #2399, opened 2020-03-24
Update Adding-a-Workflowtype.md

thomasandresen:
The example does not work for v8, thus updating the doc with a working one


sofietoft:
Thanks for getting this sample updated @thomasandresen ! 😄 👏 

Could I get you to do one more thing? 😇 
I can see you've completely removed the sample for Umbraco 7.
Instead, it would be great if you could add the Umbraco 7 sample to an article with filename: `Adding-a-Workflowtype-v7.md`, and then in that article also add the following YAML tag to the top of the page:
```
---
versionFrom: 7.0.0
--
```

That way, we'd have samples for both Umbraco 8 and Umbraco 7.
How does that sound? 😁

thomasandresen:
@sofietoft Done :)

sofietoft:
Brilliant, thanks @thomasandresen !

I saw the the Umbraco 7 version was still part of this article, so I took the liberty to remove it, now that you've made the Umbraco 7 article as well 💪 😁

Issue #2129, opened 2019-12-05
Editor's Manual for Umbraco 8


sofietoft:
The [Umbraco 7 Editor's Manual](https://our.umbraco.com/packages/website-utilities/umbraco-7-editors-manual/) has been converted and added to the Umbraco Docs:
https://our.umbraco.com/documentation/Tutorials/Editors-Manual/

This needs to be updated for Umbraco 8!
It's a pretty big section, so I've divided it into smaller bits:

- [ ] [Landing page](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/) + [Introduction](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Introduction/)
- [x] [Getting Started with Umbraco](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Getting-Started-With-Umbraco/)
- [x] [Working with content](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Working-With-Content/)
- [ ] [Version Management](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Version-Management/)
- [ ] [Media Management](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Media-Management/)
- [ ] [Tips and Tricks](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Tips-and-Tricks/)

Think some of the sections are a bit too large? Let me know, and let's see if we can divide them into even smaller bits 😄 

We might also want to consider adding more sections, related to the new features in Umbraco 8:
* Language variants
* Infite editing
* Nested Content
* More?

OwainWilliams:
Hey @sofietoft 
I'm happy to start on the Getting Started with Umbraco section and if anyone else wants to help with subsections in there then that would be great.  :)

OwainWilliams:
@sofietoft  - How do I create a U8 version of the docs? Do I just need change the markdown?
to this?


---
versionFrom: 8.0.0
needsV8Update: "false"
---

sofietoft:
Great question @sofietoft !

We have some docs on it here: https://our.umbraco.com/documentation/Contribute/#multi-version-documentation
Basically, what you'll need to do:
- Copy the current file
- Rename the current file to fx index-v7.md (replace "index" with the name of the file!) and remove the `needsV8Update` label
- The new copy should take the name of the original file, and have only this YAML tag: `versionFrom: 8.0.0`

[...]

PR #2416, opened 2020-04-03
Updated comapring versions

erikjanwestendorp:
Updated comapring versions for v8


Jeavon:
@erikjanwestendorp the new images don't seem to be working, I think maybe GitHub is case sensitive?

erikjanwestendorp:
@Jeavon You're right! I changed the file name to lower case but still an issue with it. It is probably my stupid mistake but it seems GitHub looks in to the master branch but the images are not there yet of course, see image below. Any suggestions? 

image

sofietoft:
Looks like the image is working 😄 

Let's get this merged, and see what happens.
Thanks for updating the article @erikjanwestendorp ! 😁

PR #2420, opened 2020-04-08
Reflect public class requirement for overriding Default Umbraco Forms Providers.

robmaas:
When overriding a Umbraco Forms Default Provider it is required that the custom class is public, otherwise the Reflection code won't pick up the overriden class.

For example:
`public class CustomOverridenSendEmail : Umbraco.Forms.Core.Providers.WorkflowTypes.SendEmail`

works, but a internal class does not:

`internal class SendEmailNoFrom : Umbraco.Forms.Core.Providers.WorkflowTypes.SendEmail`

The current documentation states that there is only a requirement that the WorkflowType Id should be the same, which is incorrect.

### DIFF
--- a/Add-ons/UmbracoForms/Developer/Extending/Adding-a-Type-v7.md
+++ b/Add-ons/UmbracoForms/Developer/Extending/Adding-a-Type-v7.md
@@ -95,7 +95,7 @@ needed. Also look in the reference chapter for complete class implementations of
 
 ## Overriding default providers in Umbraco Forms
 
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks for testing, verifying and updating this @robmaas ! 😄 
Much appreciated.

I'll make sure it's reviewed as soon as possible 💪

sofietoft:
I've had one of our Forms experts give this a quick review, and it looks like we're good to go 😄 

I'll get this merged right away. Thanks again for putting in the effort to get this updated 💪

PR #2428, opened 2020-04-11
Feature docs/back office search

bielu:
I noticed one of the features is missing in docs, here is short docs about how to use that.
This is draft PR as I am sure it will need more work on that.

### DIFF
--- a/Extending/Backoffice-Search/index.md
+++ b/Extending/Backoffice-Search/index.md
@@ -1,0 +1,198 @@
+---
+keywords: Backoffice tours
+versionFrom: 8.6.0
+meta.Title: "Backoffice Tours"
+meta.Description: "A guide configuring backoffice tours in Umbraco"
+---
+
+# Backoffice tours
+
+Backoffice tours are a way to create helpful guides for how to work in the Umbraco backoffice.
+
+They are managed in a JSON format and stored in files on disk. The filenames should end with the `.json` extension.
+
+## Tour file locations
+
+The tour functionality will load information from multiple locations.
+
+### Core tours and custom tours
+
+The tour files that ship with Umbraco are stored in `/Config/BackofficeTours`. This is also the recommended location for storing your own tour files.
+
+### Plugin tours
+
+When you want to include a tour with your custom plugin/package you can store the tour file in `/App_Plugins/<YourPlugin>/backoffice/tours`. It is recommended that you place the tour files in this location when you are [creating a package](../Packages/Creating-a-Package).
+
+## The JSON format
+
+A tour file contains an array of tour configuration JSON objects. It's possible to have multiple, (un)related tours in one file.
+
+```json
+[
+    {
+        // tour configuration object
+    },
+    {
+        // tour configuration object
+    }
+]
+```
+
+## The tour configuration object
+
+A tour configuration JSON object contains all the data related to a tour.
+
+Example tour configuration object:
+
+```json
+{
+    "name": "My Awesome tour",
+    "alias": "myUniqueAlias",
+    "group": "Get things done!!!",
+    "groupOrder": 1,
+    "allowDisable": true,
+    "culture" : "en-US",
+    "contentType" : "homePage",
+    "requiredSections": ["content","media"],
+    "steps": []
+}
+```
+
+Below is an explanation of each of the properties on the tour configuration object:
+
+### name
+
+This is the name that is displayed in the help drawer for the tour.
+
+![Tour name highlighted](images/tourname-v8.png)
+
+### alias
+
+The unique alias of your tour. This is used to track the progress a user has made while taking a tour. The progress information is stored in the TourData column of the UmbracoUsers table in the database.
+
+### group
+
+The group property is used to group related tours in the help drawer under a common subject (e.g. Getting started).
+
+![Tour group highlighted](images/tourgroup-v8.png)
+
+### groupOrder
+
+This is used to control the order of the groups in the help drawer. This must be an integer value.
+
+### allowDisable
+
+A boolean value that indicates if the "Don't show this tour again" should be shown on the tour steps. If the user clicks this link the tour will no longer be shown in the help drawer.
+
+![Tour allow disable link highlighted](images/tourallowdisable-v8.png)
+
+### culture
+
+You can set a culture (e.g. nl-NL) and this tour will only be shown to users that have set this culture in their profile. If omitted or left empty the tour will be shown to all users.
+
+### contentType
+
+Use this property if you want to limit the tour to a specific content type. E.g. if you want to make a tour that's specific to content nodes using the Home Page Document Type, use the alias of the Document Type, `homePage`, to set the `contentType` property.
+
+The `contentType` property can also be used to limit the tours to content types that are using a specific composition. This will show the tour on all nodes that are using a specific composition.
+
+![Content Type specific tours](images/contentTypespecific.png)
+
+In the image above, two tours are avaibable on the *Welcome* node:
+
+1. "Setup the Welcome page" is available because the tour is limited to the `homePage` content type and
+2. "Setup the SEO" is available because the content type uses the `SEO` composition, which is associated with a specific tour.
+
+When the `contentType` property is set, the tour will **not** show as part of any groups.
+
+### requiredSections
+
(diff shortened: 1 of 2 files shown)


Extending/Backoffice-Search/index.md
```diff
@@ -0,0 +1,84 @@
...
+Backoffice search is usefull tool, which support you in finding your content. Umbraco by default will allow you to search using only few fields: 
+
+| Node Type    | Propagated Fields    |
+| ------------ | -------------------- |
+| All Nodes    | UmbracoFileFieldName |
+| Media Nodes  | File path            |
+| Member Nodes | email, loginName     |
+|              |                      |
+
+Umbraco 8.6 introduce new easy way for extend/override that fields, which is called IUmbracoTreeSearcherFields. This service exposed hardcoded internal list of fields and allow for modification on them.
```
marcemarc, line 19:
```suggestion

However, a specific Umbraco implementation may have additional custom properties that it would be useful to be considered in a Backoffice Search, for example perhaps there is an 'Organisation Name' property on the Member Type, or the 'Main Body Text' property of a Content item. 

Umbraco 8.6 introduced a new easy way to extend/override the default fields by implementing `IUmbracoTreeSearcherFields`. This service exposes the list of default internal search fields for each section and allows modification of them.
```

Extending/Backoffice-Search/index.md
```diff
@@ -0,0 +1,84 @@
...
+#### All Node types
+
+```
+public class CustomInternalSearchConstants : InternalSearchConstants,     IInternalSearchConstants
+{
+    public List<string> GetBackOfficeFields()
+
+    {
+        var list = base.GetBackOfficeFields();
+        list.Add("parentID");
```
marcemarc, line 32: I think with the examples we should provide more real worldy things, such as in the suggested text above, eg searching Members by an Organisation or Company Name property - searching content by BodyText alias?

Extending/Backoffice-Search/index.md
```diff
@@ -0,0 +1,84 @@
...
+|              |                      |
+
+Umbraco 8.6 introduce new easy way for extend/override that fields, which is called IUmbracoTreeSearcherFields. This service exposed hardcoded internal list of fields and allow for modification on them.
+
+### Adding custom properties to backoffice search
+
+#### All Node types
+
+```
+public class CustomInternalSearchConstants : InternalSearchConstants,     IInternalSearchConstants
```
marcemarc, line 26: InternalSearchConstants and IInternalSearchConstants, have been renamed to be UmbracoTreeSearcherFields and IUmbracoTreeSearcherFields

Extending/index.md
```diff
@@ -40,6 +41,10 @@ Backoffice tours is the integrated and interactive help for the backoffice.
 
 A library of API Reference documentation.
 
+## [Backoffice Search](Backoffice-Search/)
+
+Backoffice search is integrated search service which allow find nodes.
```
marcemarc, line 46:
```suggestion
Extending the Umbraco Backoffice search, customising the fields searched - replacing the searching mechanism.
```

Extending/Backoffice-Search/index.md
```diff
@@ -0,0 +1,84 @@
...
+        var list = base.GetBackOfficeMembersFields();
+        list.Add("parentID");
+        return list;
+    }
+}
+```
+
+## More advanced extensions
+
+To make more customization on back-office search check [ISearchableTree](https://our.umbraco.com/Documentation/Extending/Section-Trees/Searchable-Trees/ "https://our.umbraco.com/Documentation/Extending/Section-Trees/Searchable-Trees/")
```
marcemarc, line 82:
```suggestion
For further extensibility of the Umbraco Backoffice search implementation check [ISearchableTree](https://our.umbraco.com/Documentation/Extending/Section-Trees/Searchable-Trees/ "https://our.umbraco.com/Documentation/Extending/Section-Trees/Searchable-Trees/")
```

Extending/Backoffice-Search/index.md
bielu: true, forgot to update my example code, will upgrade PR soon

Extending/Backoffice-Search/index.md
bielu: @marcemarc here I was using fields which are in all solution and are not related to any on specific starter kits/demo sites

Extending/Backoffice-Search/index.md
```diff
@@ -0,0 +1,84 @@
...
+meta.Description: "A guide to customization of Backoffice Search"
+---
+
+# Backoffice Search
+
+Backoffice search is usefull tool, which support you in finding your content. Umbraco by default will allow you to search using only few fields: 
+
+| Node Type    | Propagated Fields    |
+| ------------ | -------------------- |
+| All Nodes    | UmbracoFileFieldName |
```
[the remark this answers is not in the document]
bielu: Yeah it would, I think when I was editing table I screw up that, will fix that soon

Extending/Backoffice-Search/index.md
marcemarc: ahh I see, to ensure the example works for everyone - I was thinking it would be cool to show the possibilities though - perhaps with a comment above explaining that the property being added was a custom one, whereas I couldn't think why you would want to search by ParentId?, so would it confuse people as to what the purpose of the example was for... maybe creatorName or writerName, or __NodeTypeAlias would be useful? but difficult, because you want it to be useful text that might contain keywords to be searched... so maybe showing custom properties being used isn't so bad...

bielu: @marcemarc when I was implementing that tool my client request possibility of searching by parentID :)

marcemarc: hehe

marcemarc: Think 'we could' have one other different example to parentid? - I don't mind adding it in after we merge... it would be good to show the difference between GetBackOfficeFields and the other specific three Getters, in that if you wanted to add ParentId to all three, you'd only need to add it to the GetBackOfficeFields one...?

Also I think we should add an example of registering the new CustomUmbracoTreeSearcherFields in a composer, so people can see they need to do that step too.

bielu: @marcemarc 
 >>if you wanted to add ParentId to all three, you'd only need to add it to the GetBackOfficeFields one...?
Yes you need only add to GetBackOfficeFields
about registering composer, hmm I have mixed feeling, but yeah probably a good idea as someone could be not familiar with Umbraco

marcemarc: Yep, I think I feel the same thing, eg not repeating everything everywhere, but I don't think people will necessarily know to use RegisterUnique

bielu: Probably we should get input from doc team @sofietoft, can you help us here or point to a person which should be involved:)?

marcemarc: @bielu we are in the Documentation meeting now, so I am discussing with the rest of the team!

bielu: @marcemarc cool :)

sofietoft:
Hi @bielu !
Thanks for putting in the work to make this PR 😄 
Much appreciated.

We, the docs team, have just had a meeting, where one of the things we discussed was this PR.
We talked about having different examples of the field to add into the search, instead of using the same field in all examples. Perhaps instead, we could add a field more specific to each type? 🤔  
So, for the Member type, add something specific to that type, and the same for Document Type and Media Type.
What do you think about that?

We also talked about the benefits of adding an example of how to register this in a composer - I think it would be good to have the example. Not much information would be needed though, as I think one could simply link to where this is explained in more details 🤔 
That can of course always be added afterwards - if you do not feel it's something you're up to.

Finally, what do you think about adding a short explanation to the examples for each of the types? 
Like for the "All Node types" section add something like "Register something like this if you want to add a field to all searches on the site, including Documents, Media and Members..".

bielu:
@sofietoft Sure will add some more description when I will have little more time :)

sofietoft:
Hi @bielu ,
It's been a while since we heard back from you, so we've gone ahead and merged the PR.
Thanks for putting in the work to make this article 💪 Much appreciated.

If you're up for it, feel free to create another PR with the changes we've discussed!

bielu:
Hi @sofietoft, 
I will when I will be less busy with other staff :)

PR #2429, opened 2020-04-11
Meta data updates for SEO

crgrieve:
During Hackathon we realised there were some parts of documentation that are hard to find when googling. We thought additional SEO meta could help.

Thanks,
Carole


sofietoft:
This is awesome @crgrieve ! 😄 
Thanks a bunch for adding these meta tags - much appreciated.

Let's get it merged 💪

PR #2438, opened 2020-04-15
Add link to Latch from config transforms

Jeavon:


sofietoft:
Great idea, getting a link to Latch in this article. Make a whole lot of sense 💪 

Thanks for adding it @Jeavon !
I'll merge it right away! 😄

PR #2441, opened 2020-04-17
Added SEO Metadata to Contribute section.

poornimanayar:
Hi,

I have added SEO Metadata to contribute section. I have removed the `versionFrom` on some docs too in the same section. Let me know if its okay :-)

Poornima


sofietoft:
Looks great @poornimanayar ! 😄 

I think removing the `versionFrom` tags is a good idea - how to contribute should be the same no matter which version of Umbraco it's related to! 💪

PR #2447, opened 2020-04-19
SEO Metadata added

poornimanayar:
Hi,

More SEO Metadata added :-)

Poornima


sofietoft:
Looks great @poornimanayar ! 😄 

Thanks for adding the tags 💪

PR #2450, opened 2020-04-20
SEO Info on Property Value Converters

poornimanayar:
I think I missed out on this with my last PR #2447 

Poornima


sofietoft:
Ah, looks like it!

Thanks for adding it @poornimanayar !
Much appreciated 😁

PR #2453, opened 2020-04-21
Update the price for Additional domains

Aaen:


Aaen:
You can find the price here

sofietoft:
Good catch @Aaen ! 😄 

PR #2460, opened 2020-04-22
Update Index.md

koenvanras:
There was a small mistake in the code snippet.
In the snippet the code creates a media type Image, in stead of a Folder.


sofietoft:
Ah, good catch @koenvanras ! 😄 
Thanks for getting this updated.

I'll merge it right away.

PR #2461, opened 2020-04-24
Change Relate to original warning

KevinJump:
Update Warning box: as this is not how relate to original works :( - it creates a relation in a table, so the two pages can be linked (e.g languages) but no coping goes on between them, after the first time.


sofietoft:
Ah, good catch @KevinJump ! Thanks for getting this corrected 💪 

Will get it merged right away.

PR #2466, opened 2020-04-28
Mention resource for picking a license

jmayntzhusen:
We set the default license when creating a package, but would be good to have the author really consider which license to use and not just go with MIT without thinking about it.

### DIFF
--- a/Extending/Packages/Creating-a-Package/index.md
+++ b/Extending/Packages/Creating-a-Package/index.md
@@ -35,7 +35,7 @@ This section contains mostly meta data about the package and the creator. We wil
 | Author | Your name | Here you get to take credit for your awesome work! |
 | Author URL | Your website or maybe Twitter tag | Will link to this from the author name certain places in the backoffice |
 | Contributors | _Blank_ | Here you can add the names of other contributors if you have any |
-| License | MIT License | Will be set to MIT by default. If you are planning on selling your package you can change the license here |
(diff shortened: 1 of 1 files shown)


Extending/Packages/Creating-a-Package/index.md
```diff
@@ -35,7 +35,7 @@ This section contains mostly meta data about the package and the creator. We wil
 | Author | Your name | Here you get to take credit for your awesome work! |
 | Author URL | Your website or maybe Twitter tag | Will link to this from the author name certain places in the backoffice |
 | Contributors | _Blank_ | Here you can add the names of other contributors if you have any |
-| License | MIT License | Will be set to MIT by default. If you are planning on selling your package you can change the license here |
+| License | MIT License | Will be set to MIT by default. Please consider how you want your package licensed, if in doubt when deciding an open source license there are [good resources available](https://choosealicense.com/licenses/) |
```
sofietoft, line 38:
```suggestion
| License | MIT License | Will be set to MIT by default. Please consider how you want your package licensed. If in doubt when deciding an open-source license there are [good resources available](https://choosealicense.com/licenses/). |
```

jmayntzhusen: Thought I could finally sneak in a long sentence 😁

sofietoft: Never! 😎 😄

PR #2468, opened 2020-04-29
Updates point in time restore retention

RobertCopilau:
We have recently migrated all customer databases to keep 35 days of point in time to restore backups, updating docs to reflect that.


RobertCopilau:
My pleasure :D

PR #2471, opened 2020-05-01
Load balancing updates

Shazwazza:
* Describes what MainDom is and why it's needed and adds version specific information
* Changes headline `File system replication` to `Server configuration` and moves this section up because it's actually very important and updates this section with explicit links to each type of server design, this makes it much easier to find azure specific docs... these links were extremely hard to find.

We are still missing any docs/links to 'shared' file systems such as NAS/SAN. We don't mention this at all in the new v8 docs since everything now just points to this new `file-system-replication` document. I've linked the 'shared' file system to the file-system-replication#standalone-file-system link because that seems like its correct ... but then reading that it doesn't seem correct since i'm unsure why blob storage would be needed in that case. We basically have deleted this whole section https://our.umbraco.com/Documentation/Getting-Started/Setup/Server-Setup/Load-Balancing/files-shared-v7 for v8. I've left a question for @Jeavon in another PR about that.


Shazwazza:
I'm merging this in because it's vitally important that we make it clear to people how to set this up 'now' until a fix is out for the SqlMainDomLock setting for load balancing.

PR #2476, opened 2020-05-03
Cross link to Route Hijacking information

marcemarc:
I've noticed a pattern where people want to use a custom MVC controller, but follow the scent of 'custom routes', and end up on this page... which then seems to suggest you need to map a route to a custom MVC controller, which if you are coming from MVC background doesn't seem unusual, and you are not going to guess or search for the existence of route hijacking, so I've just added a note underneath mapping a route using a virtual node route handler to point in the direction of the by convention approach... so people can discover it is a thing.

### DIFF
--- a/Implementation/Custom-Routing/index.md
+++ b/Implementation/Custom-Routing/index.md
@@ -100,7 +100,7 @@ namespace Umbraco8.Components
 See: [Custom routing documentation](../../Reference/Routing/custom-routes)
 
 :::note
-This is an approach for mapping a custom route to a custom MVC controller, for existing content page routes you can use a custom MVC controller to handle the request by naming convention: see [Custom Controllers - Route Hijacking](../../Reference/Routing/custom-controllers). 
+This is an approach for mapping a custom route to a custom MVC controller. For creating routes for existing content pages you can use a custom MVC controller to handle the request by naming convention: see [Custom Controllers - Route Hijacking](../../Reference/Routing/custom-controllers). 
 :::
 
 ### PublishedRequest.Prepared event
--- a/Implementation/Custom-Routing/index.md
+++ b/Implementation/Custom-Routing/index.md
@@ -99,6 +99,10 @@ namespace Umbraco8.Components
 
 See: [Custom routing documentation](../../Reference/Routing/custom-routes)
 
+:::note
+This is an approach for mapping a custom route to a custom MVC controller, for existing content page routes you can use a custom MVC controller to handle the request by naming convention: see [Custom Controllers - Route Hijacking](../../Reference/Routing/custom-controllers). 
+:::
+
 ### PublishedRequest.Prepared event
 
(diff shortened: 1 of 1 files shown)


Implementation/Custom-Routing/index.md
```diff
@@ -99,6 +99,10 @@ namespace Umbraco8.Components
 
 See: [Custom routing documentation](../../Reference/Routing/custom-routes)
 
+:::note
+This is an approach for mapping a custom route to a custom MVC controller, for existing content page routes you can use a custom MVC controller to handle the request by naming convention: see [Custom Controllers - Route Hijacking](../../Reference/Routing/custom-controllers).
```
sofietoft, line 103:
```suggestion
This is an approach for mapping a custom route to a custom MVC controller. For creating routes for existing content pages you can use a custom MVC controller to handle the request by naming convention: see [Custom Controllers - Route Hijacking](../../Reference/Routing/custom-controllers). 
```

marcemarc:
> Added a small suggestion to update the wording a bit - and make shorter sentences 😉
> 
> Also, this note you've replaced, that referring to the section above it, right?
> Wouldn't the information in the note, be worth knowing BEFORE reading that section? 🤔 Just a thought - let me know what you think!

I did think about this, it's largely about the journey that people are taking to get here... this journey is serving people who are searching for Custom Routing, - Route Hijacking isn't really changing the Route, it's more about handling the Route, so I felt it was more of an aside, in case having read the custom routing information you thought this was the only way to 'handle an existing route' in a custom way... but certainly in 'complexity' terms route hijacking is simpler and would normally be presented before mapping a custom route... it's just they are different things in this context ?

sofietoft:
Yeah, alright 🤔 I see why you would put it below..

I still think though, that we should put it at the top. Just to prevent people from wasting their time on something, that in fact is much more complex than what they really want to do... if that makes sense? 🤔

marcemarc:
Yes, I understand what you are saying, but perhaps we'd need to word it differently... with these kinds of 'subject introducing' docs, we want to service the quick lookup, for returning users to the docs, and then expand further for the 'first read' - so I was thinking having a note about a different (but related) subject up front, might get in the way of people who are just looking up how to do custom routing... it's always a trade off... so just saw it as a bit of an aside to try and catch people's attention, as it's an Umbraco thing rather than an MVC thing.

...The problem is the heading of the page is already Custom Routing, hijacking routes isn't a way to do custom routing, and also Custom Mvc Routes isn't the only way to do custom routing so it's hard to justify it having too much focus...

... but

Custom MVC routes
In addition to handling routes with custom MVC Controllers (see route hijacking), you can also specify your own custom MVC routes to map to an custom MVC Controller within the Umbraco pipeline. 

??? does that sneak it in ?

sofietoft:
Right! So it actually is more two different things 🤔 

Let's just leave it in the bottom then 😅 
If people don't find the guide useful or don't think it looks like it's what they need - then they will scroll past it anyways, and then find the note!

PR #2480, opened 2020-05-05
Re-added information about URL Tracking disabling if some packages detected.

NikRimington:
The change made in this [commit](https://github.com/umbraco/UmbracoDocs/commit/f745095f476fdc266454483b676225dcc068cdc6) removed important information that applies to v7 sites without creating multi-versions of the page. Adding this back in with a note that this doesn't apply to v8 sites.


sofietoft:
Looks great @NikRimington ! 😄 
Super with the note as well.

Let's get this merged back into the docs!

PR #2481, opened 2020-05-06
Fallback fix

craigs100:
Fixes issue with Fallback.ToAncestors() not working when dropdown not saved (=null) or saved at 1st option (="").

### DIFF
--- a/Tutorials/Creating-an-XML-Site-Map/index.md
+++ b/Tutorials/Creating-an-XML-Site-Map/index.md
@@ -118,7 +118,14 @@ This is a great candidate for a [Razor Helper](https://weblogs.asp.net/scottgu/a
 ```csharp
 @helper RenderSiteMapUrlEntry(IPublishedContent node)
 {
-    var changeFreq = node.HasValue("searchEngineChangeFrequency", fallback: Fallback.ToAncestors) ? node.Value("searchEngineChangeFrequency", fallback: Fallback.ToAncestors) : "monthly";
+	string changeFreq = string.Empty;
+	if(String.IsNullOrEmpty(node.Value<string>("searchEngineChangeFrequency", fallback: Fallback.ToAncestors))) {
+        changeFreq = node.Parent.Value<string>("searchEngineChangeFrequency", fallback: Fallback.ToAncestors);
+    } else {
+        changeFreq = node.Value<string>("searchEngineChangeFrequency", fallback: Fallback.ToAncestors);
+    }
+    changeFreq = String.IsNullOrEmpty(changeFreq) ? "monthly" : changeFreq;
+    
     // with the relative priority, this is a per page setting only, so we're not using recursion, so we won't set Fallback.ToAncestors here and we'll default to 0.5 if no value is set
     var priority = node.HasValue("searchEngineRelativePriority") ? node.Value<string>("searchEngineRelativePriority") : "0.5";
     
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for putting in the work to update thie @craigs100 ! 😄 

marcemarc:
@craigs100 thanks for raising this!, I think the tutorial was originally written for V7, and then has been over time updated several times for V8... firstly in V8 there is a better syntax for checking for an ancestor value and then falling back to a default value... the fallback: option enables you to 'chain' fullbacks... so this would be the 'V8' recommended way:

` var changeFreq = node.Value("searchEngineChangeFrequency", fallback: Fallback.To(Fallback.Ancestors, Fallback.DefaultValue),defaultValue:"monthly");`

However are you saying that if the dropdown is null or an empty string (the first option) you are finding that it doesn't 'fallback' to the value above automatically? 

in V7 if reading a property and having 'recurse' set to true, then it was the fact that the property was null or empty that made it go look for an ancestor value... so are you finding this is not the same in V8?

craigs100:
@marcemarc Thanks for the new code but it doesn't solve the issue. This could well be more to do with how dropdowns are implemented than fallback to be honest. With a dropdown, you always get the 1st option injected by Umbraco, uninvited. If nothing is ever selected and the page saved it's saves null for the dropdown value and Fallback.ToAncestors will work. If you select something and save the page, then select the 1st option again, it will save an empty string. If an empty string is there then that's a value and Fallback.ToAncestors won't run up the tree and find a dropdown that has a meaningful value and your sitemap will show `<changefreq/>` instead of `<changefreq>monthly</changefreq>`. So that's why I had to check for String.IsNullOrEmpty(...).

marcemarc:
Thanks @craigs100 ... I wonder if that's changed in V8, as the dropdown if 'untouched' always used to be empty... will give it a spin, and have a think - if that's the case then might need to think about a different way of managing/explaining this in the tutorial

marcemarc:
@craigs100 This appears to be a bug with the way the dropdown property editor is implemented, it stores '[]' or '[null]' in the database if you set the dropdown not to have a selected item... rather than 'null' ... this breaks IsValue and HasValue which is what Fallback.ToAncestors relies on to detect if a property has a value before falling back to the parent then parent's parent...

... I've create an issue and a PR here: https://github.com/umbraco/Umbraco-CMS/pull/8154

if we get a quick response, then we can update accordingly, but if not we should put a warning in the tutorial about it's broken state and your possible workaround!

thanks for reporting the issue!

regards

marc

sofietoft:
Hi guys!
I can see that the referenced PR has been merged.
Doesn't look like it's been released yet, though. 

In any case, does the merging of that CMS PR mean that this PR should be merged or closed? 🤔

marcemarc:
@sofietoft it means several things... :-(
1) there should be an update to the tutorial for 'how to get the' the changefreq to follow/introduce the new V8 way:
`var changeFreq = node.Value("searchEngineChangeFrequency", fallback: Fallback.To(Fallback.Ancestors, Fallback.DefaultValue),defaultValue:"monthly");`
2) there should be some screenshot updates to reflect the previous name changes - I have taken screenshots... on my machine somewhere
3) there should be a warning about the fallback for a dropdown not working correctly in sites < v8.7
- the workaround in this PR I think may still not function correctly in some circumstances eg: if it's the parent item that has also had a setting set and then reverted back to be blank, it will also not 'fall through' :-(

Not sure I've made that any clearer!

regards

Marc

sofietoft:
Hi @craigs100 !

Apologies for the long wait on this one. I have an update for you.
@marcemarc just updated the article with a few changes and a mention about the fallback issue with the dropdown (which seems to be specific to an older version).

Would you be up for giving [the updated article](https://our.umbraco.com/documentation/Tutorials/Creating-an-XML-Site-Map/) a spin? 😄

craigs100:
Hi @sofietoft 
Yes will give a spin and report back.

craigs100:
Hi @sofietoft 

I've given it a spin and find that I still get empty `<changefreq/>` nodes so I'll be keeping my code as I had it. But the new warning about the issue below 8.7 is clear enough I think.

marcemarc:
Yes, the updated version won't fix the bug for old versions - but the tutorial should work in 8.7 where the fix will be included.

sofietoft:
Thanks for testing this again @craigs100 ! Much appreciated 💪 

As Marc said, the guide doesn't fix the bug unfortunately!
Hopefully Umbraco 8.7 is released soon, so you can avoid having to do the workaround in the future.

I'll consider this one closed for now.
Thanks again for reporting 💪

PR #2489, opened 2020-05-17
SEO tags added to packages section

poornimanayar:
Added SEO metadata to packages documentation


sofietoft:
This is great @poornimanayar ! 😄 

Thanks for helping us optimizing SEO for the docs 💪

PR #2491, opened 2020-05-17
Added some more information about registering sections

abjerner:
Per a discussion on Slack, I noticed the documentation doesn't really describe how to control the order of the sections. So here is some more information about that 😉

### DIFF
--- a/Extending/Section-Trees/sections.md
+++ b/Extending/Section-Trees/sections.md
@@ -8,7 +8,7 @@ meta.Description: "A guide to creating a custom section in Umbraco"
 
 The Umbraco backoffice consists of Sections, also referred to as Applications, which contain Trees.
 
-Each section is identified by an icon in the top navigation ribbon of the Umbraco Backoffice.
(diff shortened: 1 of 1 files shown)


Extending/Section-Trees/sections.md
```diff
@@ -36,8 +36,10 @@ Create a new file in the `/App_Plugins/MyFavouriteThings/` folder and name it `p
 
 ... would create a new section in your Umbraco backoffice called 'My Favourite Things'.
 
+Registering your section this way, it is added to the end of the collection of sections. But as more `package.manifest` files may do the same, the order of the additional sections depends on the order of which the `package.manifest` files are loaded. Registering your section this way doesn't allow further control of it's order.
```
sofietoft, line 39:
```suggestion
When registering your section this way, it is added to the end of the collection of sections. But as more `package.manifest` files may do the same, the order of the additional sections depends on the order of which the `package.manifest` files are loaded. Registering your section this way doesn't allow further control of its order. 
```

abjerner: Makes sense 👍

PR #2518, opened 2020-06-06
Update install command to include version number

louisjacksonrees:
While in pre-release this should include the version number specification. Issue raised here: https://github.com/umbraco/UmbPack/issues/23


jmayntzhusen:
Thanks @louisjacksonrees ! 🎉🙌

Issue #2350, opened 2020-03-08
Add [RuntimeLevel(MinLevel = RuntimeLevel.Run)] to the composer


daanstolp:
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/master/Extending/Database/index.md***

The page describes a method for creating custom tables in the Umbraco database. However, when you are creating a new site, it is important to add the `[RuntimeLevel(MinLevel = RuntimeLevel.Run)]` to the composer. 

When you want to deploy a new website to the server, you need to clear the connection string from the Web.config file, in order to trigger the Umbraco installation. However, without the `RuntimeLevel` attribute, the composer will try to run even before the Umbraco installation. Since no connection string has been configured yet, the code in the composer that tries to execute the migration will fail. Adding this attribute prevents the composer from running too soon.

sofietoft:
Thanks for the report @daanstolp ! 😄 

So, I don't know too much about how to work with this particular scenario - what you're saying here is, that in order for the sample in the article to work, the `RunetimeLevel` attribute needs to be added - is that correct?
Would you happen to have an example of how it's worked for you? 😁

daanstolp:
The sample in the article works just fine, _when you add the component to an already installed site_. But it won't work when you are setting up a new site (that already includes this component) and still need to through the Umbraco installation.

I ran into this issue like this:
1. I created a new site using Visual Studio, installed it on localhost (just using Visual Studio). 
1. At some point, I added the component. (Note that at this time, the local database had already been set up and initialized).
1. Install the site to a server by copying the files, and clearing the connection string and `ConfigurationStatus` in `Web.config`. _This forces Umbraco to go through the installation process, so the database is properly initalized_. 

Without the attribute, this last step now fails. The component tries to run a Migration on the database, but since I cleared the connection string in `Web.config`, it cannot connect to the database and throws an Exception. And I cannot add the connection string, because I need to trigger the installation wizard. 

Adding the attribute makes sure that the component _won't_ try to execute until _after_ Umbraco has been completely installed and initialized.  

```csharp
[RuntimeLevel(MinLevel = RuntimeLevel.Run)]
public class BlogCommentsComposer : ComponentComposer<BlogCommentsComponent>
{
}
```

nul800sebastiaan:
Ah, interesting.. according to the docs at https://our.umbraco.com/Documentation/Implementation/Composing/#runtimelevel : 

[...]

PR #2521, opened 2020-06-07
Add RuntimeLevel attribute to the Composer snippet

daanstolp:
When you want to deploy a new website to the server, you need to clear the connection string from the Web.config file, in order to trigger the Umbraco installation. However, without the RuntimeLevel attribute, the composer will try to run even before the Umbraco installation. Since no connection string has been configured yet, the code in the composer that tries to execute the migration will fail. Adding this attribute prevents the composer from running too soon.

See also the discussion at https://github.com/umbraco/UmbracoDocs/issues/2350

### DIFF
--- a/Extending/Database/index.md
+++ b/Extending/Database/index.md
@@ -22,6 +22,7 @@ using Umbraco.Core.Persistence.DatabaseAnnotations;
 
 namespace Umbraco.Web.UI
 {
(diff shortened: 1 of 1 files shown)


sofietoft:
Looks great @daanstolp ! 😄 

Thanks for adding the extra addition to this article.
Much appreciated 💪

Issue #2045, opened 2019-10-25
Add Angular directives, services etc. to the documentation


abjerner:
When extending the Umbraco backoffice, we can consult the [Umbraco Backoffice UI API documentation](https://our.umbraco.com/apidocs/v8/ui/#/api) to see which Angular directives, services and so on we can use in our custom code. But the current API documentation is far from great, as doesn't show some of the options that actually are supported, and it lacks useful examples.

On way to solve this is of course to improve the existing API documentation. But instead (or in addition to), the Angular side of the backoffice could be documented as part of the main documentation - a bit similar to how the [C# service](https://our.umbraco.com/documentation/Reference/Management/Services/) are documented in the main documentation while also being part of the C# API Documentation.

I'm a bit unsure where to put something like this in the folder structure, but content wise I was thinking about something like this:

https://gist.github.com/abjerner/61a54258de9e7dd1f15ef9a1b24425e2

I know this requires some extra work, but I'd be happy to give a hand writing this part of the documentation 😉

nul800sebastiaan:
I am not sure what you're suggesting here @abjerner - just duplicating the existing generated documentation into the docs?

abjerner:
The basis of this issue is that the current API reference feels a bit limited. All options and parameters (eg. for the editor service) should be documented (which they are currently not). But ideally there should be better examples on how to use the different service and directives. Perhaps even screenshots of how a given directive would look.

My question was partly whether something like this is best done by updating the existing API reference - or introducing a new part of the main documentation that focuses on describing the Angular parts in greater detail.

With the C# services that I mentioned, they are part of the C# API reference, but also described in the main documentation. It's not exactly duplicate content, but more something that is complimenting each other.

I hope that makes sense 😉

sofietoft:
Hi @abjerner ,
I've been trying to wrap my head around this, and also had a chat with my fellow Curators about it at our meeting yesterday.
As we understand, this is two-fold:

1. There should be more details on each service in the source code - therefor also the API docs. These details should include a description of the parameters (right?)
2. There should be examples on each serivce - preferably in the Documentation on Our

Would you say, we got that all right? 😅

abjerner:
Hi @sofietoft,

For **1)**, yes. All parameters should be documented through the ngdocs.

**2)** is what I'm asking about, as I'm not sure myself.

We could either have everything in the API documentation. But having extended explanation and examples might clutter the API documentation, so it may be more ideal to have that part in the main documentation.

[...]

PR #2522, opened 2020-06-07
Added documentation for the umbLoadIndicator directive

abjerner:
Following the discussion from: https://github.com/umbraco/UmbracoDocs/issues/2045


sofietoft:
This looks great @abjerner !
I'll make sure we have it reviewed as soon as possible 💪 

In the meantime, would you be up for adding some meta data? 😇 
Would be really great if you could add the version YAML tags (https://our.umbraco.com/documentation/Contribute/Adding-Metadata/).

Also looks like a there's a lint error 🤔 All list items should start with a capital letter it looks like!

abjerner:
@sofietoft I weren't entirely sure which meta data to add. This directive has been around forever. Or 7.4 I think - I guess I could add that 😉

Should `meta.Title` always be used, or is it only when it differs from the H1 element of the page?

The linter error seems to be because `umbLoadIndicator` is lower camel casing. Since that is how it's declared in Angular, maybe it can be considered a false positive in terms of the linter? Or would it be better to refer to it as "load indicator" instead of "umbLoadIndicator"?

sofietoft:
Apologies for the delay here 🙈 Busy week.

For the `versionFrom` metatag, I think 7.4.0 is completely fine.
The `meta` tags are all optional - the most important one is the versioning one 😁 

About that lint error - yeah, I see the challenge there 🤔 
Perhaps it can be solved by saying "The `umbLoadIndicator`" instead? Or, instead of using a list, it could be an h2? Not sure what the "correct" way of doing this would be tbh..

abjerner:
@sofietoft no worries. Both things should be updated now 😉 

As the list of directives should hopefully grow, I think writing "The *" will quickly feel redundant. So I've gone with writing "Load indicator" instead.

image

sofietoft:
Hi @abjerner ! 😄 
Thanks again for putting in the work to do this!

Let's get this merged! 💪

sofietoft:
Ah. Looks like the Angular folder needs an /index.md file, and also the gif on the umbLoadIndicator article needs to be put in an /images folder 😄 

Would you be up for making those changes? 🤞

abjerner:
@sofietoft I tested the pages in my local Our installation, so it ought to work. But I did create the PR via the GitHub website, so maybe I messed something up along the way. I'll have a look 😉 

I wasn't sure about the contents of the **Angular** page, as is currently a single directive and nothing else. But I can see that it gives a 404 error, so I'll have a look at that was well ;)

abjerner:
Hmm. The image actually does work, but the link to the load indicator page is bugged:

Correct URL: https://our.umbraco.com/documentation/Reference/Angular/Directives/umbLoadIndicator/

Current, but incorrect URL: https://our.umbraco.com/documentation/Reference/Angular/Directives/_/umbLoadIndicator/

I'll check what I've messed up 😉

sofietoft:
Ah, interesting 🤔 

Hm! Thanks for looking into this.
In any case, could I get you to place to gif into an /images folder? 😇 
It might not be a necessity for it to work, but it's consistent with the rest of the structure in the docs - and I like consistency 😄

PR #2523, opened 2020-06-08
Clarify app setting location

jmayntzhusen:


sofietoft:
Looks great!

Why not use the note formatting for this, as it is a note?
Just wondering 🤔

jmayntzhusen:
Figured since it is inside a list it would be disruptive to do so, and adding it at the beginning or end could mean people miss it. I can move it / change it if you want 🙂

sofietoft:
Ah, missed that 😄 

Thanks! I'll merge it.

PR #2542, opened 2020-06-12
Correction to Composing Docs

poornimanayar:
Hi, 

I believe the `IRuntimeComposer` has been replaced with Web and Core Initial Composer. So I have made that change. I have also added some information about some of the responsibilities that are undertaken by the composers.

Poornima


sofietoft:
Thanks for the PR @poornimanayar ! 😄 

I'll make sure we get this reviewed as soon as possible 💪

Jeavon:
Thank @poornimanayar It seems this was changed a while back! I think we need a little bit more restructuring here as the InitialComposer's aren't implementing a specific interface. If it's okay I'll add to your PR?

poornimanayar:
Yes @Jeavon , I wasn't totally sure how to approach it either. Happy to help if needed too :-)

PR #2545, opened 2020-06-12
Editor's Manual - Notifications for v8

erikjanwestendorp:
- Added a v8 image

The notification option is only available if the SMTP settings are set correctly. Is this something to make a note of?


sofietoft:
Thanks for adding this @erikjanwestendorp ! 😄 

Good point about the SMTP settings.
I think it would actually be great to add some info about that.
Not sure an editor would have access to change these though 🤔 Would something like "Contact your website administrator..." work, you think?

erikjanwestendorp:
@sofietoft Sounds good! I just added a note :-)

sofietoft:
Looks great @erikjanwestendorp ! 😄 

Thanks for updating the docs - let's get this merged 🎉

PR #2548, opened 2020-06-16
updated working with folders

erikjanwestendorp:
Created a v8 add version
Added v8 images
Deleted needsV8Update


sofietoft:
Thanks for updating the article @erikjanwestendorp ! 😄 

I just realized that the JPG files don't show on Our 🙈 
I believe they should be PNGs instead - would you mind uploading some new images for this?
And if you have an extra minute, perhaps also for the [Audit Trail](https://our.umbraco.com/documentation/Tutorials/Editors-Manual/Tips-and-Tricks/Audit-Trail/) article?

erikjanwestendorp:
@sofietoft I just updated the JPG files to PNG files also for the audit trail (This was already a png file, but I forgot to change the extension in the article 🙈).

sofietoft:
Brilliant - thanks @erikjanwestendorp ! 😄 

Let's get this merged then.
Thanks again for putting in the work to get the articles updated.

erikjanwestendorp:
@sofietoft Great! 

The images are still not appearing. There is a weird thing in the image path:
`` 

which is different from the path in the article: 

``

sofietoft:
Ah, you're right of course!

Completely forgot about this. So, when PNG is written in capital letters, the image will show just fine in the article here on Github. But on Our, I think's case sensitive 🤔 

In any case, I think I've managed to fix it now, be changed it to small letters instead 🙈 
Our is such a mystery sometimes!

PR #2554, opened 2020-06-18
Add example of using the colorpicker without modelsbuilder

jmayntzhusen:
Based on a question on the Community Slack here is an example of how to get a colorpicker label and hex value when not using Modelsbuilder 🙂


sofietoft:
Looks great @jmayntzhusen ! 😄 

Thanks for adding this!

PR #2577, opened 2020-06-30
Update EditorModel Events example to use ContentSavedState

marcemarc:
Update the example on this page to use the ContentSavedState - so people can see how to set a default value ONLY when a new page is created, and not when a particular field is left blank by the editor


marcemarc:
@sofietoft ok merging!

PR #2578, opened 2020-07-01
Add link to where to find dashboard section aliases

LottePitcher:
Took me a while to get my Members dashboard working as I didn't realise that the section alias to use in the package.manifest is "member" not "members".  As this alias is the only one that differs from its English name I thought it would be helpful to add it to the examples, as well as adding a link to where you can double check what these aliases should be.


sofietoft:
That's a great suggestion @LottePitcher !
We'll take it 😎 

Thanks for adding this!

PR #2579, opened 2020-07-01
Remove usage of unnecessary and problematic terms

nul800sebastiaan:


jmayntzhusen:
Maybe an idea to also add them to this Vale rule? 🙂
https://github.com/umbraco/UmbracoDocs/blob/master/.github/valeStyle/Terms.yml

nul800sebastiaan:
@jmayntzhusen Sure! this one was quite easy since there was just no need to use `black` or `white`, but in different contexts it might be harder to do a substitute. `blacklist` could become `disallowed list` and `whitelist` could be `allowed list`. However, in this context that would make less sense, it would turn into something like: 

> This setting consists of a "disallowed list" of file extensions that editors shouldn't be allowed to upload via the backoffice.

I think most other cases (if there are any) would be fine though. If I can get a hand with the syntax I can amend this PR.

sofietoft:
Thanks for the change, and also, good point adding this to our Styleguide!

I agree that it could go into the `Terms.yml` file.
As for the syntax - I think it's simply a matter of writing the words we want to avoid, followed by what we recommend changing it to 🤔 

```
swap:
  blacklist: "'disallowed list'"
  whitelist: "'allowed list'"
```

nul800sebastiaan:
Ah.. sure thing! what about `black list` with a space in it like we had here?

nul800sebastiaan:
I've made some updates.. maybe this is enough for now since those terms with spaces were only used once.

sofietoft:
Looks great - thanks! 💪 

Hm, good point about `black list` with space... I think, let's add it in! Then we're sure a warning will how up in those cases as well, should it be used in future PRs.

nul800sebastiaan:
Just not sure of the notation, would it be:

`black list: "'deny list'"`
or
`"black list": "'deny list'"`

nul800sebastiaan:
Ah.. I see `back office: ` is in the list above, got it, fixed. Make sure to squash this one! 😉

sofietoft:
Superduber! Let's get this one merged then 💪 

Thanks again @nul800sebastiaan !

jmayntzhusen:
Just a note - you can do notation like this:
```black[ -]*list: "'deny list'"```

That will match on `blacklist`, `black list` and `black-list`

PR #2584, opened 2020-07-07
Let v7 LB article link to other v7 article instead of v8.

jmayntzhusen:
After clicking this link a few times I am now ready to fix it so I can save future clicks 😅


sofietoft:
Great, thanks for fixing the link 😄 

I committed a small change as well, to correct the word "master" to "primary" instead.

PR #2585, opened 2020-07-07
Issues upgrading from a version prior to 7.6

hjaltedaniel-umbraco:
If you upgrade your project from a version prior to 7.6 without updating your property editors to the non-obsolete version, that stores UDIs rather than IDs you might run into issues when trying to transfer content to Umbraco Cloud as these pickers will have the wrong content. It is therefore important that customers know that if they want a project prior to v7.6 on Cloud, they might need to do a db migration as well - if they're using the obsolete pickers.

### DIFF
--- a/Umbraco-Cloud/Getting-Started/Migrate-Existing-Site/index.md
+++ b/Umbraco-Cloud/Getting-Started/Migrate-Existing-Site/index.md
@@ -95,7 +95,7 @@ If you need help upgrading your project, we have some excellent [Upgrade instruc
 
 :::note
 #### Upgrading from a version prior to Umbraco 7.6.0
-With Umbraco 7.6.0 the property editors for pickers for content, media, members and related links has been updated to store UDI instead of node ID. If you're upgrading your site from a version prior to v7.6.0 or it has been upgraded earlier you might be using these pickers even though they have been marked as obsolete. 
+With Umbraco 7.6.0 the property editors for pickers for content, media, members, and related links have been updated to store UDI instead of node ID. If you're upgrading your site from a version prior to v7.6.0 or it has been upgraded earlier you might be using these pickers even though they have been marked as obsolete. 
 This can cause issues when transfering content to your Cloud project a bit down the road from here. 
 
 If you're using these obsolete pickers in the project you want on Umbraco Cloud, you will need to migrate to the newest versions of these pickers. This will include the need for a database migration. Benjamin Carleski, from ProWorks has made [this guide that take you through the process of migrating your database to use UDIs instead of IDs](https://skrift.io/issues/umbraco-migrations-made-easy/)
--- a/Umbraco-Cloud/Getting-Started/Migrate-Existing-Site/index.md
+++ b/Umbraco-Cloud/Getting-Started/Migrate-Existing-Site/index.md
@@ -93,6 +93,14 @@ You can download the latest version of Umbraco from [Our](https://our.umbraco.co
 
 If you need help upgrading your project, we have some excellent [Upgrade instructions](https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/general) you can follow. Be thorough when upgrading, as the latest upgrade might contain breaking changes and/or updated configuration.
 
+:::note
+#### Upgrading from a version prior to Umbraco 7.6.0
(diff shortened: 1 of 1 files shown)


Umbraco-Cloud/Getting-Started/Migrate-Existing-Site/index.md
```diff
@@ -93,6 +93,14 @@ You can download the latest version of Umbraco from [Our](https://our.umbraco.co
 
 If you need help upgrading your project, we have some excellent [Upgrade instructions](https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/general) you can follow. Be thorough when upgrading, as the latest upgrade might contain breaking changes and/or updated configuration.
 
+:::note
+#### Upgrading from a version prior to Umbraco 7.6.0
+With Umbraco 7.6.0 the property editors for pickers for content, media, members and related links has been updated to store UDI instead of node ID. If you're upgrading your site from a version prior to v7.6.0 or it has been upgraded earlier you might be using these pickers even though they have been marked as obsolete.
```
sofietoft, line 98:
```suggestion
With Umbraco 7.6.0 the property editors for pickers for content, media, members, and related links have been updated to store UDI instead of node ID. If you're upgrading your site from a version prior to v7.6.0 or it has been upgraded earlier you might be using these pickers even though they have been marked as obsolete. 
```

Umbraco-Cloud/Getting-Started/Migrate-Existing-Site/index.md
```diff
@@ -93,6 +93,14 @@ You can download the latest version of Umbraco from [Our](https://our.umbraco.co
 
 If you need help upgrading your project, we have some excellent [Upgrade instructions](https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/general) you can follow. Be thorough when upgrading, as the latest upgrade might contain breaking changes and/or updated configuration.
 
+:::note
+#### Upgrading from a version prior to Umbraco 7.6.0
+With Umbraco 7.6.0 the property editors for pickers for content, media, members and related links has been updated to store UDI instead of node ID. If you're upgrading your site from a version prior to v7.6.0 or it has been upgraded earlier you might be using these pickers even though they have been marked as obsolete. 
+This can cause issues when transfering content to your Cloud project a bit down the road from here. 
+
+If you're using these obsolete pickers in the project you want on Umbraco Cloud, you will need to migrate to the newest versions of these pickers. This will include the need for a database migration. Benjamin Carleski, from ProWorks has made [this guide that take you through the process of migrating your database to use UDIs instead of IDs](https://skrift.io/issues/umbraco-migrations-made-easy/)
```
sofietoft, line 101: Removed the mention of the author here, in order to keep the article here timeless.

```suggestion
If you're using these obsolete pickers in the project you want on Umbraco Cloud, you will need to migrate to the newest versions of these pickers. This will include the need for a database migration. [This guide will take you through the process of migrating your database to use UDIs instead of IDs](https://skrift.io/issues/umbraco-migrations-made-easy/).
```

hjaltedaniel-umbraco:
Suggestions committed 👍 
As far as I know, the ones listed is the updated property editors - [I simply copied this, from the breaking changes of 7.6.0](https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/760-breaking-changes#property-editors-storing-udi-instead-of-id-u4-9310).

About including it in version specific docs; it is not an issue per say, to be using the obsolete property editors. The content shows just fine, the editors are able to edit content etc. However, the issue starts when trying to transfer content between environments as Deploy will then transfer ID ints to the Cloud environment, that are probably not corresponding to the IDs of the local environment. Same if you transfer content between environments.
It would probably be a good thing to do these db migrations when upgrading from prior v7.6 to past v7.6 - however, it is not required.

sofietoft:
Thanks for clarifying about the property editors. Completely missed that they are already listed in the text you wrote.
Would it make sense to pull them out of the text and into a list instead? That makes it a little easier to quickly get an overview of which editors are impacted? 🤔 

And good point about the version specifics - sounds like it's "only" relevant on Cloud environments 👍

hjaltedaniel-umbraco:
I've pulled them out of the text, and into a list - I agree that this enhances the reading experience 😅

PR #2592, opened 2020-07-09
Purged the smallest typo

peaxe:
- Removed a small typo.


sofietoft:
Good catch @peaxe ! 😄 

I'm merging right away.

PR #2601, opened 2020-07-17
Add tip for injecting a Lazy version of IUmbracoContextFactory

BarryFogarty:
This is a gotcha when injecting `IUmbracoContextFactory` into a service which is registered as part of an `IUserComposer` as per the example on this page.  I have found that there are situations, particularly when interacting with third party extension code (e.g. Umbraco Packages), the provided example will fail with a boot error 'Recursive dependency detected: ServiceType:Umbraco.Web.IUmbracoContextFactory'.  

The tip provided fixes this issue, whereas other techniques (such as applying the ComposeBefore/After attribute) have no effect.  I feel like there are many other situations where this could crop up and therefore would be a worthwhile addition to this page.

Ref: forum discussion: https://our.umbraco.com/packages/backoffice-extensions/contentment/contentment-feedback/102944-boot-fail-when-using-an-injected-service-recursive-dependency-iumbracocontextfactory

### DIFF
--- a/Implementation/Services/Circular-Dependencies/index.md
+++ b/Implementation/Services/Circular-Dependencies/index.md
@@ -1,0 +1,29 @@
+---
+versionFrom: 8.0.0
+---
+
+# Circurlar Dependencies
+
+In some cases you might experience that a circular dependency is preventing your Umbraco installing from starting up.
+
+An example of this, could be a circular dependency on `IUmbracoContextFactory`, which would happen if your service interacts with third party code that also depends on an `IUmbracoContextFactory` instance.
+
+In this situation, you can request a lazy version of the dependency so it won't evaluate during boot, and would only be evaluated when accessed:
+
+```csharp
+public class SiteService : ISiteService
+{
+    private readonly Lazy<IUmbracoContextFactory> _umbracoContextFactory;
+    public SiteService(Lazy<IUmbracoContextFactory> umbracoContextFactory)
+    {
+        _umbracoContextFactory = umbracoContextFactory;
+    }
+     public IPublishedContent GetNewsSection()
+    {
+         using (UmbracoContextReference umbracoContextReference = _umbracoContextFactory.Value.EnsureUmbracoContext()) 
+         {
+             // Do your thing
+         }
+    }
+}
+```
--- a/Implementation/Services/Circular-Dependencies/index.md
+++ b/Implementation/Services/Circular-Dependencies/index.md
@@ -2,7 +2,7 @@
 versionFrom: 8.0.0
 ---
 
-# Circurlar Dependencies
+# Circular Dependencies
 
 In some cases you might experience that a circular dependency is preventing your Umbraco installing from starting up.
 
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -446,27 +446,9 @@ namespace Umbraco8.Services
 The second approach can seem 'different' or more complex at first glance, but it is the syntax and method names that are slightly different... it enables the registering of the service in Singleton Scope, and its use outside of controllers and views.
 
 :::tip
-Occasionally, you may face a situation where Umbraco fails to boot, due to a circular dependency on `IUmbracoContextFactory`.  This can happen if your service interacts with   third party code that also depends on an `IUmbracoContextFactory` instance (e.g. an Umbraco package).  In this situation, you can request a lazy version of the dependency so it won't evaluate during boot, and would only be evaluated when accessed:
+Occasionally, you may face a situation where Umbraco fails to boot, due to a circular dependency on `IUmbracoContextFactory`.  This can happen if your service interacts with   third party code that also depends on an `IUmbracoContextFactory` instance (e.g. an Umbraco package).
 
-```csharp
-public class SiteService : ISiteService
-{
-    private readonly Lazy<IUmbracoContextFactory> _umbracoContextFactory;
-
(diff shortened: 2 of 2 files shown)


marcemarc:
Thanks @BarryFogarty !! - I saw the issue in the forum and had a chat with @leekelleher on Friday, glad you've found a workaround for now, and also anyone googling the error in the future will find the Our post, to discover Matt's tip. 

I agree we should have an example here on this page that works in these kinds of scenarios, what I'd sort of like to get to the bottom of is what is going on, because it doesn't feel like the 'DI is great everything just works' kinda experience - but maybe it does have to be this way... in this instance.

To me (largely guessing here) the problem seems related to Contentment 'type scanning' anything that implements IContentmentListItem in the ContentmentComposer to add to the ContentmentListItems() collection.

https://github.com/leekelleher/umbraco-contentment/blob/9c6c6f1b18756fa2b415c9d63086c0dbb6387f2a/src/Umbraco.Community.Contentment/Composing/ContentmentComponentComposer.cs#L21

But your CountryDataList implements IDataListSource, which in turn implements, IContentmentEditorItem, which in turn implements IContentmentListItem so I think possibly is being scanned and added here...? but is this before your ISiteService has been composed and added to the container? in the forum post, you try composing your ISiteService 'After' the Contentment Composer, but did you also try 'before'?

That said I wouldn't expect that to trigger the message with the circular reference? and the fact that using IUmbracoContextAccessor or Lazy<IUmbracoContextFactory> seems to work around the issue, it does appear related to the dependencies of UmbracoContextFactory...

The IUmbracoContextAccessor will give you access to the available UmbracoContext - if it's not available then IUmbracoContextFactory can ensure one exists, either using the current UmbracoContext or constructing a new one if one does not exist...

Anyway, I don't know :-) but I think it's worth just flagging this with HQ before we pull the tip in, to see if there is an opinion on why this is this way... even if it just confirms the tip is the correct workaround, so will escalate this upwards...

BarryFogarty:
Thanks for the detailed explanation @marcemarc - all makes a lot of sense.  Really I was hoping this PR would start a conversation about the ins and outs of the DI implementation, so holding on this while that takes place sounds like the best option.  If it transpires that using `IUmbracoContextAccessor` or `Lazy<>` is the correct workaround we can revisit this tip for the documentation then.

FYI I did try composing my `ISiteService` before as well as after, got the circular ref error on boot in both cases.  I tried a few different things to affect the order of the composers, but the error was the same.

Jeavon:
@Shazwazza @clausjensen any comments on this PR regarding a circular dependency with IUmbracoContextFactory?

Shazwazza:
I'm unsure about adding a tip specifically here about IUmbracoContextFactory because the underlying problem will not always be that particular service, it is just going to be a circular reference which can happen with any other service (have mentioned some here https://our.umbraco.com/packages/backoffice-extensions/contentment/contentment-feedback/102944-boot-fail-when-using-an-injected-service-recursive-dependency-iumbracocontextfactory#comment-321824). @Jeavon i think your initial investigation is leading to showing where the circular dependency actually exists but i think a little deeper digging needs to be done to find it which would be good to do to understand this case. In some situations, it's difficult to avoid (i.e. like i mention in my reply, the property editor collection is a bad one for circular refs... but also, any 'collection' list can cause this issue which it seems in Contentment is also the case here). We unfortunately have to work around this sometimes in the core too with `Lazy<T>`, it's normally an indication of violating the single responsibility principal, but nobody is perfect :) 

This is not an issue with LightInject. Keep in mind that IUmbracoContextFactory implementation takes dependencies as well, this is the code: https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web/UmbracoContextFactory.cs#L38

I would be more inclined to put this tip into another area of the docs that is a little more generic and then linking to it if that's possible?

sofietoft:
Hi there!
I have an important message for you, regarding this PR:

Yesterday we tried to do some cleaning up and force-pushed to the master branch which now causes some problems with existing forks and local clones. 
We have fixed up all the open PRs so that they are still mergeable now.

With that said, if you've been working on a **fork of the repo** or in a **local clone**, those will not work anymore.
If you wish to push any new changes to this PR, you will need to delete your local clone or fork and create fresh ones.

If you have any local work in progress, then you copy it over from your old fork or clone, following these steps:

**Forks**
1. In the old clone of the old fork, switch to the branch you were working on
2. In the new clone of the new fork, create a new branch with the same name as the old one
3. Copy all the affected files from the old clone to the new clone
4. Repeat steps 1-3 if you've been working with multiple branches
5. You can now safely delete your old fork

**Clones**
1. In the old clone, switch to the branch you were working on
2. Create and checkout a new branch on the new clone, using the same name as the old clone
3. Copy all the affected files from the old clone to the new clone
4. Repeat steps 1-3 if you've been working with multiple branches
5. Delete the old clone of the docs repo

And with that, you should be all set once more.

I do apologize for the inconvenience, and please let me know if you have any questions!
/Sofie

sofietoft:
Hi all! 👋 

After some back and forth on this one - I believe we at a Docs team meeting yesterday, come to a good solution on this one.
I've committed some changes to the PR:
- The code sample has been moved to a separate article, "Circular Dependencies"
- The tip remains on the Services index page, but now instead links to the Circular Dependencies article

It would be good if someone could give the wording on the new article an extra look - just to make sure, it still makes sense!
Thanks!

sofietoft:
Hi there,
Just letting y'all know that I'll merge this one for now.
Thanks for making the PR @BarryFogarty and to the rest of you for pitching in! 😄

Issue #847, opened 2018-06-04
Missing documentation for services


abjerner:
About half of the services are not documented - at least when looking here:

https://our.umbraco.org/documentation/Reference/Management/Services/

- [x] ApplicationTreeService
- [x] AuditService
- [x] ConsentService
- [x] ContentService
- [x] ContentTypeService
- [x] DataTypeService
- [x] DomainService
- [x] FileService
- [x] EntityService
- [x] ExternalLoginService
- [x] FileService
- [x] LocalizationService
- [x] MacroService
- [x] MediaService
- [x] MemberGroupService
- [x] MemberService
- [x] MemberTypeService
- [ ] MigrationEntryService
- [x] NotificationService
- [x] PackagingService
- [x] PublicAccessService
- [x] RedirectUrlService
- [x] RelationService
- [x] SectionService
- [x] ServerRegistrationService
- [x] TagService
- [ ] TaskService
- [x] TextService
- [x] UserService (work in progress : #604)

Just posting this for reference here. The ones in the list with a checkmark already have their own page(s) - the ones without the checkmark are missing. Should I create an issue for each missing service instead?

marcemarc:
@abjerner I think it's fine for this to be a single issue, we have been in two minds when discussing the services, in that I think they should all be listed, their existence acknowledged! - and a short description of what they are and where they might be used, and any quirks, example of common usage if appropriate - but we sort of also feel we want to steer clear of duplicating the content generated in the API docs - eg https://our.umbraco.org/apidocs/csharp/api/Umbraco.Core.Services.ContentTypeService.html (but each service should have a clear signpost to the relevant api doc too) - just having each service, listed with it's own page would be a great start, so the checkbox list above is super useful

Soundug:
@marcemarc Can I take this one? 
I could add some services documentation.

marcemarc:
@Soundug Yes, absolutely, we're trying to follow the pattern that @abjerner has established for the services - take a look at the domain service for example: https://our.umbraco.com/Documentation/Reference/Management/Services/

Essentially we are not trying to duplicate the auto generated api documentation eg https://our.umbraco.com/apidocs/csharp/api/Umbraco.Core.Services.DomainService.html

but kind of create a 'landing page' for each service, which links to the api documentation, and gives an overview of what the service does / is commonly used for, with samples on how to 'get a reference to it' etc

or at least first step is to get something existing for each service!

abjerner:
[...]

PR #2606, opened 2020-07-20
ServerRegistrationService

erikjanwestendorp:
Belogs to https://github.com/umbraco/UmbracoDocs/issues/847
Add ServerRegistrationService


marcemarc:
Thanks @erikjanwestendorp !! - we have a few to work through - as it's holidays for different folk at the moment - but we'll get back to you once we've had a good read through.

erikjanwestendorp:
@sofietoft Just updated the PR. (sorry for the same mistake all the time ;-))

PR #2620, opened 2020-07-25
Added documentation for eventsService

abjerner:
This PR is still work in progress. I haven't marked it as a draft as I'm not sure whether those are reviewed, and I may need some feedback in order to continue.

The events service is currently not documented - not even in the main Angular documentation - so this PR is an attempt to do at least something about that.

The first part of the page describes the three functions of the events service with both some text and some examples. This should be good to go - unless of course the docs team has some feedback for any changes.

The second part attempts to list all events currently used and broadcasted by Umbraco. This is where I need some feedback 😄 

My idea was to have a quick example of how each event is broadcasted, but also a link to the source code. For one, this could let developers get a better overview of how the event is broadcasted. And second, linking to a specific tag - eg. `release-8.6.3` - may make it easier to maintaining the list going forward.

I'm not sure whether having the links to the source code there adds too much noise, but if so we could add the links as HTML comments so they're still there for reference.

Currently the links refer to a mix of mix of files at at specific commit and files from a specific release/tag. Ideally they should all link to a specific release, so if we decide that their is a need to show/include the links, I will go through them and make sure they all refer to the tag of the most recent release.

Some of the events may also need a bit more description or better examples than what I have so far - eg. so it's better explained which arguments an given event has. But at least this should serve as a start 😉

### DIFF
--- a/Reference/Angular/Services/eventsService/index.md
+++ b/Reference/Angular/Services/eventsService/index.md
@@ -1,0 +1,471 @@
+# Events Service
+
+The events service allows different components in Umbraco to broadcast and listen for global events.
+
+https://github.com/umbraco/Umbraco-CMS/blob/v8/contrib/src/Umbraco.Web.UI.Client/src/common/services/events.service.js
+
+### Broadcasting an event
+
+To broadcast an event, you can use the `emit` function. It takes two arguments, where the first is the name of the event - eg. `featured.updated`, and the second argument is an object or similar describing the event.
+
+The second argument is optional, so if your use case doesn't need this, feel free to skip this argument.
+
+The illustrate this function, you could have a controller with an `updated` function that is triggered by the view. In this dummy example, the function will increment the value of a property editor, and then use the events service to broadcast that the value was updated:
+
+```javascript
+angular.module('umbraco').controller("MyController", function($scope, eventsService) {
+
+    $scope.updated = function() {
+        $scope.model.value++;
+        eventsService.emit("feature.updated", { value: $scope.model.value });
+	};
+
+});
+```
+
+### Listening for an event
+
+Another controller could then listen for broadcasts of your `feature.updated` event via the events service's `on` function, which takes the name of the event as the first argument, and a callback function as the second argument.
+
+Then in the callback function, the first argument is the event it self, and the second argument is the object we pass on to the `emit` function when we're broadcasting:
+
+```javascript
+angular.module('umbraco').controller("MyOtherController", function($scope, eventsService) {
+
+    $scope.count = 0;
+    
+    // Subscribe to the event
+    var unsubscribe = eventsService.on("feature.updated", function(event, args) {
+        $scope.count = args.value;
+    });
+    
+    // When the scope is destroyed we need to unsubscribe
+    $scope.$on("$destroy", function () {
+        unsubscribe();
+    });
+
+});
+```
+
+### Unsubscribing from an event
+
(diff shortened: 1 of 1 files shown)


sofietoft:
This is super @abjerner ! Thanks for putting in all the work to write this.

I'll make sure it get's reviewed as soon as possible.
I have a few comments as well:

- Is there a reason you're jumping straight from an H1 to an H3 heading in the beginning? 
- There's a lot of links in this article, but when they're just written out without any Markdown syntax around them, I don't think they will render as links on Our 🤔 We'll need to wrap each of these links in some syntax and give them a proper link text. I realize you're not done with all the links, just thought I'd mention it still 😄

marcemarc:
@abjerner This is great!

I'm not sure about the links into the source code either - from a documentation point of view it's important to flag where these events are thrown, but you kind of have that with the 'titles' - for each event. 

If anything you want a little video showing the 'interaction' that the event relates to :-P - because it's hard to know from the names or the code! - but that might be a little over the top with the amount of events! - call it phase II.

Having the links there is really useful for updating this content moving forwards - so agree Html comments would be a way to reduce the noise but keep the reference, they will I think be stripped during the publishing process.... agree using a tag is the consistent way of keeping the particular line number link working.

And yes better examples can come later, the important thing is highlighting what you already have here, for something people can hook into when working within the backoffice.

abjerner:
> Is there a reason you're jumping straight from an H1 to an H3 heading in the beginning?

The article is still a bit work in progress, but I probably missed the H2, which I've added now.

> There's a lot of links in this article, but when they're just written out without any Markdown syntax around them, I don't think they will render as links on Our

Markdig should make sure that the raw links are made clickable as well. It shouldn't be necessary to mark it up with Markdown syntax, although than would allow us to have a custom link text. But as I mentioned, I wasn't sure how and if these links should be visible at all.

> Having the links there is really useful for updating this content moving forwards - so agree Html comments would be a way to reduce the noise but keep the reference, they will I think be stripped during the publishing process.... agree using a tag is the consistent way of keeping the particular line number link working.

HTML comments are not stripped out by Markdig, so they will also be in the output HTML on Our. But still as HTML comments, so not visible. Would that still be cool? Or is there a better way for keeping "editorial notes"?

sofietoft:
Hi there!
I have an important message for you, regarding this PR:

Yesterday we tried to do some cleaning up and force-pushed to the master branch which now causes some problems with existing forks and local clones. 
We have fixed up all the open PRs so that they are still mergeable now.

With that said, if you've been working on a **fork of the repo** or in a **local clone**, those will not work anymore.
If you wish to push any new changes to this PR, you will need to delete your local clone or fork and create fresh ones.

If you have any local work in progress, then you copy it over from your old fork or clone, following these steps:

**Forks**
1. In the old clone of the old fork, switch to the branch you were working on
2. In the new clone of the new fork, create a new branch with the same name as the old one
3. Copy all the affected files from the old clone to the new clone
4. Repeat steps 1-3 if you've been working with multiple branches
5. You can now safely delete your old fork

**Clones**
1. In the old clone, switch to the branch you were working on
2. Create and checkout a new branch on the new clone, using the same name as the old clone
3. Copy all the affected files from the old clone to the new clone
4. Repeat steps 1-3 if you've been working with multiple branches
5. Delete the old clone of the docs repo

And with that, you should be all set once more.

I do apologize for the inconvenience, and please let me know if you have any questions!
/Sofie

abjerner:
I created a new issue, so hereby closing this issue...

https://github.com/umbraco/UmbracoDocs/pull/2647

PR #2637, opened 2020-08-07
Added Umbraco Heartcore GraphQL documentation

rasmusjp:
The API docs covers how the GraphQL schema is generated, what types the Property Editors are returning and how to query the different types.

There's also a tutorial guiding you through setting up a basic site in Heartcore and querying it with GraphQL


Umbraco-Heartcore/Tutorials/Querying-With-GraphQL/index.md
```diff
@@ -0,0 +1,523 @@
...
+      ]
+    }
+  }
+}
+```
+
+
+### Using Variables
+
+You might have noticed that the `url` argument is hardcoded in the query, this means that we will always get back the Content for that url. What we want is to pass the argument to the query dynamically.
```
sofietoft, line 413:
```suggestion
You might have noticed that the `url` argument is hardcoded in the query. This means that we will always get back the Content for that url. What we want is to pass the argument to the query dynamically.
```

Umbraco-Heartcore/Tutorials/Querying-With-GraphQL/index.md
```diff
@@ -0,0 +1,523 @@
...
+  "variables": {
+    "url": "/home/"
+  }
+}
+```
+
+
+### Putting it all together
+
+Now that we have all the individual parts we can combin it all into a single query that fetches all the data needed to display the pages on our site:
```
sofietoft, line 446:
```suggestion
Now that we have all the individual parts we can combine it all into a single query that fetches all the data needed to display the pages on our site:
```

PR #2647, opened 2020-08-10
Added documentation for eventsService

abjerner:
Follow up from: https://github.com/umbraco/UmbracoDocs/pull/2620

This PR still contains visible links to usage of `eventsService`. Should these be converted to HTML comments? Or should we do something else? HTML comments are still visible in the source code when the documentation is rendered on Our.

### DIFF
--- a/Reference/Angular/Services/eventsService/index.md
+++ b/Reference/Angular/Services/eventsService/index.md
@@ -81,7 +81,7 @@ Below you'll find a list of events broadcasted by the Umbraco codebase. The list
 eventsService.emit("app.ready", data);
 ```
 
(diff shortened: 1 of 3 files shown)


dampee:
Just discussed this: yes please use HTML comments.

sofietoft:
Thanks again for the PR @abjerner !
As Damiaan mentioned, let's get the links converted to HTML comments to begin with.
Then we can get this one merged, and take it from there.
How does that sound?

abjerner:
Great. I've pushed another commit that adds HTML comments around most of the links. I've kept the initial link to `events.service.js`, as that may still be relevant.

PR #2656, opened 2020-08-13
Updated Retrieving-languages.md examples to work for Umbraoc 8

abjerner:
There is another Markdown file for Umbraco 7, and this has `versionFrom: 8.0.0` - so I've updated the link to the API docs and removed `ApplicationContext.` from the Razor example.


sofietoft:
Good catch @abjerner !

Looks like a few things missed our noses there 😅 
Thanks for getting it up to date!

PR #2657, opened 2020-08-13
Might as well have syntax highlighting

abjerner:


sofietoft:
You're absolutely right. Might as well 😄 🎉 

Thanks for updating the article!

PR #2658, opened 2020-08-14
Fixing the issue with latest release for tutorial Article Item

thejokerd3:
Fixing the issue with the Truncate Method. It prints out the <p> tags to the frontend view when using the Rich Text Box Editor. It is not supposed to do that. It has now been fixed so the tutorial is up to date with the latest releases.


sofietoft:
Thanks for the PR @thejokerd3 !

Great to have this updated 🎉

PR #2665, opened 2020-08-17
Suggestion: The fallback handling for changefreq can give invalid output

jakob-bagterp:
When using the current fallback to ancestors strategy for the changefreq variable, the view occasionally produces invalid output like this...

`<changefreq/>
`

... instead of:

`<changefreq>monthly</changefreq>
`

This is most likely due to undesired handling of empty strings (and not just null values) in the fallback function, and the suggested "old school" ternary operator method fixes the fallback handling (though the ancestors strategy would be better it worked for both null and empty strings).

### DIFF
--- a/Tutorials/Creating-an-XML-Site-Map/index.md
+++ b/Tutorials/Creating-an-XML-Site-Map/index.md
@@ -118,7 +118,7 @@ This is a great candidate for a [Razor Helper](https://weblogs.asp.net/scottgu/a
 ```csharp
 @helper RenderSiteMapUrlEntry(IPublishedContent node)
 {
-    var changeFreq = node.Value("searchEngineChangeFrequency", fallback: Fallback.To(Fallback.Ancestors, Fallback.DefaultValue),defaultValue:"monthly");
+    var changeFreq = node.HasValue("searchEngineChangeFrequency") && node.Value<string>("searchEngineChangeFrequency") != "" ? node.Value<string>("searchEngineChangeFrequency") : "monthly";
     // with the relative priority, this is a per page setting only, so we're not using recursion, so we won't set Fallback.ToAncestors here and we'll default to 0.5 if no value is set
     var priority = node.HasValue("searchEngineRelativePriority") ? node.Value<string>("searchEngineRelativePriority") : "0.5";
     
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @jakob-bagterp ! 💪 
Much appreciated.

dampee:
Hi @jakob-bagterp 

Thanks a lot for your suggestion.  The change you propose is not using the fallback anymore. So the code would work in a completely different way.  

That being said: is it possible you have not configured the same field on ancestor nodes? Because the fallback should always work and not "occasionally".

Kind regards

marcemarc:
@jakob-bagterp Just following up on this - there exists a bug in Umbraco for 8.0 to 8.6 for Dropdown Values not reverting back to 'empty' if they have been previously set.

so that if a new document is created, and the dropdown is left untouched, the fallback will work as expected, Umbraco will see the dropdown as not being set - however, if the dropdown is changed to be a value and published, and then subsequently changed back to be 'empty' - for these versions of Umbraco 8.0 to 8.6.x - Umbraco still perceives there to be a value set in the empty dropdown and the fallback doesn't work.

This is fixed in 8.7: https://github.com/umbraco/Umbraco-CMS/pull/8154

So the first question, is which version of Umbraco are you trying the tutorial with? - as it would be super useful to know there is still an issue in sites after 8.7 in case another problem has occurred.

We've put a note on the tutorial to warn about this behaviour 

image

It's a tricky decision, as the tutorial is really to show off good practice working with razor, and the sitemap, is a real world-ish example - we want to try and keep it as simple as possible, and show off the concept of the fallback - but that's not helpful if it doesn't actually work!! - if all is ok in 8.7+ then at least most people building new things will be using the latest version - but maybe we need to move the warning to the beginning of the article? so people know in advance of starting the article that there is an underlying issue!

Be great to get your thoughts on this from the perspective of somebody tackling the tutorial.

regards

Marc

sofietoft:
Hi all!

I've moved the warning about the bug in this article, above the code snippet - so, the warning will be the first thing you see, when you get to the section about rendering a site map entry.

With that, and as the bug is also resolved as of 8.7, I'll close this PR for now.
Thanks to all who pitched in 🙌

PR #2667, opened 2020-08-18
Examine - Fixed a issue with code sample

poornimanayar:
Hi,

Noticed while working today that there is an issue with the code sample https://our.umbraco.com/Documentation/Reference/Searching/Examine/quick-start/#search-children-of-a-specific-node 

While searching for children of specific node there needs to be an `And()` before the `Field()` can be used

Poornima


sofietoft:
Good catch @poornimanayar !
Thanks for amending the article 💪

PR #2671, opened 2020-08-18
Added note about ContentFinderByUrlAndTemplate

NikRimington:
Added note about ContentFinderByUrlAndTemplate  being disabled by default and linking to IContentFinder documention on how to re-add it in.

This is to address the docs due to https://github.com/umbraco/Umbraco-CMS/issues/8650 and @marcemarc 's response.


sofietoft:
Looks great @NikRimington ! 😄 

Thanks for adding this to the docs - much appreciated!

PR #2675, opened 2020-08-20
Update azure-web-apps-vpre-8.6.0.md

deMD:
There is no setting that you need to set to true


sofietoft:
Thanks for the PR @deMD ! Much appreciated.

So, if there is no setting, doesn't it make sense, to perhaps instead rewrite the sentence, so it does not mention this setting? 🤔 
Or is it simply, that it doesn't been to be set to true?

deMD:
here is a setting, as you can see at the end of the text, it's just that it isn't set to 'true' but it is instead set to a value indicating where the data should be stored
`<add key="Umbraco.Core.LocalTempStorage" value="EnvironmentTemp" />`

PR #2677, opened 2020-08-23
Update reference for SiteService example

jakob-bagterp:
Line 424...

IPublishedContentCache contentCache = umbracoContextReference.UmbracoContext.Content;

... currently gives and error:

CS0246 The type or namespace name 'IPublishedContentCache' could not be found (are you missing a using directive or an assembly reference?)


sofietoft:
Good catch @jakob-bagterp !

Thanks for updating the code sample here 💪 Much appreciated!

PR #2681, opened 2020-08-27
Add Umbraco Volatile docs

nikolajlauridsen:
Added a document about Umbraco Volatile Error, currently it shouldn't be available in the menu since it's not finished yet, however we need somewhere to link to from the error tooltip.

### DIFF
--- a/Reference/UmbracoVolatile.md
+++ b/Reference/UmbracoVolatile.md
@@ -1,0 +1,59 @@
+# Umbraco Volatile Error
+
+Umbraco Volatile Errors occurs when you are trying to use a method which *should* be internal, but is left 
+public since it's useful when testing. 
+
+Methods marked with UmbracoVolatile may break in the future, however if you want to use the method for testing, 
+you can use an assembly level attribute called ```UmbracoSuppressVolatile```, when the attribute is applied to an assembly 
+any occurrences of Umbraco Volatiel Error will be suppressed to a warning. 
+
+
+```UmbracoSuppressVolatile``` is located in the ```Umbraco.Core.CodeAnnotations```, however any attribute named UmbracoSuppressVolatile 
+will do the trick if you don't have access to that specific namespace. 
+
+## Example of suppresing Umbraco Volatile Error
+
+~~~c#
+    public class DemoClass
+    {
+	    [UmbracoVolatile]
+        public void VolatileMethod()
+        {
+           // Volatile things here...
+        }
+    }
+~~~
+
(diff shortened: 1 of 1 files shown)


sofietoft (changes requested):
This is awesome @nikolajlauridsen ! Thanks for adding this article 👏 

I've made a few suggestions here and there.
Also, have a look at this article about which [Markdown conventions we use fo inline code](https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/).

For code snippets, instead of `~~~`, you should use ```.

Also:
Could you add some [YAML versioning tags](https://our.umbraco.com/documentation/Contribute/Adding-Metadata/) at the top of the article? 🤩

Finally, as long as this is a markdown file and not a folder, it should not be added to the menu structure. Do you have some ideas, on which articles should link to this one? 🤔 

Hope it all makes sense. Happy to answer any questions you might have to any of the above!

nikolajlauridsen:
Thanks for reading my pull request :) 

I've applied your suggestion, and changed the code blocks to use backticks. 

I've added the title tag and description tag in the top, as for version we're not exactly certain for what version this will be relevant yet. The imidiate plan is to have it be applicable in the .net core version, which has no version number currently, however, there's no reason why it couldn't be used in the framework version as well, so that's why there's no version number.

The version number of Umbraco Volatile is also more related to the version of the Umbraco.Code project (which would beversion 1.1.0), but that might cause confusion? 

As for which articles should link to this one, then currently I don't think any articles should, this article is just used to have somewhere to link to from the error tooltip you get in your editor. 

Again thanks for your edits, documentation is hard :D

sofietoft:
Documentation can be hard, yes 😉 but also fun, I hope 😅  

Alright! I hear what you're saying.
So, this is actually only going to be a thing in the .NET Core version, and not something you might run into now in the current release of Umbraco?
What you really need here, is something to link to, once people start trying out the Alpha of .NET Core? 🤔 
Cause if that's the case, I think it'd be fine to release the article without a versioning number - as long as we don't link to it anywhere in the docs, people won't stumble on it accidentally...

nikolajlauridsen:
That's correct 😄

sofietoft:
Super!
I've added a little note at the very top of the article - could you let me know if what I've added there, makes sense?

nikolajlauridsen:
Looks all good to me 👍

sofietoft (approved):
Sweet.
Let's get it merged then 😄 

Thanks again for writing the article!
Looking forward to many more contribs from you 😉

PR #2685, opened 2020-08-31
Removing repetitive wording around generating meta data

hjaltedaniel-umbraco:
There are currently some wording around "again" and "double-check", when the users are asked to do an echo > deploy command. That is weird, since this is the first time the user is introduced with this.


sofietoft (approved):
Ah, that's a very good point @hjaltedaniel-umbraco !

Thanks for making the changes. Makes a lot of sense 💪

Issue #2690, opened 2020-09-04
Update all examples of subscribing to static events to include corresponding 'unsubscribing' to that event during termination


marcemarc:
eg this sort of example:

public class SubscribeToSavedEventComponent : IComponent
    {
        private readonly ILogger _logger;

        public SubscribeToSavedEventComponent(ILogger logger)
        {
            this._logger = logger;
        }
       
        public void Initialize()
        {
            //specify my event handler
            MemberService.Saved += MemberService_Saved;
        }

        private void MemberService_Saved(IMemberService sender, SaveEventArgs<IMember> e)
        {
            foreach (var member in e.SavedEntities)
            {
                
              //write to the logs every time a member is saved
              this._logger.Info<MyPublishEventComponent>("member {member} has been saved and event fired!", member.Name);
            }
        }
        public void Terminate()
        {
            // Nothing to terminate
        }

should be

public class SubscribeToSavedEventComponent : IComponent
    {
        private readonly ILogger _logger;

        public SubscribeToSavedEventComponent(ILogger logger)
        {
            this._logger = logger;
        }
       
        public void Initialize()
        {
            //specify my event handler
            MemberService.Saved += MemberService_Saved;
        }

        private void MemberService_Saved(IMemberService sender, SaveEventArgs<IMember> e)
        {
            foreach (var member in e.SavedEntities)
            {
                
              //write to the logs every time a member is saved
              this._logger.Info<MyPublishEventComponent>("member {member} has been saved and event fired!", member.Name);
            }
        }
        public void Terminate()
        {
             //unsubscribe to event during shutdown
            MemberService.Saved -= MemberService_Saved;
        }

See this PR for context: https://github.com/umbraco/Umbraco-CMS/pull/8815

marcemarc:
PR here: https://github.com/umbraco/UmbracoDocs/pull/2692/files

nul800sebastiaan:
Fixed in https://github.com/umbraco/UmbracoDocs/pull/2692

PR #2692, opened 2020-09-04
Update Examples that show subscribing to a Static Event in a component to show unsubscribing from the event in Terminate()

marcemarc:
See issue: https://github.com/umbraco/UmbracoDocs/issues/2690

This PR updates examples that subscribe to static events in a component, to also include an example of unsubscribing

### DIFF
--- a/Reference/Searching/Examine/examine-events.md
+++ b/Reference/Searching/Examine/examine-events.md
@@ -66,6 +66,7 @@ public class ExamineEvents : IComponent
 
     public void Terminate()
     {
+ indexProvider.TransformingIndexValues -= IndexProviderTransformingIndexValues;
     }
 }
 ```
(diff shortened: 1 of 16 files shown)


Reference/Searching/Examine/examine-events.md
```diff
@@ -66,6 +66,7 @@ public class ExamineEvents : IComponent
 
     public void Terminate()
     {
+ indexProvider.TransformingIndexValues -= IndexProviderTransformingIndexValues;
```
sofietoft, line 69:
```suggestion
    {
        //unsubscribe during shutdown
       indexProvider.TransformingIndexValues -= IndexProviderTransformingIndexValues;
```

sofietoft (commented):
Looks great @marcemarc ! 😄 

I added a single suggestion to add a comment - just to keep all the samples consistent.
Let me know if it doesn't make sense!

sofietoft (approved):
Thanks @marcemarc ! 😄 

Just noticed that something was a bit off with the syntax highlighting in the Examine Events article, so I updated it.

PR #2693, opened 2020-09-04
Small corrections and notes in regards to content apps

Jay-umbr:
Added a note which explains which icon pack we use, and added an explanation on where to put the cs file when making a content app the C# way


sofietoft (approved):
Looks great!

Thanks for adding the extra tips to make this tutorial even easier to follow 🙌

PR #2694, opened 2020-09-08
Added video links for articles - Heartcore

Jay-umbr:
(API keys, deployments, GraphQL playground, intro to cloud, intro to backoffice)


sofietoft:
Looks great @Jay-umbr !

I think it would be great with a heading for each of the videoes.
Should be something simple like "Video" and perhaps some text that says "Watch a video tutorial on how to ... "

What do you think?

Jay-umbr:
Not sure, I followed what I saw on some other doc pages - where videos are pretty much inline with content, without headers. If you feel like that would be better I can add the headers too 👍

sofietoft:
Alright. Let's see how it looks. We can always add heading afterwards.

Thanks again for the PR.

PR #2695, opened 2020-09-08
Extra info on installing full version of ModelsBuilder

tristanjthompson:
I struggled a bit this week after reading this to understand the best way to install the full version of models builder as nuget doesn't seem to be referenced in the ModelsBuilder repository, and it took me a while to figure out which nuget packages to install.

So I thought I'd submit a PR with more detailed instructions in case anyone else has the same problem!  I also wanted to include a link to the package on Our, but I couldn't find it?  Only the old version of ModelsBuilder?

### DIFF
--- a/Reference/Templating/Modelsbuilder/Control-Generation.md
+++ b/Reference/Templating/Modelsbuilder/Control-Generation.md
@@ -6,7 +6,7 @@ versionFrom: 8.5.0
 
 Generation of models can be controlled through various C# attributes in the custom files.
 
-These features are only available on the full version of the [Models Builder](https://github.com/modelsbuilder/ModelsBuilder.Original).
+These features are only available on the full version of the [Models Builder](https://github.com/zpqrtbnk/Zbu.ModelsBuilder).
 
 ## Installing the Full Version of Models Builder
 The full version can be installed with the following nuget packages.  
--- a/Reference/Templating/Modelsbuilder/Control-Generation.md
+++ b/Reference/Templating/Modelsbuilder/Control-Generation.md
@@ -6,4 +6,25 @@ versionFrom: 8.5.0
 
 Generation of models can be controlled through various C# attributes in the custom files.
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @tristanjthompson !

I think it's a good idea having some steps on how to install the full version of the Models Builder.
However, we should put the steps to do it in their own article.
Would you be up for doing that? We have some useful docs around about [file structure](https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/) and some about [file names](https://our.umbraco.com/documentation/Contribute/File-Naming-Conventions/) as well, for reference.

Also, if you have a look here: https://our.umbraco.com/documentation/reference/templating/modelsbuilder/
We have a link to the full Models Builder package 💪

sofietoft:
Hi @tristanjthompson !

I've taken the liberty to add the correct link to the ModelsBuilder package.

I'll merge this PR now, and make sure to add the steps to install to it's own article instead, as I feel that would be a better fit.

Thanks again for providing the steps and for making this PR! Much appreciated 💪

tristanjthompson:
Hi @sofietoft  - I'm so sorry I didn't get back to you - life overtook me with wedding planning and I haven't had time to look at this yet!

The link I put it is actually the same repo - if you go to `https://github.com/zpqrtbnk/Zbu.ModelsBuilder` you'll see that it redirects to `https://github.com/modelsbuilder/ModelsBuilder.Original` which is why I changed it - both links end up in the same place though so it's all good!

Thanks for looking through this 🎉 I'll keep it on my to-do list to submit a PR to the ModelsBuilder repo at some point!

sofietoft:
No worries at all @tristanjthompson ! 😄 
I'm just doing a bit of cleaning up here, so we can be ready for Hacktoberfest madness which starts tomorrow.

Ah! Good points about the link - didn't even notice that 😅 

Cheers!

PR #2697, opened 2020-09-08
Remove 15 hostnames mention

ArnoldV:
Removed the mention that you can add up to 15 hostnames to a cloud plan. This is now dependent on the cloud plan you have purchased.


sofietoft:
Good catch @ArnoldV ! 💪 

Thanks for setting this right! Much appreciated.

PR #2698, opened 2020-09-09
Added note for including document type name in the nested content template

poornimanayar:
Hi,

I dont know whether this is useful, but I wanted to use the document type name as the template and figured out that it can be done by using `{{alias}}` as the template. Hence raising a PR.

Poornima


sofietoft:
This is a great addition @poornimanayar !

Thanks for adding this to the article 💪 Much appreciated.

PR #2700, opened 2020-09-11
Update index.md - Added a headline

jesperlange:
Customer said it was a bit lackluster, and I agreed it was a bit short, he didn't see the fix and he skimmed for it.

I agreed and said it could perhabs use a headline that states "here is how to fix it"


sofietoft (approved):
That's a great observation actually 💪 Thanks for adding this heading @jesperlange ! Much appreciated.

PR #2705, opened 2020-09-11
"Block List" proofreading corrections

hfloyd:


sofietoft (approved):
Thanks for catching and correcting all these isses @hfloyd ! 😄 
Much appreciated!

PR #2707, opened 2020-09-11
provide a couple of tips

kenburcham:
I gave just a couple of tips that would have been helpful to me!

### DIFF
--- a/Reference/Security/authenticate-with-AD.md
+++ b/Reference/Security/authenticate-with-AD.md
@@ -64,7 +64,9 @@ The `ActiveDirectoryBackOfficeUserPasswordChecker` will look in appSettings for 
     <add key="ActiveDirectoryDomain" value="mydomain.local" />
 </appSettings>
 ```
-Note: one way to find your Active Directory Domain if you are logged into your domain is to open a command prompt and run "set logon" and use the value returned as the LOGONSERVER (not including any slashes).
+:::tip
+One way to find your Active Directory Domain if you are logged into your domain is to open a command prompt and run `set logon` and use the value returned as the `LOGONSERVER` (not including any slashes).
+:::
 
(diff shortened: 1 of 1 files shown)


Reference/Security/authenticate-with-AD.md
```diff
@@ -64,6 +64,7 @@ The `ActiveDirectoryBackOfficeUserPasswordChecker` will look in appSettings for
     <add key="ActiveDirectoryDomain" value="mydomain.local" />
 </appSettings>
 ```
+Note: one way to find your Active Directory Domain if you are logged into your domain is to open a command prompt and run "set logon" and use the value returned as the LOGONSERVER (not including any slashes).
```
sofietoft, line 67:
```suggestion
:::tip
One way to find your Active Directory Domain if you are logged into your domain is to open a command prompt and run `set logon` and use the value returned as the `LOGONSERVER` (not including any slashes).
:::
```

sofietoft (commented):
Thanks for adding the tips @kenburcham !
Makes a lot of sense!

I've suggested converting the note to a tip instead.
That way, it'll be rendered as a blue box with the tip inside it - let me know what you think!

sofietoft (approved):
Thanks for taking the suggestion! A box here, will make the tip stand out and almost impossible to miss 😄 

Thanks again for the PR! 
I can see that this is the first time you're contributing to the Umbraco Docs! That's awesome #H5YR 😎 
Do you have an account on [Our](https://our.umbraco.com/)? If you do, I can grant you a **C-Trib** badge for contributing. Let me know!

kenburcham:
Yes, I’m on our: https://our.umbraco.com/members/kenburcham/

Thanks, I’d love a badge! ☺

sofietoft:
Woohoo! You now have a new, shiny **C-Trib** badge on your Our profile 🎉 

Thanks again for helping out.
Looking forward to see more from you in the future!

PR #2722, opened 2020-09-23
Update index.md

HalldorLyngmo:
The timeouts were increased to 20 minutes in v3.5.0. You can see the release notes mention it here - https://our.umbraco.com/packages/developer-tools/umbraco-deploy/
 Since the code example had 1200 seconds I increased it to 1800 so it would be 30 minutes if people are copying it straight from here.


sofietoft (approved):
Thanks for adding the updated numbers @HalldorLyngmo !
Much appreciated 💪

PR #2725, opened 2020-09-23
An old reference link deleted.

busraparnell:


sofietoft:
Ah. Good catch @busraparnell !

How about instead of deleting the link entirely, we update it instead?
Not sure if the video is at all relevant still - but I found the video here: https://umbraco.23video.com/video/635188/codegarden-10-web-search-with

Let me know what you think!

PR #2729, opened 2020-09-26
More detail about creating custom backoffice views

sussexrick:
When using custom edit views with inline editing the `block.edit()` command does nothing, so it should be stated that custom views are incompatible with inline editing.

When you add an image to a block editor element type you only get back the UDI. It's not obvious how to display the image, so I've linked to a forum message explaining this.

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Block-List-Editor/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Block-List-Editor/index.md
@@ -190,11 +190,11 @@ Example:
 
 You can choose to customize your editing experience by implementing a custom view for each Block.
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @sussexrick !

Great catch with the custom view and inline editing! Thanks for updating that.

Regarding the link to the forum - I believe what Marc is explaining there, is very similar to what we have in the tutorial on creating a Custom Property Editor. Perhaps it would make sense to link to that one as well? 🤔 
Just a thought.

sussexrick:
I've had a look at the property editor pages and I'm not sure it's helpful to link to those. Yes, the configuration is the same but the property editor pages also talk about a lot of other options that I don't think are relevant here, and would just be confusing, eg a different way to specify CSS that's not how the Block List Editor works. Marc's post is focused on only the information that's relevant here.

sofietoft:
Sure! That's also a valid point 🤔 

Would be good to have the example in the docs as well eventually - but for now, I think the link would be OK.

I'll get this merged.
Thanks again for updating the article!

PR #2732, opened 2020-09-28
Adds docs for ExportMemberGroups Deploy Setting in 3.4.0+

### LINKED ISSUE #2708: exportMemberGroups option is not documented
The [Umbraco Deploy releases page](https://our.umbraco.com/packages/developer-tools/umbraco-deploy/) refers to an `exportMemberGroups` setting introduced in Deploy 3.4.0, but it's not documented where this goes. Is it an element? An attribute? Is in in `UmbracoDeploy.Settings.config` or `UmbracoDeploy.config`? What's the default value?

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Umbraco-Cloud/Deployment/Deploy-Settings/index.md***

warrenbuckley:
Adds docs for **ExportMemberGroups** settings
Fixes #2708

### DIFF
--- a/Umbraco-Cloud/Deployment/Deploy-Settings/index.md
+++ b/Umbraco-Cloud/Deployment/Deploy-Settings/index.md
@@ -80,3 +80,16 @@ In order for your Cloud project to handle Forms data as content, you'll need to 
 :::note
 This only needs to be done if your Cloud project was created **before** June 18th 2019. Make sure to [follow the instructions](../Umbraco-Forms-on-Cloud/#did-you-create-your-cloud-project-before-june-18th-2019) to succesfully change this setting on all environments.
 :::
+
+
+## ExportMemberGroups
+This is available in Deploy 3.4.0+ version and newer. This setting is to be enabled only if you are using an external membership provider for your members and thus it does not make sense to export member groups which would no longer be managed by Umbraco but by an external membership provider.
+
+Setting the `exportMemberGroups` to false will no longer export member groups to disk UDA entities, by default if this setting is not present its value will automatically be set to true as most sites use Umbraco's built in membership provider and thus will want the membership groups exported.
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<settings xmlns="urn:umbracodeploy-settings">
+    <deploy exportMemberGroups="false" />
+</settings>
+```
--- a/Umbraco-Cloud/Deployment/Deploy-Settings/index.md
+++ b/Umbraco-Cloud/Deployment/Deploy-Settings/index.md
@@ -86,7 +86,7 @@ This only needs to be done if your Cloud project was created **before** June 18t
 
 This is available in Deploy 3.4.0+ version and newer. 
 
-This setting is to be enabled only if you are using an external membership provider for your members. You will not want to export member groups that would no longer be managed by Umbraco but by an external membership provider.
+This setting is to be defined and set to false only if you are using an external membership provider for your members. You will not want to export member groups that would no longer be managed by Umbraco but by an external membership provider.
 
 Setting the `exportMemberGroups` to false will no longer export member groups to disk UDA entities. By default if this setting is not present its value will automatically be set to true as most sites use Umbraco's built-in membership provider and thus will want the membership groups exported.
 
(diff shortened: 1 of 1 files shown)


Umbraco-Cloud/Deployment/Deploy-Settings/index.md
```diff
@@ -80,3 +80,16 @@ In order for your Cloud project to handle Forms data as content, you'll need to
 :::note
 This only needs to be done if your Cloud project was created **before** June 18th 2019. Make sure to [follow the instructions](../Umbraco-Forms-on-Cloud/#did-you-create-your-cloud-project-before-june-18th-2019) to succesfully change this setting on all environments.
 :::
+
+
+## ExportMemberGroups
+This is available in Deploy 3.4.0+ version and newer. This setting is to be enabled only if you are using an external membership provider for your members and thus it does not make sense to export member groups which would no longer be managed by Umbraco but by an external membership provider.
+
+Setting the `exportMemberGroups` to false will no longer export member groups to disk UDA entities, by default if this setting is not present its value will automatically be set to true as most sites use Umbraco's built in membership provider and thus will want the membership groups exported.
```
sofietoft, line 88:
```suggestion
Setting the `exportMemberGroups` to false will no longer export member groups to disk UDA entities. By default if this setting is not present its value will automatically be set to true as most sites use Umbraco's built-in membership provider and thus will want the membership groups exported.
```

Umbraco-Cloud/Deployment/Deploy-Settings/index.md
```diff
@@ -80,3 +80,16 @@ In order for your Cloud project to handle Forms data as content, you'll need to
 :::note
 This only needs to be done if your Cloud project was created **before** June 18th 2019. Make sure to [follow the instructions](../Umbraco-Forms-on-Cloud/#did-you-create-your-cloud-project-before-june-18th-2019) to succesfully change this setting on all environments.
 :::
+
+
+## ExportMemberGroups
+This is available in Deploy 3.4.0+ version and newer. This setting is to be enabled only if you are using an external membership provider for your members and thus it does not make sense to export member groups which would no longer be managed by Umbraco but by an external membership provider.
```
sofietoft, line 86:
```suggestion
## ExportMemberGroups

This is available in Deploy 3.4.0+ version and newer. 

This setting is to be enabled only if you are using an external membership provider for your members. You will not want to export member groups that would no longer be managed by Umbraco but by an external membership provider.
```

sofietoft (commented):
Thanks for updating the article @warrenbuckley ! 😄 

I've made a few suggestions.

Also, you write in the first part: "This setting is to be enabled only..."
But, as I read it, the setting is enabled - set to `true` - by default. Is that correct?
If that is the case, wouldn't it make more sense to in the first part say "This setting is to be disables only..." instead? 🤔

Umbraco-Cloud/Deployment/Deploy-Settings/index.md
```diff
@@ -80,3 +80,19 @@ In order for your Cloud project to handle Forms data as content, you'll need to
 :::note
 This only needs to be done if your Cloud project was created **before** June 18th 2019. Make sure to [follow the instructions](../Umbraco-Forms-on-Cloud/#did-you-create-your-cloud-project-before-june-18th-2019) to succesfully change this setting on all environments.
 :::
+
+
+## ExportMemberGroups
+
+This is available in Deploy 3.4.0+ version and newer. 
+
+This setting is to be enabled only if you are using an external membership provider for your members. You will not want to export member groups that would no longer be managed by Umbraco but by an external membership provider.
```
warrenbuckley, line 89:
```suggestion
This setting is to be defined and set to false only if you are using an external membership provider for your members. You will not want to export member groups that would no longer be managed by Umbraco but by an external membership provider.
```

Good spot, what I was trying to say was more like if this setting exists or is present, but I can understand the confusion.

Let me know what you think of this @sofietoft

sofietoft: Super! Makes sense 💪 Thanks for updating it!

sofietoft (approved):
All looks good! 👍 

Thanks again for updating the article with the new setting @warrenbuckley ! 🎉

PR #2733, opened 2020-09-29
Update index.md

chris-evansnz:
As per this forum post: https://our.umbraco.com/forum/using-umbraco-and-getting-started/103577-the-cancellationtokensource-has-been-disposed-umbraco-864  

it seems that Umbraco requires the path /umbraco/backoffice/ to be available for certain things to work.

### DIFF
--- a/Reference/Security/Security-hardening/index.md
+++ b/Reference/Security/Security-hardening/index.md
@@ -14,11 +14,13 @@ The prerequisite of this to work is that you’re using [IISRewrite](../../Routi
 
 If you’ve made sure that you’ve installed this on your server we can start locking down our Umbraco-folders. This can be down by following these three steps.
 
-1. We are going to lock down /Umbraco/, but because API-controllers and Surface-controller will use the path /umbraco/api/ and /umbraco/surface/ these will also be locked down. Our first rule in the IISRewrite.config will be used to make sure that these are not locked by IP-address.
+1. We are going to lock down /Umbraco/, but because API-controllers and Surface-controller will use the path /umbraco/api/ and /umbraco/surface/ these will also be locked down. Some internal methods also use the path /umbraco/backoffice/. 
+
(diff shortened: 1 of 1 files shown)


chris-evansnz:
on second review the /backoffice url is presumably meant to be only accessible by approved IP addresses in this case; our issue was also resolved by explicitly whitelisting the external IP of the server to grant it access to these URLs

chris-evansnz:
Should the documentation maybe be updated to warn users of this fact? We only became aware of it after encountering the errors found in the forum post above.

sofietoft:
Thanks for raising the question @chris-evansnz !

Is it correctly understood, that the confusion here was caused by the IP whitelisting for access to `/umbraco/` wasn't mentioned in this particular article?

We have a sample of restricting access to the backoffice using rewrite in the Cloud section (https://our.umbraco.com/documentation/Umbraco-Cloud/Security/#restrict-backoffice-access-using-ip-filtering) - perhaps it would make sense to add something like that in this Security Hardening article as well? 🤔 What do you think?

chris-evansnz:
The issue was down to the fact that the security hardening docs correctly mention you should add your client IPs to the regular expression to allow them to get access to the Umbraco admin area;  but it doesn't mention that some activities in the admin area require the webserver _itself_ to be able to access restricted paths (like /umbraco/backoffice/...) 

Maybe the docs should be updated to point that out? Our example was on an Azure App service, but it may affect other hosting setups as well like VMs, dedicated physical servers, etc.

Jeavon:
@chris-evansnz which version of Umbraco did you find this on? I would like to replicate it if possible. Thanks!

chris-evansnz:
For sure. It was v8.6.4 and we noticed the issue primarily on adding or deleting media - although presumably any other call to a /umbraco/backoffice url would also be blocked.

PR #2740, opened 2020-10-01
Fix link in faqs section to point to the correct one 

arpitpandey0209:


sofietoft (approved):
Looks great @arpitpandey0209 !
Thanks for fixing the link 💪

PR #2745, opened 2020-10-02
changed the default mb mode

RyuLindow:
PureLive is the default. At least in 8.6, 7 and 8


sofietoft (approved):
Good catch @RyuLindow !

Thanks for updating the article. Much appreciated 💪

PR #2748, opened 2020-10-05
Warning message to use Umbraco.Web.Models

beumsk:
I often forget about this. I guess it would help others.


sofietoft:
Thanks for the PR @beumsk !

The `Umbraco.Web.Models` is already added in the code snippet 🤔 
I don't think this should be a warning - perhaps adding is at "regular" text would suffice.
What do you think?

beumsk:
I know but I tend to focus on the core of the code and miss the first line.
I'm not sure regular text will be visible enough.

sofietoft:
Ah, I see 🤔 

Well, let's try to add a warning here then. Would definitely help people remember it!
Thanks again for the PR! Much appreciated.

PR #2749, opened 2020-10-05
Deploying deletes

clausjensen:
Fixed up a few errors. There's some mention of a "database entry referencing a document type" .. the database *entry* is actually what the document type *is*.

Also made it a bit clearer what the consequence of automatically deploying a delete is, to ensure it is clear why we do not want to do this.

Fixed up the description of what to do to completely remove a document type from a project.


sofietoft (approved):
Thanks for the PR @clausjensen ! 🎉 

The changes you've made, makes a lot of sense 💪 🤓 
I'll make sure this is merged right away.

PR #2750, opened 2020-10-06
Update index.md

miguelcrpinto:
fixed a little typo


sofietoft:
Good catch @miguelcrpinto !
Thanks for fixing up the typo 💪

Issue #2747, opened 2020-10-05
Missing tips regading "useLegacyEncoding" in webconfig. 


natanguitz:
Missing documentation when having useLegacyEncoding enabled in Umbraco 7 
https://our.umbraco.com/documentation/Getting-Started/Setup/Upgrading/migrating-to-v8

It looks like my project have "useLegacyEncoding"  enabled. 
After installing the clean Umbraco 8 site i could not login. 

A solution that work for me :  Go back to 7+ site, disable "useLegacyEncoding" and restore the password. 

I could continue with Step 2: Migrating content to Umbraco 8 only after running password restore.

sofietoft:
Thanks for reporting this issue @natanguitz ! Much appreciated.

I think it makes a lot of sense to add a note about the `useLegacyEncoding` to this article.

Do you have any thoughts on where in article. by any chance?
Perhaps somewhere in the first steps? 🤔

natanguitz:
Yes, happy to help :)  
I think another warning before the text "Once it is upgraded and you have verified everything is working, move on to step 2. "  it may be the best so nobody miss it.

nul800sebastiaan:
There is a warning here that tells you to copy the provider configuration over:

![image](https://user-images.githubusercontent.com/304656/95098499-098c9e80-072f-11eb-8329-cfbe5d113bfb.png)

This would include everything you had on the `UsersMembershipProvider` before (including the `useLegacyEncoding` attribute). 

So most of all, I would think that warning needs clarification.

natanguitz:
Yes i guess it would be enough with a clarification , adding the attribute's name and little earlier in the article. 
I think is possible that my case was a unique case where it was not enough to just copy & paste.

sofietoft:
Updating the warning to mention also the `useLegacyEncoding` value sounds like a good plan.

Would you be up for updating the article yourself @natanguitz ? 😄 
We have some docs here, on how you can make the PR: https://our.umbraco.com/documentation/Contribute/

natanguitz:
Yes, I can give it a try :)

natanguitz:
#2753 
I think I did it the correct way :) 
cheers!!

sofietoft:
Brilliant @natanguitz , thanks! 😄

nul800sebastiaan:
Only mentioning `useLegacyEncoding` won't be enough tough, there's many more attributes that need to be copied over. I would update this to read more like: 

[...]

PR #2753, opened 2020-10-06
Update migrating-to-v8.md

natanguitz:
Add warning notes

### DIFF
--- a/Getting-Started/Setup/Upgrading/migrating-to-v8.md
+++ b/Getting-Started/Setup/Upgrading/migrating-to-v8.md
@@ -76,6 +76,12 @@ The updated properties are:
 You can see if your site is using the obsolete properties by the `(Obsolete)` prefix in their name.
 :::
 
+:::warning
+If you have customized your `UsersMembershipProvider` on your Umbraco 7 site you will need to copy that over to the 8.1 web.config as well. Additionally you will need to update the `type` attribute to be `type="Umbraco.Web.Security.Providers.UsersMembershipProvider, Umbraco.Web"`.
+This also includes the attribute `useLegacyEncoding` value. In some cases you need to go back to your **Umbraco 7.14 site** , disable it and run a password restore after that in order to login succeed in your **Umbraco 8.1 site**. 
+
+:::
+
 Once it is upgraded and you have verified everything is working, move on to step 2.
 
 ### Step 2: Migrating content to Umbraco 8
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for adding this information @natanguitz ! 🙌 

I do have one concern though - you've moved the warning up to the section about upgrading to the latest version of Umbraco 7. And therefore also **before** anything is migrated to Umbraco 8 🤔

As I understood the case here, it can be solved by one of two ways:
1. Making sure that the new Umbraco 8 site uses the same `enableLegacyEncoding` value as the Umbraco 7 site or
2. Disabling the value on the Umbraco 7 site and then run a password-recover

And I believe, if you do (1) then you will not need to do (2) 🤔 
Does that sound correct to you?

natanguitz:
I just tried to move it in the order i think it can be better but of course I can be wrong. 

1. In the case that just copy the settings works,  yes, it didnt for me.

sofietoft:
Right. That's a good point 🤔 

Hm. I still think that since it's currently placed **before** an Umbraco 8 site is created... Perhaps move it down here: 
image

Then there's still time to set the setting **before** anything is migrated.
What do you think about that instead?

sofietoft:
Did you see my message here @natanguitz ? 😄 
And also, I noticed Sebastiaan adding some more comments on the Issue: https://github.com/umbraco/UmbracoDocs/issues/2747

natanguitz:
Hello @sofietoft 

I didn't miss it, I just forgot to answer :) 

I made the change you suggested.

Getting-Started/Setup/Upgrading/migrating-to-v8.md
```diff
@@ -84,6 +85,12 @@ First thing to do is spin up a fresh new Umbraco 8.1+ site. Make sure it all wor
 
 ![Fresh 8.1 site](images/fresh-8_1-site.png)
 
+:::warning
+If you have customized your `UsersMembershipProvider` on your Umbraco 7 site you will need to copy that over to the 8.1 web.config as well. Additionally you will need to update the `type` attribute to be `type="Umbraco.Web.Security.Providers.UsersMembershipProvider, Umbraco.Web"`.
+This also includes the attribute `useLegacyEncoding` value. In some cases you need to go back to your **Umbraco 7.14 site** , disable it and run a password restore after that in order to login succeed in your **Umbraco 8.1 site**.
```
sofietoft, line 90:
```suggestion

This also includes the attribute `useLegacyEncoding` value. Make sure that this setting is copied into your new Umbraco 8 site, as it is needed in order for you to be able to log in.
```

sofietoft (commented):
Thanks for moving the note @natanguitz !

I've updated the notes about the `useLegacyEncoding` setting. Instead of adding steps to resolve once it has failed, I think it makes sense to add preventive information - that way people will not have to fix it, as they will not run into it in the first place.

I hope it makes sense! Happy to answer any questions this might evoke.

sofietoft:
Thanks for adding the suggestions @natanguitz !
And once again, thanks a lot for helping add this to the Docs - much appreciated 😄 

I say it's time we merge this in.
And I see that this is the first time you're contributing to the Umbraco Docs - that's awesome!!
Do you have an account on [Our](https://our.umbraco.com/)? If you do, I can grant you a C-Trib badge for your efforts - let me know!

natanguitz:
Yay!   😄 yes  here => https://our.umbraco.com/members/natanguitz/   Thank you !!

PR #2755, opened 2020-10-08
Remove broken styleguide rule

jmayntzhusen:
Just copied the Vale rules into a different repo and noticed this App Plugins rule a bunch of places. The idea to capitalize it is good, but this ruleset is meant more for wrong terms or spaces, hyphens, etc. 

The rule has this line: `ignorecase: true`, which means this rule will always trigger and will never be "correct".

If you want a seperate rule to check for correct casing of terms I can add that, just let me know 🙂

### DIFF
--- a/.github/valeStyle/Terms.yml
+++ b/.github/valeStyle/Terms.yml
@@ -6,7 +6,6 @@ ignorecase: true
 swap:
   back office: "'backoffice'"
   back-office: "'backoffice'" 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for correcting this, @jmayntzhusen !

I had noticed it a few times but never considered the reason.
I do not think it would be necessary to create a new rule for this - but I'll let you know if I change my mind 😉

PR #2764, opened 2020-10-10
Fix broken link pointing to YAML metadata

Bharat123rox:
Fixed broken link pointing to YAML metadata (probably due to folder migration?)

### DIFF
--- a/Contribute/File-Naming-Conventions/index.md
+++ b/Contribute/File-Naming-Conventions/index.md
@@ -30,7 +30,7 @@ For SEO reasons it is not necessary to change a file name when a feature becomes
 
 ## Adding meta data
 
-It is the [YAML meta data](adding-metadata.md) in the document itself, that will be used as the "point of truth" for when a version applies from and to.
+It is the [YAML meta data](../Adding-Metadata/index.md) in the document itself, that will be used as the "point of truth" for when a version applies from and to.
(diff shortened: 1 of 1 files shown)


Contribute/File-Naming-Conventions/index.md
```diff
@@ -30,7 +30,7 @@ For SEO reasons it is not necessary to change a file name when a feature becomes
 
 ## Adding meta data
 
-It is the [YAML meta data](adding-metadata.md) in the document itself, that will be used as the "point of truth" for when a version applies from and to.
+It is the [YAML meta data](../Adding-Metadata/index.md) in the document itself, that will be used as the "point of truth" for when a version applies from and to.
```
sofietoft, line 33:
```suggestion
It is the [YAML meta data](../Adding-Metadata) in the document itself, that will be used as the "point of truth" for when a version applies from and to.
```

sofietoft (changes requested):
Thanks for the PR @Bharat123rox ! 😄 

As we do not prefer to add the filename - `index.md` in this case - to the link, I've made a tiny suggestion to remove it 😄

sofietoft (approved):
Brilliant! Thanks for applying the suggestions @Bharat123rox .

Let's get this merged in then.
Thanks again for helping out 😄

PR #2765, opened 2020-10-10
Fix faulty "center" and use HTML center alignment

Bharat123rox:
Uses HTML center alignment, since the `:::center` attribute is not supported in GitHub markdown.

### DIFF
--- a/Contribute/Code-samples/index.md
+++ b/Contribute/Code-samples/index.md
@@ -102,9 +102,9 @@ Each item is treated as a standard `IPublishedElement entity`, which means you c
 
 The markdown above will output the following:
 
-:::center
-![Example of inline code](images/inline-sample.png)
-:::
+<p align="center">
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the suggestion @Bharat123rox !

[The `:::center` is actually something we've set up](https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/#center-images), in order to be able to center the images on Our, were our Documentation is presented: https://our.umbraco.com/.
You're correct that it does not work here on Github, but as we prefer people reading our Docs on Our, we're gonna stick with this formatting.

I'll have to close this PR.
Hope you understand!

Bharat123rox:
No problem @sofietoft and sorry for the inconvenience. I just wasn't aware that `:::center` worked on the [Our site](https://our.umbraco.com/)

PR #2773, opened 2020-10-14
G maps API key clarification

RyuLindow:


sofietoft (approved):
Makes sense! Thanks for adding this @RyuLindow 💪

PR #2787, opened 2020-10-19
Improve vale: add more condescending words

dampee:
Removed also the link as it is not resolving


sofietoft:
I generally agree with all these words.
Will make sure they are merged when I have time to run through our current docs, so new contributors won't run into these warnings unnecessarily 😁

PR #2788, opened 2020-10-20
Adds documentation for new external login provider functionality and cleans up a bit of the existing docs

Shazwazza:
With 8.9 there's a bunch of new functionality that I've documented. This functionality is also listed here https://github.com/umbraco/Umbraco-CMS/pull/8623#issue-466447539

I think i've got the versioning of these things ok, but please adjust as you see fit.

### DIFF
--- a/Reference/Security/index-v7.md
+++ b/Reference/Security/index-v7.md
@@ -30,6 +30,10 @@ The Umbraco ASP.NET Identity implementation can be extended by using the [Umbrac
 
 The Umbraco backoffice supports external login providers (OAuth) for performing authentication of your users. This could be any OpenIDConnect provider such as Azure Active Directory, Identity Server, Google or Facebook.
 
+### [BackOfficeUserManager](backoffice-user-manager-v7.3.0.md) and Events
+
+The [`BackOfficeUserManager`](backoffice-user-manager.md) is the ASP.NET Identity [UserManager](https://docs.microsoft.com/en-us/previous-versions/aspnet/dn613290(v=vs.108)) implementation in Umbraco. It exposes APIs for working with Umbraco User's via the ASP.NET Identity including password handling.
+
 ### Custom password check
 
 In most cases [External login providers (OAuth)](external-login-providers.md) will meet the needs of most users when needing to authenticate with external resources but in some cases you may need to only change how the username and password credentials are checked.
--- a/Reference/Security/index-v7.md
+++ b/Reference/Security/index-v7.md
@@ -32,7 +32,7 @@ The Umbraco backoffice supports external login providers (OAuth) for performing 
 
 ### [BackOfficeUserManager](backoffice-user-manager-v7.3.0.md) and Events
 
-The [`BackOfficeUserManager`](backoffice-user-manager.md) is the ASP.NET Identity [UserManager](https://docs.microsoft.com/en-us/previous-versions/aspnet/dn613290(v=vs.108)) implementation in Umbraco. It exposes APIs for working with Umbraco User's via the ASP.NET Identity including password handling.
+The [`BackOfficeUserManager`](backoffice-user-manager-v7.3.0.md) is the ASP.NET Identity [UserManager](https://docs.microsoft.com/en-us/previous-versions/aspnet/dn613290(v=vs.108)) implementation in Umbraco. It exposes APIs for working with Umbraco User's via the ASP.NET Identity including password handling.
 
 ### Custom password check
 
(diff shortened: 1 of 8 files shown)


jmayntzhusen:
The thing you are mentioning with the https://our.umbraco.com/Documentation/Reference/Security/ page is because we are doing versioning a little bit differently throughout the documentation currently, there are 2 ways as far as I can see:

1. We have a full document copied and then changed for a new version. So fx a 7.14 document with all the info we have, and then if there is a small addition in 7.15 we'd copy the 7.14 document and add the change in 7.15 as well.
2. We have a major version document with all the info that's valid in the initial release, and then minor version variations added that only show the difference (as is the case with the one you linked).

Don't know if the docs team has a specific strategy going forwards, as having 1 way of doing it would probably be preferred! I am partial to #2 as it feels a lot more useable to not have to check all prior minor versions when looking for info! 
If we move to DocFX v3 in the future this would no longer matter as you'd be able to version specific sections of the same file, so hopefully that project gets finished and released for the public in not too long 🤞

**As for how the versioning actually works**:
To "trigger" the version box on articles you need to have 2 files in the same folder with a naming convention of:
`name.md`
`name-v*.md`

This will cause Our to pick up on the fact that there are multiple versions of the same file, to then determine how the versioning numbers are represented on the box:
image

These values come from the yaml tags in the files. There are 2 version tags:
`versionFrom`
`versionTo`

Most documents don't specify a versionTo only a versionFrom. Say you have 3 articles:

Article1 works for 7.3-7.x
Article2 works for 8.0-8.9
Article3 works for 8.9-8.x

You probably version them like this:
1:
`versionFrom: 7.3.0`
2:
`versionFrom:8.0.0`
`versionTo:8.9.0`
3:
`versionFrom:8.9.0`

That would give a box saying:
```
8.9.0 +
8.0.0 - 8.9.0
7.0.0 +
```

You may also notice the versionbox having a `(current)` at the end of one version sometimes, that is for the file that doesn't have -v* in the filename as we expect the current version to also be the index file as that is where you would be autorouted to.

Reference/Security/index-v7.md
```diff
@@ -20,38 +20,31 @@ Don't forget to [configure your Umbraco when using HTTPS](use-https.md).
...
+This is typically a legacy approach to validating credentials with external resources but it is possible.
 
-You are able [check the username and password against your own credentials store](custom-password-checker-v7.3.0.md) by implementing a new  `IBackOfficeUserPasswordChecker`.
+You are able check the username and password against your own credentials store by implementing a [`IBackOfficeUserPasswordChecker`](custom-password-checker-v7.3.0.md).
 
-### Authenticating with Active Directory credentials
+#### Authenticating with Active Directory credentials
 
-You want to [connect the backoffice to Active Directory](Authenticate-With-AD-vpre8.md)? Should be pretty straight forward with the `ActiveDirectoryBackOfficeUserPasswordChecker`.
+If you are using a network based Azure Directory (not Azure Active Directory), we have setup a guide on how to [connect the backoffice to Active Directory](Authenticate-With-AD-vpre8.md). It should be pretty straight forward with the `ActiveDirectoryBackOfficeUserPasswordChecker`.
```
sofietoft, line 47:
```suggestion
If you are using a network-based Azure Directory (not Azure Active Directory), we have set up a guide on how to [connect the backoffice to Active Directory](Authenticate-With-AD-vpre8.md). It can be done using the  `ActiveDirectoryBackOfficeUserPasswordChecker`.
```

Reference/Security/index.md
```diff
@@ -32,39 +32,33 @@ When your project is hosted on Umbraco Cloud, you might be interested in more de
...
+This is typically a legacy approach to validating credentials with external resources but it is possible.
 
-Read more about [auto linking](auto-linking.md)
+You are able check the username and password against your own credentials store by implementing a [`IBackOfficeUserPasswordChecker`](custom-password-checker.md).
 
-### Custom password check
+#### Authenticating with Active Directory credentials
 
-You are able [check the username and password against your own credentials store](custom-password-checker.md) by implementing a new  `IBackOfficeUserPasswordChecker`.
+If you are using a network based Azure Directory (not Azure Active Directory), we have setup a guide on how to [connect the backoffice to Active Directory](authenticate-with-AD.md). It should be pretty straight forward with the `ActiveDirectoryBackOfficeUserPasswordChecker`.
```
sofietoft, line 59:
```suggestion
If you are using a network based Azure Directory (not Azure Active Directory), we have set up a guide on how to [connect the backoffice to Active Directory](authenticate-with-AD.md). It can be done using the  `ActiveDirectoryBackOfficeUserPasswordChecker`.
```

sofietoft (changes requested):
Thanks for updating all these articles @Shazwazza ! 🙌 

I've made some suggestions here and there.
Mainly warnings on some long sentences and the use of `back office` instead of `backoffice`.
I hope it all makes sense - let me know if it doesn't!

hemraker:
Great update to the docs @Shazwazza 💪 I added a few tweaks in #2796

Shazwazza:
Hi, i won't be able to make further changes to this as I have too much on and leave for holidays tomorrow for 3 weeks so please feel free to make all changes you would like and merge. Cheers!

PR #2790, opened 2020-10-20
Add info on property replacement

jmayntzhusen:
This will be an upcoming feature in Umbpack v0.9.6 from https://github.com/umbraco/UmbPack/pull/38

Would appreciate a review but **don't merge!** If it's approved I can just merge and deploy myself when we release the new version 🙂

### DIFF
--- a/Extending/Packages/UmbPack/index.md
+++ b/Extending/Packages/UmbPack/index.md
@@ -153,6 +153,21 @@ umbpack pack .\package.xml -v 1.9.9
 umbpack pack .\package.xml -n MyPackageWithBadVersioning_FirstVersion.zip
 ```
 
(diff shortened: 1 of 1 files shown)


jmayntzhusen:
Maybe @mattbrailsford has any thoughts on this as he's built the feature 🙂

Extending/Packages/UmbPack/index.md
```diff
@@ -153,6 +153,21 @@ umbpack pack .\package.xml -v 1.9.9
...
+    <package>
+      <name>$Name$</name>
+      <version>$Version$</version>
+```
+
+```
+umbpack pack .\package.xml -p Name=MyPackage;Version=1.0.0
+```
+
+This would pack your package and update your package.xml - note this happens before the automatic naming based on name and version as well. This is especially useful in CI/CD deployments where you may not know all your values before the run!
```
sofietoft, line 169:
```suggestion
This would pack your package and update your package.xml - note this happens before the automatic naming based on name and version as well. This is especially useful in CI/CD deployments where you may not know all your values before the run.
```

sofietoft (commented):
Looks good to me. Except maybe the exclamation point at the end :)

PR #2792, opened 2020-10-20
Updating docs for custom database tables.

clausjensen:
Adding to docs for custom database tables explaining why it is important to not reuse schema classes.
Adding details on handling data in custom database tables.

### DIFF
--- a/Extending/Database/index.md
+++ b/Extending/Database/index.md
@@ -6,7 +6,7 @@ meta.Description: "A guide to creating a custom Database table in Umbraco"
 
 # Creating a custom Database table
 
-In Umbraco it is possible to add custom database tables to your site if you want to store additional data. Below you will find an example that sets up a database table by registering a [component in a composer](../../Implementation/Composing/index.md) and then creating a migration plan and running the plan to add the database table to the database. The end result looks like this:
+In Umbraco it is possible to add custom database tables to your site if you want to store additional data that should not be stored as normal content nodes. Below you will find an example that sets up a database table by registering a [component in a composer](../../Implementation/Composing/index.md) and then creating a migration plan and running the plan to add the database table to the database. The end result looks like this:
 
 ![Database result of a migration](images/db-table.png)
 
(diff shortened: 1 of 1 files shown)


Extending/Database/index.md
```diff
@@ -117,8 +117,20 @@ namespace Umbraco.Web.UI
 
 ## Schema class and migrations
 
-**Important!** It is important to note that the `BlogCommentSchema` class is purely used as a database schema representation and should not be used as a Data Transfer Object (DTO) to access the table data. Equally, you shouldn't use your DTOs to define the schema used by your migration, instead you should create a duplicate snapshot as demonstrated above specifically for the purpose of creating your database table.
+**Important!** It is important to note that the `BlogCommentSchema` class nested inside the migration is purely used as a database schema representation class and should not be used as a Data Transfer Object (DTO) to access the table data. Equally, you shouldn't use your DTO classes to define the schema used by your migration, instead you should create a duplicate snapshot as demonstrated above specifically for the purpose of creating or working with your database tables in the current migration. The name of the class is not important as you will be overriding it using the TableName attribute, so choose a name that makes it clear for you and everyone else that this class is purely for defining the schema in this migration.
```
sofietoft, line 120: Added a few suggestions to shorten the length of the sentences a bit.

```suggestion
**Important!** It is important to note that the `BlogCommentSchema` class nested inside the migration is purely used as a database schema representation class and should not be used as a Data Transfer Object (DTO) to access the table data. Equally, you shouldn't use your DTO classes to define the schema used by your migration. Instead you should create a duplicate snapshot as demonstrated above specifically for the purpose of creating or working with your database tables in the current migration. The name of the class is not important as you will be overriding it using the TableName attribute. So you should choose a name that makes it clear for you and everyone else that this class is purely for defining the schema in this migration.
```

Extending/Database/index.md
```diff
@@ -117,8 +117,20 @@ namespace Umbraco.Web.UI
...
+
+When adding further migrations it is also important to note that if you need to reuse the schema class, it can be a good idea to duplicate this again in those particular migrations. You want the migrations to be immutable, so having separate classes in separate namespaces, reduces the risk of modifying a schema class from your initial migration.
+
+## Data stored in custom database tables
+
+When storing data in custom database tables, this is by default not manageable by Umbraco at all. This can be great for many purposes such as storing massive amounts of data that you do not need to edit from inside the Umbraco backoffice. Decoupling part of your data from being managed by Umbraco as content can be a way of achieving better performance for your site, as it will no longer take up space in indexes, caches and the Umbraco database which may not have the best structure for your type of data.
```
sofietoft, line 130:
```suggestion
When storing data in custom database tables, this is by default not manageable by Umbraco at all. This can be great for many purposes such as storing massive amounts of data that you do not need to edit from inside the Umbraco backoffice. Decoupling part of your data from being managed by Umbraco as content can be a way of achieving better performance for your site. That way, it will no longer take up space in indexes and caches, and the Umbraco database which may not have the best structure for your type of data.
```

Extending/Database/index.md
```diff
@@ -117,8 +117,20 @@ namespace Umbraco.Web.UI
...
+
+This however also means that if you actually do need to edit or display this data, it is up to you to implement the underlying functionality supporting this.
```
sofietoft, line 132:
```suggestion
This however also means that if you do need to edit or display this data, it is up to you to implement the underlying functionality supporting this.
```

sofietoft (changes requested):
Great addition to the article @clausjensen !
I've made a few suggestions - mainly due to long sentences.
Let me know if something doesn't add up!

clausjensen:
@sofietoft all of it looks good :) Thanks!

sofietoft (approved):
Brilliant! Thanks for applying the suggestions 😄 

I'll make sure this is merged in asap.

PR #2803, opened 2020-10-23
fixed typo 'the the'

prjseal:


sofietoft:
Another good catch! Thanks @prjseal ! 😄

PR #2807, opened 2020-10-23
fixed typo 'can can'

prjseal:


sofietoft:
I can't believe we had so many word-duplications in the docs 😅 
Thanks for the help getting rid of them @prjseal ! Much appreciated!

PR #2808, opened 2020-10-23
Duplicated word removal

NikRimington:
Removed duplicated "can can".


sofietoft:
Looks like someone actually beat you do it, and has already removed the duplicate word 😅 

The PR is still mergable though, so I'll go ahead and do that.
Thanks for helping out with the docs 😄

PR #2811, opened 2020-10-23
Updated examples for eventsService

abjerner:
The existing documentation only explained how to listen for events from a controller.

I've now added an example for how to listen for events on a more global level - eg. if you need to listen for an event during the entire lifetime of the application, and not just during the lifetime of a component.


sofietoft (commented):
Looks great @abjerner !

How about adding a heading to this new section?
You'd know better than me, what the heading should be 😄 - but I think it would make sense to have one.
What do you think?

abjerner:
@sofietoft I'm not entirely sure a heading does that much of a difference, but I've now added a H4 called **Listening for events globally** 😉

sofietoft:
Great, thanks! 😄 

Perhaps it doesn't do much, but it still helps give the article a little more structure, so people can quickly get an overview of what they can find in it 😄 
I'll merge this PR now. Thanks again for adding it! 👏

PR #2813, opened 2020-10-26
Added detail to how different "Lifetimes" work

tristanjthompson:
I recently used `Lifetime.Request`, assuming that this is would create a new instance per web-request.  I remember looking at the docs and having to make a guess, and went with `Lifetime.Request` because it sounded the most similar to "web request".  It turned out I actually should have used `Lifetime.Scope`.

This is because the terminology is based on Lightinject (which uses "Request" and "Scope" lifetimes), which is confusing in the context of a web application - this was [highlighted and clarified with Lightinject](https://github.com/seesharper/LightInject/issues/494) by @Shazwazza last year.  It looks like other people have also made the same mistake as me judging by [a forum post I came across](https://our.umbraco.com/forum/umbraco-8/98389-iusercomposer-with-lifetimerequest-works-as-lifetimetransient) which helped me solve my issue.

This PR is to update the IoC docs to add more information around the differences between the different lifetimes for those unfamiliar with the nuances in the terminology (and also for "future me" when I come to look into this again next time I'm setting up the DI for a new project!).

### DIFF
--- a/Implementation/Services/index.md
+++ b/Implementation/Services/index.md
@@ -346,28 +346,30 @@ namespace Umbraco8.Composers
     {
         public void Compose(Composition composition)
         {
-            // if your service makes use of the current UmbracoContext, eg AssignedContentItem - register the service in Request scope
+            // if your service makes use of the current UmbracoContext, eg AssignedContentItem - register the service using the 'Request' or 'Scope' lifetimes
             // composition.Register<ISiteService, SiteService>(Lifetime.Request);
-            // if not then it is better to register in 'Singleton' Scope
+            // if not then it is better to register using the 'Singleton' lifetime
             composition.Register<ISiteService, SiteService>(Lifetime.Singleton);
         }
     }
 }
 ```
 
-#### Lifespans
+#### Lifetimes
 
 **"Transient"** services can be injected into "Transient" and below ⤵. (i.e. "Transient" services can be injected anywhere)
 
 - "Transient" means that anytime this type is needed a brand new instance of this type will be created.
 
+**"Request"** services can be injected into "Request"/"Scope" based lifetimes only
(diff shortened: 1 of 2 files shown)


sofietoft (commented):
Thanks a lot for adding these clarifications to the article @tristanjthompson ! 😄 
Sure sounds like it can help a lot of people.

However, as it's not really my expertise, I'll send for some help to get a proper review 💪

marcemarc:
Hi @tristanjthompson 

Thanks for updating the docs with this clarification, yes agree it's really confusing, the scent of the word 'request' is powerful here!

Do you think we should make a similar clarification on this page?

https://our.umbraco.com/Documentation/Implementation/Services/#lifespans

regards

Marc

tristanjthompson:
Thanks for looking at this @sofietoft & @marcemarc!

Ooh, interesting, I hadn't come across that page - makes sense to update it there too as it looks like it only mentions `Request` lifetimes and not `Scope` lifetimes.

I'll try to find some time in the next week to update that too and add it to this PR 👍

tristanjthompson:
Hi both - apologies it took a little longer to do this than I'd hoped.  I've updated the page that @marcemarc suggested with a few details of `Scope` lifetimes and updated a few things around it for consistency (e.g. using "Lifetime" instead of "Lifespan" consistently, etc).

Let me know if that all makes sense or if there are any other udpates you'd like me to make!

marcemarc:
@tristanjthompson brilliant, thanks for taking the time to update this, I think it's much clearer now, and I'm pretty sure I've fallen foul of the Scope/Request subtle difference too...

regards

Marc

tristanjthompson:
Glad I could help - thanks for checking over it!

PR #2815, opened 2020-10-26
Added notes about culture and hostnames to variants 

poornimanayar:
The documentation for variants covers culture and hostnames in the video but not in the notes. I have added that.


sofietoft:
Thanks for adding this PR @poornimanayar !
I agree that this should be covered in the written docs as well.

I'm thinking though, that some extra context needs to be added.
Like, perhaps give an example of a URL? Could be something like `https://yourwebsite.com/dk`

What do you think?

poornimanayar:
Yes, Some more context needs to be given. I totally agree to that. Happy to give this suggestion a thought. I am reviewing a PR on the CMS now, will come back to this shortly :-)

sofietoft (approved):
Thanks for adding a bit more context to this part @poornimanayar ! 💪 

I'll make sure it's merged then 😄

PR #2816, opened 2020-10-26
Upate emailaddress datatype

erikjanwestendorp:
Belongs to https://github.com/umbraco/UmbracoDocs/issues/1528 

- Fix image path
- Add ModelsBuilde example.


sofietoft:
Thanks for updating the article @erikjanwestendorp !
Great to have an example in there with the use of ModelsBuilder 💪

PR #2822, opened 2020-10-28
Proposed Improvements for Document Types page

mattou07:
Grammatical and highlighting to Document Types page https://our.umbraco.com/documentation/Tutorials/Creating-Basic-Site/Document-Types/

### DIFF
--- a/Tutorials/Creating-Basic-Site/Document-Types/index.md
+++ b/Tutorials/Creating-Basic-Site/Document-Types/index.md
@@ -6,7 +6,7 @@ versionFrom: 8.0.0
 ## Data first
 ### nothing in = nothing out!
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for adding these updates @mattou07 !

I'm not entirely sure about adding all the extra bold text 🤔 
I believe that the idea with the [ ] is to highlight buttons or text where you need to click when following the guide. And the text in bold are the names of types and sections in the backoffice.

So, I think either all the mentions of buttons that are in [ ] should be bold, or none of them 🤔 

What do you think?

mattou07:
@sofietoft I made the text bold so its easier to read/identify and skim over. Happy to remove the bold text on the buttons.

mattou07:
@sofietoft Sorry, are you saying I missed some buttons in my PR? Or are you saying all pages in the docs need to bold the buttons?

sofietoft:
Yes, sorry! I think what confused me, when I reviewed this earlier, was the fact that not all text in [ ] was made bold 🤔 
Just wondering if there was a reason for that, or if it was a simple oversight?

Also, I do agree with you, that using bold, does make it easier to skim the important parts!

mattou07:
@sofietoft this is now fixed in my latest commit.

sofietoft:
Thanks @mattou07 !

Looks great! I'll merge it in.

PR #2782, opened 2020-10-17
Added documentation for the changeTitle

RachBreeze:
Hello,

Thank you for responding to my question here https://github.com/umbraco/UmbracoDocs/issues/2770 
I've added missing documentation for the changeTitle directive.  This is my first PR to the docs, I am really keen to get feedback to ensure I've followed the preferred structure and documented the directive correctly.

Thank you in advance for your time

(Created on CanConHack Day)

PR #2830, opened 2020-10-29
Move changeTitle documentation

RachBreeze:
This PR updates the change documentation as per the feedback this pr https://github.com/umbraco/UmbracoDocs/pull/2782 :-)


sofietoft (commented):
Looks great @RachBreeze !

This is what you were suggesting, @abjerner - right? 😁

abjerner:
Yes, something like that 👍 

But if I have to be real picky, the part in the overall page about the event service is now located under the H3 for **Directives**. The event is emitted from a few directives, but also from `init.js`, so perhaps we can create a new H3 at the bottom of the page called **Other**?

RachBreeze:
Hi @abjerner  and @sofietoft this isn't a problem, I can move it to "other".

sofietoft:
Apologies for the delay on this one!
Thanks for the feedback @abjerner , and thanks @RachBreeze for adding this to the documentation.

Looks like it's all fallen into place, so I'll go ahead and merge it in 😄 Thanks again!

PR #2839, opened 2020-10-31
Change meta Umbraco v of the permissions

Rockerby:
There are currently two sets of permissions, https://github.com/umbraco/UmbracoDocs/blob/main/Getting-Started/Setup/Server-Setup/permissions-v7.md and this one which has the v7 folders removed.


sofietoft:
Ah! Good catch @Rockerby ! 😄

PR #2847, opened 2020-11-02
Update Label Property Editor

erikjanwestendorp:
Add ModelsBuilder example.

Belongs to:
https://github.com/umbraco/UmbracoDocs/issues/1528


sofietoft (approved):
Great work @erikjanwestendorp ! Thanks for adding this.
It's super to have more than one example on these articles 💪 

I'll it right in!

PR #2849, opened 2020-11-02
Fix media link case

kmeilander:
MediaLink doesn't exist on the object.  it should be mediaLink (lower m).


sofietoft (approved):
Good catch @kmeilander ! 💪 
Thanks for amending the code sample - much appreciated.

PR #2851, opened 2020-11-03
Fix deeplink

jmayntzhusen:
This link didn't work as it tried to find this v7 headline on the v8 document. This change should do it 🙂


sofietoft:
Good catch @jmayntzhusen !

I'll merge in right away, so we can test whether it works.
Thanks for updating the link 💪

sofietoft:
Seems to be working 💪 Thanks again!

PR #2855, opened 2020-11-05
Update index.md

jesperlange:
A Customer reached out and said his settings was missing the "Contact Form Recaptcha" as shown on the image, so I thought a friendly note could clear it up


sofietoft:
Thanks for adding the PR @jesperlange !

How about adding the note under the image instead, and then use the formatting that we have, to create a "green" note instead - see here, how that works: [Markdown conventions](https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/#adding-notes-warnings-tips).

Oh, and instead of mentioning that the note is friendly - wouldn't it be enough to make sure that the note itself is actually friendly, so it's not necessary to say that it is? 😅 

Let me know what you think!

sofietoft:
Looks great @jesperlange ! 💪 
Great update to the article.

I'm gonna merge right away 🎉

PR #2859, opened 2020-11-06
Update Radio Button List Property Editor

erikjanwestendorp:
Delete hardcoded string form modelsbuilder example. 

Belongs to:
https://github.com/umbraco/UmbracoDocs/issues/1528


sofietoft (approved):
Good catch @erikjanwestendorp !

Thanks for help keeping these examples consistent 💪

PR #2863, opened 2020-11-06
Update wrong Editor Alias for Umbraco.Integer

tkjaergaard:
Sees there's the Editor Alias of the Numeric field is wrong. It should be `Umbraco.Integer` and not `Umbraco.Numeric`.

https://our.umbraco.com/documentation/getting-started/backoffice/property-editors/built-in-property-editors/Numeric/


sofietoft:
Thanks for pointing this out @tkjaergaard !

I'll have our Heartcore experts give it a quick look 💪

sofietoft:
Looks like we're all good here 😄 
I'll merge it right away.

Thanks again @tkjaergaard , for correcting the article. Much appreciated.

PR #2864, opened 2020-11-08
Update index.md

enkelmedia:
Having self closing element did not work, the header was not closed so updated source here.


sofietoft:
Thanks for the PR @enkelmedia !

I could have sworn this worked perfectly the last time I tested this guide 😅 
But great to get it amended if that is no longer the case! 🙏

enkelmedia:
This was on 8.8.1, maybe something changed with the directive since it was written =D

PR #2871, opened 2020-11-11
More information regarding Ignorer user start node

nielslyngsoe:


sofietoft (commented):
Thanks for the PR @nielslyngsoe .

I'm not entirely sure I understand what this checkbox does.
Is the following correct:
When "Ignore user start nodes" is enabled, a user who doesn't normally have access to the media selected as "Start Node" (/Design in this case), can access the media when using this particular Media Picker?

nielslyngsoe:
@sofietoft exactly. I updated this cause I didn't know for sure myself, so I went ahead and tested it and thought it would be good to update documentation.

You might want to adjust the formulation so it's even more clear :-)

sofietoft:
Thanks for clarifying @nielslyngsoe !

I've added an extra setting in there, to make it even more clear.
I'll get it merged in 💪

Issue #2829, opened 2020-10-29
Storing form data in Azure blob storage


mattou07:
## What type of issue is it? (Choose one - delete the others)
Missing documentation

## What article/section is this about?
https://our.umbraco.com/Documentation/Extending/FileSystemProviders/Azure-Blob-Storage/index-v7

## Describe the issue
This page should include information on how to pull forms data from Azure Blob Storage instead of just media. 

If its documented somewhere else feel free to close this issue.

Happy to take this on myself.

sofietoft:
Thanks for creating the issue here @mattou07 !

Just to be sure I understand - you think we should also have an article on how to work with forms data on Azure blob, as currently the articles we have, are primarily media files and the image processor.

Is that correct? 😄

mattou07:
@sofietoft I was thinking of having it on the same page as an added bonus to using Azure Filesystem providers. If you use blob storage container for the Umbraco Forms data folder. Which would hold forms, templates and workflows. It allows you to use features like Azure Web App slot swaps without you having to manually copy the forms data folder to the other slot as they all use the same container.

sofietoft:
Thanks for explaining and sharing your thoughts on this @mattou07 !

So, is the setup for adding Forms to Axure Blob fairly the same as with media?
If so, then I think it would be perfectly fine to add a section to the existing article... If it's different - then I think another article would be the way to go. Not everyone uses Umbraco Forms, so it might not make sense for all that it's mentioned in this article 🤔

sofietoft:
[PR related to this](https://github.com/umbraco/UmbracoDocs/pull/2875) has been merged, so I'll close this issue as well.

PR #2875, opened 2020-11-12
How to store forms data in Azure Storage

mattou07:
Updated "Setup Your Site to use Azure Blob Storage for Media and Image Processor Cache" showing how to store forms in Azure blob storage


mattou07:
Related to issue I raised here: https://github.com/umbraco/UmbracoDocs/issues/2829

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
```diff
@@ -2,19 +2,19 @@
...
 
-For Umbraco sites there are some scenarios when you may want, or need, to consider using Azure Blob Storage for your media.  Particularly if your site contains large amounts of media.  Having your site’s media in Azure Storage can also help your deployments complete more quickly and has the potential to positively affect site performance as the Image Processor cache is moved to Azure Blob Storage.  It also allows you to serve your media from the Azure CDN.
+For Umbraco sites there are some scenarios when you may want, or need, to consider using Azure Blob Storage for your media and forms.  Particularly if your site contains large amounts of media.  Having your site’s media in Azure Storage can also help your deployments complete more quickly and has the potential to positively affect site performance as the Image Processor cache is moved to Azure Blob Storage.  It also allows you to serve your media from the Azure CDN. Moving your forms data to Azure Storage allows you to have multiple environments share the same forms data folder. This particularly useful when using Web App Slots in Azure App Service.
```
sofietoft, line 7:
```suggestion
For Umbraco sites there are some scenarios when you may want or need, to consider using Azure Blob Storage for your media and forms.  Particularly if your site contains large amounts of media.  Having your site’s media in Azure Storage can also help your deployments complete more quickly and have the potential to positively affect site performance as the Image Processor cache is moved to Azure Blob Storage.  It also allows you to serve your media from the Azure CDN. Moving your forms data to Azure Blob Storage allows you to have multiple environments share the same forms data folder. This is particularly useful when using Web App Slots in Azure App Service.
```

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
```diff
@@ -68,10 +70,44 @@ If you are using IISExpress (as with Visual Studio) you’ll need to add a stati
     </location>
   </configuration>
 ```
+### Forms
+To clarify we are not storing form submissions into Azure storage but the JSON files used to map the layout of the forms such as the templates used and workflows. Forms submissions are stored in the database.
```
sofietoft, line 74:
```suggestion
We are not storing form submissions into Azure Blob Storage but the JSON files used to map the layout of the forms such as the templates used and workflows. Forms submissions are stored in the database.
```

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
```diff
@@ -68,10 +70,44 @@ If you are using IISExpress (as with Visual Studio) you’ll need to add a stati
...
+
+To configure Umbraco forms data its very similar to how media is setup above. You just need to provide another provider as a sibling to the media provider. In the snippet below I have added a provider with the alias of forms along side with the media provider.
```
sofietoft, line 76:
```suggestion
Configuring Umbraco Forms data is very similar to how media is setup above. You need to provide another provider as a sibling to the media provider. In the snippet below I have added a provider with the alias of "forms" alongside the media provider.
```

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
```diff
@@ -68,10 +70,44 @@ If you are using IISExpress (as with Visual Studio) you’ll need to add a stati
...
+
+```xml
```
sofietoft, line 78: I am not entirely sure I understand this code snippet here.
There is ONLY a provider for the forms here, right?
Looks like the media provider is closed in the same line as it's started.

I might be wrong though 🤔
Perhaps you could help me understand exactly what you're doing here?

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
```diff
@@ -68,10 +70,44 @@ If you are using IISExpress (as with Visual Studio) you’ll need to add a stati
...
+    <add key="maxDays" value="365"/>
+    <!--
+        When true this allows the VirtualPathProvider to use the default "media" route prefix regardless
+        of the container name.
+    -->
+    <add key="useDefaultRoute" value="true"/>
+    <!--
+        When true blob containers will be private instead of public what means that you can't access the original blob file directly from its blob url.
+    -->
+    <add key="usePrivateContainer" value="true"/>
```
sofietoft, line 102: So, setting this to `true` means that the folder can't be accessed? Is that correct?

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
```diff
@@ -147,6 +183,10 @@ Any media files you already have on your site will not automatically be added to
 
 Any new media files you upload to the site, will automatically be added to the Blob Storage.
 
+## Exisiting Forms files
+
+Any Umbraco forms data files will not automatically be added to blob storage. You will need to upload the **contents** of the Umbraco forms data folder to storage. This can be found in `~/App_Plugins/UmbracoForms/Data`. Again you need the **contents** of the data folder so the **forms**, **templates** and **workflows** folders should be the only folders in the container. Do not upload the whole data folder!
```
sofietoft, line 188:
```suggestion
Any existing Umbraco Forms data files will not automatically be added to the blob storage. You will need to upload the **contents** of the Umbraco Forms data folder to the storage. This can be found in `~/App_Plugins/UmbracoForms/Data`. Again you need the **contents** of the Data folder so the **forms**, **templates** and **workflows** folders should be the only folders in the container. Do not upload the whole Data folder. 
```

sofietoft (changes requested):
Thanks for updating this article @mattou07 !
I've made some suggestions and added some comments here and there.

Also, I notice that you've updated the Umbraco 7 version of this article, and not for the latest version.
Does that mean, that this does not work on Umbraco 8? 🤔

mattou07:
> Thanks for updating this article @mattou07 !
> I've made some suggestions and added some comments here and there.
> 
> Also, I notice that you've updated the Umbraco 7 version of this article, and not for the latest version.
> Does that mean, that this does not work on Umbraco 8? 🤔

I haven't worked with Umbraco 8 enough to be sure. A lot of the time I am dealing with sites running 7.

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
mattou07: To keep the snippet short I removed all the xml code inside the media provider to just show the bare bones of it along side the forms provider.

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
mattou07: Yes forms data shouldn't be publicly accessible unlike media. By setting this to true we confirm that we are using private blob containers.

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
mattou07: "the blob storage" and "the storage" makes sense but sounds wrong. I guess changing it to be this would be better: "automatically be added to Azure Blob storage" "Umbraco Forms data folder to Blob storage". 

I am not sure how to change your suggestion. "Add suggestion to batch" is greyed out for me

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
sofietoft: Alright! Makes sense 💪 
Perhaps that could be explained in [a note](https://our.umbraco.com/documentation/Contribute/Markdown-Conventions/#adding-notes-warnings-tips) before or after the snippet?
What do you think?

Extending/FileSystemProviders/Azure-Blob-Storage/index-v7.md
sofietoft: I see your point 🤔
I've tried to update the wording a bit..
Instead of saying "the Blob storage", I've left it out all together, so it now says: "You will need to upload the contents of the Umbraco Forms data folder manually."

I think that works better. What do you think?

sofietoft:
Thanks for clarifying about the versioning.
Would be able to test this on a Umbraco 8 set up? I think it would be great to have it tested and updated for the latest version as well.

sofietoft:
Thanks for adding the suggestions @mattou07 !

I was talking with the rest of the documentation team about this yesterday, at it turns out we already have a guide for storing Forms data in Azure Blob Storage. It's hidden away in the Forms documentation: https://our.umbraco.com/Documentation/Add-ons/UmbracoForms/Developer/IFileSystem/ 
Sorry for not being aware of this before!!

I think what you've made here is great! Perhaps what we could do, is to link between the two?
Then there's a Forms specific one, and also one for both 🤔 
What do you think?

mattou07 (commented):
> Thanks for adding the suggestions @mattou07 !
> 
> I was talking with the rest of the documentation team about this yesterday, at it turns out we already have a guide for storing Forms data in Azure Blob Storage. It's hidden away in the Forms documentation: https://our.umbraco.com/Documentation/Add-ons/UmbracoForms/Developer/IFileSystem/
> Sorry for not being aware of this before!!
> 
> I think what you've made here is great! Perhaps what we could do, is to link between the two?
> Then there's a Forms specific one, and also one for both 🤔
> What do you think?

Sure I will see what I can do thanks!

sofietoft:
Hi @mattou07 !

I went ahead and added a link in this article to the article in the Forms section. Hope that's OK.
I'll also go ahead and merge in this PR, så people can start benefitting from the guide 💪 

Thanks again for updating the article! Much appreciated.

sofietoft (approved):
Approving PR.

PR #2881, opened 2020-11-16
Update index-vpost8.6.0.md

HalldorLyngmo:
Saw a user try to ping the keepalive URL from an external source. Adding some details about the OnlyLocalRequests attribute to clarify that this is not possible. :) 

Discussion and a link to the OnlyLocalRequests attribute
- https://github.com/umbraco/Umbraco-CMS/pull/7164
- https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/WebApi/Filters/OnlyLocalRequestsAttribute.cs

### DIFF
--- a/Reference/Config/umbracoSettings/index-vpost8.6.0.md
+++ b/Reference/Config/umbracoSettings/index-vpost8.6.0.md
@@ -16,3 +16,8 @@ Here's the default configuration of the `keepAlive` node:
 ```xml
 <keepAlive disableKeepAliveTask="false" keepAlivePingUrl="{umbracoApplicationUrl}/api/keepalive/ping" />
 ```
+
+:::note
+There is an `[OnlyLocalRequests]` attribute which verifies if the request to the KeepAliveController.Ping action came from the local machine.
+If the request is made from an external source, Umbraco will return a HTTP 404 response.
+:::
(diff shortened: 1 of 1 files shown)


Reference/Config/umbracoSettings/index-vpost8.6.0.md
```diff
@@ -16,3 +16,8 @@ Here's the default configuration of the `keepAlive` node:
 ```xml
 <keepAlive disableKeepAliveTask="false" keepAlivePingUrl="{umbracoApplicationUrl}/api/keepalive/ping" />
 ```
+
+:::note
+There is an `[OnlyLocalRequests]` attribute which verifies if the request to the KeepAliveController.Ping action came from the local machine.
+If the request is made from an external source, Umbraco will return a HTTP 404 response.
```
sofietoft, line 22:
```suggestion
There is an `[OnlyLocalRequests]` attribute which verifies if the request to the `KeepAliveController.Ping` action came from the local machine.

If the request is made from an external source, Umbraco will return an HTTP 404 response.
```

sofietoft (changes requested):
Looks good @HalldorLyngmo !

I've added a suggestion - mind taking a quick look?

sofietoft (approved):
Thanks for applying the suggestions 😄 

I'll merge it in right away.

PR #2886, opened 2020-11-18
Added missing bracket and semicolon

FransdeJong:
Added a missing bracket and semicolon in the basic controller example


dampee:
Good catch Frans!  Thanks for making the effort to do the PR!

Issue #2714, opened 2020-09-18
Code samples: Creating property editors programmatically


sofietoft:
In the Getting Started/Backoffice section of the docs, we have articles for most of the property editors in Umbraco. This documentation currently includes details on the implementation part.

We would like to have examples of **how to create each of the property editors programmatically** as well.
Some are simple and straightforward; others might be a bit more complicated.

[Link to examples will be posted as soon as possible.]

Property Editors that need code samples:
----
- [ ] Checkbox (Umbraco.TrueFalse)
- [ ] Checkbox list (Umbraco.CheckBoxList)
- [ ] Color Picker (Umbraco.ColorPicker)
- [ ] Content Picker (Umbraco.ContentPicker)
- [ ] Date (Umbraco.Date)
- [ ] Date/Time (Umbraco.DateTime)
- [ ] Decimal (Umbraco.Decimal)
- [ ] Dropdown (Umbraco.DropDown.Flexible)
- [ ] Email address (Umbraco.EmailAddress)
- [ ] File upload (Umbraco.UploadField)
- [ ] Grid layout (Umbraco.Grid)
- [ ] Image Cropper (Umbraco.ImageCropper)
- [ ] Label (Umbraco.NoEdit)
- [ ] List view (Umbraco.ListView)
- [ ] Markdown editor (Umbraco.MarkdownEditor)
- [ ] Media Picker (Umbraco.MediaPicker)
- [ ] Member Group Picker (Umbraco.MemberGroupPicker)
- [ ] Member Picker (Umbraco.MemberPicker)
- [ ] Multi Url Picker (Umbraco.MultiUrlPicker)
- [ ] Multinode Treepicker (Umbraco.MultiNodeTreePicker)
- [ ] Nested Content (Umbraco.NestedContent)
- [ ] Numeric (Umbraco.Integer)
- [ ] Radio button list (Umbraco.RadioButtonList)
- [ ] Repeatable textstrings (Umbraco.MultipleTextstring)
- [ ] Rich Text Editor (Umbraco.TinyMCEv3)
- [ ] Slider (Umbraco.Slider)
- [ ] Tags (Umbraco.Tags)
- [ ] Textarea (Umbraco.TextArea)
- [ ] Textbox (Umbraco.TextBox)
- [ ] User picker (Umbraco.UserPicker)

sahil9001:
[...]

PR #2895, opened 2020-11-20
Update Textbox

erikjanwestendorp:
Add example how to add a value
Belongs to: #2714

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textbox/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textbox/index.md
@@ -56,20 +56,26 @@ Textbox is an HTML input control for text. It can be configured to have a fixed 
 }
 ```
 
-## Add value
+## Add values programmatically
 
 See the example below to see how a value can be added or changed programmatically. To update a value of a property editor you need the [Content Service](../../../../../Reference/Management/Services/ContentService/index.md).
 
 ```csharp
 @{
+    // Get access to ContentService
     var contentService = Services.ContentService;
 
+    // Create a variable for the GUID of the page you want to update
     var guid = new Guid("32e60db4-1283-4caa-9645-f2153f9888ef");
 
+    // Get the page using the GUID you've defined
     var content = contentService.GetById(guid); // ID of your page
+	
+	// Set the value of the property with alias 'pageTitle'
     content.SetValue("pageTitle", "Umbraco Demo");
 	
-    contentService.SaveAndPublish(content);
+	// Save the change
+    contentService.Save(content);	
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for another one of these @erikjanwestendorp ! 😄 

Now, I hate to do this, as we really do appreciate all the work you're putting into our docs!
But, would you mind holding off with these "adding a value programmatically" samples for a bit? 🙈 
There is some confusion in the documentation team, as to how exactly we want these samples to look and what we want them to show... There have been talks about not using GUID/IDs at all, we've also discussed whether we also need to add more validations, and so forth!

So! I do not want you to waste your time writing all these great samples if we at one point then decide we need to add more to them or to change them up completely.

I hope you understand! And please do let us know if you have any more ideas for how we can make these samples even more "real life"!

sofietoft:
OK! I've made a decision - so sorry for the confusion 😅 

I've also myself been playing around with adding values programmatically to various properties - quite fun actually!
Anywho! Here's what I think we should do:

The samples you've made a really great! They're simple and easy to understand 👏 There are a few things we can add, to make them even more friendly:

- Let's update the heading of the section from "Add value" to "Add values programmatically" - it's a tat more explanatory.
- Also, let's use the "Save" method instead of "SaveAndPublish" - a point was made, that sometimes you might want to "just" save first, and then set up some validation before actually publishing the content!
- Finally - I really like comments in code! - should we add some more of these to make it simpler to understand for new users?

Here's a sample of what I've done with the Textbox property that I've played around with:
```
@{
    // Get access to ContentService
    var contentService = Services.ContentService;

    // Create a variable for the GUID of the page you want to update
    var guid = new Guid("796a8d5c-b7bb-46d9-bc57-ab834d0d1248");

    // Get the page using the GUID you've defined
    var content = contentService.GetById(guid);
    // Set the value of the property with alias 'pageTitle'
    content.SetValue("pageTitle", "text for the text box/string.");

    // Save the change
    contentService.Save(content);
}
```

What do you think?

erikjanwestendorp:
@sofietoft I get your point about using hard coded GUID/Ids, I hate that too. But on the other hand if you don't want to use them you need to develop a whole scenario for each property editor. For now I update the PR, add some code comments, use Save instead of SaveAndPublish and change the title. If you have any updates about the way you would like to see the samples please let me know 😄

sofietoft:
Thanks for updating the article @erikjanwestendorp 💪 

I agree. Not using GUID/ID would require a bit more complicated sample - also why I think it's OK for now, to go with these more static values.
It might make sense at some point, to have a complete sample or guide somewhere, with a full sample of a real-life scenario.
But that's for another day 😄 

I think we've found a great format for these snippets for now!
Thanks so much for helping out with these 💪 
I'll get it merged in!

PR #2898, opened 2020-11-23
Changing trailing slash redirect to a better approach

hjaltedaniel-umbraco:
With the current version, if you have a content node called eg. "Mediacenter" it is not gonna work.
With this correction all urls with a ".something" will be negated.

### DIFF
--- a/Umbraco-Cloud/Set-Up/Manage-Hostnames/Rewrites-on-Cloud/index.md
+++ b/Umbraco-Cloud/Set-Up/Manage-Hostnames/Rewrites-on-Cloud/index.md
@@ -91,7 +91,7 @@ For example, the following rule will redirect all requests for `https://mysite.c
   <conditions>
     <add input="{REQUEST_FILENAME}" matchType="IsDirectory" negate="true" />
     <add input="{REQUEST_FILENAME}" matchType="IsFile" negate="true" />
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @hjaltedaniel-umbraco !

If I understand correctly, the change you've made here, now negates anything after Umbraco.com as long as it starts with a dot.
Is that correct?
And how does that help with the Azure Blob Storage situation?

hjaltedaniel-umbraco:
It negates everything that has a file extension (.jpg, .svg, .png).
It helps the blob storage situation as it then does not add the trailing slash, and the file storage provider is able to find the file in the blob container.

sofietoft:
Ah, alright. Thanks for explaining.

What will this mean for .css and .js files?
I'm assuming it just means that those files will no longer have a trailing slash - which should be OK, right?

hjaltedaniel-umbraco:
They will be affected by this negate:
`<add input="{REQUEST_FILENAME}" matchType="IsFile" negate="true" />`
So it is only the media files from blob that will be affected by the negate, as they're not considered files by the RewriteModule

sofietoft:
Makes sense! Thanks for explaining 💪 

I'll merge these updates in.

PR #2902, opened 2020-11-24
Update True/False

erikjanwestendorp:
Update title, add code comments, use Save instead of SaveAndPublish and add a ModelsBuilder Example


sofietoft (approved):
Looks great @erikjanwestendorp ! 😄 
Thanks for updating this so it matches the rest of the samples. Much appreciated.

I'll merge it right in!

PR #2905, opened 2020-11-26
Properly closing comment in tinymce table config example

lars-erik:


sofietoft:
Good catch @lars-erik ! Thanks for fixing it 😄 

I'll merge it right in!

PR #2907, opened 2020-11-27
Update UserPicker PropertyEditor

erikjanwestendorp:
Add a value programmatically

Belongs to: #2714

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/User-Picker/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/User-Picker/index.md
@@ -68,8 +68,8 @@ See the example below to see how a value can be added or changed programmaticall
 	// Get the page using the GUID you've defined
     var content = contentService.GetById(guid); // ID of your page
 	
-	// Set the value of the property with alias 'userPicker'
-    content.SetValue("userPicker", 1);
+	// Set the value of the property with alias 'userPicker'. The value is the specific ID of the user
+    content.SetValue("userPicker", -1);
             
     // Save the change
     contentService.Save(content);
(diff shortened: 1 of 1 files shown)


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/User-Picker/index.md
```diff
@@ -53,3 +53,43 @@ Please note that getting the Value of the property will return the user ID - pro
...
+    var contentService = Services.ContentService;
+
+    // Create a variable for the GUID of the page you want to update
+    var guid = new Guid("796a8d5c-b7bb-46d9-bc57-ab834d0d1248");
+    
+	// Get the page using the GUID you've defined
+    var content = contentService.GetById(guid); // ID of your page
+	
+	// Set the value of the property with alias 'userPicker'
+    content.SetValue("userPicker", 1);
```
sofietoft, line 72: The ´1´ in this line of code is the ID of a specific user, right?
Perhaps that should be explained in the comments or before/after the code snippet? 🤔 
What do you think?

sofietoft (commented):
I've added a single comment - mind taking a look?

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/User-Picker/index.md
erikjanwestendorp: @sofietoft '1' is de specific user indeed. I changed it to -1 because it always exists. Also added a comment in code :-)

sofietoft (approved):
Ah, super! Thanks for clarifying the User ID 💪 

I'll merge it right in.
Thanks again for updating these articles 🎉

PR #2911, opened 2020-11-30
Update index.md

busraparnell:
Fall into some missing info about prerequisites and added one line to make it clearer.

### DIFF
--- a/Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
+++ b/Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
@@ -32,6 +32,7 @@ Below is a screenshot of our recommendation on how the projects should be config
 
 * Visual Studio 2017 v15.9.6 or later
 * Git and/or Git Credential Manager for Windows
+* If you're using GitKraken, SourceTree etc. you will still need to have Git CLI installed and available in the path
 
 ## Video tutorial
 
--- a/Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
+++ b/Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
@@ -32,7 +32,10 @@ Below is a screenshot of our recommendation on how the projects should be config
 
 * Visual Studio 2017 v15.9.6 or later
 * Git and/or Git Credential Manager for Windows
-* If you're using GitKraken, SourceTree etc. you will still need to have Git CLI installed and available in the path
+
+:::note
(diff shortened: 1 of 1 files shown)


Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
```diff
@@ -32,6 +32,7 @@ Below is a screenshot of our recommendation on how the projects should be config
 
 * Visual Studio 2017 v15.9.6 or later
 * Git and/or Git Credential Manager for Windows
+* If you're using GitKraken, SourceTree etc. you will still need to have Git CLI installed and available in the path
```
sofietoft, line 35: The step above, already mentions that you need to have Git on your machine. Perhaps this prerequisite should instead mention that you might want a Git Client of some sort, to use when you push changes back to the Cloud.

What do you think?

```suggestion
* If you're using GitKraken, SourceTree etc. you will still need to have Git CLI installed and available in the path
```

sofietoft (commented):
Thanks for the PR @busraparnell ! 😄 

I've added a single comment - mind taking a look?

Umbraco-Cloud/Set-Up/Working-With-Visual-Studio/index.md
busraparnell: I had GitKraken, Github desktop to actually download the solution locally but I couldn't run UaaSClone.cmd because I didn't have Git CLI. I needed Git for my command line as well just to download first. That's why I wanted to make it a bit more clearer :)

sofietoft: Aha! Thanks for explaining.
For this guide here though, you do not need to clone down the project using Git. When you use the UaaS.cmd tool, it will clone down everything you need. You then need to use Git to push changes back up to the Cloud.
Does that make sense?

busraparnell: It does indeed thanks Sofie!
But to be able to run UaaS.cmd in the beginning you do need Git for the command line. Otherwise, it throws an error.

sofietoft: Right 🤔 Perhaps that should be a bit clearer.

How about, instead of adding the new bullet to the prerequisites list, that we instead add a note below it?
Something along the lines of "If you're used to using a Git client like GitKraken or SourceTree, you will still need to make sure that you have Git CLI installed. Git CLI is used by the UaaS.cmd tool to clone down your Cloud project."

What do you think?

busraparnell: That's perfect :)

sofietoft (approved):
Thanks for adding the suggestions 😄 

PR #2913, opened 2020-11-30
Update true/false property editor

erikjanwestendorp:
Bool instead of string


sofietoft (approved):

PR #2914, opened 2020-12-01
 small note about URL management redirects when migrating

jonat123:


sofietoft (approved):
Looks good 😄 💪 

I'll merge it right in.

PR #2915, opened 2020-12-01
Add a tip on how to control the colour of the content app icon

poornimanayar:
Hi,

I  have added a tip on how to control the colour of the content app icon.

Poornima

### DIFF
--- a/Extending/Content-Apps/index.md
+++ b/Extending/Content-Apps/index.md
@@ -263,7 +263,7 @@ You will still need to add all of the files you added above but, because your `C
 ```
 
 :::tip
-You can also have a coloured icon for your Content App by specifying the icon in the format `icon-[name of icon] color-[name of color]`.
+You can also have a coloured icon for your Content App by specifying the icon in the format `icon-[name of icon] color-[name of color]`. For eg, an indigo colored icon can be specified for your content app by specifying the icon as  `"icon-calculator color-indigo"` in your Content App class or *package.manifest* .
 :::
 
 ## Notification badges
--- a/Extending/Content-Apps/index.md
+++ b/Extending/Content-Apps/index.md
@@ -262,6 +262,10 @@ You will still need to add all of the files you added above but, because your `C
 }
 ```
 
+:::tip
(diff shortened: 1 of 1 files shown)


Extending/Content-Apps/index.md
```diff
@@ -262,6 +262,10 @@ You will still need to add all of the files you added above but, because your `C
 }
 ```
 
+:::tip
+You can also have a coloured icon for your content app by specifying the icon in the format *icon-[name of icon] color-[name of color]*.
```
sofietoft, line 266:
```suggestion
You can also have a coloured icon for your Content App by specifying the icon in the format `icon-[name of icon] color-[name of color]`.
```

sofietoft (commented):
Thanks for adding the note @poornimanayar !

I've added a few suggestions.
Also, I'm a bit unsure as to where I would add these colors? Perhaps you could add some more details on where these should be specified?

poornimanayar:
Hi @sofietoft ,

These colours are already in the core. I believe all the colours available when you choose an icon for a doc type are supported. Having said that I believe there is a need to list them out. Do we have a page for supported icons and colours?

Poornima

sofietoft:
Ah! I'm sorry, I should have been more clear.
I meant, where you would add `icon-[name of icon] color-[name of color]` when creating a Content App.
Like, in what file/piece of code would you add these values, in order to set the color and the icon?

About whether we have a page for supported colors and icons - I do not believe that we do.
I'll have a look around and see what I can find.

poornimanayar:
Yes, got it. I will amend acccordingly

sofietoft (commented):
Thanks for the updates @poornimanayar !

I see that I might have misunderstood this.
The color is added along with the icon - right? 
So, in this article, in line 65 (the packageManifest sample), that's where you would add the color together with the icon?
Is that correct?

poornimanayar:
Hi @sofietoft ,

Thats right.

Poornima

sofietoft:
Perfect! Thanks for helping me understand, and for adding it to the article 💪 Much appreciated!

I'll get it merged in!

PR #2917, opened 2020-12-01
Update Member Picker Property Editor

erikjanwestendorp:
Add Value example

Belongs to: #2714

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Member-Picker/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Member-Picker/index.md
@@ -60,10 +60,10 @@ See the example below to see how a value can be added or changed programmaticall
     var content = contentService.GetById(guid); // ID of your page
 
     // Create a variable for the GUID of the member ID
-    var author = Guid.Parse("ed944097281e4492bcdf783355219450");
+    var authorId = Guid.Parse("ed944097281e4492bcdf783355219450");
 
     // Set the value of the property with alias 'author'. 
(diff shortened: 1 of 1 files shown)


sofietoft (commented):
Thanks for adding the sample @erikjanwestendorp ! 💪 

One comment:
I found it slightly confusing that the alias is "author", and the variable with the Member ID is also author. Makes it a little hard to distinguish between the two.
Perhaps instead the variable could be something like `authorId`? What do you think? 🤔

erikjanwestendorp:
@sofietoft Good point! 😄 I committed the change!

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Member-Picker/index.md
```diff
@@ -43,3 +43,48 @@ The member picker opens a panel to pick a specific member from the member sectio
...
+}
+```
+
+If Modelsbuilder is enabled you can get the alias of the desired property without using a magic string:
+
+```csharp
+@{
+
+    // Set the value of the property with alias 'author'
+    content.SetValue(Home.GetModelPropertyType(x => x.Author).Alias, author);
```
sofietoft, line 88:
```suggestion
    content.SetValue(Home.GetModelPropertyType(x => x.Author).Alias, authorId);
```

sofietoft (commented):
Thanks for the updates.

One more suggestion, and I think we're good here 💪

erikjanwestendorp:
@sofietoft oops.. my bad. Just committed your suggestion 🙂

PR #2922, opened 2020-12-03
Delete 8.8 page due to error in Url example

OwainWilliams:
The example I added yesterday was incorrect in this example. link.Url is valid and correct.

Sorry.


sofietoft (approved):
Ah 🙈 
Sorry for not picking up on that either.

And thanks for fixing it! 😄 Much appreciated.

PR #2924, opened 2020-12-03
Added content as an option for package

Aaen:
A package can also contain content so I have added this to the list of options.


jmayntzhusen:
Good point, not sure how we missed that one 😅 

Thanks Dennis!

PR #2931, opened 2020-12-08
Guide on moving away from the Latch service

sofietoft:


Umbraco-Cloud/Set-Up/Manage-Hostnames/Move-away-from-Latch/index.md
```diff
@@ -0,0 +1,48 @@
+---
+versionFrom: 8.0.0
```
jonat123, line 2: Is this only available from version 8? And should we also not add both a meta title and meta description? :)

sofietoft: Good point 🙈 I've gotten so used to "just" writing `8.0.0`...
And I'll add some meta stuff as well - thanks for the reminder!

PR #2933, opened 2020-12-08
Added small guide to check if you are using latch

jonat123:


sofietoft (changes requested):
Looks great @jonat123 !

I've added a comment about the steps in the warning on the Latch article.

Also, consider not using words like "simple" - what seems simple to you, might not be simple to others ;)

sofietoft (approved):
Awesome! Thanks for applying the suggestions.

I'll merge this in.

PR #2945, opened 2020-12-15
Update Checkbox-List

erikjanwestendorp:
Set value programmatically
Belongs to: #2714


sofietoft:
Thanks for yet another of these @erikjanwestendorp ! 😄 

Do you think it would also make sense to add the JSON using statement to this one perhaps?

erikjanwestendorp:
@sofietoft Good point. I just updated de PR with a new commit. I will look to the previous PR's as well. I think some using statements should be added in more places.

PR #2947, opened 2020-12-16
Update label

erikjanwestendorp:
Set label value programmatically

Belongs to: #2714


sofietoft (approved):
And another one 😄 
Seems you've found quite a good morning ritual here - sending in a PR just before work!

We really do appreciate the work you put into this! 🙌 
Thanks @erikjanwestendorp !

erikjanwestendorp:
@sofietoft Yea, Creating a PR is indeed part of my morning ritual 😄

PR #2949, opened 2020-12-17
Update label

erikjanwestendorp:
add string quotes


sofietoft:
Ah, whoops!
Seems I was a bit too quick with the merge in that one 😅 

Thanks for updating the sample 💪

PR #2952, opened 2020-12-17
Confirm where to apply the Maindom.Lock setting

marcemarc:
Keeps coming up in the forum that it's not clear now in 8.4+ in these docs whether the setting is applied just to Primary, or to Replica servers or to both


Jeavon (approved):

Getting-Started/Setup/Server-Setup/Load-Balancing/azure-web-apps.md
```diff
@@ -36,7 +36,7 @@ Several file system based services in Umbraco such as the Published Cache and Lu
 ```xml
 <add key="Umbraco.Core.MainDom.Lock" value="SqlMainDomLock" />
 ```
-
+Apply this setting to both the __PRIMARY__ Administrative server and the __REPLICA__ scalable public facing servers.
```
marcemarc, line 39:
```suggestion
Apply this setting to both the __MASTER__ Administrative server and the __REPLICA__ scalable public-facing servers.
```

PR #2953, opened 2020-12-18
Update Color Picker

erikjanwestendorp:
Set value programmatically

Belongs to: #2714

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Color-Picker/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Color-Picker/index.md
@@ -47,3 +47,44 @@ The Color picker allows you to set some predetermined colors that the editor can
     }
 }
 ```
+
+## Add values programmatically
+
+See the example below to see how a value can be added or changed programmatically. To update a value of a property editor you need the [Content Service](../../../../../Reference/Management/Services/ContentService/index.md).
+
+```csharp
+@{
+	// Get access to ContentService
+	var contentService = Services.ContentService;
+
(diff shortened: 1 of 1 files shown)


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Color-Picker/index.md
```diff
@@ -46,4 +46,45 @@ The Color picker allows you to set some predetermined colors that the editor can
...
+	var contentService = Services.ContentService;
+
+	// Create a variable for the GUID of the page you want to update
+	var guid = Guid.Parse("32e60db4-1283-4caa-9645-f2153f9888ef");
+
+	// Get the page using the GUID you've defined
+	var content = contentService.GetById(guid); // ID of your page
+
+	// Set the value of the property with alias 'color'. 
+	content.SetValue("color", "38761d");
```
sofietoft, line 67: Should we mention that the value set here, needs to be one of the prevalues on the Color Picker in order to work? 🤔
It might be different when you have it in a complete setup, but testing this sample, it won't set anything unless it's one of the prevalues I've set.

What do think?

sofietoft (commented):

I've left a comment - take a look when you have the time.

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Color-Picker/index.md
erikjanwestendorp: @sofietoft Good point. I added a comment

sofietoft (approved):
Great! Thanks for applying the suggestions.

I'll make sure this is merged in!

PR #2957, opened 2020-12-21
Link to same major version

jmayntzhusen:
The v7 version of the flexible LB docs link to advanced v8 features instead of advanced v7. It would be nice to link on to the same major 😊


sofietoft:
Yes! That's a very good point 💪 
Thanks for updating the link @jmayntzhusen !

PR #2962, opened 2020-12-23
Add sample to content service

erikjanwestendorp:
Add a content programmatically


sofietoft:
Thanks for the PR @erikjanwestendorp 😄

I'll make sure it is reviewed as soon as possible.

sofietoft:
Apologies for the delay on this @erikjanwestendorp !
I've been struggling a bit with it, so I brought it to the team this afternoon.

We talked about how the sample you've made doesn't really fit on the ListView property editor page.
I understand why you've added it, but the sample is more general and shows how to add a new content item under a parent page. So, nothing really specific to the ListView property...
It's a really great sample though! Simple, and easy to follow 👏 
We thought it would be great to have it as a sample under [ContentService](https://our.umbraco.com/documentation/Reference/Management/Services/ContentService/) instead - it's a great real-life sample of what you can do with the ContentService, right?

As for a sample for the ListView property editors - we can't really think of anything that makes sense. Like, are there any values in a ListView that you would wanna update programmatically? 🤔 
So, we thought that it would be OK to not have a sample for this particular property editor.

I hope it all makes sense - Happy to answer any question or try to resolve any confusion my long answer might have caused 😄

erikjanwestendorp:
@sofietoft Thanks for you feedback! 😄 I can certainly imagine that it is not quite appropriate. I've moved the example to the content service article. Perhaps the description here is a bit brief, if you have any suggestions, please let me know!

sofietoft (approved):
Thanks for moving the sample into a new article here💪

PR #2966, opened 2020-12-29
Update Grid Layout

erikjanwestendorp:
Set value programmatically

Belongs to: #2714


sofietoft:
Thanks for the PR @erikjanwestendorp 😄 

I'll make sure it is reviewed as soon as possible.

sofietoft (commented):
That sure is a lot of brackets to get through 😅 
But great sample!

Would it perhaps make sense to add a little section in the top of article on which kind of "settings" are used for the Grid in the sample? 🤔 
I'm guessing it's "just" the default one that Umbraco ships with that's used?

What do you think?

erikjanwestendorp:
@sofietoft Sounds good!! 😄 I added an section 👍

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Grid-Layout/Add-Value-Programmatically.md
```diff
@@ -0,0 +1,96 @@
...
+versionFrom: 8.0.0
+---
+
+# Add values programmatically
+
+## 	Configuration
+
+For this example, the default grid configuration has been used.
+
+![Grid configuration](images/Grid-configuration.jpg)
```
sofietoft, line 11:
```suggestion
![Grid configuration](Images/Grid-configuration.jpg)
```

sofietoft (commented):
Thanks for applying the suggestions @erikjanwestendorp ! 😄 

I noticed that the image wasn't visible in the preview - might be because the images folder in this case is actually capitalized... 🤔 
I tried suggestion to update the path to the image - let's see if that does the trick, yeah? 🤞

erikjanwestendorp:
@sofietoft Thanks for the suggestion. I don't think it's because it's case sensitive but because the image isn't in the expected place yet. I think it will be fixed after merge. See the difference between:

https://github.com/umbraco/UmbracoDocs/pull/2966/files?short_path=73ba3a0#diff-73ba3a0c212e37d634ce61c5997644b22eff2469350375194634b0ae3e9e3cff

https://github.com/erikjanwestendorp/UmbracoDocs/blob/update-grid-layout-property-editor/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Grid-Layout/Add-Value-Programmatically.md

sofietoft (approved):
Looks like that did the trick!

Thanks again for adding this article @erikjanwestendorp !
I'll get it merged in 💪

sofietoft:
Ah. Merged before I saw your last comment there 🙈 

But of course it looks like you were right 😄 
The image (and the article) is live! 💪

Issue #2759, opened 2020-10-08
Document HasChildren method on ContentTypeService


Jeavon:
Both Guid and Int examples need to be added

https://github.com/umbraco/UmbracoDocs/blob/main/Reference/Management/Services/ContentTypeService/Retrieving-content-types.md

umbrabot:
Hi @Jeavon,

We're writing to let you know that we would love some help with this issue. We feel that this issue is ideal to flag for a community member to work on it. Once flagged here, folk looking for issues to work on will know to look at yours. Of course, please feel free work on this yourself ;-).  If there are any changes to this status, we'll be sure to let you know. 

For more information about issues and states, have a look at [this blog post](https://umbraco.com/blog/the-umbraco-issue-tracker-process/)

Thanks muchly, from your friendly Umbraco GitHub bot :-)

PR #2967, opened 2020-12-30
Update Retrieving-content-types.md

erikjanwestendorp:
HasChildren

Belongs to https://github.com/umbraco/UmbracoDocs/issues/2759

### DIFF
--- a/Reference/Management/Services/ContentTypeService/Retrieving-content-types.md
+++ b/Reference/Management/Services/ContentTypeService/Retrieving-content-types.md
@@ -82,7 +82,3 @@ Although the use of a GUID is preferable, you can also use it's numeric ID:
 // Check if there are children
 var hasChildren = contentTypeService.HasChildren(1234);
 ```
-
-
-
-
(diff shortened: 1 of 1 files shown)


Reference/Management/Services/ContentTypeService/Retrieving-content-types.md
```diff
@@ -67,3 +67,18 @@ IEnumerable<IContentType> contentTypes = contentTypeService.GetChildren(Guid.Par
 
 ```
 
+## Check whether a content type has children
+
+In some cases it can be useful to check if a content type has children. The `HasChildren` method can be used to check whether a content type has children.
+
+```C#
```
sofietoft, line 74:
```suggestion
```csharp
```

Reference/Management/Services/ContentTypeService/Retrieving-content-types.md
```diff
@@ -67,3 +67,18 @@ IEnumerable<IContentType> contentTypes = contentTypeService.GetChildren(Guid.Par
...
+// Check if there are children
+var hasChildren = contentTypeService.HasChildren(Guid.Parse("2b54088e-d355-4b9e-aa4b-5aec4b3f87eb"));
+```
+
+Although the use of a GUID is preferable, you can also use it's numeric ID:
+
+```C#
```
sofietoft, line 81:
```suggestion
```csharp
```

sofietoft (commented):
I've made a few suggestions.

Also, would it make sense to mention what type of data the method returns? I'm assuming it's a boolean - true/false.. that might also be implied already 🤔 
What do you think?

erikjanwestendorp:
@sofietoft Thanks for you review. Just updated the PR.

sofietoft (approved):
Looks great! I'll get it merged it 💪

PR #2968, opened 2020-12-31
Update Cache/index-v7.md

erikjanwestendorp:
Remove needsV8Update

### DIFF
--- a/Reference/Cache/index-v7.md
+++ b/Reference/Cache/index-v7.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
-needsV8Update: "true"
 ---
 
 # Cache & Distributed Cache


sofietoft:
So the reason we haven't yet removed the "needV8Update" label from this article, is that only very little about caching in Umbraco 8 has been documented so far.
We've kept it there, to let people know that not all contents on the page have yet been verified for Umbraco 8 - which I know, can be very confusing.
As you might see - https://our.umbraco.com/Documentation/Reference/Cache/ - the index page for caching on Umbraco 8 is very empty at the moment.

I think we should keep the label there for now!

erikjanwestendorp:
@sofietoft That make sense. I'm closing this one, and will be working on updating the article.

PR #2970, opened 2021-01-01
Update Index.md

erikjanwestendorp:


sofietoft:
Good catch @erikjanwestendorp !
Thanks for fixing this excessive use of the word 'use' 😄

PR #2971, opened 2021-01-06
Added the "get children" menu link

RyuLindow:


sofietoft (approved):
Good catch @RyuLindow ! Thanks.
Will merge it right in 💪

PR #2974, opened 2021-01-07
Updating whitelist to allowlist

MMasey:
Renamed whitelist to allowlist for improved inclusivity

### DIFF
--- a/Umbraco-Cloud/Set-Up/External-Services/index.md
+++ b/Umbraco-Cloud/Set-Up/External-Services/index.md
@@ -1,15 +1,15 @@
 ---
 versionFrom: 7.0.0
-keywords: whitelist allowlist ip internal
+keywords: whitelist allow list allowlist ip internal
 ---
 
 # Using external services with Umbraco Cloud
 
 In some cases Umbraco Cloud might not be the only service you are working with. You might need to work with other services as well - this could be either internal or third party services. In either case, it will be services external to Umbraco Cloud.
 
-When you are working with an external service that is behind a firewall and that service needs to communicate with your Umbraco Cloud project you need to make sure that the Umbraco Cloud Server IPs are allowed to bypass the firewall.
+When you are working with an external service that is behind a firewall and that service needs to communicate with your Umbraco Cloud project, you need to make sure the Umbraco Cloud Server IPs are allowed to bypass the firewall.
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for adding this PR @MMasey !
I agree that we should definitely strive towards not using words and phrases like whitelist and blacklist.

I the cases where "whitelist" is used in this article, it's used as a verb and not a noun, in which case I do not think that "allowlist" words.
How about we rephrase the sentences instead?

So, for example with line 10 that currently says:
"_When you are working with an external service that is behind a firewall and that service needs to communicate with your Umbraco Cloud project you need to allowlist the Umbraco Cloud Server IPs (other services may refer to it as "whitelist")_"

We could rephrase it to
"_When you are working with an external service that is behind a firewall and that service needs to communicate with your Umbraco Cloud project you need to make sure that the Umbraco Cloud Server IPs are allowed to bypass the firewall._"

And then something similar for the other mentions in the article.
What do you think? 🤔

MMasey:
I know what you mean, it's an interesting grammatically as in theory the term "allowlist" follows the same concept as "whitelist". They mean the same thing, and should be usable in the same way, but I think we're it's not been adopted in the same way as "whitelist" is feels a little odd (and I do agree it feels a little odd). I just noticed that my spellchecker also doesn't recognise "allowlist" as a word too 🤷‍♂️. 

I'm happy to reword if if it'll make it easier to for more people to understand. Would you like me to look at how the other uses of it in on the page could be changed too?

Thanks

sofietoft:
Yeah, I hear what you're saying! The word "allowlist" is also kinda growing on me, to be honest. At least as a noun... As you also say, it is the same concept as "whitelist", it just hasn't really caught on as a "real word" yet.
I think it's when it's being used as a verb like it is in this article - that's when it becomes a bit confusing and odd-sounding.

If you're up for it, it would be awesome if you would take a shot at rephrasing the sentences a bit here, to avoid using "allowlist" as a verb! 💪

MMasey:
Hi @sofietoft, i've finally got round to making those ammendments, let me know what you think :)

sofietoft (approved):
It looks great @MMasey !

I think it's a nice way to introduce the "allow list" concept, but referring to how other services might call it "whitelist" 💪 
Thanks again for updating this! Much appreciated 🙌

PR #2979, opened 2021-01-11
removed unused line

dampee:
the sample had unused code, so I removed the unnecesary line


sofietoft (approved):
Good catch @dampee !

I'll merge it in right away.

Issue #2972, opened 2021-01-06
Creating a property editor with c#


erikjanwestendorp:
## What type of issue is it? 
Missing documentation

## What article/section is this about?
[Creating a property editor](https://our.umbraco.com/documentation/Tutorials/creating-a-property-editor/)

## Describe the issue
Is it an idea to add an example of how to create a property editor with c#?

sofietoft:
Thanks for the suggestion @erikjanwestendorp !

I think that would be a great idea actually.
Can't believe we don't already have one 😅 But I've had a look and all I could find that uses C# to add a property editor is the one made for the Block Editor (https://our.umbraco.com/documentation/Extending/Property-Editors/Build-a-Block-Editor).

I think it should be a tutorial in the Tutorials section. We already have a guide in there on how to create a property editor by creating a folder for it manually and adding a package manifest - but I think it would be great to offer two ways of doing it!

erikjanwestendorp:
@sofietoft I updated the existing article with a C# example 😄! Is this the way you prefer it? If so, I can update the next steps as well (add configuration to a property editor etc.).

sofietoft:
That's great @erikjanwestendorp !

I just left a review actually 😄

umbrabot:
Hiya @erikjanwestendorp,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #2981, opened 2021-01-13
Add property editor with c#

erikjanwestendorp:
Belongs to https://github.com/umbraco/UmbracoDocs/issues/2972

### DIFF
--- a/Tutorials/Creating-a-Property-Editor/index.md
+++ b/Tutorials/Creating-a-Property-Editor/index.md
@@ -74,7 +74,7 @@ Inside this package manifest, we add a bit of JSON to describe the property edit
 ```
 
 ## Setting up a property editor with C#
-You can also create a property editor with C# instead of defining it in a `package.manifest`. Create a `MarkdownEditor.cs` file  file in `/App_Code/` to register a editor this way.
+You can also create a property editor with C# instead of defining it in a `package.manifest`. Create a `MarkdownEditor.cs` file in `/App_Code/` to register the editor this way.
 
 ```csharp
 using Umbraco.Core.Logging;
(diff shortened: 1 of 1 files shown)


sofietoft:
Great to have a sample of how this property editor can be added using C# as well 💪 

Going through the sample, I'm not entirely sure where exactly to add the C# code that adds the property editor.
Would that be in a file in the `/App_Plugins/MarkDownEditor/` folder? And does that file need to have a specific name?
What do you think?

I think it would be good to be as specific as possible here - think about those who might be doing something like this for the very first time following this guide! 😄 

I hope it makes sense - do let me know if something doesn't!

erikjanwestendorp:
@sofietoft First of all sorry for the delay on this. I updated the PR. Explained where to add c# code 😄

Tutorials/Creating-a-Property-Editor/index.md
```diff
@@ -73,6 +73,42 @@ Inside this package manifest, we add a bit of JSON to describe the property edit
 }
 ```
 
+## Setting up a property editor with C#
+You can also create a property editor with C# instead of defining it in a `package.manifest`. Create a `MarkdownEditor.cs` file  file in `/App_Code/` to register a editor this way.
```
sofietoft, line 77:
```suggestion
You can also create a property editor with C# instead of defining it in a `package.manifest`. Create a `MarkdownEditor.cs` file in `/App_Code/` to register the editor this way.
```

sofietoft (commented):
Thanks for adding the extra detail @erikjanwestendorp !

I made just a few corrections - other than that, we're good to go here 😄

erikjanwestendorp:
@sofietoft Thanks for your review! I committed your suggestion.

sofietoft:
Great! 😄 
I'll merge it right in then.

Thanks again for adding this extra sample to this tutorial - much appreciated! 💪

PR #2987, opened 2021-01-21
Removed unused instance of ICultureDictionary

Adolfi:
Turns out the local mocked instance of ICultureDictionary is unused in this test, since the actual test uses the base.CultureDictionary. Probably a result of previous refactoring.


sofietoft:
Ah! Good catch @Adolfi ! 💪 
Thanks for fixing this.

I'll merge it in right away!

PR #2988, opened 2021-01-21
Fixing broken link to Examining Examine article

OwainWilliams:
This link might work is it's just `/Documentation/Reference/Searching/Examine/quick-start/index-v7` but I wasn't 100% sure. 
Happy to update it if that would work.


sofietoft:
Thanks for catching this link @OwainWilliams !

I see you've added an `https://` link - we do prefer using relative links in the docs, actually.
Looks like the **Overview Explanation** article is placed directly under `/Examine`, which is one level higher up, than what the current link is set to.
I know this can be a bit confusing to figure out - I **always** use a tool where I can see the filestructure of the docs, when fixing these links! Gives the visual overview of it, which is a great help!

I believe the correct link would be `../overview-explanation.md` 🤔
Would you mind try updating it to that instead?

OwainWilliams:
Had issue with my fork. I'll resubmit this as a seperate PR.

Issue #2992, opened 2021-01-22
Tutorial / Creating a custom dashboard


Jay-umbr:
## What type of issue is it?

Code sample might need updating

## What article/section is this about?
https://our.umbraco.com/documentation/tutorials/Creating-a-Custom-Dashboard/

## Describe the issue
After following the guide and implementing the dashboard fully (section: once we try to do "we know what you edited last week"), we get a bunch of errors in the backoffice when navigating to the dashboard. The dashboard is still functional but will throw errors every single time you open it.
It seems to happen after you edit some templates in the backoffice - that is the only instance I found to be causing issues.

Tested on 8.11 Release Candidate.

`Possibly unhandled rejection: {"errorMsg":"Failed to retrieve entity data for id 2","data":{"Message":"An error has occurred.","ExceptionMessage":"The Umbraco.Web.Editors.EntityController does not currently support data for the type Language","ExceptionType":"System.NotSupportedException","StackTrace":"
   at Umbraco.Web.Editors.EntityController.GetResultForId(Int32 id, UmbracoEntityTypes entityType)\r\n
   at lambda_method.lambda_method(Closure , Object , Object[] )\r\n
   at System.Web.Http.Controllers.ReflectedHttpActionDescriptor.ActionExecutor.<>c__DisplayClass6_2.<GetExecutor>b__2(Object instance, Object[] methodParameters)\r\n
   at System.Web.Http.Controllers.ReflectedHttpActionDescriptor.ExecuteAsync(HttpControllerContext controllerContext, IDictionary 2 arguments, CancellationToken cancellationToken)\r\n--- End of stack trace from previous location where exception was thrown ---\r\n
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()\r\n
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n
   at System.Web.Http.Controllers.ApiControllerActionInvoker.<InvokeActionAsyncCore>d__1.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---\r\n
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()\r\n
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)\r\n
[...]

PR #2996, opened 2021-01-25
Dashboard Tutorial: Update JS samples to define supported entity types

sofietoft:
Based on comments here: #2992

### DIFF
--- a/Tutorials/Creating-a-Custom-Dashboard/index.md
+++ b/Tutorials/Creating-a-Custom-Dashboard/index.md
@@ -541,4 +541,8 @@ You can create your own custom angular services/resources, to interact with your
 
 ## The end
 
+With all of the steps completed, you should have a functional dashboard that will let the logged-in user see the changes they made!
+
+Please note this sample also includes document type changes, which is something editors do not usually do. As a result they might not be able to navigate to the document type by clicking the link on the dashboard.
+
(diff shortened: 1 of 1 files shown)


Jay-umbr (commented):
Looks good! No more errors. Though doctype links on the dashboard do not allow you to navigate to the doctype you edited, unlike content/media. Not sure how important that is.

sofietoft:
Brilliant! Thanks for running the tests 💪 

I believe the intentions of the dashboard built with this tutorial, is to show **editors** the stuff they've been working on recently.
Usually editors do not have access to Document Types or Templates anyways, so I'd say it's fine that those links do not work.

Perhaps it should be made a bit more clear in the "[End result](https://our.umbraco.com/documentation/Tutorials/Creating-a-Custom-Dashboard/#the-end-result)" section, what exactly it is, that the finished dashboard will do. What do you think?

Tutorials/Creating-a-Custom-Dashboard/index.md
```diff
@@ -525,4 +541,8 @@ You can create your own custom angular services/resources, to interact with your
 
 ## The end
 
+With all of the steps completed, you should have a functional dashboard that will let the logged-in user see the changes they made!
+
+Please note this sample also includes document type changes, which is something editors do not usually do. As a result they might not be able to navigate to the document type by clicking the link on the dashboard.
```
sofietoft, line 546: I'm not entirely sure I agree with the comments you added here, @Jay-umbr 🤔 

They will not see any Document Types if they do not have access to edit them.
I think instead, it would make sense to add a note like this further up in the article.
And then instead, have the note mention that the dashboard created in this tutorial is meant to show which content and media files the editors have been editing.

What do you think about that instead?

Issue #2292, opened 2020-02-19
Verify docs for Umbraco 8: "Events" section under Reference


sofietoft:
In the Events section under Reference, there's a handful of articles that still needs to be verified against Umbraco 8.

First of all, there's these three articles listing different events for some services.
Would be great to have these verified, and to have them follow the same format as our "Services" articles - Read more here: https://github.com/umbraco/UmbracoDocs/pull/2227#issuecomment-581955282
- [ ] [Content Type Service Events](https://our.umbraco.com/documentation/Reference/Events/ContentTypeService-Events)
- [ ] [FileService events](https://our.umbraco.com/documentation/Reference/Events/fileservice-events)
- [ ] [LocalizationService events](https://our.umbraco.com/documentation/Reference/Events/LocalizationService-Events)

Other articles that needs to be updated:
- [ ] [Customizing the links box](https://our.umbraco.com/documentation/Reference/Events/EditorModel-Events/Customizing-the-links-box) (example based on the [EditorModel Events article](https://our.umbraco.com/documentation/Reference/Events/EditorModel-Events/))
- [ ] [Extending umbracoApplication](https://our.umbraco.com/documentation/Reference/Events/extending-umbracoapplication)
- [ ] [Hot vs Cold restarts](https://our.umbraco.com/documentation/Reference/Events/hot-vs-cold-restarts)
- [ ] [Troubleshooting slow startup](https://our.umbraco.com/documentation/Reference/Events/troubleshooting-slow-startup)
- [ ] [Understanding bootmanagers](https://our.umbraco.com/documentation/Reference/Events/understanding-bootmanagers)

umbrabot:
Hiya @sofietoft,

Just wanted to let you know that we noticed that this issue got a bit stale and might not be relevant any more.

We will close this issue for now but we're happy to open it up again if you think it's still relevant (for example: it's a feature request that's not yet implemented, or it's a bug that's not yet been fixed). 

To open it this issue up again, you can write `@umbrabot still relevant` in  a new comment as the first line. It would be *super* helpful for us if on the next line you could let us know why you think it's still relevant.

For example:

>@umbrabot still relevant
>This bug can still be reproduced in version `x.y.z` 

This will reopen the issue in the next few hours.

Thanks, from your friendly Umbraco GitHub bot :robot: :slightly_smiling_face:

PR #2999, opened 2021-01-27
V8/2292 extending umbracoapplication

marcemarc:
Updated the example for binding to UmbracoApplicationBase.ApplicationInit event to use a V8 component to gain access and subscribe to the event in an IComponent

This addresses one item (extending UmbracoApplication) on this issue: https://github.com/umbraco/UmbracoDocs/issues/2292

originally raised on this issue: https://github.com/umbraco/UmbracoDocs/issues/1579

### DIFF
--- a/Reference/Events/Extending-UmbracoApplication-v7.md
+++ b/Reference/Events/Extending-UmbracoApplication-v7.md
@@ -1,0 +1,57 @@
+---
+versionFrom: 7.0.0
+versionTo: 8.0.0
+---
+
+## Binding to HttpApplication events
+
+It is not necessary to override and replace the [UmbracoApplication](https://our.umbraco.com/apidocs/v7/csharp/api/Umbraco.Web.UmbracoApplication.html).  If you consider changing the default `global.asax`, read on!
+
+Umbraco allows you to bind directly to HttpApplication events which is very handy since normally you would require an HttpModule to bind to these types of events.
+
+The HttpApplication events are listed here: [https://msdn.microsoft.com/en-us/library/system.web.httpapplication_events.aspx](https://msdn.microsoft.com/en-us/library/system.web.httpapplication_events.aspx)
+
+In order to bind to these events you need to first listen to the `UmbracoApplicationBase.ApplicationInit` event. Here is an example:
+
+```csharp
+using Umbraco.Core;
+using Umbraco.Core.Events;
+using Umbraco.Core.Logging;
+using Umbraco.Core.Models;
+using Umbraco.Core.Services;
+
+namespace MyProject.EventHandlers
+{
+    public class RegisterEvents : ApplicationEventHandler
+    {
+        protected override void ApplicationStarted(UmbracoApplicationBase umbracoApplication, ApplicationContext applicationContext)
+        {
+            //Listen for the ApplicationInit event which then allows us to bind to the
+            //HttpApplication events.
+            UmbracoApplicationBase.ApplicationInit += UmbracoApplicationBase_ApplicationInit;
+        }
+
+        /// <summary>
+        /// Bind to the events of the HttpApplication
+        /// </summary>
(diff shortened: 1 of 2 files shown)


Reference/Events/Extending-UmbracoApplication.md
```diff
@@ -1,52 +1,71 @@
...
 
-It is not necessary to override and replace the [UmbracoApplication](https://our.umbraco.com/apidocs/v7/csharp/api/Umbraco.Web.UmbracoApplication.html).  If you consider changing the default `global.asax`, read on!
+It is not necessary to override and replace the [UmbracoApplication](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Web.UmbracoApplication.html?q=UmbracoApplication).  If you consider changing the default `global.asax`, read on!
 
 Umbraco allows you to bind directly to HttpApplication events which is very handy since normally you would require an HttpModule to bind to these types of events.
 
 The HttpApplication events are listed here: [https://msdn.microsoft.com/en-us/library/system.web.httpapplication_events.aspx](https://msdn.microsoft.com/en-us/library/system.web.httpapplication_events.aspx)
 
-In order to bind to these events you need to first listen to the `UmbracoApplicationBase.ApplicationInit` event. Here is an example:
+In order to bind to these events you need to first listen to the `UmbracoApplicationBase.ApplicationInit` event, which you can gain access to in a custom `IComponent`, here is an example:
```
sofietoft, line 13:
```suggestion
In order to bind to these events you need to first listen to the `UmbracoApplicationBase.ApplicationInit` event, which you can gain access to in a custom `IComponent`. Here is an example:
```

Reference/Events/Extending-UmbracoApplication-v7.md
```diff
@@ -0,0 +1,57 @@
...
+            //Do something...
+        }
+    }
+}
+```
+
+## Related Links
+* [Troubleshooting Slow Startup](Troubleshooting-Slow-Startup.md)
+* [More information about BootManager](Understanding-Bootmanagers.md) (EXPERT)
+* [Adding startup logic and plugin on c# events](Application-Startup.md) (EXPERT)
```
sofietoft, line 57:
```suggestion
* [Adding startup logic and plugin on c# events](Application-Startup-v7.md) (EXPERT)
```

sofietoft:
Thanks for the PR @marcemarc !

I'm wondering about the "Related links" at the bottom of the articles.
Two of them
https://our.umbraco.com/documentation/Reference/Events/Troubleshooting-Slow-Startup 
and
https://our.umbraco.com/documentation/Reference/Events/Understanding-Bootmanagers
haven't been updated for Umbraco 8 yet.

Shouldn't we remove the link to them from the v8 version here? Otherwise I think it'll be confusing that something for v8 links to something that's for v7, but doesn't say that it's ONLY for Umbraco 7 😅
(oh, how lovely complex our versioning system is!!!)

marcemarc:
Hi @sofietoft 

Yes, it's a dilemma...

If the links remain and the pages are updated in the future to V8, then the links will still work...

if the links  are removed, and at a later date the pages are updated to V8, will we remember to go back in and add the links back in?

also if nobody finds those pages will anyone make the effort to update them?

The V7 pages should have the bit at the top saying this hasn't been checked for V8?

It's a very obscure area, somebody in the forum was trying to handle the init event in V8, which prompted the update, but I'm not sure on the value of these pages generally...

regards

Marc

sofietoft:
Yeah, I see your point about the articles getting "lost" if we simply remove the links.

I think, what we should do instead, is add a note to the top of the v7 articles mentioning that they haven't been updated for v8 yet. The reason the auto-note isn't there, is because the files aren't `index.md` files 🙈 

I'll merge this one in, and then I'll add the note to the two articles 💪

marcemarc:
yep that feels like a good strategy!

PR #3000, opened 2021-01-27
Timeouts more grouped

HamDerAndrew:
I've added them a little closer together as these two subjects are very much the same.
However, I haven't merged them into one (yet) because I can see that having them seperate, but close/grouped, is good for two reasons:
1. If a user just needs to know about modifying the Timeout, we can link to that heading/section
2. If a user has timeout issues, then it has its seperate sections as of what you can do about it.

The scenario might be a bit too vague, but I am open to suggestions 😄


sofietoft (approved):
Looks great @HamDerAndrew ! 💪 

Makes sense just moving them closer together, as I can see now that they topic for the sections have a slightly different agenda.

Thanks for making the changes!
I'll merge it in right away 👏

PR #3002, opened 2021-01-29
Added --force to (dotnet new umbraco --UseSqlCe -n MyCustomUmbracoSol…

Adolfi:
Without a **--force** flag you get the following message and the install wont proceed until you add it:

```
Creating this template will make changes to existing files:
  Overwrite   appsettings.Development.json
  Overwrite   appsettings.json
  Overwrite   Program.cs
  Overwrite   Startup.cs
  Overwrite   UmbracoUnicoreSolution.csproj
  Overwrite   Properties/launchSettings.json

Rerun the command and pass --force to accept and create.
```

### DIFF
--- a/UmbracoNetCoreUpdates.md
+++ b/UmbracoNetCoreUpdates.md
@@ -112,14 +112,16 @@ If you have already installed the Umbraco `dotnet new` template, you will need e
 
 ### Steps to create an Umbraco solution using the `dotnet new` template
 
-1. Create a new empty Umbraco solution using MS SQL Azure/Server (first option) or SQL CE (second edition):
+1. Create a new empty Umbraco solution using MS SQL Azure/Server:
 
     ```none
     dotnet new umbraco -n MyCustomUmbracoSolution
     ```
 
+    Or if you prefer to using SQL CE:
+
     ```none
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @Adolfi ! 💪 

I'll run it past our Net Core team, just to make this is indeed how we intend it to work!

bergmania:
This error only happens, if you install a new project over an existing. I don't think this should be part of the documentation.

Adolfi:
@bergmania ah my bad. I ran both `dotnet new umbraco -n MyCustomUmbracoSolution` and `dotnet new umbraco --UseSqlCe -n MyCustomUmbracoSolution` because I missed the "**or**" comment. That probably explains why I needed the --force to override it on the second command.

Maybe if I missed that comment then other might miss it also? Could we make it clearer that you should use one of them and not both? I will make a suggestion and update this PR.

Adolfi:
@bergmania updated. What do you think?

bergmania:
> @bergmania updated. What do you think?

Looks good to me 💪

Adolfi:
@bergmania thanks! Have a great monday!

bergmania:
@Adolfi Thanks for improving our docs :)

PR #3006, opened 2021-02-01
More test samples

Adolfi:
* Added test sample for ContentModel predefined property .Url.
* Added test sample for Apicontroller that returns Json instead of just Enumerable.
* Added test sample for UmbracoContentQuery ContentAtRoot.

### DIFF
--- a/Implementation/Unit-Testing/index.md
+++ b/Implementation/Unit-Testing/index.md
@@ -256,6 +256,16 @@ public class ProductsController : UmbracoApiController
     {
         return new[] { "Table", "Chair", "Desk", "Computer", "Beer fridge" };
     }
+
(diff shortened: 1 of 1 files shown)


sofietoft:
Apologies for the delay on this one @Adolfi !
Looks great - super with more samples on this one 😄  I'll make sure it's merged in!

I noticed that you've also created a solution which holds all the various tests.
Would it be OK, if we add a link to that repo from the docs here on Our? Let me know what you think!

Adolfi:
Absolutely, I would be honoured!
I can create a pull request for that as well.

Cheers Sofie, and thanks!

sofietoft:
Super!! 👏 
And yes, that would be excellent!

PR #3008, opened 2021-02-02
Rename Create-content-programmatically.md to create-content-programma…

kjsguld:
…tically.md

Renamed to lowercase to fix path issues.


sofietoft:
Thanks for the PR @kjsguld !

It is actually the `Index.md` files that needs to be renamed here.
So, instead of `Index.md`, is should be with a lowercase 'i', so `index.md`.
When it's uppercase, it will add Index to the path, which will mess up any links added to the page.

You might need to remove and add the article again, in order to rename it - Git is weird like that sometimes 🤔 

Let me know if you have further questions!

kjsguld:
Thanks for the additional guidance on sorting these routes out @sofietoft. I've renamed the additional files in /ContentService to fit the convention as seen elsewhere on the .our repo.

sofietoft (approved):
Looks great @kjsguld ! 💪 

Thanks for sticking with me on this - sure isn't the most obvious of issues 😉 

I'll get it merged in now - let's make sure that link works again!

PR #3019, opened 2021-02-11
Add information about handling events with variants in play

marcemarc:
Further to this issue:

https://github.com/umbraco/Umbraco-CMS/issues/5410

It seems events with variants can be fired not how you might first expect them to work, but if you think about it, it makes sense - action of the issue is to document the things you can do with variants in events, and this PR tries to provide this information


sofietoft (approved):
Looks great @marcemarc !
Apologies for the delayed reply.

I'll make sure it's merged in!

PR #3023, opened 2021-02-15
Updates code language to xml

liamlaverty:
I'd missed the `xml` part of the language file in my previous PR, so the config file isn't showing on the front end:
https://our.umbraco.com/documentation/add-ons/The-Licensing-model/ 

I think this fixes it 🤔


sofietoft:
Ah, you're right! Completely missed that as well 🙈 

The changes you've made here should indeed fix it 💪 
I'll merge it in right away.

PR #3024, opened 2021-02-16
Adding HQ IPs to backoffice restriction section

hjaltedaniel-umbraco:
In order for Cloud users to be able to get support, with their backoffice restricted by IP filtering, they'll need to allow HQ IPs access.
With this change the IP addresses for the network at HQ as well as the Umbraco HQ VPN is added.


sofietoft:
Thanks for the PR @hjaltedaniel-umbraco !

We're currently looking into whether having our IP addresses public on our docs could pose any kind of security (or other) risks.
I'll keep you posted on where it lands.

sofietoft:
Hi @hjaltedaniel-umbraco .

After some investigation we've decided to close this PR for now.
We are not interested in having our IP addresses out there in the public at the moment.

I hope you understand.

PR #3026, opened 2021-02-19
Adding info on custom certs that the pfx file can only have one cert

hjaltedaniel-umbraco:


sofietoft:
Looks good @hjaltedaniel-umbraco !
Thanks for adding this 💪

PR #3027, opened 2021-02-19
Use constant

erikjanwestendorp:
Use constant for Writer User Group


sofietoft:

I'll make sure this is reviewed as soon as possible!

sofietoft:
Looks like we're all good here 😄 
Thanks again for adding the PR @erikjanwestendorp .

I'll merge it in right away!

PR #3030, opened 2021-02-23
Fix typo in Members Group Service code example

MMasey:
The code example was referencing the `IMemberService` when it should be `IMemberGroupService`.

Thanks


sofietoft (approved):
Good catch @MMasey !
Thanks for correcting this 💪 

I'll merge it right in!

PR #3032, opened 2021-02-24
Updated the Backoffice IP filtering examples with two variations

sitereactor:
The first variation is relevant when you have a reverse proxy like Cloudflare in front of the website, which will be the common approach on Cloud going forward. The second variation is the standard way of using rewrite rules for restricting access to the backoffice based on the IP of the visitor. This will be relevant to old Cloud sites and anything running outside of Cloud.

### DIFF
--- a/Umbraco-Cloud/Security/index.md
+++ b/Umbraco-Cloud/Security/index.md
@@ -121,12 +121,12 @@ Please note these two different variations, which differ if you have a reverse p
 :::note
 Since December 8th, 2020 all Umbraco Cloud sites uses Cloudflare for DNS, so new and updated projects should use the Reverse Proxy version.
 
-If you are ensure whether your Cloud project uses Cloudflare or not, get in touch with the friendly support team, and they will help you out.
+If you are unsure whether your Cloud project uses Cloudflare or not, get in touch with the friendly support team, and they will help you out.
 :::
 
 **Reverse Proxy version (eg. Cloudflare)**
 
-When using Cloudflare, which is the default setup for all Cloud projects, the project will use reverse proxy which gets the IPs from the `X-Forwarded-For` header. In this case, which is most cases, use the first variation here to restrict access to your backoffice using IP filtering.
+When using Cloudflare, which is the default setup for all Cloud projects, the project will from behind a reverse proxy which gets the IPs from the `X-Forwarded-For` header. In this case, which is most cases, use the first variation here to restrict access to your backoffice using IP filtering.
 
 ```xml
 <rule name="Backoffice IP Filter" enabled="true">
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for adding this @sitereactor !

A few questions about the "naming of things".
So, the first variation is what people should use on Cloud projects going forwards - so, with the default/standard set up that our hosting provides.
The second variation is what you should use when you have not yet been migrated to Cloudflare - or if your project is not hosted on our Cloud service.

Is that correct?
I think we should make that a little clearer in the article, as calling the second variation the "Standard version" seems to be a bit misleading if that's not the approach people should use on Cloud going forward.
What do you think? 🤔

sitereactor:
Yes, agree. I'm not sure what a better name would be though? Maybe "non-cloud and old-cloud" 🙈 ... I'm open to suggestions. I'm afraid that explicitly stating that its for 'Cloud projects who are not yet using the updated dns settings from December 8th 2020' becomes a bit much.

sofietoft:
True 🤔 

How about
`Default Umbraco Cloud version` and
`Custom version - also usable off Cloud`

So, we mention that if you're using the default setup that Cloud ships with, then use the first one.
If you on the other hand use custom certs or other, then use the other.

Does that work, you think?
Not sure if that becomes too generic - especially the "Custom version" 🤔 
I want to try to avoid mentioned dates and version numbers - doesn't really seem relevant for a Cloud product (even though I'm very much aware that in our case, that's very relevant at times 😅 )

sitereactor:
What about "Reverse Proxy (like Cloudflare)"-version and "Non Reverse Proxy"-version and then make a Note saying something like "Since December 8th, 2020 all Umbraco Cloud sites uses Cloudflare for DNS, so new and changed projects should use the Reverse Proxy version" ?

sitereactor:
And when we know all hostnames are using Cloudflare DNS we remove the note.

sofietoft:
Yes, I think that's a good idea 💪 

I see you are very keen on using the technical terms here, which makes a lot of sense, as that's what it's called!
Thinking of those who might not be as versed in those terms (someone like me fx!) - would it make sense to add a little description for each of the versions, just to make it perfectly clear when to use which? I'd be happy to make those additions of course - just let me know!

sofietoft:
Hi @sitereactor ,
Based on our back-and-forth yesterday, I've tried updating the naming and added some text as well, before each of the variations.

Please let me know if I got all the details correct, or if I need to make some adjustments. Thanks!

sitereactor:
@sofietoft thanks! Looks good to me. I made one commit with a minor change. So happy with this outcome 👍

sofietoft:
Great! I'll merge it in 💪

PR #3033, opened 2021-02-24
Updated with wildcard example

sitereactor:
Adding an example of a wildcard CAA record for clarification


sofietoft (approved):
Looks great @sitereactor ! 💪 Thanks!

I'll merge it right in.

PR #3034, opened 2021-02-25
Added a note about the old Latch DNS records

RyuLindow:


Umbraco-Cloud/Set-Up/Manage-Hostnames/index.md
```diff
@@ -57,6 +57,9 @@ Cloudflare is a popular DNS provider, which offers a variety of different servic
 
 When creating a CNAME or A-record for your hostname in Cloudflare you need to start with Proxy Status set to 'DNS Only'. Once your hostname is marked with "Protected" under the Hostname page for your Umbraco Cloud project and you can access your website through the hostname, you can choose to change Proxy Status to 'Proxying'. This is mostly relevant when you need to use specific Cloudflare services like Page Rules, Workers, or the like. If you keep the Proxy Status set to 'DNS Only' Umbraco Cloud will handle the automatic TLS setup to ensure that your hostname is always protected with HTTPS.
 
+:::note
+Do note that if you've used our old TLS provider Latch/Let's Encrypt then you have to remove the old DNS records.
```
sofietoft, line 61: I am not sure this note relevant, is the note in lines 48-51 explain the move from Latch to Cloudflare, and links to an article that explains how to go about the move.

Should this note perhaps be added to the "Move away from Umbraco Latch" article? 🤔 
What do you think?

RyuLindow: I've added this note based on a case I had yesterday and that person's feedback.

Also, we have a lot of customers who reach out about DNS atm, so it should be here and not in the move away from one, as it's not the one that we link to in the mail that was sent out :)

sofietoft: Ah, alright. That's a good point.

So, just to make sure I understand it.
The note should make it clear that when moving to the new hostnames set up, they will need to create new DNS records and also make sure to remove the old ones - the ones pointing at the project URL or the "old" IPs.
Is that correct?

RyuLindow: Yep, 100% correct :)

sofietoft: OK. And it this **only** relevant for the customers who are using Cloudflare as their hostname provider?

RyuLindow: Honestly, this is relevant for all kinds of DNS providers.

Maybe we should move this note a bit higher 🤔

sofietoft: I agree!
How about moving this note under the warning in around lines 28-30? 🤔

PR #3038, opened 2021-03-01
Updated Public Access note

HamDerAndrew:
I see people getting confused as to why they cannot enable or disable Basic Auth on their Live environment when they are on a Starter Plan. I tried to add a little clarification as to what it means when you are on a Starter Plan and what Basic Auth can do 😄

### DIFF
--- a/Umbraco-Cloud/Set-Up/Project-settings/index.md
+++ b/Umbraco-Cloud/Set-Up/Project-settings/index.md
@@ -37,7 +37,9 @@ All Staging and Development environments on Umbraco Cloud projects are by defaul
 :::note
 Access to disabling/enabling basic authentication requires your project to be on the Standard plan or higher. 
 
-On **Trial** and **Starter** projects, the basic authentication is enabled on the Live environment and cannot be removed unless the project is upgraded to a higher plan. On a **Starter** plan this means that your Live environment is accessible publicly and does not require a log in, however you cannot enable Basic Auth to require log in on the Live environment again.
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for adding this @HamDerAndrew !
It appears that what's in this article is actually not correct 🤔 

The mention of the **Starter** plan here should be removed.
Then the note would say:

```
Access to disabling/enabling basic authentication requires your project to be on the Standard plan or higher. 

On **Trial** projects, the basic authentication is enabled on the Live environment and cannot be removed unless the project is upgraded to a paid plan.
```

Would that make it clearer, you think? 😅

HamDerAndrew:
I think it would 😃

Although, I'm not sure if we should add a little note like "note: if your project still has Basic Auth enabled on Live after upgrading to a Starter, contact support".

I've experienced a bug, where Basic Auth was not disabled on my Live environment after upgrading from Trial to Starter, two times in February.

One could say that people would probably reach out to Support if they experience this anyway, but I'm not sure 😅

sofietoft:
Aha! That's a very good point.
Would make sense to add a little note on that, yes!

You up for updating the PR with the changes we've talked about? 😁

HamDerAndrew:
Sorry for the delayed response Sofie 😅 

I see you have added it though 👍 

Cheers!

sofietoft:
No worries 😄 I thought it quicker to just apply the changes we discussed 💪 

Let's hope it's a little clearer now how this works.
Thanks again! 😄

PR #3040, opened 2021-03-02
Added documentation for forms features introduced in V8.7.0.

AndyButland:
I've added in this PR some notes on configuration and new field types available in the upcoming 8.7.0 release of the Forms add-on.


Add-ons/UmbracoForms/Developer/Configuration/index.md
```diff
@@ -34,10 +37,12 @@ Enable this feature ONLY if you do understand the security implications.
 :::
 
 ### RecaptchaPublicKey & RecaptchaPrivateKey
-Both of these configuration values are needed in order to use ReCaptcha V2 from Google. You can obtain both of these values after signing up to create a ReCaptcha key here - https://www.google.com/recaptcha/admin
+Both of these configuration values are needed in order to use the "*Recaptcha2*" field type implementing legacy ReCaptcha V2 from Google. You can obtain both of these values after signing up to create a ReCaptcha key here - https://www.google.com/recaptcha/admin
 
 Google has renamed these recently and the `Site Key` refers to `RecaptchaPublicKey` and `Secret Key` is to be used for `RecaptchaPrivateKey`
 
+### RecaptchaV3SiteKey & RecaptchaV3PrivateKey
+Both of these configuration values are needed in order to use the "*reCAPTCHA V3 with Score*" field type implementing ReCaptcha V3 from Google, which is available in Umbraco Forms from v8.7.0. You can obtain both of these values after signing up to create a ReCaptcha key here - https://www.google.com/recaptcha/admin
```
sofietoft, line 45:
```suggestion
Both of these configuration values are needed in order to use the "*reCAPTCHA V3 with Score*" field type implementing ReCaptcha V3 from Google. This field type is available in Umbraco Forms from v8.7+. 

You can obtain both of these values after signing up to create a ReCaptcha key here:  https://www.google.com/recaptcha/admin.
```

Add-ons/UmbracoForms/Developer/Configuration/index.md
```diff
@@ -48,11 +53,13 @@ This setting needs to be a `True` or `False` value and will enable the ASP.NET A
...
 
 ### StoreUmbracoFormsInDb
-
 This setting needs to be set to `True` if you want your Forms data to be stored in the database instead of the .json files in the `App_Data/UmbracoForms` directory in the file system.
 
 For more information on this, read the [Forms in the Database](../Forms-in-the-Database) article.
 
+### UseLegacyPageService 
+In version 8.7.0 an amend was made to improve performance of a service responsible for retrieving the content of the Umbraco page where a form is hosted, in order to allow for the population of string placeholders with the values of properties from the page.  It's expected that the newer implementation will be the best one to use in all scenarios, as, although there are subtle behaviour changes, they aren't expected to be of concern in how the service is used.  However, if there is any reason to use the old one, it can be reinstated by setting this configuration value to `True`.
```
sofietoft, line 61: I've made some significant changes to this part.
First of all, I add a full stop in the first sentence, as it exceeded our recommendation of a max length of 40 words. Could you let me know if the sentence still makes sense?

Second, I removed most of the final part, as I think the wording sounded a little vague as if we do not really believe that the change will actually work 😅 
Happy to add more details if it's necessary - but I think keeping it short and clean is the best option. What do you think?

```suggestion
In Umbraco Forms 8.7 an update was made to improve the performance of the service responsible for retrieving the content of the Umbraco page where a form is hosted. This was done in order to allow for the population of string placeholders with the values of properties from the page.  

By setting the value of the `UseLegacyPageService` to `True` the old service can be reinstated.
```

sofietoft (commented):
Looks great @AndyButland ! 💪 
Super to have these changes ready for the upcoming Umbraco Forms release already.

I've made a few suggestions here and there.
Mainly in the "Configuration" article.
Take a look when you have a minute - and let me know if something doesn't make sense!

Add-ons/UmbracoForms/Developer/Configuration/index.md
AndyButland: Happy with most of these changes but I think it's now a little unclear in the second sentence, as it sounds now that the update was made for the reason described, but actually I was trying to explain what the updated service is used for.

So perhaps change the first paragraph to the following, and ths second as you have suggested.:

```
In Umbraco Forms 8.7 an update was made to improve the performance of the service responsible for retrieving the content of the Umbraco page where a form is hosted. This service is used the population of string placeholders - or "magic strings" - with the values of properties from the page.
```

sofietoft: Thanks for this @AndyButland !
Seems that some words might have gotten lost in that sentence
`This service is used the population of string placeholders - or "magic strings" - with the values of properties from the page.`

Should it instead be
`This service is using the population of string ...`
or perhaps
`This service is used to populate the string ...` ? 🤔

AndyButland: What I meant was `This service is used in the population of....".  Sorry!  But your second suggestion if fine (probably better).

sofietoft: Great! Thanks 😄 
I'll update the article 💪

sofietoft (approved):

warrenbuckley (approved):
From a tech perspective I am all happy so down to you & @sofietoft to sort out any niggles

AndyButland:
@sofietoft @warrenbuckley - just to note, I've added a further page to the docs for 8.7.0, covering the changes that will come in [this PR](https://github.com/umbraco/Forms/pull/435).

warrenbuckley:
Docs read good to me and includes the SQL files too which is nice. Luckily they won't get updated a lot or if ever luckily. Otherwise it would be something else to maintain & ensure is up to date any time a change is made

AndyButland:
No - I see this as something that should be a one-off for this release.  I think we've caught all the places where we probably want to have database constraints in place but currently don't, so it shouldn't need doing again.

Add-ons/UmbracoForms/Developer/Healthchecks/index.md
```diff
@@ -0,0 +1,117 @@
...
+As well as in the log files, such issues will be visible via the health check and will need to be resolved by applying scripts directly to the database.
+
+To support this, we provide the following SQL scripts:
+
+- Apply database integrity schema changes for 8.7.0 - [8.7.0-apply-keys-and-indexes.sql](scripts/8.7.0-apply-keys-and-indexes.sql)
+- Apply database integrity schema changes for 8.7.0 (forms in database tables) - [8.7.0-apply-keys-and-indexes-forms-in-db.sql](scripts/8.7.0-apply-keys-and-indexes-forms-in-db.sql)
+
+The first of these provides the SQL statements required to apply the schema updates for 8.7.0 to the common Umbraco Forms tables. The second applies just to those tables used for when forms are stored in the database, and hence only need to be applied if that option is configured.
+
+Before running any scripts or queries, please be sure to have a database backup in place.
```
sofietoft, line 49: Added this to a note block in an attempt to highlight the fact that it's important to remember to have a backup in place.

```suggestion
:::note
Before running any scripts or queries, please be sure to have a database backup in place.
:::
```

Add-ons/UmbracoForms/Developer/Healthchecks/index.md
```diff
@@ -0,0 +1,117 @@
...
+To find duplicate 'Key' field value in this table you can run the following SQL statement:
+
+```
+SELECT [Key]
+FROM UFForms
+GROUP BY [Key]
+HAVING COUNT(*) > 1
+```
+
+Which will list out the 'Key' fields that are duplicated in the table.
```
sofietoft, line 82:
```suggestion
Running the statement above will list out the 'Key' fields that are duplicated in the table.
```

sofietoft (commented):
Looks great @AndyButland !
I've added a few suggestions; primarily adding some more text between some headings.
Hope it all makes sense - otherwise, let me know.

I do have 2 questions about the SQL scripts:

- Do we have any idea how these will look on Our when people click the link? 😅 Will it perhaps just act as a "download link", so it will start downloading a file, when they click on the link? 🤔 Not sure how we can test this.
- The SQL scripts are all named using Umbraco Forms version 8.7.0 - how can we ensure that this will continue to be relevant in 2 years, when it will be called Forms 10 (or an even higher number)? 🤔 I understand that there will not be any changes to how these scripts should look, but shouldn't we then also be careful about using the version number in the naming of the files/scripts?

AndyButland:
Thanks @sofietoft - I agree with and have added your suggestions (with a couple of tiny amends).

To answer your questions:

- I'm fairly sure it'll just open in the browser (does in my tests and can't see it would be different on our).  Someone can right-click and "Save link as..." if they wanted to download it.
- - I think the 8.7 prefix is probably OK.  It's useful as these are the database schema changes introduced in this version, and it's possible we'd do similar in a later version.  If we did, having those prefixed with the version would be useful to separate them.

sofietoft:
As the Umbraco Forms 8.7 RC is being released today, I will merge this PR as well, as agreed with @AndyButland 👏

PR #3042, opened 2021-03-03
Remove incorrect note on password reset for Cloud

tcmorris:
In response to this tweet: https://twitter.com/sitereactor/status/1367011171530928129


sofietoft:
Thanks for this @tcmorris !
We're actually in the process of updating the text in this note instead of removing it entirely.
We're also adding a bit more information about Umbraco ID to make this new login flow a little clearer.

You can see the process over here: https://github.com/umbraco/UmbracoDocs/pull/3044 
I'll close this one for now.
Happy to hear your thoughts on the changes being made in the other PR, if you're up for it!

PR #3043, opened 2021-03-03
Typo

HamDerAndrew:
Just fixed a typo. I think it's meant to say "issued" and not "issues" 😃


sofietoft (approved):
Good catch @HamDerAndrew ! 🎉 
Thanks for correcting this 💪

PR #3047, opened 2021-03-04
Moving note about upgrading Cloud sites higher

hjaltedaniel-umbraco:
It is a bit too easy to miss the note about the upgrade flow for Cloud sites is different. Moving this up - and putting it into a note, will help make it more visible.


sofietoft (approved):
Ah. Very good point @hjaltedaniel-umbraco !
Thanks for moving the note.

I'll merge it right in.

PR #3048, opened 2021-03-04
Removing section about querystring parameter in Heartcore GraphQL

hjaltedaniel-umbraco:
... As it is not available 😅


rasmusjp:
Thanks 💪 It must've been a copy paste error 😅  as we've never supported the querystring parameter for the GraphQL endpoint

rasmusjp (approved):

PR #3051, opened 2021-03-05
A few corrections regarding Media Service

Jay-umbr:
Explained why we need to include the Umbraco.Core, and suggested different approached to creating folders programmatically


sofietoft (approved):
Looks great @Jay-umbr !

Good additions to the article 💪

PR #3057, opened 2021-03-09
Document how to upgrade projects unattended

sofietoft:
Related to: https://github.com/umbraco/Umbraco-CMS/pull/9531 

Also included a "Table of contents" to the General Upgrades article.

### DIFF
--- a/Getting-Started/Setup/Upgrading/general.md
+++ b/Getting-Started/Setup/Upgrading/general.md
@@ -26,7 +26,7 @@ In this article you will find instruction of 3 different ways of upgrading:
 
 * [Upgrade using NuGet](#upgrade-using-nuget)
 * [Upgrade manually from a Zip file](#upgrade-manually-from-a-zip-file)
-* [Run an unattended upgrade](#run-an-unattended-upgrade)
+* [Run an unattended upgrade (v8.12+)](#run-an-unattended-upgrade)
 
 ## Upgrade using NuGet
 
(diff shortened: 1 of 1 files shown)


rasmusjp (approved):
Looks good 💪 

We might want to add a note about all request made to the site will wait for the upgrade to finish before any content is returned. What do you think?

sofietoft:
Yes. I think that's a great idea.
Should I added as part of the note about composers/components in the last part of the guide, or should it be a separate note somewhere else? 🤔

rasmusjp:
Hmm... Good question 😄 Maybe add a new note after `With the correct configuration...` and then add a new headline before `The Runtime level will use...` called `Boot order` or something like that given it's a more technical explanation on how the boot process is working with unattended upgrades enabled.

Getting-Started/Setup/Upgrading/general.md
```diff
@@ -83,12 +96,59 @@ The installer will do two things:
...
+
+With the correct configuration applied, the project will be upgraded on the next boot.
+
+:::note
+While the upgrade processes are running, any requests made to the site will be "put on hold", meaning that no content will be returned before the upgrade is complete.
+:::
+
+#### Boot order
+
+The Runtime level will use `Run` instead of `Upgrade` in order to allow the website to continue to boot up directly after the migration is run, instead of initiating the otherwise required restart.
```
sofietoft, line 138: @rasmusjp - I've tried adding a note mentioning the requests to the site, and also added one more heading, to sorta split up this section a bit.

I know the phrasing in the new note is a bit strange, but I couldn't really come up with another way of saying it, without using "put on hold" 🤔 Any suggestions are very welcome!!
Let me know what you think!

PR #3060, opened 2021-03-10
Updated the Umbraco cloud user profile info

jonat123:
Updated so it reflects how it currently looks like on cloud


sofietoft (approved):
Looks great @jonat123 ! 💪

PR #3065, opened 2021-03-12
Added a note to clarify the approach with a newer version of Examine …

Jay-umbr:
…on v7 CMS in Azure web apps. The description in the docs was a bit ambiguous.


sofietoft (approved):
Looks great @Jay-umbr !

I'll merge it right in.
Thanks for adding this note!

PR #3067, opened 2021-03-12
Reformatted error msg on deployment config transforms issue

Jay-umbr:
What was once whole, is now separated for readability.


sofietoft (approved):
Looks good 😁

PR #3071, opened 2021-03-13
Changes to ownerID and SKU tags

RyuLindow:
Added a line on how to create v8 projects
Also, changed the ownerID value to make this feature more approachable


sofietoft:
Hi @RyuLindow !

Thanks for adding this.
After a chat with our Cloud team leads, we've agreed that we do not want to update the API docs to include the option to create Umbraco 8 project.
It is not something that we should recommend doing, as the entire API will most likely be rendered obsolete within the next 6-8 months due to our move to a new Cloud infrastructure.

That's also why I'll close this one for now.
Hope you understand, and do let me know if you have further questions.

PR #3078, opened 2021-03-23
Add example using SendingMemberModel

bjarnef:
Adding a basic example of how `SendingMemberModel` can be using to set default member group for a specific member type.
I have suggested in this feature request to be able to set a `MemberGroup` property https://github.com/umbraco/Umbraco-CMS/issues/10032 but for now this works 👍

### DIFF
--- a/Reference/Events/EditorModel-Events/index.md
+++ b/Reference/Events/EditorModel-Events/index.md
@@ -95,6 +95,12 @@ public class SubscribeToEditorModelEvents : IComponent
 
     private void EditorModelEventManager_SendingMemberModel(System.Web.Http.Filters.HttpActionExecutedContext sender, EditorModelEventArgs<Umbraco.Web.Models.ContentEditing.MemberDisplay> e)
     {
+        bool isNew = !int.TryParse(e.Model.Id?.ToString(), out int id) || id == 0;
+        
+        // Skip if entity not is new
+        if (isNew == false)
+            return;
+               
         // Set a default value member group for the member type `Member`
         if (e.Model.ContentTypeAlias == "Member")
         {
--- a/Reference/Events/EditorModel-Events/index.md
+++ b/Reference/Events/EditorModel-Events/index.md
@@ -98,13 +98,16 @@ public class SubscribeToEditorModelEvents : IComponent
         // Set a default value member group for the member type `Member`
         if (e.Model.ContentTypeAlias == "Member")
         {
+            // Find a specific member group
             var group = _memberGroupService.GetByName("Customer");
             if (group == null)
                 return;
-
+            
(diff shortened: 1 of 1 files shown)


sofietoft (commented):
Thanks for this @bjarnef !
Samples are always much appreciated 👏 😄 

marcemarc:
@bjarnef Does this update the Member Group everytime the Member is opened in the backoffice? there doesn't seem to be the equivalent of ContentState that exists in the ContentSending model, that can tell you if it's being created for the first time...

... I guess you could check for MemberId? if it's 0, it's first time?

Or am I reading the code wrong!

If it's the intention to always reset the Member to that Member Group, even if they've been previously manually removed, or added to other groups, then it might be worth adding a comment to that effect in the example...

But I might be totally misunderstanding!!

bjarnef:
@marcemarc yes, it actually does. In my specific use-case it probably isn't a big deal as the specific member types would use a specific member group - the member type to have specific properties and the member groups to grant different permissions.
So yes if selecting a second member group or selecting another member group it resets to this default on save.

I will to check if the entity is new (probably using id or guid).

bjarnef:
@marcemarc I have updated this in https://github.com/umbraco/UmbracoDocs/pull/3078/commits/3ee27c6fae25429b7375dd099f9c1c9f246d9464

Reference/Events/EditorModel-Events/index.md
```diff
@@ -70,6 +70,61 @@ namespace My.Website
...
+    public void Terminate()
+    {
+        EditorModelEventManager.SendingMemberModel -= EditorModelEventManager_SendingMemberModel;
+    }
+
+    private void EditorModelEventManager_SendingMemberModel(System.Web.Http.Filters.HttpActionExecutedContext sender, EditorModelEventArgs<Umbraco.Web.Models.ContentEditing.MemberDisplay> e)
+    {
+        bool isNew = !int.TryParse(e.Model.Id?.ToString(), out int id) || id == 0;
+        
+        // Skip if entity not is new
```
marcemarc, line 100:
```suggestion
        // We only want to set the default group when the member is initially created, eg doesn't have an Id yet
```

marcemarc:
@bjarnef awesome! - fully expecting you to jump to the 'other repo' and now sort out why 'state' is null for Members in this event :-P - I've made a suggestion on the comment above the skipping line... but it's a great example to have in the docs for sure!

marcemarc (approved):

sofietoft (approved):
Looks like we're good to merge this one 😄 

Thanks again for adding the sample @bjarnef ! 
And thanks @marcemarc for the review 💪

PR #3081, opened 2021-03-25
Add csharp code formatting

bjarnef:
Oh I should have added this as well @sofietoft @marcemarc 🙈🤦‍♂️


sofietoft:
Yes! Good catch 😄 👏

PR #3084, opened 2021-03-25
Minor formatting in SubscribeToEditorModelEvents

bjarnef:
Just some minor formatting and it seems the code higtlighting mark `group` with blue color, probably because it is a query keyword: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/group-clause

Although it works in this case it have just renamed to variable to `mg` instead 😄


Jeavon:
Thanks, @bjarnef A very keen eye to spot that!

PR #3097, opened 2021-03-29
Added documentation for Forms configuration key MaxNumberOfColumnsInFormGroup

AndyButland:
This was added in dealing with some feedback from the forms 8.7 RC.

Please can it be reviewed and then merged in on 6th April?  If it goes earlier doesn't really do any harm though, as I've flagged that it's only available from 8.7,


jonat123:
@AndyButland Looks good to me, I can get it merged in on the 6th :-)

PR #3098, opened 2021-03-29
Updating the Hiding the Default umbraco.io Url rewrite rule

hjaltedaniel-umbraco:
After we switched to Cloudflare the rewrite rule to hide the default umbraco.io url will have to use {HTTP_X_Forwarded_For} rather than {HTTP_HOST} to not cause a redirect loop, effectively making the website unavailable.


jonat123:
Looks good @hjaltedaniel-umbraco i'll merge it in :)

PR #3105, opened 2021-04-07
added a section about manual usage check

RyuLindow:


RyuLindow:
I want to link to this section here - https://our.umbraco.com/documentation/Umbraco-Cloud/Getting-started/migrate-existing-site/#1-requirements too :)

jonat123:
@RyuLindow Thanks for the PR!

I think that clarifying what counts as a content node is a good addition, however not sure why users would want to check manually when they can see their usage on cloud already, can you clarify why you think we would need a guide for that? 🙂

RyuLindow:
Sure :)

1. Usage on Cloud can only be used to check the number of nodes on the live env.
2. You can see an actual list of all content nodes in the DB and filter it further or search for something specific if you need to do so.
3. You can use that to check the node count on your on premise site that you want to move to Cloud hence my [comment ](https://github.com/umbraco/UmbracoDocs/pull/3105#issuecomment-814919554)

jonat123:
@RyuLindow thanks for clarifying!

However, I don't think that is something that customers that are looking at their usage should worry about checking.

As you say we are only counting nodes for the live site while the usage limits are also only for the live site and the other environments are not counted in that, so I think that they should not worry about the node count they have on their staging or dev environment when they are checking the usage ​🙂

So I do think that it could be a good thing to have in the migration guide so customers can check their node count to see which plan might be best suited for them before migrating, however when users have a site on Cloud, it is not something that they should worry about checking when they can see the relevant data from the project portal 🙂

Let me know what you think 🙂

Umbraco-Cloud/Set-Up/Usage/index.md
```diff
@@ -14,6 +14,17 @@ You can also get an overview of the usage limitations for your Umbraco Cloud pro
 
 ## Usage limits
 
+First, to clarify what content nodes are:
+
+1. Published nodes
+2. Unpublished nodes
+
+Here's what they are not:
+1. Thrashed nodes
```
jonat123, line 23:
```suggestion
1. Nodes in the recycle bin
```

PR #3106, opened 2021-04-07
Media Picker 3 documentation

nielslyngsoe:
Documentation for the new Media Picker. Lets try to ensure that its not misunderstood what is Media Picker (v2, to become the old) and what is Media Picker 3 (the new one).

Notice for v8, both version will be available, In v9 I hope we can remove Media Picker v.2.


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+## Multiple disabled without Modelsbuilder
+
+```csharp
+@{
+    var typedMediaPickerSingle = Model.Value<MediaWithCrops>("media");
+    if (typedMediaPickerSingle != null)
+    {
+        <p>@typedMediaPickerSingle.MediaItem.Url</p>
```
madsrasmussen, line 81: I think we should remove the paragraph with url
```suggestion

```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+## Multiple disabled with Modelsbuilder
+
+```csharp
+@{
+    var typedMediaPickerSingle = Model.FeaturedBanner;
+    if (typedMediaPickerSingle is MediaWithCrops mediaEntry)
+    {
+        <p>@mediaEntry.MediaItem.Url</p>
```
madsrasmussen, line 94: I think we should remove the paragraph
```suggestion

```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,157 @@
...
+        <img src="@typedMediaPickerSingle.MediaItem.Url" style="width:200px" alt="@typedMediaPickerSingle.MediaItem.Value("alt")" />
+    }
+}
+```
+
+## Multiple disabled with Modelsbuilder
+
+```csharp
+@{
+    var typedMediaPickerSingle = Model.FeaturedBanner;
```
madsrasmussen, line 91: I think we should use the same alias as the example above
```suggestion
    var typedMediaPickerSingle = Model.Media;
```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,157 @@
...
+    foreach (var entry in Model.Medias)
+    {
+        <img src="@Url.GetCropUrl(entry.LocalCrops, "cropAlias")"/>
+    }
+}
+```
+
+## Use global crops
+
+Global image croppings are croppes stored on the Media Item, by the Property Editor `Image Cropper`, making it shared between all usages of the media Item.
```
madsrasmussen, line 128:
```suggestion
Global image croppings are crops stored on the Media Item, by the Property Editor `Image Cropper`, making it shared between all usages of the media Item.
```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,155 @@
...
+### Ignorer user start nodes
+
+Use setting to overrule user permissions, to enable any user of this property to pick any Media Item of the choosen Start node.
+
+When this setting is enabled, a user who doesn't normally have access to the media selected as "Start Node" (/Design in this case), can access the media when using this particular Media Picker. If no Start node has been defined for this property any content can be viewed and selected of this property.
+
+
+### Image Croppings
+
+Define local image croppings, this data will stay locally, meaning that this will be stored on the document in this property.
```
sofietoft, line 42: I'm having a hard time understanding what is meant by this sentence.

Would you be able to help me understand?

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,155 @@
...
+@{
+    var typedMultiMediaPicker = Model.Medias;
+    foreach (var entry in typedMultiMediaPicker)
+    {
+        <img src="@entry.MediaItem.Url" style="width:200px" />
+    }
+}
+```
+
+## Multiple disabled without Modelsbuilder
```
sofietoft, line 74:
```suggestion
### Multiple disabled without Modelsbuilder
```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,155 @@
...
+@{
+    var typedMediaPickerSingle = Model.Value<MediaWithCrops>("media");
+    if (typedMediaPickerSingle != null)
+    {
+        <img src="@typedMediaPickerSingle.MediaItem.Url" style="width:200px" alt="@typedMediaPickerSingle.MediaItem.Value("alt")" />
+    }
+}
+```
+
+## Multiple disabled with Modelsbuilder
```
sofietoft, line 86:
```suggestion
### Multiple disabled with Modelsbuilder
```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,155 @@
...
+    if (typedMediaPickerSingle is MediaWithCrops mediaEntry)
+    {
+        <img src="@mediaEntry.MediaItem.Url" style="width:200px"/>
+    }
+}
+```
+
+## Using local crops
+
+Local image croppings are stored and retrived different than global crops, below is two examples of how to retrive local crops.
```
sofietoft, line 100:
```suggestion
Local image croppings are stored and retrieved differently than global crops. Below are two examples of how to retrieve local crops.
```

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
```diff
@@ -0,0 +1,155 @@
...
+@{
+    foreach (var entry in Model.Medias)
+    {
+        <img src="@Url.GetCropUrl(entry.MediaItem, "cropAlias")"/>
+    }
+}
+
+## Add values programmatically
+
+```This is not part of the current Media Picker v3, feel free to provide such.```
```
sofietoft, line 155: What exactly does this part mean? 🤔
Does it mean that it is not possible to add values to this property programmatically?

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/index.md
```diff
@@ -98,6 +98,12 @@ This control gives the same functionality as the standard listview, but allows y
 
 The markdown editor will be interpreted by the Models Builder. Behind the scenes, Umbraco uses the [Markdown NuGet package](https://www.nuget.org/packages/Markdown/).
 
+## [Media Picker 3](Media-Picker-3/index.md)
+
+`Alias: Umbraco.MediaPicker3`
+
+Media picker 3 is the latest Media Picker, which enables to select one or more medias from the Media Section. This Property Editor can be configured to only pick certain Media Types. And its possible to define Local Image Crops.
```
sofietoft, line 105:
```suggestion
Media Picker 3 is the latest Media Picker, which enables you to select one or more media items from the Media Section. This Property Editor can be configured to only pick certain Media Types and it's possible to define Local Image Crops.
```

sofietoft (commented):
Thanks for the PR @nielslyngsoe !

I've made some suggestions here and there, and added some comments as well.

 I think we need to make it a little clearer that there are now 2 different Media Pickers. Perhaps on the "old version" we link to the new, and somehow make it clearer that we recommend using the new one? What do you think?

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
nielslyngsoe: The existing Image Cropper feature is stored together with the Image-file in the Image Cropper Property Editor. Default is that its used on the Media Type 'Image'. When setting Image Crops on the Image Cropper, they are then shared by all usages of that Media Item. aka. Crops made on Photo of Umbraco HQ, will be the same everywhere that Media Item is used.

The Local Image Crops Feature of MediaPicker 3, is a feature where the crops are stored together with the pick, not on the Media Item itself. Enabling differentiation, as each media item will have custom crops for each usage.

I did reformulate a bit, but please help me out, its rather complex to get everyone's understanding right :-)

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
nielslyngsoe: I just don't know how that's done, or whether more code needs to be made to enable such. But its part of MP2, so i thought to point that out.

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
sofietoft: Ah! Alright. I understand it now. I like the new wording you've added 💪 Let's stick with that.

Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker-3/index.md
sofietoft: Good point! I've removed it for now. I will add it to the Docs issue tracker, in case someone in the community wants to help out with adding the section.

sofietoft:
Thanks for applying the suggestions to the article @nielslyngsoe !

About the different snippets in the bottom of the article, about how to fetch the global and local crops... Would it make sense to move that up above the MVC samples? 🤔 
Also, does it make sense to have 7 different samples on how to fetch values from the property? Seems to me like it could be a bit confusing trying to figure out which one to use... What do you think?

filipbech:
Need to update so it says 8.14 not 8.13

nielslyngsoe:
@sofietoft this manner was inspired by the current Media Picker documentation, I think we need to ask a C# developer on whether it's necessary or useful. if not we should correct both.

sofietoft (commented):
I've given this a second review. Added some suggestions.

Also, I still think it's a lot with all those different MVC samples... 🤔 I'll try cleaning them up, once this has been merged in.

Regarding the crops.. I think it would be great if there was some more explanation about the two different type, local and global. Right now, it's only mentioned in the samples at the very bottom, and not really explained.
Would it make sense to talk a bit about the local and global crops somewhere further up in the article? There's already a little section about "Image crops" in the article. Perhaps it could be added there?

Does the addition to this article relate at all to this: https://our.umbraco.com/documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Image-Cropper/ ? 🤔 Should that article also be updated, or?

Also, I think it would be worth linking to the new Media Picker 3 article from the original Media Picker article, with a short note about how we recommend using the new one!

nielslyngsoe:
@sofietoft good point about mentioning Media Picker 3 in Image Cropper.

I also added the notice to the current Media Picker.

We would need some C# person to make sense of the MVC samples, I don't know if there is an audience for all of it and whether it's understandable for C# developer. Its a general problem in many of our property editors documentation.

sofietoft (approved):
Looks great @nielslyngsoe !

Thanks for adding the suggestions.
I'll make sure we review all the property editor articles at some point, so we can clean up all of these many different samples, that we list for each of them.

I'll make sure this is merged Thursday morning, so it'll be ready for the 8.14 RC.

dampee:
I looked at a few C# samples. And they make some sense, but could be improved in the future.

nielslyngsoe:
@dampee thanks, would be nice to improve them... Currently, they require quite a high skill set to understand why and which one to go with, which is bad for newcomers. So if you or anyone else has ideas on how to communicate this so it becomes easier to learn Umbraco, would just be awesome :-)

I´m not super skilled with C#, so it would be nice with someone show knows what is right to have a look at this. :-)

dampee:
Launching just an idea. 

I don't know if it fits in our agenda of the docs workshop upcoming friday (11am - 1pm). The idea is that we are working on the docs at that time. I'll join the meeting from 11am till 12 as I have another meeting at 12:00. 

Maybe we can work on it together?

nzdev:
Does this support using a different image per crop? Sometimes you need to use a different image if the original does not look good when cropped for some crops.

nielslyngsoe:
Hi @nzdev 
I see the case but it's not a feature of this. From a UX perspective that would be rather complex to establish in a way where it's comfortable and easy for content editors to set up and navigate.
I think the best option, for now, would be to have additional upload fields for example the Image Media Type? Maybe create a new Multi-File Image Media Type, of which you have several Image Croppers (like one for wide-crop tall-crop, etc).

If the crops have to be local, I would recommend having several image pickers? And If you are dealing with a scenario where you need to pick multiple images, it might be an idea to make a Block List of an Element Type that holds those Media Pickers.

I hope that can inspire you to find a good solution for your case.

Interesting thought and it really got my mind started on thinking about different UI solutions.

How do you imagine it working. Like how would you set a different image for one or multiple crops? And how does another Content Editor spot that there is a different image for some crops?

nzdev:
Hi @nielslyngsoe thanks for the reply. I would imagine you pick an image, then adjust the crops, see that for a certain local crop the image does not look good when cropped. Then on the local crop you can choose to select another image. Perhaps with a replace button that then let's you browse for another image. Then to revert have a reset button.

PR #3117, opened 2021-04-13
Use entity HasIdentity property

bjarnef:


sofietoft:
Thanks for the PR @bjarnef ! 💪 

Issue #3076, opened 2021-03-17
Important notes to add to the document


thenexus00:
Hi Team,
I think it would be good to add an important notes on using this feature for people:

- While in the file I think the page should also state that if debug in the web.config is on this will not work
- You can not add url parameters into your solution:
Html.RequiresCss("/css/styles.css?v=12", 6);
This will break the compiling and minifying
- While you can add external files in the require Html.RequiresCss("https://cdnjs.cloudflare.com/ajax/libs/animate.css/3.7.2/animate.min.css", 6);
Anything like this will render outside the compiled dependancy(After). This can have effect on projects as something ordered 3 will actually come after everything inside the css require rendering.

Through several guides, forum posts as well I have seen the call for:

@Html.RenderCssHere()
AS
@Html.Raw(Html.RenderCssHere());

In Umbraco 8 at least this also stops the proper client dependancy from compiling and minifying.

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Getting-Started/Design/Stylesheets-Javascript/index.md***

sofietoft:
Hi @thenexus00 !

Thanks for the report 💪 Much appreciated.

As I understand it, there are several things you'd like to improve about the "Stylesheets and Javascripts" article - and I agree! It could really do with more details and information.

I'm not 100% sure how all this works, so I'll have to run it by some of our experts as well.
Meanwhile, you're more than welcome to take a shot at updating the article yourself, if you want!

marcemarc:
Hi @thenexus00, @sofietoft 

Yes, they are like 'gotchas' with using the Client Dependency, that we should certainly mention here, eg Debug=true and the problem with adding querystrings...

I've created a PR here to start the ball rolling - https://github.com/umbraco/UmbracoDocs/pull/3118

It's hard to know how much to include, as the Client Dependency Framework has its own documentation here: https://github.com/Shazwazza/ClientDependency/wiki but certainly these are all common problem that people hit when using it in Umbraco so super helpful to stop people pulling their hair out for hours to mention.

[...]

PR #3118, opened 2021-04-13
Add a couple of notes around gotchas with the Client Dependency Framework

marcemarc:
see https://github.com/umbraco/UmbracoDocs/issues/3076

### DIFF
--- a/Getting-Started/Design/Stylesheets-Javascript/index.md
+++ b/Getting-Started/Design/Stylesheets-Javascript/index.md
@@ -103,7 +103,7 @@ You can bundle and minify as follows in a view template file.
 ```
 
 :::note
-If you are running your site locally and in your Web.Config the compilation setting for debug is set to true - the Client Dependency Framework will not minify or combine your js/css. This is by design to help when debugging issues in a local environment - but also means if there is a problem minifying a file - you'll only encounter the issue when you deploy - so do test minification and bundling locally by setting the debug setting to false for compilation.
+If you are running your site locally and the compilation setting for debugging in your `web.config` is set to true, the Client Dependency Framework will not minify or combine your js/CSS. This is by design to help when debugging issues in a local environment. However, it also means if there is a problem minifying a file you will only encounter the issue when you deploy. To test minification and bundling locally, set the debug setting to false for compilation.
 :::
 
 :::note
--- a/Getting-Started/Design/Stylesheets-Javascript/index.md
+++ b/Getting-Started/Design/Stylesheets-Javascript/index.md
@@ -107,7 +107,9 @@ If you are running your site locally and the compilation setting for debugging i
 :::
 
 :::note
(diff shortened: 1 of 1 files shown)


Getting-Started/Design/Stylesheets-Javascript/index.md
```diff
@@ -102,4 +102,13 @@ You can bundle and minify as follows in a view template file.
 </head>
 ```
 
+:::note
+If you are running your site locally and in your Web.Config the compilation setting for debug is set to true - the Client Dependency Framework will not minify or combine your js/css. This is by design to help when debugging issues in a local environment - but also means if there is a problem minifying a file - you'll only encounter the issue when you deploy - so do test minification and bundling locally by setting the debug setting to false for compilation.
```
sofietoft, line 106:
```suggestion
If you are running your site locally and the compilation setting for debugging in your `web.config` is set to true, the Client Dependency Framework will not minify or combine your js/CSS. This is by design to help when debugging issues in a local environment. However, it also means if there is a problem minifying a file you will only encounter the issue when you deploy. To test minification and bundling locally, set the debug setting to false for compilation.
```

Getting-Started/Design/Stylesheets-Javascript/index.md
```diff
@@ -102,4 +102,13 @@ You can bundle and minify as follows in a view template file.
...
+:::
+
+:::note
+When adding stylesheet references to Html.RequiresCss there is no need to add ?v=1 to specify a specific version of the asset for cachebusting (In fact doing so will break the minification and combination of your files!). The Client Dependency Framework uses the version number from /config/clientdependency.config on the querystring of the combined and minified resource. To force your updated minified and combined asset to be refreshed, increment this version number by 1 in the config file to bust any browser caching on the old version of your assets. See [Healthcheck for Client Dependency Framework](https://our.umbraco.com/packages/developer-tools/health-check-for-client-dependency-framework/) package to do this from the backoffice, (or automatically increment the version in your build and deploy process).
```
sofietoft, line 110:
```suggestion
When adding stylesheet references to `Html.RequiresCss` there is no need to add `?v=1` to specify a specific version of the asset for cache-busting. Doing so will break the minification and combination of your files. The Client Dependency Framework uses the version number from `/config/clientdependency.config` on the querystring of the combined and minified resource. 

To force your updated, minified and combined assets to be refreshed, increment this version number by 1 in the config file to bust any browser caching on the old version of your assets. See the [Healthcheck for Client Dependency Framework](https://our.umbraco.com/packages/developer-tools/health-check-for-client-dependency-framework/) package to do this from the backoffice, or you can automatically increment the version in your build and deploy process. 
```

sofietoft (commented):
Thanks for adding these notes @marcemarc ! 💪 

I've made a few suggestions here and there - mostly to eliminate those looong sentences 😄 

Also, any reason it's split into two notes, instead of keeping it in one? 🤔 Just a though...

marcemarc:
@sofietoft ace thanks! I did sort of dash it off without a care in the world... :-P

I think it's worth having as two different notes, because they are two very different gotchas... depending on what we think should happen over the external resource, it might make sense to convert from notes, into a 'Troubleshooting' section in it's own right...

sofietoft (approved):
Good point about why they are two separate notes! Let's leave it like that then 💪

sofietoft:
Quick question @marcemarc - Would merging this PR close #3076 ? 🤞

marcemarc:
@sofietoft not completely... it addresses the first two bullet points, merging it now would at least make that information available to people.

I've replied on the PR about the other points!

sofietoft:
Ah, OK. Cool! 💪 
I'll merge this now then - but not the issue as of yet.

Thanks again! 😄

PR #3120, opened 2021-04-14
Updated indentation to avoid breaking csharp code formatting.

Adolfi:
Look at the second code snippet on this page: https://our.umbraco.com/documentation/Extending/Health-Check/

The csharp code snippet is not formatted properly an I think it is because the second row of the content of the robots.txt file does not have any indentation which confuses the opening/closing csharp block.

### DIFF
--- a/Extending/Health-Check/index.md
+++ b/Extending/Health-Check/index.md
@@ -250,9 +250,11 @@ namespace Umbraco.Web.HealthCheck.Checks.SEO
         {
             var success = false;
             var message = string.Empty;
-            const string content = @"# robots.txt for Umbraco
-User-agent: *
-Disallow: /umbraco/";
+            
+            const string content = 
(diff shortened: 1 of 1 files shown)


sofietoft:
Ah! Good catch @Adolfi ! 
Looks like the indentations fixes it - let's see when get it live on Our 🤞

sofietoft:
Hm. Doesn't look like that did the trick 🤔 
Not entirely sure why the code formatting doesn't work here...

Adolfi:
Hmm strange.. I was sure it was due to the indentation but I guess not then.. :/

sofietoft:
Yeah, it's really odd!
I tried adding the snippet to a .cs file in VS Code, and there it says that the formatting is all correct 🤷‍♀️
Not really sure what to do here, tbh 😓

Issue #3070, opened 2021-03-13
In the js controller call to contentResource.getById(vm.CurrentNodeId) gives error


onlinegroeicoach:
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Extending/Content-Apps/index.md***

In the js controller call to contentResource.getById(vm.CurrentNodeId) gives error

I'm using Umbraco 8.9.1. perhaps there is a breaking change in Umbraco, don't know how to solve this.

nul800sebastiaan:
Hi there @onlinegroeicoach, can you please add some more of an explanation of what you tried, why it didn't work, errors, etc? That way we have something to go on 👍

onlinegroeicoach:
Part of description is missing indeed, sorry, this is the page https://our.umbraco.com/documentation/extending/Content-Apps/

I've implented the WordCount Content App. When the executing this content app, I see that this is executed

angular.js?cdv=15311759:13630 GET https://localhost:44343/umbraco/backoffice/UmbracoApi/Content/GetById?id=1290

Which gives a 404 error.

onlinegroeicoach:
Plus a:

angular.js?cdv=15311759:15635 Possibly unhandled rejection: {"errorMsg":"Failed to retrieve data for content id 1290","data":"","status":404}

onlinegroeicoach:
Wait I get it, I get this when I'm in settings editting a document type, in Content it works fine.

So guess the example kind of works, it just needs an extra check that the word count doesn't get executed when within a document type edit screen.

onlinegroeicoach:
Is solved by

`		"show": [
				"+content/*", // show in Content section
                "-contentType/*" // hide in Settings section -> Documenttypes => REQUIRED FOR THIS APP
            -`

Wasn't clear a first that a content app would run (and this case crash) in the DocumentType screen.

sofietoft:
Thanks for the report @onlinegroeicoach , and glad to hear you're managed to solve the issue.

As I understand the case here, there was some confusion as to where the Content Apps can actually be used.
You were using the Content App in the Settings section, which caused you to run into the errors you mentioned above.
Is that correct?

onlinegroeicoach:
Yes, if you don't explicitly disable this sample Content App from being able to run in the Document Type section, it will cause errors there. Correct.

________________________________

sofietoft:
Aha! That's a very good observation @onlinegroeicoach ! 
Thanks for making us aware of this.

[...]

PR #3121, opened 2021-04-14
Updated ContentTypes info

eshanrnh:
Updated "Limiting according to type" section

### DIFF
--- a/Extending/Content-Apps/index.md
+++ b/Extending/Content-Apps/index.md
@@ -165,11 +165,10 @@ Here is an example where all types are taken intro consideration when limiting a
             "show": [
                 "-content/homePage", // hide for content type 'homePage'
                 "+content/*", // show for all other content types
-                "+media/*", // show for all media types
+                "+media/*", // show for all other media types
                 "-member/premiumMembers", // hide for Member type 'premiumMembers'
                 "+member/*", // show for all other Member types
-                "-contentType/textPage", // hide on Content Type with alias 'textPage'
-                "+contentType/*", // show for all other Content types
+                "-contentType/textPage", // hide for Content Type with alias 'textPage'
                 "-contentType/*", // hide for all other Content types
             ]
         }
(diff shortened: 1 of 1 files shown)


Extending/Content-Apps/index.md
```diff
@@ -167,6 +170,7 @@ Here is an example where all types are taken intro consideration when limiting a
                 "+member/*", // show for all other Member types
                 "-contentType/textPage", // hide on Content Type with alias 'textPage'
                 "+contentType/*", // show for all other Content types
+                "-contentType/*", // hide for all other Content types
```
sofietoft, line 173: When adding the `"-contentType/*", // hide for all other Content types`, I think it makes sense to remove the two other lines referencing the contentType 🤔 
What do you think?

sofietoft (commented):
Looks great @eshanrnh !

I've added a single comment - take a look, and let me know what you think!

Extending/Content-Apps/index.md
eshanrnh: @sofietoft yes, I think it makes sense to remove the **_"+contentType/*", // show for all other Content types_** since we are mentioning hide. 

I think we should leave the alias one  **_"-contentType/textPage", // hide on Content Type with alias 'textPage'_** since it acts as an alias example and it will be consistent with the first two lines

Let me know your thoughts and I'll make the changes accordingly :)

sofietoft: Right! That's a good point 🤔 I think keeping the specific one for the textPage is a good idea!

Let's just get rid of `"+contentType/*"` 💪

eshanrnh: @sofietoft Done.

sofietoft (approved):
Looks great @eshanrnh ! 💪 Thanks for updating the article.
Feel free to merge it in.

Closes #3070

PR #3122, opened 2021-04-14
Update documentation for HasValue()

bjarnef:
The documentation here doesn't seem to be correct: https://our.umbraco.com/Documentation/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/File-Upload/

The extension exists as a method here:
https://github.com/umbraco/Umbraco-CMS/blob/d428a4543f33bb7094cf7db5f6b6fdc2d1de3063/src/Umbraco.Web/PublishedCache/PublishedElementPropertyBase.cs#L39

ModelsBuilder  return the value as a `string` property in File Upload and the `HasValue()` isn't an string extension method.


sofietoft:
Ah! Good catch @bjarnef ! Thanks for updating this article 💪 
Much appreciated!

I'll merge it right in.

Issue #3115, opened 2021-04-13
Changing field value types using this method is unpredictable (and not technically supported?)


JasonElkin:
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Reference/Searching/Examine/Indexing/index.md***

I recently ran into an issue on a site that's using this method to define a custom field:

```
provider.FieldDefinitionCollection.AddOrUpdate(new FieldDefinition("eventDate_Long", FieldDefinitionTypes.Long));
```
After starting up I was getting the following error when trying to run a search.

![image](https://user-images.githubusercontent.com/7151793/114525912-69b05000-9c3e-11eb-8e98-bfd3d92481d2.png)

I could fix this by changing my `IUserComposer` like so:

```
// before - throws error
composition.Components().Append<ExamineEventsComponent>();

// after - works OK
composition.Components().InsertAfter<ExamineComponent, ExamineEventsComponent>();
```
This didn't make much sense to me until I looked under the hood at how this works in examine. Examine uses a `FieldValueTypeCollection` for searching. This collection is immutable and **lazily** created from the `FieldDefinitionCollection`.

I.e. this only works if we happen to modify the `FieldDefinitionCollection` _before_ the `FieldValueTypeCollection` is initialised  and in this particular site it turns out that the `FieldValueTypeCollection` is being created before we've modified the `FieldDefinitionCollection`.

I think in this instance (I'm still diagnosing) this is because another package is accessing the index in such a way that's causing the `FieldValueTypeCollection` to be created before my component fires. However, to my mind a package _should_ be allowed/able to do that. The fact it's creating a side-effect that's breaking this methodology implies that it's this process that's wrong.

Simply updating the docs to `InsertAfter<ExamineComponent,...` is likely to stop this being an issue but it strikes me that this is still not the _proper_ way to do it and that there should be a way to modify the `FieldDefinitionCollection` that the ExternalIndex uses before it is created (I will raise that as a separate issue and link to it when I have).

marcemarc:
Hi Jason

Thanks for raising this issue!

What's the other package you have installed that's possibly also accessing the index? 

Agree if it's as simple as warning people to compose after the initial ExamineComponent, to avoid the issue, then that's easy to add to the page as a warning, but what you're saying hints at possibly something that should be addressed in the core architecture, so we can see if HQ have any thoughts on this.

But knowing the additional package will give people the chance to try to replicate the problem.

regards

Marc

[...]

PR #3124, opened 2021-04-14
Update CustomizeIndexComponent to compose/insert directly after ExamineComponent

JasonElkin:
Also added a note to help mitigate against #3115

### DIFF
--- a/Reference/Searching/Examine/Indexing/index.md
+++ b/Reference/Searching/Examine/Indexing/index.md
@@ -28,8 +28,16 @@ using Examine;
 using Umbraco.Core;
 using Umbraco.Core.Composing;
 
-// register the component
-public class CustomizeIndexComposer : ComponentComposer<CustomizeIndexComponent> { }
+// register the component after the ExamineComposer
+[ComposeAfter(typeof(ExamineComposer))]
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @JasonElkin ! 💪 

Jeavon:
@JasonElkin I believe this issue might have been resolved in 8.15 by https://github.com/umbraco/Umbraco-CMS/pull/10182 Can you confirm?

JasonElkin:
@jeavon I need to re-create a minimal reproducible repro to test again but I'm not sure this will cover the problem I was having.

My problem wasn't that the index was being created after 5 seconds (I was well within that) it was that other processes were causing the index to init before my component.

marcemarc:
@JasonElkin I'm being urged to close this issue by the powers of staleness - is that ok? is it still an issue in V9 - does it need some further investigation? if you know it's still a thing, and it would be helpful to have some words around it in the docs, then feel free to reopen this and shout...and we'll put it back in the very big list of things. But thanks for raising it at the time, and sorry it didn't all get resolved sweetly and in a timely fashion - #h5yr!

PR #3130, opened 2021-04-19
Replace constructor with SetUp in test sample

Adolfi:
Sorry, my bad. Correcting my PR from yesterday.
Been working with xUnit for so long now that I forgot you need to use [SetUp] in NUnit.

Cheers.


sofietoft:
Great catch! And thanks for updating the docs here as well @Adolfi ! 💪 

PR #3131, opened 2021-04-19
Updating the Package Manifest to include Dashboards and Sections

GabrielIonut:
I have updated the docs to include dashboards and sections since both are root elements and can be registered using a package.manifest file.


sofietoft:
Thanks for updating the article @GabrielIonut ! I'll make sure it's merged right in 💪 🎉

PR #3138, opened 2021-04-24
Azure Storage Explorer Interface Change

RachBreeze:
Version 1.19.0 of Microsoft Azure Storage Explorer has had an interface change, selecting "storage account or service" rather than "Blob container" reports an error with the access signature when trying to connect. This is not an issue with Umbraco Environment variables but is caused by Microsoft Azure Storage user users picking the wrong settings on "Select Resource" screen (it took me a while to track that one down :-) ). I have updated the docs to help prevent other users getting caught out by this


sofietoft (approved):
Great catch @RachBreeze ! Thanks for finding the culprit here and updating the guide - Much appreciated 🙌 😄 

I'll make sure it's merged in!

PR #3139, opened 2021-04-26
Include a Umbraco ID section.

PratheesRavindrakumarUmbraco:
Some customers think they can use SMTP to configure user invite and forgotten password emails to be sent out through the SMTP server. However, user authentication on Umbraco cloud is all managed through https://identity.umbraco.com, and therefore these emails come from Umbraco's SMTP server and therefore not able to do it via SMTP.

There is only a "note" regarding Umbraco ID and some customers would like more clarification on this. So a Umbraco ID section would be helpful.


jonat123:
@PratheesRavindrakumarUmbraco Thanks for the feedback, as you have made a issue for now I'll close this pr down 🙂

PR #3147, opened 2021-04-28
Update UmbracoNetCoreUpdates.md

nathanwoulfe:
Fixed typo (shot => short), then removed short since there's no need to quantify the example length (and is it short? It's the longest example on the page 😄 )


sofietoft (approved):
Makes sense to get rid of the mention that the example is "shot" 😅 

PR #3148, opened 2021-04-28
Update UmbracoNetCoreUpdates.md

nathanwoulfe:
fixes two broken anchors (to the beta release and known issues in the beta)


sofietoft (approved):
Good catch @nathanwoulfe ! 💪 
Thanks for fixing these links - much appreciated!

PR #3152, opened 2021-04-30
Incorrect code in CE example

ault:
`-ce` = Invalid input switch
should be  `--UseSqlCe `


sofietoft:
Thanks for the PR @ault !

I'll make sure we review this as soon as possible 💪

bergmania:
Thanks @ault.

Sadly, an error has occurred, so our custom shortnames (`-ce`) is not picked up, so the CLI just generated some defaults (`-U`). I think it is a good idea the code examples uses the full names anyways :)

PR #3158, opened 2021-05-04
Removing warning about rewrite not working on Cloudflare

hjaltedaniel-umbraco:
I suggest removing this warning, because my latest testing has shown that the rewrite actually also works on Cloudflare.


sofietoft (approved):
Thanks for taking care of this @hjaltedaniel-umbraco ! 💪 
Much appreciated.

Let's get it merged in!

Issue #3141, opened 2021-04-26
Code example is wrong if you use the member picker on the front end


Attackmonkey:
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Member-Picker/index.md***

In this example code, it will set the picker to the member, and the picked member will display in the CMS, however, in your models on the front end, the IPublishedContent model for the picked member will always be null, as it looks like it should be the UDI for the member, not just the key.

If you pick a member using the back office UI, it is set as a UDI and the IPublishedContent model is correctly populated. We should update this example accordingly!

sofietoft:
Thanks for the report @Attackmonkey !
We'll make sure to investigate as soon as possible 💪

busrasengul:
Hey @Attackmonkey,
As you stated the picker needs to have UDI instead of the key in order to correctly populated. 
Thank you for raising this up. 

Would you be kind to supply the code example changes with a PR maybe? ✌
No pressure though if not we'll get it done as soon as possible 😊

Attackmonkey:
Of course! I'll try and get that sorted for you over the weekend :)

bjarnef:
Just fixed this in https://github.com/umbraco/UmbracoDocs/pull/3161 👍

PR #3161, opened 2021-05-04
Set correct udi value in member picker example

bjarnef:
This PR updates the example for member picker setting the correct udi value and fixes https://github.com/umbraco/UmbracoDocs/issues/3141


busrasengul:
Thank you @bjarnef ✌
I'll merge it in right now!

PR #3177, opened 2021-05-11
Fix property editors

sofietoft:
Many samples in our Property Editors articles were using `HasValue()` the wrong way.
I've updated the ones affected.

### DIFF
--- a/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Rich-Text-Editor/index.md
+++ b/Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Rich-Text-Editor/index.md
@@ -55,23 +55,23 @@ See the example below to see how a value can be added or changed programmaticall
 
 ```csharp
 @{
-	// Get access to ContentService
-	var contentService = Services.ContentService;
+    // Get access to ContentService
+    var contentService = Services.ContentService;
 
-	// Create a variable for the GUID of the page you want to update
-	var guid = Guid.Parse("32e60db4-1283-4caa-9645-f2153f9888ef");
+    // Create a variable for the GUID of the page you want to update
+    var guid = Guid.Parse("32e60db4-1283-4caa-9645-f2153f9888ef");
 
-	// Get the page using the GUID you've defined
-	var content = contentService.GetById(guid); // ID of your page
+    // Get the page using the GUID you've defined
+    var content = contentService.GetById(guid); // ID of your page
 
-	// Create a variable for the desired value
-	var htmlValue = new HtmlString("Add some text <strong>here</strong>");
+    // Create a variable for the desired value
+    var htmlValue = new HtmlString("Add some text <strong>here</strong>");
 
-	// Set the value of the property with alias 'richText'. 
-	content.SetValue("richText", htmlValue);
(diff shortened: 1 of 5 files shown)


Getting-Started/Backoffice/Property-Editors/Built-in-Property-Editors/Textbox/index.md
```diff
@@ -43,7 +43,7 @@ Textbox is an HTML input control for text. It can be configured to have a fixed
 ```csharp
 @{
     // Perform an null-check on the field with alias 'pageTitle'
-    if (Model.PageTitle.HasValue())
+    if (Model.PageTitle != null)
```
marcemarc, line 46: @sofietoft are you sure this is correct?

It's highly likely that PageTitle is a String

and therefore an 'empty string' if somebody didn't fill in a page title if it was non mandatory would mean that the value of Model.PageTitle would be "" and not null

so the check here for null would pass for an empty string, and the markup inside the if statement would be written out..

eg `<p></p>`

But really you don't want that empty markup if there is no PageTitle specified!!!

So you'd either check if (string.IsNullOrWhiteSpace(Model.PageTitle))

or I have a feeling that 'HasValue' or a variation of that is a wrapper for doing that?

sofietoft: I'm sire that `Model.PageTitle.HasValue()` doesn't work, and that `Model.PageTitle != null` works when the property has a value.

Would of course be worth testing when the property is empty.

sofietoft: I've corrected this in #3180

marcemarc: @sofietoft The other syntax that might work here is:

```
if (Model.HasValue(Model.GetModelPropertyType(p=>p.PageTitle).PropertyTypeAlias)){
<p>@Model.PageTitle</p>
}
```
Basically, IPublishedContent/IPublishedElement that Modelsbuilder models inherit from DOES have a HasValue() method to check whether the different type of property eg grid, textstring, date has a value or not - but it requires the alias of the property to do the check...

... so a non-modelsbuilder convert, would be writing

```
if (Model.HasValue("pageTitle")){
<p>@(Model.Value<string>("pageTitle"))</p>
}
```
What I've typed above kind of the strongly-typed version of doing the same with Modelsbuidler to avoid having to type 'pageTitle' - but I'm not sure it's that 'neat' or the 'defined' way: or whether at some point .HasValue() on the property became the 'short hand' for that... but now is lost.

The reason using HasValue is preferred over string.IsNullOrWhiteSpace, is it's a lot more 'friendly', but also then consistent across all the property types...eg not just for strings...

Be interesting to see what's taught on the fundamentals?

sofietoft: I'm actually not sure what they're doing in the fundamentals course 🤔 Would be interesting to find out.

I get that used the "hasValue" is better or friendlier. I can update the samples again.
I'll put it on my list.

Thanks for the input! Much appreciated 💪

PR #3183, opened 2021-05-13
Create signalR.md

Migaroez:
Didn't know where to put this and @dampee told me to put it anywhere, so here you go :)

### DIFF
--- a/Implementation/Custom-Routing/signalR.md
+++ b/Implementation/Custom-Routing/signalR.md
@@ -1,0 +1,138 @@
+---
+versionFrom: 9.0.0
+product: "CMS"
+meta.Title: "Adding a signlR hub"
+meta.Description: "Umbraco ships with signalR installed, find out how to add your own hub(s) to the existing setup"
+
+---
+
+# Todo 
+- Create a hub
+- Define a custom route for it
+- Add the routing to the umbraco composer
+- Test the setup with some javascript
+
+## Create a hub
+We are going to go for the most simple implementation possible, a simple status ping. So create a new class with the following code
+```csharp
+using Microsoft.AspNetCore.SignalR;
+
+public class TestHub : Hub
+{
+    public async Task Ping()
+    {
+        await Clients.All.SendAsync("Pong");
+    }
+}
+```
+
+### Define a custom route
+Next up is defining a custom route, for this we are going to use a `IAreaRoutes` and the base umbrace backend path so we dont have to reserve another path in the settings
+```csharp
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+using Umbraco.Cms.Core.Composing;
+using Umbraco.Cms.Core.DependencyInjection;
+using Umbraco.Cms.Web.Common.ApplicationBuilder;
+using Umbraco.Extensions;
+
+public class TestHubRoutes : IAreaRoutes
+{
+    private readonly IRuntimeState _runtimeState;
+    private readonly string _umbracoPathSegment;
+
+    public TestHubRoutes(
+        IOptions<GlobalSettings> globalSettings,
+        IHostingEnvironment hostingEnvironment,
+        IRuntimeState runtimeState)
+    {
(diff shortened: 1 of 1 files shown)


busrasengul:
Hey @Migaroez !
Thanks for your PR!! It looks great 🎉.

For my reference, you've added versionfrom 9.0.0 on the YAML tag, but is this article for just V9, or can be also used for V8 as well?

Migaroez:
Its v9 only as the signalR setup for .net core or .net framework is different

busrasengul:
That's great! Thank you for clarifying @Migaroez 🙌
We'll get this reviewed asap!

Implementation/Custom-Routing/signalR.md
```diff
@@ -0,0 +1,138 @@
...
+using Umbraco.Cms.Core.DependencyInjection;
+using Umbraco.Cms.Web.Common.ApplicationBuilder;
+using Umbraco.Extensions;
+
+public class TestHubComposer : IComposer
+{
+    public void Compose(IUmbracoBuilder builder)
+    {
+        builder.Services.AddUnique<TestHubRoutes>();
+        builder.Services.AddSignalR();
```
bergmania, line 87: I'm in doubt if this should be called.. It is already called by BackOffice (we use Signal-R for the previews). Calling this twice could potentially be a bad thing.

Alternatively, should you check if it is already added. We do something similar in the CMS a couple of places..

Migaroez: Hey Bjarke, I have looked around, but I can't find how to check if something has been added to the services yet. Could you help me out?

bergmania: I agree it is never easy with build in stuff, but AddSignalR `seems to be safe to call multiple times, as it uses statements like:
```
services.TryAddSingleton(typeof(IHubContext<>), typeof(HubContext<>));
```

Under the hood, it does something similar to this, but for each of the types, it adds to the container, so it is not super-efficient.

Maybe doing something like this will be the best compromise, as we only iterate the collection once:
 ```
 if (!builder.Services.Any(x => x.ServiceType == typeof(IHubContext<>)))
{
    builder.Services.AddSignalR();
}
```

File: Implementation/Custom-Routing/signalR.md [hunk 2]
```diff
@@ -0,0 +1,138 @@
...
+        .build();
+
+    // start the connection
+    connection.start().then(function () {
+        console.info("signalR connection established");
+
+        // connection is established => register our callbacks when the hub sends us something
+        connection.on("Pong", function () {
+            console.log("Pong");
+        });
```
nikolajlauridsen, line 127: You don't have to wait for the connection to be established before you register callbacks, I'd suggest moving this part outside of start, otherwise, you might end up having an absolutely massive start.then() block 😄

nikolajlauridsen:
Hey @Migaroez, first of all, great idea for a SignalR article, I think this is great since it shows some tricks required to run SignalR in Umbraco which might not be entirely obvious 😄 

I tried running through your guide and can confirm that it works, I've added a small suggestion as in inline comment, but if I may, I'd also recommend showing that you can add an interface for your client-side SignalR client since this is generally the recommended way to do it.

For instance, in this example we can add a client interface with a method Pong: 

```C#
public interface ITestHubClient
{
    // Define the methods implemented by the client
    public Task Pong();
}
```

And then we can make the hub of type `Hub<ITestHubClient>`, this way we have a typed client and don't have to rely on a string literal to call the method: 

```C#
public class TestHub : Hub<ITestHubClient>
{
    public async Task Ping()
    {
        await Clients.All.Pong();
    }
}
```

Migaroez:
@bergmania thank you for the suggestion. Updated the code snippet

@nikolajlauridsen I initially wanted to keep the example as basic as possible. But yeah, typed hubs are the way to go. Added it in, also made me think about the wording, so updated/added some comments.
And the callback registration was a stupid oversight. Thank you for pointing that out.

PR #3186, opened 2021-05-15
Update index.md

onlinegroeicoach:
The article doesn't mention it is still working or has been tested in Umbraco 8, I just did it and can confirm that it still works in 8 (tried it in version 8.9.1).

### DIFF
--- a/Extending/Macro-Parameter-Editors/index.md
+++ b/Extending/Macro-Parameter-Editors/index.md
@@ -185,3 +185,6 @@ var imagePosition = Model.MacroParameters["imagePosition"];
 var imagePosition = Model.GetParameterValue<string>("imagePosition","full-width");
 }
 ```
+### Umbraco 8 compatibility
+
+Just followed this page and implemented the sample in Umbraco 8.9.1 and although these instruction may have been written for version 7 this still works in 8.
--- a/Extending/Macro-Parameter-Editors/index.md
+++ b/Extending/Macro-Parameter-Editors/index.md
@@ -186,6 +186,3 @@ var imagePosition = Model.MacroParameters["imagePosition"];
 var imagePosition = Model.GetParameterValue<string>("imagePosition","full-width");
 }
 ```
-### Umbraco 8 compatibility
-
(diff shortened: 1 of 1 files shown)


busrasengul:
Hey @onlinegroeicoach !!
Thank you so much for updating the docs. 
We're using YAML tags for version controlling of every article. If you have already tested this for v8.9.1 let us try the other V8.x compatibility and create a tag instead of commenting on the article. By doing this, whoever visits the page would know it works on every V8 too. Does this sound good to you?

sofietoft:
Hi @onlinegroeicoach !

Thanks again for the PR!
So, instead of the note, I've added some extra metadata to this article, which should make it a little clearer that the contents of the article is valid from 7 to 8.15.
It will show up like this: 
image

What do you think of that approach? 🤔

onlinegroeicoach:
Thanks for the update, think I wasn’t able to update the meta data (or didn’t know how) but this way it certainly better.

I might have some extra input on how to make the a re-usable implementation and how to also have a dropdownlist in addition to the radiobuttonlist.

sofietoft:
No worries 😉 Our versioning system is not super transparent at the moment, I'm afraid!

Sounds great that you're looking into adding even more stuff! Would you be adding that to this PR as well?

busrasengul:
Hey @onlinegroeicoach 
Sorry for the delay in this. I've tweaked your PR and now it's merged in! Would you mind giving me your our.umbraco.com account details so I can grant you with C-Trib badge 🎉

onlinegroeicoach:
Hi Busra,

Ok, great ! I login into our.umbarco.com with my ***@***.*** e-mail address.

I might have an addition for this page, in the sample a macro paramater editor for a radiobuttonlist is created.

I've done the same for also a dropdownlist and a checkboxlist (multiselect), stil have a small challenge with the later (checkboxlist) but when fixed I can give 2 more examples of parameter editors.

Personally the use of macro with parameters is really helpfull, for instance I have hunderds of photos on my website and I display them using I macro, now the google guidelines have changes I only have one place were I need to make changes (calculate width/height and add srcset attribute) to every image. When I would have used the default Image control functionality this would have been a maintenance nightmare.

Regards,
Patrick

________________________________

PR #3188, opened 2021-05-16
Rename Index.md to index.md

jemayn:
The links at the bottom of this page 404s because this one is named "Index" instead of "index", this causes Our to fail and add "Index" as an additional URL path item, so links will for example become 
https://our.umbraco.com/Documentation/Reference/Management/Services/ContentTypeService/Index/Retrieving-content-types
instead of 
https://our.umbraco.com/Documentation/Reference/Management/Services/ContentTypeService/Retrieving-content-types

Renaming this file will fix that 🙂


busrasengul:
Thank you @jemayn 
I'm merging this right in 👍

PR #3192, opened 2021-05-17
Update example code of section trees

ronaldbarendse:
Besides fixing some spacing/styling in the examples, I've ensured the `GetMenuForNode` method has only a single exit point: this makes it easier to read and adjust (after copying the example code to your own solution).


sofietoft:
Looks good @ronaldbarendse !
And apologies for the long wait!

Just making sure; are these changes you've made here also relevant for the upcoming Umbraco 9 release?
If they are, would you be up for making the relevant changes to the v9 version of this article as well? 😇 
You can find it here: https://github.com/umbraco/UmbracoDocs/blob/update-trees-examples/Extending/Section-Trees/trees-v9.md

ronaldbarendse:
@sofietoft There is no hurry, as it's mostly minor styling fixes in example code 😉 I'm not very familiar with this part of the Umbraco 9 codebase, but I know it would at least need completely different example code for handling events: probably better to do that as part of another PR...

sofietoft:
Sounds good!
Alright - the article has actually already been updated for v9 - samples and all.
So it might "just" be a matter of testing with your new additions here!

I'll pass it on to whoever created the v9 version of it 😄 Thanks again!

ronaldbarendse:
Ah yes, PR https://github.com/umbraco/UmbracoDocs/pull/3213 already updated the v9 documentation and some of the examples are 1:1 copies that can also get the same minor clean-up/fixes! I will add those changes to this PR 👍🏻

ronaldbarendse:
@sofietoft I've updated the v9 version 😄

Maybe @nikolajlauridsen can double check whether throwing this exception (instead of returning an empty collection) in `GetTreeNodes` won't cause any issues:
https://github.com/umbraco/UmbracoDocs/blob/2a02cea95d97ab2a19a9630041101c58145b6af6/Extending/Section-Trees/trees-v9.md#L123-L124

nikolajlauridsen:
Just had a look at it, gotta say though that I've barely touched trees before. 

But throwing the sample code in and testing it, it seems that throwing a NotSupportedException is not a great idea, since it will result in a weird pop up, that makes very little sense:
image

I'm not sure what's the best thing to return, but at the end of the day, since this is a custom section the error shouldn't occur, however I found two options, but there might be a third, better option.

#### Option 1: 

As you mentioned you can just return an empty collection: 

```C#
return new TreeNodeCollection();
```

And nothing will really happen, in this example, i think it's the best alternative, since the root level should normally still be rendered.

#### Option 2: 

If you want there to be an error, you can return an `UmbracoErrorResult` like so: 

```C#
return new UmbracoErrorResult(HttpStatusCode.InternalServerError, "Only root node is supported");
```

This will result in a popup like this 
image

Which looks a bit weird, but at least it conveys what went wrong. 

Another thing I also noticed is that the example doesn't have a namespace, even though it's the "full file", I don't know if this is intentional, at the end of the day people will have to change it to whatever their namespace is regardless, but personally I think it would be easier to just edit the existing namespace to whatever you want instead of having to create it. 

I know that most, if not all, editors have refactoring tools for this, but someone might not know that, and then it would be kinda annoying to have to cut, add namespace and paste again.

The changes looks fine to me though 😄

nul800sebastiaan:
Sorry, didn't mean to close this one, I did something weird to the git repo yesterday that accidentally deleted a few too many branches. Reopened now!

ronaldbarendse:
@nikolajlauridsen Returning the empty collection (option 1) is the least intrusive, so lets update that and have this merged!

jonat123:
@ronaldbarendse great! I'll merge this in right away 😄

PR #3194, opened 2021-05-17
Remove mentions of umbracoUseDirectoryUrls

soreng:
The usage of umbracoUseDirectoryUrls has been removed in v8 - but not from the docs :)


busrasengul:
Thank you @soreng 
It looks right you're right! But I'll make sure to get this reviewd asap!!

#h5yr

dampee (approved):
Overlook it at the rewrite i think  :)  Good catch!

PR #3195, opened 2021-05-17
Change -P to -p

skttl:
According to dotnet new umbraco --help, it has to be a lowercase p


busrasengul:
Thank you @skttl !!
This is getting reviewed asap! 
#h5yr!

jonat123:
@skttl Thanks for the PR! 
Just tested it out with a uppercase P instead of a lower case and it indeed fails, when trying to run the command.
It works fine with a lowercase p.

Good spotted!! 

I'll get this PR merged in right away :-)

PR #3197, opened 2021-05-18
Create a basic site tutorial

eshanrnh:
Notes on platform limitations:
Things we need to double-check on Our once it's been merged.
- You've used "{: data-fancybox}" a couple of times - not sure how this will look on Our. We will need to double-check once it's been merged.
- "Prev" and "Next" options - I don't think the current format will work on Our, so we might need to consider something different!


jonat123 (commented):
Looks good, added two small comments :)

PR #3202, opened 2021-05-19
Backoffice-Sections

eshanrnh:


sofietoft (commented):
There are a few links in this article, that might need to be updated at a later point.

Would it make sense to add the ["updated-links" metadata](https://our.umbraco.com/documentation/Contribute/How-to-add-a-new-verion/#metadata-requirements) to the article? That way we can identify the articles when we're going through all links.

eshanrnh:
Done. Added "updated-links" metadata.

PR #3204, opened 2021-05-19
Update index-v9.md

erikjanwestendorp:
Fix image path


dampee:
hey @umbrabot 

It's a very friendly notice, but I think the Documentation curators can handle this.  🙄 😄

dampee (changes requested):
Hi @erikjanwestendorp 

I don't think this change is correct.  If you look in the code base the file IS with an upper case extention. 

afbeelding

In case we move this to a case sensitive file filesystem, it has to match the filename.
So, we can close the PR or we can update the filename too.

What was the reason you suggested this change?

erikjanwestendorp:
@dampee I suggested the changes because the image doen't appear on the v9 [page](https://our.umbraco.com/documentation/Extending/Database/index-v9). The image is displayed correctly on the v8 [page](https://our.umbraco.com/documentation/Extending/Database/). On the v8 version the image extension is lower case. https://github.com/umbraco/UmbracoDocs/edit/main/Extending/Database/index.md

dampee (approved):
feels cleaned up :-)

PR #3207, opened 2021-05-19
Add sendingcontentnotification example

jemayn:
A v9 example of automatically setting values when creating a new node.

Used to be an EditorModelEvent, now it's a SendingContentNotification event.

### DIFF
--- a/Reference/Events/EditorModel-Events/index-v9.md
+++ b/Reference/Events/EditorModel-Events/index-v9.md
@@ -1,0 +1,73 @@
+---
+keywords: EditorModelEventManager setting default values defaultvalue SendingContentNotifications
+versionFrom: 9.0.0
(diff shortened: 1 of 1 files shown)


busrasengul:
🎉🎉🎉
Great article @jemayn 

I'll make sure we review this ASAP then merge in!

busrasengul:
Hey @jemayn 
This article is great but we already have one under Notifications for V9. It'd be great if you could check your observations against that and maybe make a new PR?

Thank you 🙌

jemayn:
Hey @busrasengul 

I made this on one of the v9 learnathon meetups because there wasn't any information on the new way of doing things, I see that a lot more comprehensive docs were added shortly after so makes sense to just get rid of this one 🙂

PR #3208, opened 2021-05-19
Sections docs updates based on learn-a thon

crgrieve:
* Adding full code snippet including usings.
* Explaining we also need a constructor when we mention the methods we need to implement as part of inheriting TreeController.
* Changing the legacy "developer" setting reference as it may confuse folks.


busrasengul:
This was very much needed!!

Thank you Carole!
I'll get this merge in now💪🏻

PR #3214, opened 2021-05-20
Documentation updates for Forms V9.

AndyButland:
I think this covers everything that has changed in V9.  The main things are:

- Developer documentation for extending via creating custom field types, workflow types etc.  All the code samples needed some amends, so I've created "v9" versions of all of these.
- Configuration is quite different in .NET Core even though most of the settings have carried over, so I've created a V9 version of this.

Otherwise have read through and made a few other small amends, where something isn't available in V9 or is just subtly different to not justify a new version specific page.

Given the UI hasn't changed, I don't expect the editor documentation to need updating.

### DIFF
--- a/Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
+++ b/Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
@@ -11,9 +11,7 @@ With Umbraco Forms it's possible to customize the functionality with various con
 
 ## Editing configuration values
 
-Configuration in version 9 has moved from the `UmbracoForms.config` file used in previous versions, replaced by standard .NET configuration.
-
-By default this is held in an `appSettings.json` file found in the root of your Umbraco website.  If configuration has been customized to use another source, then the same keys and values discussed in this article can be applied there.
+All configuration for your Umbraco Forms is held in an `appSettings.json` file found at the root of your Umbraco website.  If the configuration has been customized to use another source, then the same keys and values discussed in this article can be applied there.
 
 The convention for Umbraco configuration is to have package based options stored as a child structure below the `Umbraco` element, and as a sibling of `CMS`.  Forms configuration follows this pattern, i.e.:
 
--- a/Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
+++ b/Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
@@ -196,6 +196,6 @@ Google has renamed these recently and the `Site Key` refers to `RecaptchaPublicK
 
 #### SiteKey & PrivateKey
 
-Both of these configuration values are needed in order to use the "*reCAPTCHA V3 with Score*" field type implementing ReCaptcha V3 from Google. This field type is available in Umbraco Forms from v8.7+. 
+Both of these configuration values are needed in order to use the "*reCAPTCHA V3 with Score*" field type implementing ReCaptcha V3 from Google.
 
 You can obtain both of these values after signing up to create a ReCaptcha key here:  https://www.google.com/recaptcha/admin.
--- a/Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
+++ b/Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
@@ -31,9 +31,9 @@ The convention for Umbraco configuration is to have package based options stored
 }
 ```
 
-All configuration for Forms is optional.  In other words, all values have defaults that will be applied if no configuration is available for a particular key.
+All configuration for Forms is optional. In other words, all values have defaults that will be applied if no configuration is available for a particular key.
 
-For illustration purposes though, the following structure represents the full set of options for configuration of Forms, along with the default values.  This will help when you need to provide a different setting to understand where it should be applied.
+For illustration purposes, the following structure represents the full set of options for configuration of Forms, along with the default values. This will help when you need to provide a different setting to understand where it should be applied.
 
 ```json
   "Forms": {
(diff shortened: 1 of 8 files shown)


Add-ons/UmbracoForms/Developer/Extending/Adding-a-Type-v9.md
```diff
@@ -0,0 +1,162 @@
...
+}
+```
+
+Also look in the reference chapter for complete class implementations of workflows, fields and export types.
+
+## Overriding default providers in Umbraco Forms
+
+It is possible to override and inherit the original provider, be it a Field Type or Workflow etc. The only requirement when inheriting a fieldtype that you wish to override is to ensure you do not override/change the Id set for the provider, and make sure your class is public.
+
+Here is an example of overriding the Textarea field aka Long Answer that is taken from Per's CodeGarden 17 talk, which has been updated for Forms 9.
```
sofietoft, line 125: Shortened this sentence a bit, as the info about Per's Codegarden talk seemed a bit out of place!

```suggestion
Here is an example of overriding the Textarea field aka Long Answer.
```

Add-ons/UmbracoForms/Developer/Extending/Adding-an-Event-Handler-v9.md
```diff
@@ -0,0 +1,79 @@
+---
+versionFrom: 9.0.0
+meta.Title: "Adding Event Handlers in Umbraco Forms"
+meta.Description: "See an example of validating a form server-side"
+state: complete
+verified-against: beta001
+---
+
+# Adding a server-side event handler to Umbraco Forms
```
sofietoft, line 9: I see "Events" referenced quite a few places in this article - also in the code snippet...
It's my understanding that everything Events in v8 has been replaced by Notifications in v9 - might be different for Forms, but I though I'd raise it anyways.
Should it be "server-side Notifications" instead of "server-side even handlers"? 🤔

Add-ons/UmbracoForms/Developer/Forms-in-the-Database/index.md
```diff
@@ -8,6 +8,8 @@ As of Umbraco Forms version 8.5.0 it is possible to persist all forms data in th
 
 In this article you will find instructions on how to migrate your Umbraco Forms data to the database.
 
+As of Umbraco Forms version 9.0.0 it is *only* possible to store form data in the database.  As such, if upgrading to Umbraco 9 and using Forms, you should first migrate the forms to the database using Forms 8.
```
sofietoft, line 11: Does this article even make sense for Umbraco 9?
Wouldn't it be better to instead have an article (or even just a section somewhere) that explains that all forms data is stored in the database.
And then have some notes in migration/upgrading guides on how to migrate the forms to the database? 🤔

Add-ons/UmbracoForms/Developer/IFileSystem/index.md
```diff
@@ -4,7 +4,7 @@ needsV8Update: "true"
 ---
 
 # Storing Form Files with IFileSystem
-Umbraco Forms available in version 4.4.0 and newer allows you to use an IFileSystem to abstract how and where the physical JSON files such as Forms, Workflows and PreValues.
+Umbraco Forms available in version 4.4.0 and up to version 8 allows you to use an `IFileSystem` to abstract how and where the physical JSON files such as Forms, Workflows and PreValues.
```
sofietoft, line 7: As I understand this, with Umbraco Forms 9 you can no longer store form files in with the IFileSystem - is that correct?

In that case, we should add the `versionRemoved` metadata to this article, with value 9.0.0.

sofietoft (commented):
Looks great @AndyButland !

I've taken the liberty to update some grammar here and there.
Also added some comments and suggestions! Let me know if something doesn't make sense.

Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
```diff
@@ -0,0 +1,180 @@
...
+---
+
+# Configuration
+With Umbraco Forms it's possible to customize the functionality with various configuration values.  
+
+## Editing configuration values
+
+Configuration in version 9 has moved from the `UmbracoForms.config` file used in previous versions, replaced by standard .NET configuration.
+
+By default this is held in an `appSettings.json` file found in the root of your Umbraco website.  If configuration has been customized to use another source, then the same keys and values discussed in this article can be applied there.
```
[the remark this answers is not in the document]
AndyButland: I think your change is fine.  There's plenty of other places where changes to configuration between .NET Framework and .NET Core are discussed.

Add-ons/UmbracoForms/Developer/Configuration/index-v9.md
```diff
@@ -0,0 +1,203 @@
...
+
+Google has renamed these recently and the `Site Key` refers to `RecaptchaPublicKey` and `Secret Key` is to be used for `RecaptchaPrivateKey`
+
+### reCAPTCHA v3 field type configuration 
+
+#### SiteKey & PrivateKey
+
+Both of these configuration values are needed in order to use the "*reCAPTCHA V3 with Score*" field type implementing ReCaptcha V3 from Google. This field type is available in Umbraco Forms from v8.7+. 
+
+You can obtain both of these values after signing up to create a ReCaptcha key here:  https://www.google.com/recaptcha/admin.
```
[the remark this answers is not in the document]
AndyButland: They are both available yes (I think it's possible people can have keys for the older version of reCAPTCHA, so there didn't seem a reason to remove it).  I think they can decide which to use if they have the option - we probably don't need to take a view on that in the configuration documentation.  Agree no need for mentioning V8 though - will remove.

Add-ons/UmbracoForms/Developer/Extending/Adding-an-Event-Handler-v9.md
AndyButland: Agreed, I'll update the titles and other references (but keep the file name so it's clear it's a version specific change).

Add-ons/UmbracoForms/Developer/Forms-in-the-Database/index.md
AndyButland: I think this is probably OK - i.e. if you read this, and you are on V9, there's just this particular note to be aware of.  But feel free to re-work this if you prefer.

Add-ons/UmbracoForms/Developer/IFileSystem/index.md
AndyButland: That's right, I've added.

PR #3215, opened 2021-05-21
Ipub Content Collections. Updated for v9.

Jay-umbr:
-ContentType replaced DocumentTypeAlias
-GetProperty replaced GetPropertyValue

### DIFF
--- a/Reference/Querying/IPublishedContent/Collections/index-v9.md
+++ b/Reference/Querying/IPublishedContent/Collections/index-v9.md
@@ -1,0 +1,39 @@
+---
+state: complete
+updated-links: true
+verified-against: alpha-3
+versionFrom: 9.0.0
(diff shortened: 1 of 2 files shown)


sofietoft (commented):
Thanks for the PR @Jay-umbr !

Something looks a little off here...
You've created an index-v9 in `PublishedContent/Collection` with the heading **IPublishedContent**, 
and then you've updated the index-v9 article in `IPublishedContent/` with the heading **IPublishingContent Collections**... 🤔 
Aren't the files places in the wrong locations? Shouldn't the file with **IPublishedContent Collection** heading be in the `IPublishingContent/Collections` folder? 🤔

Jay-umbr:
Oops. Should be OK now @sofietoft

PR #3217, opened 2021-05-21
Updating config folder

crgrieve:
Hello,

I noticed a few of the articles suggest we update files in the /config folder, but in v9 we don't have that as standard. It looks like those needed in: https://our.umbraco.com/documentation/Extending/Section-Trees/trees are inside /umbraco/config instead, and this works when running through the tutorial. So I have updated these in this PR. However...

I noticed this article: https://our.umbraco.com/documentation/Extending/Health-Check/index-v9 says _"Don't add the translations to ~/Umbraco/Config/Lang files, the correct location is ~/Config/Lang/*-*.user.xml"_, so maybe this should just be /config after all? If so I think we should mention that it won't exist as default anymore and folks will need to add it.

Feedback appreciated :)

Thanks,
Carole


busrasengul:
That's a good idea Carole! 
Thank you so much..
#h5yr

PR #3222, opened 2021-05-26
Events --> Notifications

sofietoft:
Moved all updated Events articles to a new Notifications section in the Reference section.

Also added notes to the Events articles with links to the v9 equivalent.


nikolajlauridsen (approved):
Looks all good to me 👍

PR #3223, opened 2021-05-26
add nc example

jemayn:
Added an example on rendering nested content elements based on partial views. A guy was asking on the community slack and I was a bit surprised we didn't show any examples that supports multiple different elements, so here is one 🙂


busrasengul:
Thanks so much @jemayn 🎉
Just to make sure, is your example is for V7 only?

jemayn:
@busrasengul 

Yea a few changes would have to be made for v8, but the concepts should work. Don't have a v8 example at hand though, it is probably enough to change `Model.GetPropertyValue` to `Model.Value` 🤔

busrasengul:
Thanks for confirmation @jemayn 
We'll get this reviewed asap ✌

Issue #3218, opened 2021-05-25
Use Deploy dashboard as the default way of performing Umbraco Deploy commands on Cloud


hjaltedaniel-umbraco:
## What type of issue is it? (Choose one - delete the others)
Discussion

## What article/section is this about?
https://our.umbraco.com/documentation/Umbraco-Cloud/Set-Up/Power-Tools/Manual-extractions/
https://our.umbraco.com/documentation/Umbraco-Cloud/Set-Up/Power-Tools/generating-uda-files/
https://our.umbraco.com/documentation/Umbraco-Cloud/Troubleshooting/
(Just a few examples, there are probably more around different corners of the docs)

## Describe the issue
With Umbraco Deploy 4, the long-awaited Deploy dashboard was introduced. This gives the user an easy way of extracting schema to the database, exporting deployment artifacts, and clearing cached signatures, without having the un-friendly experience of writing somewhat arbitrary commands in a tool you might not know (Kudu).

With this issue, I propose a discussion of re-writing the docs, so that the Deploy dashboard is the default way of performing the different operations that are possible to do with Umbraco Deploy. The dashboard itself is already fairly well documented in the [deploy section of the docs ](https://our.umbraco.com/documentation/Add-ons/Umbraco-Deploy/deployment-workflow/#deploy-dashboard).

jonat123:
@hjaltedaniel-umbraco I agree that with the new deploy dashboard it makes complete sense that it should be the default way we recommend using the different operations like ehco>deploy/Export and etc. rather than kudu on Umbraco Cloud 🙂

Just tagging @sofietoft so she can chime in as well when she is back tomorrow :)

sofietoft:
I think it's a great idea to recommend people using the Deploy Dashboard instead of the commands through KUDU! 💪

Now, I just wanna make sure I've understood correctly...
You're volunteering to update the articles, @hjaltedaniel-umbraco ? 😉 😄

hjaltedaniel-umbraco:
I'll look into submitting a PR when I find time for it, then we can take it from there 👍

PR #3229, opened 2021-05-27
Creating new section about deploy operations available from the deploy dashboard

hjaltedaniel-umbraco:
This PR addresses #3218

Creating a "Deploy operations" section in the documentation under "Umbraco Cloud / Deployments".
With the new Deploy dashboard, it does not make sense to just change things under "Power tools" in my opinion.

I have also changed the wording around the different deploy operations to match what the operations are called in the dashboard.


sofietoft:
Thanks for thr PR @hjaltedaniel-umbraco ! Much appreciated 🙌 

I'll make sure we review it as soon as possible.
Quick question to the structure of the articles: You're adding a bunch of new articles, which is really nice.
What do you suggest we do with the already existing articles, that are documenting more or less the same thing?

hjaltedaniel-umbraco:
My thought about it is that this is just the beginning when it comes to addressing #3218. As far as I can see, there are still some dependencies on the old articles about this subject in the documentation. I suggest that these dependencies are resolved when these articles are published. A suggestion in the interim period could be to add an infobox on the old articles, suggesting that you can use the Deploy dashboard instead?

Another note is that the Deploy dashboard is only relevant to users that have their Deploy version updated to 4+. This should be mostly everyone on Cloud. However, AFAIK there are still some users that are dependent on older versions of Deploy - especially users running their v7 projects on Cloud, though also a few on v8.

jonat123:
@hjaltedaniel-umbraco  Looks good to me.

One thing just to be sure that we are on the same page, what you are suggesting with the old articles that mentions how to access kudu and do operations for deploy, that we should add a note in them stating that from Deploy 8.4.x you can use the Deploy dashboard instead? :)

hjaltedaniel-umbraco:
That sounds like a good interim solution 😅

jonat123:
Cool! Would you be up for adding those notes in this PR? 😇

hjaltedaniel-umbraco:
@jonat123 Done ✅

jonat123:
Perfect @hjaltedaniel-umbraco ! I'll merge it in right away! :D

PR #3232, opened 2021-05-27
Update reset-admin-password.md

TwoMoreThings:
I'm suggesting a different way which has been accepted a number of occasions recently on Slack discussions. I think it's less technical and easier/quicker to implement, and safer as it doesn't result in the site being in a vulnerable state during the process.

### DIFF
--- a/Reference/Security/reset-admin-password.md
+++ b/Reference/Security/reset-admin-password.md
@@ -4,34 +4,74 @@ versionFrom: 8.0.0
 
 # Reset admin password
 
+This guide assumes you can't use the 'Forgotten password?' link in the login screen to receive a password reset email, or that the Forgotten password link isn't present on the login screen. 
+
+This also assumes access to the website database if you need to confirm the admin user's email address and also that you have access to change configuration settings in the `web.config` file. Usually this would be done running on a local PC.
+
+We are not going to send an email, but instead generate the email and use the link from the email to reset the password.
+
+This technique will also work for Umbraco versions 7.5.15 and above
+
+## Step one: Confirm the default admin user email address
+
+*Skip this step if you already know any admin user's email address*
+
 There is one default admin user in any Umbraco installation. This is the first user of the system.
 
-While there are multiple blogposts and even a [package](https://our.umbraco.com/packages/developer-tools/umbraco-admin-reset/) to reset the password of the admin user. There is an [official way](https://twitter.com/Shazwazza/status/1141594930550206464) to reset credentials from your lost admin password.
+Open the database and go to the `umbracoUser` database table.
+Search for the user where the column `id` equals `0`.
+Note the email address from the `userEmail` column.
 
-## Step one: Clear the configuration status in the web.config
+## Step two: Configure the SMTP settings
 
-The first step is to clear the configuration status in the `web.config` file. This is done to trigger the installation wizard.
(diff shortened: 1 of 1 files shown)


busrasengul:
Hey @TwoMoreThings 
Thank you so much for your PR!! 🙌
I remember some of those Slack conversations, and your suggestion definitely looks great. 
I'll be sure one of us reviews this asap!
#h5yr!!

ronaldbarendse:
@TwoMoreThings The currently documented steps (clearing the `Umbraco.Core.ConfigurationStatus` setting, changing the users password to `default` and running the installer) work great, so lets at least make sure this isn't removed from the documentation 😄 

Using the 'forgotten password' feature might indeed be easier to use though, especially if you setup your projects to always use localhost as SMTP-server in your codebase and transform/update this to the correct SMTP-settings when deploying to production/live.

Just to nit-pick: I wouldn't call the SMTP programs _proxies_: they are local/development SMTP servers.

Reference/Security/reset-admin-password.md
```diff
@@ -4,34 +4,74 @@ versionFrom: 8.0.0
...
-That means that in your web.config file the appsettings should look like this:
 
-```csharp
-<appSettings>
-    <add key="Umbraco.Core.ConfigurationStatus" value="" />
-</appSettings>
+### Method A: Using a pickup directory
+
+Open the web.config file at the root of your site.
+locate the section `<mailsettings>` and configure the `deliveryMethod` and specify an existing location for the email to be saved.
```
sophie-neale, line 36:
```suggestion
Locate the section `<mailsettings>` and configure the `deliveryMethod` and specify an existing location for the email to be saved.
```

sophie-neale (commented):
Suggested a tiny change for a tiny typo 👀 🙂

TwoMoreThings:
> Just to nit-pick: I wouldn't call the SMTP programs _proxies_: they are local/development SMTP servers.

TLDR; I'm happy for that to be changed. 

I think I picked up the 'proxy' term from docs from one of the two programs mentioned. An alternative might be to call it a 'dummy email server', or 'fake email server' or something like that. I specifically didn't want to call it an email server (even a local email server) as that might sound like a security risk or too complex for someone who's a less technical reader, and I wanted it to be clear that it wouldn't actually send any emails. I also wanted to avoid the terms stub or mock as too technical

dampee:
> @TwoMoreThings The currently documented steps (clearing the `Umbraco.Core.ConfigurationStatus` setting, changing the users password to `default` and running the installer) work great, so lets at least make sure this isn't removed from the documentation 😄

As far as I remember, this is the 'official' way of resetting a password. Using the e-mail forgot password is a workaround. It should definitely not be removed from the documentation!

sofietoft:
Hi @TwoMoreThings  👋 

After chatting with the Docs team, I have a suggestion for you:
How about you add the methods you suggest to a separate article? That way we keep the official way of doing this, while offering other ways as well.

How does that sound?

sofietoft:
Hi @TwoMoreThings !

It's been a while since there has been any activity on this one, so I'll be closing it down for now.
If you're still up for it, it would be awesome if you could add your steps here to a new article and make a new PR as well 💪 

Thanks again for helping us keeping the docs up to date! 💪

PR #3236, opened 2021-05-31
Backoffice-Login

eshanrnh:
1. Created v9 version
2. Updated file names and paths.

### DIFF
--- a/Fundamentals/Backoffice/Login/index-v9.md
+++ b/Fundamentals/Backoffice/Login/index-v9.md
@@ -32,16 +32,20 @@ The login screen features a greeting, which you can personalize by changing the 
 </area>
 ```
 
-You can customize other text in the login screen as well, grab the default values from `~/Umbraco/Config/Lang/en.xml` and copy the keys you want to translate into your `~/Umbraco/Config/Lang/MYLANGUAGE.xml` file.
+You can customize other text in the login screen as well, grab the default values from `~/umbraco/config/lang/en.xml` and copy the keys you want to translate into your `~/umbraco/config/lang/MYLANGUAGE.xml` file.
 
 ## Password reset
 
-The **Forgotten password?** link allows your backoffice users to reset their password. To use this feature, you will need to add the following key to the `"security"` section in the `appsettings.json` file:
+The **Forgotten password?** link allows your backoffice users to reset their password. To use this feature, you will need to add the following key to the `Umbraco.Cms.Security` section in the `appsettings.json` file:
 
 ```json
-"Security": { 
+"Umbraco": {
+    "CMS": {
+      "Security": { 
         "AllowPasswordReset": true
-      },
+      }
+   }
+}
 ```
 
 Set it to `true` to enable the password reset feature, and `false` to disable the feature.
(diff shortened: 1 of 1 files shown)


Fundamentals/Backoffice/Login/index-v9.md
```diff
@@ -0,0 +1,100 @@
...
+    <key alias="greeting1">Happy manic Monday </key>
+    <key alias="greeting2">Happy tubular Tuesday</key>
+    <key alias="greeting3">Happy wonderful Wednesday</key>
+    <key alias="greeting4">Happy thunderous Thursday</key>
+    <key alias="greeting5">Happy funky Friday</key>
+    <key alias="greeting6">Happy Caturday</key>
+</area>
+```
+
+You can customize other text in the login screen as well, grab the default values from `~/Umbraco/Config/Lang/en.xml` and copy the keys you want to translate into your `~/Umbraco/Config/Lang/MYLANGUAGE.xml` file.
```
sofietoft, line 35: Not sure how case sensitive these paths are, but changes them to lower case as a suggestion here - just in case! 😅 

```suggestion
You can customize other text in the login screen as well, grab the default values from `~/umbraco/config/lang/en.xml` and copy the keys you want to translate into your `~/umbraco/config/lang/MYLANGUAGE.xml` file.
```

Fundamentals/Backoffice/Login/index-v9.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Password reset
+
+The **Forgotten password?** link allows your backoffice users to reset their password. To use this feature, you will need to add the following key to the `"security"` section in the `appsettings.json` file:
```
sofietoft, line 39: I think it makes sense that we add all the sections/groups that a config key goes into.
Not sure about the format though - what do you think of doing it like this: `Umbraco.Cms.Security` ? 🤔 

```suggestion
The **Forgotten password?** link allows your backoffice users to reset their password. To use this feature, you will need to add the following key to the `Umbraco.Cms.Security` section in the `appSettings.json` file:
```

Fundamentals/Backoffice/Login/index-v9.md
```diff
@@ -0,0 +1,100 @@
...
+      },
+     "Security": { 
+        "AllowPasswordReset": true
+      },
+      "Content": {
+        "LoginBackgroundImage": "assets/img/login.jpg"
+      }
+    }
+ }
+```
```
sofietoft, line 100: I don't think there is a need to add the full config file in this article.
The configs and settings described in this article, are all something you can do independently from each other.
Showing each individual config setting including keys and groups should be enough.

What do you think?

Fundamentals/Backoffice/Login/index-v9.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Password reset
+
+The **Forgotten password?** link allows your backoffice users to reset their password. To use this feature, you will need to add the following key to the `"security"` section in the `appsettings.json` file:
+
+```json
+"Security": { 
+        "AllowPasswordReset": true
+      },
+```
```
sofietoft, line 45: I think it makes sense if we for each of these configs, show all the groups/sections as well.

So, it would be like this instead:

```
"Umbraco": {
    "CMS": {
      "Security": { 
        "AllowPasswordReset": true
      }
   }
}

It shows all sections, and then the key with the correct value.
What do you think of doing it like that instead?
```

sofietoft (commented):
Looks great @eshanrnh !

I've made some suggestions to updating the code snippets and also, to how we mention the various sections in the config keys need to be placed into.

Let me know what you think!

Fundamentals/Backoffice/Login/index-v9.md
eshanrnh: Yes, makes sense.

Fundamentals/Backoffice/Login/index-v9.md
eshanrnh: Yeah, Good idea :)

PR #3239, opened 2021-05-31
Update index.md

idseefeld:
fix two typos


busrasengul:
Good catch @idseefeld 🙌
Merging rigth in!

PR #3240, opened 2021-05-31
Update index.md

idseefeld:
fix example


busrasengul:
A good catch again! 
Thank you @idseefeld

PR #3244, opened 2021-06-03
Add outgoing IPs for new Cloud Infrastructure

mbentancour:
When we migrate projects to the new infrastructure, the outgoing IPs will change. Hence, we need to have the new IPs documented to use as reference.


busrasengul:
So many new IPs to add!! 
Thanks @mbentancour 
We'll get this checked then merged asap!

RobertCopilau (approved):
IPs have been checked and are valid.
Merging

PR #3251, opened 2021-06-08
Update to regex to handle mp* files

Frost117:


jonat123:
@Frost117 Looks good, i'll merge it in right away!

PR #3252, opened 2021-06-08
Verified against umbraco 8

erikjanwestendorp:
I saw in the comments field of this [sheet ](https://docs.google.com/spreadsheets/d/1YyARbCbLKhqR12cfbATxyCEZ6fYYxwBXeOw-jNUyWXc/edit#gid=0)that this article was not yet verified against Umbraco 8. I think this article also fits Umbraco 8, so I removed the needsV8Update tag. Maybe the image could use an update 😉. In addition, the link to Stephan's slides is somewhat outdated.

### DIFF
--- a/Reference/Routing/Request-Pipeline/index.md
+++ b/Reference/Routing/Request-Pipeline/index.md
@@ -1,6 +1,5 @@
 ---
 versionFrom: 7.0.0
-needsV8Update: "true"
 ---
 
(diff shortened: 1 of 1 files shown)


busrasengul:
Thank you @erikjanwestendorp 
I'll merge this in for now and create a new file for the older versions for outdated content :)

PR #3261, opened 2021-06-09
Update index.md

kmeilander:
Added details for how to override image processor security settings via an appsetting so you can override these values in the azure portal.

### DIFF
--- a/Extending/FileSystemProviders/Azure-Blob-Storage/index.md
+++ b/Extending/FileSystemProviders/Azure-Blob-Storage/index.md
@@ -125,7 +125,7 @@ You have to manually add `prefix="media/"` to the service element, otherwise Ima
 You have now successfully setup Azure Blob Storage with your Umbraco site.
 
 ### Application Setting Override
-After manually updating the security.config file, you can override the `Container` and `Host` values for your blob via an appsetting value.  This allows you to set these values in the Azure Portal.    Be sure that your security.config has some value configured for the container, and the host value is a valid Uri or else the site wont load. 
+After manually updating the security.config file, you can override the `Container` and `Host` values for your blob via an appsetting value.  Moving to application settings can be useful in some scenarios, for example allowing these values to be set in the Azure Portal applications settings rather than stored in the security.config file.    Be sure that your security.config has some value configured for the container, and the host value is a valid Uri or else the site wont load. 
 
 ```
 <appSettings>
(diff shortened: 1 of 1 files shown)


busrasengul:
Thank you @kmeilander 

I'll make sure one of us reviews this asap!

Extending/FileSystemProviders/Azure-Blob-Storage/index.md
```diff
@@ -124,6 +124,16 @@ You have to manually add `prefix="media/"` to the service element, otherwise Ima
 
 You have now successfully setup Azure Blob Storage with your Umbraco site.
 
+### Application Setting Override
+After manually updating the security.config file, you can override the `Container` and `Host` values for your blob via an appsetting value.  This allows you to set these values in the Azure Portal.    Be sure that your security.config has some value configured for the container, and the host value is a valid Uri or else the site wont load.
```
Jeavon, line 128: I would suggest a rephrase here "Moving to application settings can be useful in some scenarios for example allowing these values to be set in the Azure Portal applications settings rather than stored in web.config"

Extending/FileSystemProviders/Azure-Blob-Storage/index.md
```diff
@@ -124,6 +124,16 @@ You have to manually add `prefix="media/"` to the service element, otherwise Ima
 
 You have now successfully setup Azure Blob Storage with your Umbraco site.
 
+### Application Setting Override
```
Jeavon, line 127: I'm not sure Override is the correct term for this, how about "Moving settings to Application Settings"?

Jeavon (changes requested):

Extending/FileSystemProviders/Azure-Blob-Storage/index.md
kmeilander: I thought you still need to set the values in the security.config, probably so the service gets added.  

That's why I said override since you need to set it in both places.  But I haven't had the time to try that out to confirm that is true.

Extending/FileSystemProviders/Azure-Blob-Storage/index.md
kmeilander: Phrase added. Let me know if this is what you meant

Jeavon (approved):

Jeavon:
Thank you @kmeilander

PR #3264, opened 2021-06-10
Unbreak the link to the correct paragraph

JeffreyPerplex:


busrasengul:
Great catch @JeffreyPerplex 
Merging right in! 🙌

PR #3276, opened 2021-06-15
Update index.md

mistyn8:
content.SetValue(String Alias, System.Guid Guid);  doesn't automatically set the propData to a UDI as suggested. Need to take the interim step and generate a UDI.

Although the property editor can work out that if only Guid stored then correct member is displayed...  if after programmatic create you try to access the property it returns member picked == null. (until you visit the document and publish via the back office, at which point the database reflects a udi reference, not guide.)

image


busrasengul:
Thank you @mistyn8 
I know about the null field in the database when programmatically created members. 
We will review your code asap then merge it in!
#h5yr!! ✌

PR #3278, opened 2021-06-16
V9 Configuration documentation

nikolajlauridsen:
This PR adds reference documentation for the configuration pattern used in V9, as well as documents all currently available configuration options.

This is kind of a big one since it's more of a rewrite since the configuration of V9 has drastically changed. 

I added a page for every configuration section, totalling 24 at the moment, so there's been added quite a few new documents, I'm not super up to speed with how the folder structure correlates to the page in the left-hand side menu. I think it's a good idea that every configuration section has its own entry, but I don't know if it will get too cluttered since many of these will be mixed in with the V8 pages. So I don't know if the solution could be to add a "V9 Config" under the reference section or something similar?


jonat123:
@nikolajlauridsen thanks for the PR! Will take a look at it :) 

Regarding the structure, I am not 100% sure how it should be done... I think it might be best to maybe create a new structure of the config section so it is easier to know which version of Umbraco you are looking at configs for.

Currently in the menu on the side is it is a mix of configs from between V6-8.

An idea could maybe be to create a new folder under Config Called V9 configs (or something similar) where we put the V9 config articles would be the best solution, then it should also be shown in the side menu, however not all of the new configs.

Let me know what you think 🙂

jonat123:
@nikolajlauridsen  i like the idea, the problem is that the our platform does let us go more than two levels down in the side menu structure...

What i think is the best solution is to create under the reference section a V9 config folder, just like the config folder we have now and then add the V9 config documentation under there, then users will be able to see the articles underneath there.

So it will look like this:

- Reference
   - V9 config
      - Global settings
      - Content settings
      - Imaging settings
      - etc...
      - etc...

nikolajlauridsen:
Ahh I see, I've moved the new documents to a "V9-Config" folder under Reference, I'm not sure how folder naming works? Is whitespace allowed? If so it might be better to use that.

jonat123:
You need to add the - in the title for it to work properly so no white space so it is perfect that you named it V9-Config :)

Reference/V9-Config/ContentSettings/index-v9.md
```diff
@@ -0,0 +1,195 @@
...
+
+This allows you to customize the preview badge being shown when you're previewing a node.
+
+```json
+"PreviewBadge": "<![CDATA[<a id=\"umbracoPreviewBadge\" style=\"position: absolute; top: 0; right: 0; border: 0; width: 149px; height: 149px; background: url('{1}/preview/previewModeBadge.png') no-repeat;\" href=\"{0}/endPreview.aspx?redir={2}\"><span style=\"display:none;\">In Preview Mode - click to end</span></a>]]>"
+```
+
+### Macro errors
+
+This setting allows toy to specify how errors in macros should be handled.
```
jonat123, line 109:
```suggestion
This setting allows you to specify how errors in macros should be handled.
```

Reference/V9-Config/ContentSettings/index-v9.md
```diff
@@ -0,0 +1,195 @@
...
+
+Options:
+* Inline - Default Umbraco behaviour, show an inline error within the macro bu allow the page to continue rendering.
```
jonat123, line 112:
```suggestion
* Inline - Default Umbraco behavior, show an inline error within the macro but allow the page to continue rendering.
```

Reference/V9-Config/ContentSettings/index-v9.md
```diff
@@ -0,0 +1,195 @@
...
+
+By default this is set to false. To make the obsolete data types visible in the dropdown change the value to **true**
+
+### Login background image
+
+You can specify your own background image for the login screen here. The image will automatically get an overlay to match backoffice colors. This path is relative to the `wwwroot/umbraco` path. The default location is: `wwwroot/umbraco/assets/img/installer.jpg`.
+
+### Login logo image
+
+You can specify for own image for the login screen logo. This path is relative to the `wwwroot/umbraco` path. The default location is: `wwwroot/umbraco/assets/img/application/umbraco_logo_white.svg`.
```
jonat123, line 137: Not quite sure what the first sentence means here 🤔

Is there missing a word?

Reference/V9-Config/ContentSettings/index-v9.md
```diff
@@ -0,0 +1,195 @@
...
+
+This is a separated list of accepted image formats
+
+### AutoFillImageProperties
+
+You can define what properties should be automatically updated when an image is being uploaded. This means that if you decide to rename the default **umbracoWidth** and **umbracoHeight** properties to **width** and **height** then the values in **`"WidthFieldAlias"`** and **`"HeightFieldAlias"`** need to be updated with the new property aliases. This needs to happen in order to automatically populate the values when the image is being uploaded.
+
+### Custom media document
+
+If you need to create a custom media document type to handle images called something like "Custom Image" width an alias of **customImage** then you need to add another object where the **alias** is set to **customImage**. Like below. Note that the width and height attributes has also been changed in this example.
```
jonat123, line 173:
```suggestion
If you need to create a custom media document type to handle images called something like "Custom Image" with an alias of **customImage** then you need to add another object where the **alias** is set to **customImage**. Like below. Note that the width and height attributes have also been changed in this example.
```

Reference/V9-Config/HealthChecks/index-v9.md
```diff
@@ -0,0 +1,98 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+This config will disable the macro errors check, and enable notifications to run the check and notify via email if a check fails, running the first time five minutes are the site as booted, and this once every day.
```
jonat123, line 45: Not sure I understand the last part of this sentence 🤔

Reference/V9-Config/HealthChecks/index-v9.md
```diff
@@ -0,0 +1,98 @@
...
+
+There is also a `DisabledOn` key representing the date the health check was disabked and a `DisabledBy` key containing the ID of the user that disabled the health check, however these values are currently not used.
+
+## Notification
+
+Settings relating to running the health checks automatically and sending out notifications.
+
+### Enabled
+
+Allows you to disable orenable all notifications methods, if set to false, the health checks will not automatically run.
```
jonat123, line 63:
```suggestion
Allows you to disable or enable all notifications methods, if set to false, the health checks will not automatically run.
```

Reference/V9-Config/HealthChecks/index-v9.md
```diff
@@ -0,0 +1,98 @@
...
+
+### First run time
+
+This will configure when to run the health checks for the first time, if the value is not configured the health checks will run imideately after the site boots for the first time. This value is specified as a string in crontab format, so in this example the health checks will run 5 minutes after the site is first booted.
```
jonat123, line 67:
```suggestion
This will configure when you run the health checks for the first time, if the value is not configured the health checks will run immediately after the site boots for the first time. This value is specified as a string in crontab format, so in this example, the health checks will run 5 minutes after the site is first booted.
```

Reference/V9-Config/ContentSettings/index-v9.md
nikolajlauridsen: Yeah I think I did an oopsie, I've tried updating it to something that hopefully makes a bit more sense 😄

Reference/V9-Config/ImagingSettings/index-v9.md
```diff
@@ -0,0 +1,66 @@
...
+
+Specifies how long an image may be stored in the browser cache before it gets renewed, formatted as a timespan. The default is seven days.
+
+### Cache max age
+
+This setting allows you to specify how long an image may be stored in the local cache before it gets renewed. Defaults to one year (365 days).
+
+### Cache name length
+
+Whenever an image is cached it will be remamed with a randomly generated name, this setting allows you to change how long that name will be, by default 8 characters. It's worth mentioning here that cached images will be but in a series of sub folders based on their name, for instance a file with the name `abc1` will be put in `mediacahce/a/b/c/1`. This is done to mitigate the potential performance hit of having a very large amount of files in the same folder, so a longer cache name length will result in a deeper folder structure.
```
jonat123, line 50:
```suggestion
Whenever an image is cached it will be renamed with a randomly generated name, this setting allows you to change how long that name will be, by default 8 characters. It's worth mentioning here that cached images will be but in a series of subfolders based on their name, for instance, a file with the name `abc1` will be put in `mediacahce/a/b/c/1`. This is done to mitigate the potential performance hit of having a very large amount of files in the same folder, so a longer cache name length will result in a deeper folder structure.
```

Reference/V9-Config/HealthChecks/index-v9.md
nikolajlauridsen: I can see why hehe, I managed to butcher that sentence, I've tried updating it 😄

Reference/V9-Config/RequestHandlerSettings/index-v9.md
```diff
@@ -0,0 +1,151 @@
...
+meta.Title: "Umbraco Request Handler Settings"
+meta.Description: "Information on the request handler settings section"
+state: complete
+verified-against: beta-3
+update-links: true
+---
+
+# Request handler settings
+
+The options in the request handler let us do some quite useful things, like  deciding whether or not to use trailing slashes and setting URL replacement for special characters.
```
jonat123, line 12:
```suggestion
The options in the request handler let us do some quite useful things, like deciding whether or not to use trailing slashes and setting URL replacement for special characters.
```

Reference/V9-Config/RichTextEditorSettings/index-v9.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+## Plugins
+
+Allows you to specify what plugins should be enabled for the rich text editor as a comma seperated list of the plugin names.
+
+## Valid elements
+
+Allows you to specify the valid HTML elements for the rich text editor as a well as the allowed properties, for for instance, in the config above the `a` element is allowed, and it's allowed to have `onclick`, `href`, and many more attributes.
```
jonat123, line 56:
```suggestion
Allows you to specify the valid HTML elements for the rich text editor as well as the allowed properties, for instance, in the config above the `a` element is allowed, and it's allowed to have `onclick`, `href`, and many more attributes.
```

Reference/V9-Config/RuntimeMinificationSettings/index-v9.md
```diff
@@ -0,0 +1,41 @@
...
+
+* Version - Caches will be busted every time the version string changes.
+* AppDomain - Caches will be busted when the app restarts.
+* Timestamp - Caches will be busted based on a timestamp.
+
+## Automatically generated settings
+
+You might see a setting `"version"` with a long number after it, this setting is generated if you use the `Version` type cache buster, generally you don't need to change this. However. if you're making some front end changes, and not seeing the change, then you can increase this number by one, clearing the cache and now you should see the change.
+
+Another setting you might see be populated is the `"dataFolder`" setting, this setting specifies what folder Smidge will use for its temporary data, it should not be nececary to change this.
```
jonat123, line 41:
```suggestion
Another setting you might see be populated is the `"dataFolder`" setting, this setting specifies what folder Smidge will use for its temporary data, it should not be necessary to change this.
```

Reference/V9-Config/SecuritySettings/index-v9.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+## Root level settings
+
+At the root level of security you can configure the following
+
+### Keep user logged in
+
+If set to false a user will be logged out after a specific amount of time as passed with no activity, you can specify this time span in the [global settings](../GlobalSettings/index-v9.md) with the `TimeOut` key.
```
jonat123, line 55:
```suggestion
If set to false a user will be logged out after a specific amount of time has passed with no activity, you can specify this time span in the [global settings](../GlobalSettings/index-v9.md) with the `TimeOut` key.
```

Reference/V9-Config/SecuritySettings/index-v9.md
```diff
@@ -0,0 +1,115 @@
...
+
+The authentication cookie which is set in the browser when a backoffice user logs in, and defaults to `UMB_UCONTEXT`.
+
+### Auth cookie domain
+
+The authentication cookie which is set in the browser when a backoffice user logs in is automatically set to the current domain.
+
+### Username is email
+
+This setting specifies whether the username and email address are separate fields in the backoffice editor. When set to "false", you can specify an email address and username, only the username can be used to log on. When set the "true" (the default value) the username is hidden and always the same as the email address.
```
jonat123, line 76:
```suggestion
This setting specifies whether the username and email address are separate fields in the backoffice editor. When set to "false", you can specify an email address and username, only the username can be used to log on. When set to "true" (the default value) the username is hidden and always the same as the email address.
```

Reference/V9-Config/Serilog/index-v9.md
```diff
@@ -0,0 +1,79 @@
...
+meta.Title: "Umbraco Serilog Settings"
+meta.Description: "Information on the serilog settings section"
+state: complete
+verified-against: beta-3
+update-links: true
+---
+
+# Serilog settings
+
+Umbraco uses Serilog as its logging library, this means that the configuration of logging is offloaded to Serilog instead of the CMS. This means that logging specific configuration is no the `Umbraco.Cms` section, but instead the Serilog section.
```
jonat123, line 12:
```suggestion
Umbraco uses Serilog as its logging library, this means that the configuration of logging is offloaded to Serilog instead of the CMS. This means that logging specific configuration is not in the `Umbraco.Cms` section, but instead the Serilog section.
```

Reference/V9-Config/ToursSettings/index-v9.md
```diff
@@ -0,0 +1,22 @@
...
+meta.Title: "Umbraco Tours Settings"
+meta.Description: "Information on the tours settings section"
+state: complete
+verified-against: beta-3
+update-links: true
+---
+
+# Tours settings
+
+The tours section is simple section that only contains one value `"EnableTours"` which allows you to enable or disable tours, by default tours are enabled. If you wish to disable tours set the `"EnableTours"` to false, like so:
```
jonat123, line 12:
```suggestion
The tours section is a simple section that only contains one value `"EnableTours"` which allows you to enable or disable tours, by default tours are enabled. If you wish to disable tours set the `"EnableTours"` to false, like so:
```

Reference/V9-Config/TypeFinderSettings/index-v9.md
```diff
@@ -0,0 +1,22 @@
...
+meta.Title: "Umbraco Type Finder Settings"
+meta.Description: "Information on the type finder settings section"
+state: complete
+verified-against: beta-3
+update-links: true
+---
+
+# Type finder settings
+
+The type finder settings allow you to specify assemblies that accep load exceptions when they are type scanned, for multiple assemblies seperate them with a comma (`,`). To accept load exceptions for all assemblies use an asterisk (`*`), like so:
```
jonat123, line 12:
```suggestion
The type finder settings allow you to specify assemblies that accept load exceptions when they are type scanned, for multiple assemblies separate them with a comma (`,`). To accept load exceptions for all assemblies use an asterisk (`*`), like so:
```

Reference/V9-Config/index-v9.md
```diff
@@ -0,0 +1,120 @@
...
+meta.Title: "Umbraco configuration"
+meta.Description: "Information on configuring Umbraco"
+state: complete
+verified-against: beta-3
+update-links: true
+---
+
+# Configuration Files
+
+In V9, we have moved away from the previous configuration using .config files, now instead using the netcore built in configuration pattern. This means that there is no longer seperate files for different configuration, the configuration is now primarily done from the `appsettings.json` file.
```
jonat123, line 12:
```suggestion
In V9, we have moved away from the previous configuration using .config files, to instead using the netcore built-in configuration pattern. This means that there is no longer separate files for different configuration, the configuration is now primarily done from the `appsettings.json` file.
```

Reference/V9-Config/index-v9.md
```diff
@@ -0,0 +1,120 @@
...
+
+For more in depth information on the configuration pattern see Microsofts [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0) article.
+
+## Managing configuration
+
+You might not always want to have the configuration stored in the `appsettings.json` file, for instance you might not want to have the admin password in the file if using the unattended feature. You might also want to use a specific set of configurations when developing your solution, fortunately the `IConfiguration` pattern.
```
jonat123, line 18:
```suggestion
You might not always want to have the configuration stored in the `appsettings.json` file, for instance, you might not want to have the admin password in the file if using the unattended feature. You might also want to use a specific set of configurations when developing your solution, fortunately, the `IConfiguration` pattern.
```

Reference/V9-Config/index-v9.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. The `appsettings.json` file
+2. An `appsettings.{environment}.json` file
+3. UserSecrets (Only when in development)
+4. Environment variables
+5. Command line arguments
+
+This list is in order of precedence, so the values from `appSettings.json` will only be used if they're not also define in the environment variables, if they are then the environment variable will be used instead.
+
+There isone caveat, to this precedence though, the `appSettings.{environment}.json` file will only be used if the current environment matches the name of the config file, for instance the `appsettings.Development.json` file will only be used when the environment is set to development.
```
jonat123, line 30:
```suggestion
There is one caveat, to this precedence though, the `appSettings.{environment}.json` file will only be used if the current environment matches the name of the config file, for instance, the `appsettings.Development.json` file will only be used when the environment is set to development.
```

jonat123:
@nikolajlauridsen  pheeeewww that was a long read 🤓

I can see you have implemented most of my suggestions, that is great, i have added a few more, and once they have been fixed, then we are golden and i think this is ready to be merged in :D

nikolajlauridsen:
Hehe yes, we have quite a bit of settings 😄 I've applied your suggestions, to I think we're ready to roll 😄

jonat123:
Perfect! I am merging it in :D

PR #3280, opened 2021-06-17
Migrate creating forms to v9

nikolajlauridsen:
This migrates the "Creating Forms" to v9, I've primarily just updated the code examples to make them work on V9. Pretty much the only thing I changed in regards to the text is recommending creating an empty class instead of a controller template since we don't use any of the functionality from the controller template.


jonat123:
@nikolajlauridsen  Looks good, just tested it out and works like a charm :D 

I will merge this in right away :)

PR #3282, opened 2021-06-17
Creating a form

eshanrnh:
1. Verified against version 9 beta 3.
2. Structured the document and added images


jonat123:
@eshanrnh Looks good!

I don't have any thing to add, so lets merge it in :D

PR #3283, opened 2021-06-18
Update reference management models contenttype for v8

erikjanwestendorp:
Update [ContentType article](https://our.umbraco.com/documentation/Reference/Management/Models/ContentType) for v8


busrasengul:
Thank you @erikjanwestendorp 
Let's merge this in!

PR #3286, opened 2021-06-18
Migrate models builder docs to v9

nikolajlauridsen:
This migrates the remaining docs to V9.

I've updated the code examples and removed some claims which are no longer true, such as "If the custom partial class inherits from a base class, it will be detected and the generated model will not inherit from anything," since I found this to no longer be the case, neither from experimentation nor when looking through the code.

There is one page that I have not updated, and that is the "Install the full version of Models Builder" one, since the full version of models builder haven't been upgraded for V9 yet


Reference/Templating/Modelsbuilder/Using-Interfaces-v9.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Using Interfaces
+
+When using compositions, Models Builder generates an interface for the composed model, which enables us to not have to switch back to using `Value()` for the composed properties.
+
+A common use-case for this is if you have a separate composition for the "SEO properties" `Page Title` and `Page Description`.
+
+You would usually use this composition on both your `Home` and `Textpage` document types. Since both `Home` and `Textpage` will implement the generated `ISeoProperties` interface, you will still be able to use the simpler models builder syntax (e.g. `Model.PageTitle`).
+
+However, you wont be able to the the nice models builder syntax on any master template, since a master template need to be bound to a generic `IPublishedContent`. So you'd have to resort to the *ever-so-slightly* clumsier `Model.Value("pageTitle")` syntax to render these properties. It is possible to solve this issue of master templating, by using partial views, to render the SEO specific properties.
```
jonat123, line 18:
```suggestion
However, you won't be able to use the nice models builder syntax on any master template, since a master template needs to be bound to a generic `IPublishedContent`. So you'd have to resort to the *ever-so-slightly* clumsier `Model.Value("pageTitle")` syntax to render these properties. It is possible to solve this issue of master templating, by using partial views, to render the SEO specific properties. 
```

jonat123:
@nikolajlauridsen looks good, i have made a small suggestion as it seemed like there was missing a word :)

nikolajlauridsen:
> @nikolajlauridsen looks good, i have made a small suggestion as it seemed like there was missing a word :)

Ope, my bad, seems like my brain was faster than my typing for a second there 😄

Reference/Templating/Modelsbuilder/index-v9.md
```diff
@@ -0,0 +1,29 @@
...
+meta.Description: "Modelsbuilder reference"
+state: complete
+verified-against: beta-3
+update-links: false
+---
+
+
+# Umbraco Models Builder
+
+:::note
```
jonat123, line 13: I think it would make sense to say that the full modelsbuilder currently is not working with Umbraco 9, as this is the first page they end up on and they might think that it works with V9.

jonat123:
@nikolajlauridsen hehe it happens ;) 

Just added another comment about the landing page.

let me know what you think!

Reference/Templating/Modelsbuilder/index-v9.md
nikolajlauridsen: Ahh that's a very good point, I've changed the note to say that it hasn't been migrated yet, but when it is that's where you'll find it.

jonat123:
@nikolajlauridsen perfect, looks good to me, iøll merge it in :D

PR #3293, opened 2021-06-22
Added v9 docs for load balancing.

bergmania:
Added docs for v9 regarding load-balancing...

Note that most text is reused. But especially the area around Data Protection is hard to give one recommendation about. Instead, I pointed to Microsoft's documentation, as there are so many different ways to set this up. And this is required for all ASP.NET Core load-balanced environments. This is not specific for Umbraco,

### DIFF
--- a/Fundamentals/Setup/Server-Setup/Load-Balancing/index-v9.md
+++ b/Fundamentals/Setup/Server-Setup/Load-Balancing/index-v9.md
@@ -107,8 +107,12 @@ ASP.NET Core supports multiple ways to share keys. Use the [official docs](https
 
 
 ### Session State and Distributed Cache
+It is required to setup a distributed cache, like `DistributedSqlServerCache` or an alternative provider (see [https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed](https://msdn.microsoft.com/en-us/library/aa478952.aspx) for more details).
+The distributed cache is used by the session in your application, which is used by the default TempDataProvider in MVC.
 
-* If you use Session in your application, or are using the default TempDataProvider in MVC (which uses Session) then you will need to configure your application to use the `DistributedSqlServerCache` or an alternative provider (see [https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed](https://msdn.microsoft.com/en-us/library/aa478952.aspx) for more details).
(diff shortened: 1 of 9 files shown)


Fundamentals/Setup/Server-Setup/Load-Balancing/index-v9.md
```diff
@@ -0,0 +1,150 @@
...
+The replacement for Machine Keys in ASP.NET Core are called Data Protection.
+You will need to setup data protection to the same keys on all servers,
+without this you will end up with view state errors, validation errors and encryption/decryption errors since each server will have its own generated key.
+
+ASP.NET Core supports multiple ways to share keys. Use the [official docs](https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview) to find a description that fits your setup the best.
+
+
+### Session State and Distributed Cache
+
+* If you use Session in your application, or are using the default TempDataProvider in MVC (which uses Session) then you will need to configure your application to use the `DistributedSqlServerCache` or an alternative provider (see [https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed](https://msdn.microsoft.com/en-us/library/aa478952.aspx) for more details).
```
Jeavon, line 111: @bergmania does v9 use TempData or ViewData in any of the front end methods, forms, surface controllers etc? If so that would require a distributed cache configuring when load balancing?

bergmania: As far as I know CMS do not use TempData for the frontend but Umbraco Forms does. I would always set up the distributed cache when configuring when load balancing.

Jeavon: @bergmania cool, I think in v8 the BeginUmbracoForm and associated methods used TempData which was the reason session state was required

bergmania: You are right.. Seems like we use it in a couple of places
- All the member snippets use TempData
- The `RedirectToUmbracoUrlResult` and `RedirectToUmbracoPageResult` implements `IKeepTempDataResult`.

Guess you should always set up a distributed cache when using load balancing..

PR #3298, opened 2021-06-23
Reference management models language

erikjanwestendorp:
Add language article

### DIFF
--- a/Reference/Management/Models/Language.md
+++ b/Reference/Management/Models/Language.md
@@ -45,7 +45,7 @@ return language.CultureName;
 
 ### .FallbackLanguageId
 
-Gets or sets the identifier of a fallback language. The fallback language can be used in multi-lingual scenarios, to help  define fallback strategies when a value does not exist for a requested language.
+Gets or sets the identifier of a fallback language. The fallback language can be used in multi-lingual scenarios, to help define fallback strategies when a value does not exist for a requested language.
 
 ```csharp
 var language = new Language("en-US");
--- a/Reference/Management/Models/Language.md
+++ b/Reference/Management/Models/Language.md
@@ -25,4 +25,57 @@ Constructor for creating a new `Language` object where the necessary parameter i
 
 ## Properties
 
+### .CultureInfo
+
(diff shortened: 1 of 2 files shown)


Reference/Management/Models/Language.md
```diff
@@ -22,6 +19,61 @@ All samples in this document will require the following using statement:
...
+Gets or sets a value indicating whether the language is the default language.
+
+```csharp
+var language = new Language("en-US");
+return language.IsDefault;
+```
+
+### .IsMandatory
+
+Gets or sets the identifier of a fallback language. The fallback language can be used in multi-lingual scenarios, to help define fallback strategies when a value does not exist for a requested language.
```
busrasengul, line 65: I think this sentences is the same with .FallbackLanguageId was it intentional?

Reference/Management/Models/Language.md
```diff
@@ -22,6 +19,61 @@ All samples in this document will require the following using statement:
...
+Gets or sets the culture name of the language.
+
+```csharp
+var language = new Language("en-US");
+return language.CultureName;
+```
+
+### .FallbackLanguageId
+
+Gets or sets the identifier of a fallback language. The fallback language can be used in multi-lingual scenarios, to help  define fallback strategies when a value does not exist for a requested language.
```
busrasengul, line 48:
```suggestion
Gets or sets the identifier of a fallback language. The fallback language can be used in multi-lingual scenarios, to help define fallback strategies when a value does not exist for a requested language.
```

busrasengul (changes requested):
Hey @erikjanwestendorp 
I've add little comments would you mind checking them?
Thank you :)

Reference/Management/Models/Language.md
erikjanwestendorp: @busrasengul  Yes your right. I updated the description.

erikjanwestendorp:
@busrasengul Thanks foy your review. Just updated the PR 😄

busrasengul:
Thanks for updating them @erikjanwestendorp 
Looks good to me!
Merging it now :)

PR #3300, opened 2021-06-25
v9 article for IsHelpers

Jay-umbr:
and updated v8 one about specific config based on the issue:
https://github.com/umbraco/Umbraco-CMS/issues/10391
(template methods)


eshanrnh:
Looks Good, @Jay-umbr 😀

PR #3302, opened 2021-06-28
Update Extending-UmbracoApplication.md

mistyn8:
Added the applicationComposer in order to register this component.
Updated the using statements to include all those required.
Not sure why Webprofiller and profile were here so removed and replaced with Umbraco native logger.

The related link https://github.com/umbraco/UmbracoDocs/blob/main/Reference/Events/Application-Startup.md covers a similar topic, though this one which isn't EXPERT has the included SessionState? should the two be merged?


busrasengul:
Thank you for your PR @mistyn8 
We'll review your PR soon and let you know! 🎉

PR #3303, opened 2021-06-28
Change MyCustomUmbracoSolution to MyCustomUmbracoProject

Adolfi:
The installer does not create a solution, it creates a project. 
Therefor it should not be called MyCustomUmbraco**Solution**, but instead MyCustomUmbraco**Project**.


busrasengul:
Thank you @Adolfi 
MyCustomProject or MyCustomSolution is the name you give already so it's not changing things. But you're right, the installer creates a project. So I'll merge this in! 🙌

PR #3304, opened 2021-06-28
Links

eshanrnh:
Updated page links


busrasengul:
Looks like the links are not working with -v9 extensions @eshanrnh 
Would you mind double checking them?
See here also [#3305](https://github.com/umbraco/UmbracoDocs/issues/3305)

eshanrnh:
Fixed the issue @busrasengul. **versionTo** metadata had a **+** which was causing an issue in rendering the folder.

PR #3306, opened 2021-06-29
Reference management models media type v8

erikjanwestendorp:
Update MediaType for v8


busrasengul:
Thank you @erikjanwestendorp 
We'll get this reviewed soon! 🙌

PR #3307, opened 2021-06-30
Reference management models relation type v8

erikjanwestendorp:
Update RelationType for v8


busrasengul:
Thank you @erikjanwestendorp 
We'll review this soon :)

PR #3265, opened 2021-06-11
Routing custom controllers v9

erikjanwestendorp:
Belongs to https://github.com/umbraco/UmbracoDocs/issues/3145

PR #3312, opened 2021-07-02
V9: Routing docs

nikolajlauridsen:
This PR migrates the Reference/Routing section, I'm not done yet, but decided to make a draft PR still, so it's visible that this is being worked on.

I'm not going to work on the "Custom controllers (hijacking routes)" section though, since there's an open PR for it #3265

~~When working on this, I also found a couple of bugs/missing things from the CMS, so not all of this will work for beta4, surface controllers are routed different in beta4, and `UmbracoAuthorizedController` is not part of beta4~~ This is part of RC001

I've chosen not to migrate the IIS Rewrite Rules section since V9 is no longer coupled to IIS, people can host Umbraco with whatever web server they want, for instance, kestrel, and how you handle the URL rewrites will depend on that, so I didn't think it made much sense to have this in the Umbraco documentation, since it's no longer tied to Umbraco.


Reference/Routing/Request-Pipeline/IContentFinder-v9.md
```diff
@@ -0,0 +1,221 @@
...
+}
+
+```
+:::note
+In Umbraco7 there existed an IContentFinder that would find content and display it with an 'alternative template' via a convention. This could be to avoid the ugly `?alttemplate=blogfullstory` appearing on the querystring of the url when using the alternative template mechanism. Instead the Url could follow the convention of `/urltocontent/altemplatealias`. 
+
+Eg: `/blog/my-blog-post/blogfullstory` would 'find' the `/blog/my-blog-post` page and display using the `blogfullstory` template. 
+
+In Umbraco 8 this convention has been removed from the default configuration of Umbraco. You can reintroduce this behaviour by adding the `ContentFinderByUrlAndTemplate` ContentFinder back into the ContentFinderCollection, using an `IUserComposer`, or Umbraco builder extension (see above example).
+:::
```
jonat123, line 146: @nikolajlauridsen Im a little confused about this note here 🤔 You mention that it was something in V7 which have been removed in v8 but can be re-introduced, however this is a v9 article, is this also something that can be done in v9? If so then it might need to be made clear :)

nikolajlauridsen: Ah yes, this is something that is removed by default in both V8 and V9, but you should be able to add it back in V9 as well, I'll double-check and then update the note 👍

Reference/Routing/Request-Pipeline/outbound-pipeline-v9.md
```diff
@@ -0,0 +1,512 @@
...
+}
+```
+### How the Default Url provider works
+
+- If the current domain matches a root domain of the target content.
+  - Return a relative Url.
+  - Else must return an absolute Url.
+- If the target content has only one root domain.
+  - Use that domain to build the absolute Url.
+- If the target content has more that one root domain.
```
jonat123, line 219:
```suggestion
- If the target content has more than one root domain.
```

Reference/Routing/Request-Pipeline/IContentFinder-v9.md
nikolajlauridsen: I've updated it so it says "In Umbraco 9 this convention has been removed", I don't know if there's a better way to explain it since it's the case for both v8 and v9.

jonat123: It looks fine and makes it clear that it have been removed in v9 as well :)

jonat123:
@nikolajlauridsen Looks good, i can see that you have implemented my suggestions, I don't have anything else than that, so I will get this bad boy merged in :D

PR #3326, opened 2021-07-08
Creating a basic site tutorial

eshanrnh:
1. Updated image and articles for v9.


jonat123:
@eshanrnh Looks good!

There is a few things that we might need to change when V9 have been released, mostly in the getting started file, but for now it is fine 🙂

PR #3327, opened 2021-07-08
Extending Umbraco with Forms

eshanrnh:
1. Updated the Installing Umbraco Forms section.
2. Updated the images.


Add-ons/UmbracoForms/Installation/Install/index-v9.md
```diff
@@ -0,0 +1,38 @@
...
+
+# Extending Umbraco with the Forms Section
+
+Since Umbraco v7.2, Umbraco contains the **Forms** section, by default. You will see a similar interface, when you click on the **Forms** section in the **Umbraco Backoffice**.
+
+![Form section in backoffice](images/Forms_Section_Backoffice.png)
+
+## Installing Umbraco Forms
+
+To install the Umbraco Forms package (**Umbraco.Forms.9.0.0-beta001**), follow these steps:
```
AndyButland, line 16: Would suggest not mentioning the version here, otherwise we'd need to update the docs every time a new version is released.  Think we can omit this, and people will get the latest version, and if they want a specific one they can add the version parameter.

```suggestion
To install the Umbraco Forms package (**Umbraco.Forms**), follow these steps:
```

Add-ons/UmbracoForms/Installation/Install/index-v9.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. Run the following command on a command prompt of your choice:
+
+    ```cli
+    dotnet add package Umbraco.Forms --version 9.0.0-beta001
```
AndyButland, line 21:
```suggestion
    dotnet add package Umbraco.Forms
```

AndyButland (approved):
Just left a comment about referencing version numbers, that we probably don't need or want.

Add-ons/UmbracoForms/Installation/Install/index-v9.md
eshanrnh: Got it, removing the version

PR #3330, opened 2021-07-11
Fix to custom health check code

LottePitcher:
I copied the custom 'SEO' health check sample code into my project from https://our.umbraco.com/documentation/Extending/Health-Check/ (8.0.0). 

To compile, the class needed to inherit from `HealthCheck` not  `HealthCheck.HealthCheck`.

This was in a v8.14 project.


busrasengul:
Thank you @LottePitcher 
I'll be sure to check this with early versions as well then merge your PR 🙌
#h5yr

LottePitcher:
Ah yes, good point!

I happen to have both a 8.1.6 site and a 8.5.1 site open right now so can confirm that both these versions also raise this compilation error if I copy in the sample code. Both expect the class to inherit from `HealthCheck` not `HealthCheck.HealthCheck`

busrasengul:
That's great and enough for now! I'm merging this 💪🏼

PR #3336, opened 2021-07-13
Cleared up some confusion around this feature.

RyuLindow:
Added a note about a required header if the Content Delivery API is set to protected.


jonat123:
@RyuLindow Looks good :D I'll merge it in right away :D

PR #3345, opened 2021-07-15
Added documentation on the Deploy configuration option ignoreBrokenDependencies

AndyButland:
This setting was references as being an "undocumented feature" in [an issue](https://github.com/umbraco/Umbraco.Deploy.Issues/issues/5) I've been looking into, and doesn't seem to be any reason why it shouldn't be documented, so I think that's just an oversight.

Hence this PR to document it for V4 and V9.


jonat123:
@AndyButland  thanks for the PR! Looks good to me, so I'll merge it in right away :)

PR #3353, opened 2021-07-15
Updating install by CLI guide based on RC

crgrieve:
Removing beta note as it's now RC, updating the help output to be accurate to what is returned, updating mysql command to match one in RC release blog post


busrasengul (approved):
Thank you @crgrieve 
I'm merging this right in!

PR #3359, opened 2021-07-16
please, update this

RyuLindow:
It causes miscommunication...
either change it to what I wrote or change the date because people are getting confused and I don't blame them


busrasengul:
Let's update it to what you wrote now @RyuLindow to prevent any further confusion. When we have any date we can update again :)

RyuLindow:
Thank you :)

PR #3361, opened 2021-07-17
Reference management models content small issues

erikjanwestendorp:
Fix some small issues in the Model articles. 

- [x] remove needsV8Update in index files
- [x] remove double use of using
- [x] All usings in code block to be consistent


busrasengul:
Thank you @erikjanwestendorp 🙌🏻

PR #3367, opened 2021-07-24
Update trees-v9.md

cornehoskam:
The documentation speaks of adding an example Tree to the Settings section, but tries to register it in an unknown section with alias "myFavouriteThings".
Changing this to the settings Alias as also used in the rest of the documentation fixes this issue!


jonat123:
@cornehoskam  thanks for the PR :) We will get to review it asap 🙂

jonat123:
@cornehoskam Just had a look at it, and you are correct, good catch!

I'll merge it in right away :D

PR #3379, opened 2021-08-06
Updated create a database table article for V9

AndyButland:
It came up in the gold partner "learn-a-thon" today that there's a couple of ways with V9 to do what's described in this article.  So here I've documented the second method, and fixed up a few things with the first for code changes that have been introduced since it was first reviewed and created.

@bergmania - just wondering if you have any thoughts about what the "best" method here is, and whether we should be advocating one, or just as I have document both.

### DIFF
--- a/Extending/Database/index-v9.md
+++ b/Extending/Database/index-v9.md
@@ -284,6 +284,14 @@ namespace MyNamespace
 
 ```
 
+## Which to use?
+
+In short, it's up to you.  If you are migrating from V8 and want the quickest route to getting running with V9, then using a component makes sense.
+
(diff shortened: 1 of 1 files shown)


bergmania:
@AndyButland, The components are initialized just before the `UmbracoApplicationStartingNotification` are published, so what you do in a component can also be done in that notification. 

Personally, I like using the notification the most, but I don't really have good arguments for that.  The main reason is that notifications are a pattern used elsewhere, where components are a one-off thing.
On the other hand, `Components` offer both `Initialize` and `Terminate` methods, where you will need to use two unrelated notifications to do the same (`UmbracoApplicationStartingNotification` and `UmbracoApplicationStoppingNotification`).

AndyButland:
Thanks - I've added a paragraph giving some guidance on which to used including these points.

bergmania (approved):

jonat123:
@AndyButland  Looks good!

I'll get this merged in right away 🙂

PR #3385, opened 2021-08-10
Rename TagService to CacheTagService

erikjanwestendorp:
Belongs to #3190 

Rename TagService to CacheTagService


busrasengul:
Thank you @erikjanwestendorp 
We'll get this review 👍

PR #3390, opened 2021-08-11
Better formatting of this page

JeffreyPerplex:
Listitems are now displayed as listitems instead of one long string


busrasengul:
Thank you @JeffreyPerplex 
We'll get this review :)

PR #3395, opened 2021-08-19
Fix a type and add a word

BatJan:
Fixed type in "Intellisense" and added "to" in the sentence "having **to** configure..."


busrasengul:
Thank you @BatJan 
This is getting merged in now!

PR #3396, opened 2021-08-19
Add tilde pattern to paths in config example

BatJan:
I've added "~/" to the path examples in the config to align it with how it's done elsewhere in the config documentation like in https://our.umbraco.com/documentation/Reference/V9-Config/GlobalSettings/


busrasengul:
Good catch @BatJan 
I'm merging this in!

PR #3398, opened 2021-08-19
Added versionTo metadata for Tag support

eshanrnh:


Extending/Property-Editors/tag-support.md
```diff
@@ -1,14 +1,14 @@
 ---
 versionFrom: 7.0.0
```
jonat123, line 2: If the version is only to V7 the versionfrom the tag should be removed and the the versionTo tag should say versionTo: 7.0.0 :-)

jonat123:
@eshanrnh looks good, merging it in!

PR #3402, opened 2021-08-22
Example for when Max items is set to 1

abhamla:
Add example for when Max items is set to 1 in property editor


busrasengul:
Thank you @abhamla 
We'll get this reviewed! 
🙌

PR #3404, opened 2021-08-23
Add filesystem providers to V9 Config section

nikolajlauridsen:
This creates the V9 version of the `fileSystemProviders` section: https://our.umbraco.com/documentation/Reference/Config/fileSystemProviders/

I've updated the code examples, and also somewhat shortened the article, since the information is also available in the [extending filesystems section](https://our.umbraco.com/Documentation/Extending/FileSystemProviders/index-v9), so now we point to that document instead.


jonat123:
@nikolajlauridsen looks good to me 🙂

I'll merge this in right away!

PR #3419, opened 2021-08-30
Update package.manifest samples

BatJan:
I might be late to the party but recently noticed that there is a schema that can be referenced in the package.manifest files in order to provide us with some intellisense so I have updated the package.manifest samples to use it hoping it will benefit people who read the documentation 🤞🏻 😃


busrasengul:
Thank you @BatJan
This looks good to me! Let's merge this in too!

PR #3426, opened 2021-08-31
V9: Added Cache docs, updated from v7.

bergmania:
Added more docs to the Reference/Cache section. 
Most docs are updated from v7 version, but a few changes are made. *
- Removed an example with donut cache that do not make sense for v9 / .NET5+
- Removed the small individual page for ICacheRefresher<T>.

### DIFF
--- a/Reference/Cache/cache-refresher-v9.md
+++ b/Reference/Cache/cache-refresher-v9.md
@@ -1,0 +1,15 @@
+---
+versionFrom: 9.0.0
+---
+
+# ICacheRefresher
+
+_This section describes what ICacheRefresher and ICacheRefresher&lt;T&gt; are and how to use them to invalidate your cache correctly including load balanced environments_
+
+## What is an ICacheRefresher
+
+This interface has been in the Umbraco core for quite some time but has really only been used to ensure that content cache is refreshed among all server nodes participating in a load balanced scenario.
+
+With changes in 6.1, this has changed slightly, an `ICacheRefresher` is now the primary way to invalidate *any* cache that needs to be refreshed or removed regardless of whether we are in a load balanced environment.
+
+There are now a few different types of `ICacheRefreshers` in the Umbraco core and it is important to understand the differences between them and how cache invalidation works across multiple server nodes.
--- a/Reference/Cache/index-v9.md
+++ b/Reference/Cache/index-v9.md
@@ -2,9 +2,91 @@
 versionFrom: 9.0.0
 ---
 
-# Caching
+# Cache & Distributed Cache
 
-We're busy getting the documentation up to date for Umbraco 9. Below you can find a list of examples made to give you an idea about how you can work with caching in Umbraco 9.
+_This section refers to how to implement caching features in the Umbraco application in a consistent way that will work in both single server environments and load balanced (multi-server) environments. The caching described in this section relates to application caching in the context of a web application only._
+
+## IF YOU ARE CACHING, PLEASE READ THIS
+
+Although caching is a pretty standard concept it is very important to make sure that caching is done correctly and consistently. It is always best to ensure performance is at its best before applying any cache and also beware of *over caching* as this can cause degraded performance in your application because of cache turnover.
+
+In normal environments caching seems to be a pretty standard concept. If you are a package developer or developer who is going to publish a codebase to a loadbalanced environment then you need to be aware of how to invalidate your cache properly, so that it works in load balanced environments. If it is not done correctly then your package and/or codebase will not work the way that you would expect in a load balanced scenario.
+
+**If you are caching businesslogic data that changes based on a user's action in the backoffice and you are not using an *ICacheRefresher* then you will need to review your code and update it based on the below documentation.**
+
+## Retrieving and Adding items in the cache
+
+You can [update and insert items in the cache](updating-cache-v9.md).
+
+## Refreshing/Invalidating cache
+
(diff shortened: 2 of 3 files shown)


Reference/Cache/index-v9.md
```diff
@@ -2,9 +2,91 @@
...
-# Caching
+# Cache & Distributed Cache
+
+_This section refers to how to implement caching features in the Umbraco application in a consistent way that will work in both single server environments and load balanced (multi-server) environments. The caching described in this section relates to application caching in the context of a web application only._
+
+## IF YOU ARE CACHING, PLEASE READ THIS
+
+Although caching is a pretty standard concept it is very important to make sure that caching is done correctly and consistently. It is always best to ensure performance is at its best before applying any cache and also beware of *over caching* as this can cause degraded performance in your application because of cache turnover.
+
+In normal environments caching seems to be a pretty standard concept. If you are a package developer or developer who is going to publish a codebase to a loadbalanced environment then you need to be aware of how to invalidate your cache properly, so that it works in load balanced environments. If it is not done correctly then your package and/or codebase will not work the way that you would expect in a load balanced scenario.
```
jonat123, line 13:
```suggestion
In normal environments caching seems to be a pretty standard concept. If you are a package developer or developer who is going to publish a codebase to a load balanced environment then you need to be aware of how to invalidate your cache properly, so that it works in load balanced environments. If it is not done correctly then your package and/or codebase will not work the way that you would expect in a load balanced scenario.
```

jonat123, line 13: I think this section should be added as a warning that way it is more clear to people that they should be aware of the points you are making.

Reference/Cache/updating-cache-v9.md
```diff
@@ -0,0 +1,46 @@
...
+
+The example above will retrieve a strongly typed object of `MyObject` from the cache with the key of "MyCacheKey", if the object doesn't exist in the cache a new instance of `MyObject` will be added to it with the same key.
+
+There are a couple of overloads of `GetCacheItem` allowing you to customize how your object is cached from cache dependencies to expiration times.
+
+To use this generic implementation, add the `Umbraco.Extensions` namespace to your code.
+
+### Retrieving an item from the cache without a callback
+
+One of the overloads of `GetCacheItem` doesn't specify a callback, this will allow  you to retrieve an item from the cache without populating it if it doesn't exist.
```
jonat123, line 33:
```suggestion
One of the overloads of `GetCacheItem` doesn't specify a callback, this will allow you to retrieve an item from the cache without populating it if it doesn't exist.
```

jonat123:
@bergmania Looks good, I have made a few suggestions to what you have written.

Reference/Cache/index-v9.md
bergmania: Feel free to do such changes.. This is just updated from v7 as close to 1-1 as possible, while still being valid for v9.

Reference/Cache/index-v9.md
```diff
@@ -2,9 +2,91 @@
...
+
+## IF YOU ARE CACHING, PLEASE READ THIS
+
+Although caching is a pretty standard concept it is very important to make sure that caching is done correctly and consistently. It is always best to ensure performance is at its best before applying any cache and also beware of *over caching* as this can cause degraded performance in your application because of cache turnover.
+
+In normal environments caching seems to be a pretty standard concept. If you are a package developer or developer who is going to publish a codebase to a load balanced environment then you need to be aware of how to invalidate your cache properly, so that it works in load balanced environments. If it is not done correctly then your package and/or codebase will not work the way that you would expect in a load balanced scenario.
+
+**If you are caching businesslogic data that changes based on a user's action in the backoffice and you are not using an *ICacheRefresher* then you will need to review your code and update it based on the below documentation.**
```
jonat123, line 15:
```suggestion
:::warning
**Please read this if you are Caching**

Although caching is a pretty standard concept it is very important to make sure that caching is done correctly and consistently. It is always best to ensure performance is at its best before applying any cache and also beware of *over caching* as this can cause degraded performance in your application because of cache turnover.

In normal environments caching seems to be a pretty standard concept. If you are a package developer or developer who is going to publish a codebase to a load balanced environment then you need to be aware of how to invalidate your cache properly, so that it works in load balanced environments. If it is not done correctly then your package and/or codebase will not work the way that you would expect in a load balanced scenario.

**If you are caching business logic data that changes based on a user's action in the backoffice and you are not using an *ICacheRefresher* then you will need to review your code and update it based on the below documentation.**
:::
```

Reference/Cache/index-v9.md
jonat123: Ahhh sorry, I have made the change now.

Reference/Cache/cache-refresher-v9.md
```diff
@@ -0,0 +1,15 @@
...
+
+# ICacheRefresher
+
+_This section describes what ICacheRefresher and ICacheRefresher&lt;T&gt; are and how to use them to invalidate your cache correctly including load balanced environments_
+
+## What is an ICacheRefresher
+
+This interface has been in the Umbraco core for quite some time but has really only been used to ensure that content cache is refreshed among all server nodes participating in a load balanced scenario.
+
+With changes in 6.1, this has changed slightly, an `ICacheRefresher` is now the primary way to invalidate *any* cache that needs to be refreshed or removed regardless of whether we are in a load balanced environment.
```
Jeavon, line 13: I don't think we need to reference v6.1

PR #3429, opened 2021-08-31
Fix typo

greystate:
Just adding a missing letter "additinal" => "additional"


busrasengul:
Great catch @greystate 
Let's merge this in!

PR #3437, opened 2021-09-03
Update updating cache

erikjanwestendorp:
Update updating cache for v8


busrasengul:
Thank you @erikjanwestendorp 
We'll review this soon :)

PR #3441, opened 2021-09-03
add example of content cache event

jemayn:
Added an example of a cache event - the use case described is something I've been working on recently, and found that the cache events are covered in the docs at all, so figured I'd kick it off with an example 🙂


busrasengul:
Thank you @jemayn
We'll review this soon!!
✌🏻

PR #3445, opened 2021-09-08
Update namespaces in ContentService-Notifications

dinc5150:
Umbraco.Cms.Core.Service.Notifications namespace does not exist. The above docs worked when changing the namespace to Umbraco.Cms.Core.Notifications


busrasengul:
Thank you @dinc5150 
We'll review this 🙌

PR #3448, opened 2021-09-08
Redirect both old and new Umbraco Cloud url

bjarnef:
Some Umbraco Cloud projects are not migrated yet, but will probably be sometime soon in future.
If you not are added a technical contact on the project, you are not notified about when this happens in Umbraco Cloud portal.

With this redirect rule you can match both, so you don't need to update the redirect rule later when the project is migrated to new infrastructure.

Just checked on a new project launched, which is still running on `*.s1.umbraco.io`.


jonat123:
@bjarnef Thanks for this PR and sorry that you did not get a reply earlier 😕

Since all project has now been migrated to the new platform, it does not make sense to have rewrites for s1.umbraco.io anymore.

I will therefore close this PR.

PR #3465, opened 2021-09-14
Add dependency injection docs for V9

nikolajlauridsen:
This adds a V9 version of the dependency injection documentation: https://our.umbraco.com/documentation/Reference/Using-Ioc/

I've removed some sections like "Accessing LightInject container" since it doesn't make sense, and added some new ones like how to inject services into views, and how you should register dependencies in packages vs your own sites.

I've updated most links, however, two were missing the one for  UmbracoHelper and the one for Examine, I have updated all the code examples though.

### DIFF
--- a/Reference/Using-Ioc/index-v9.md
+++ b/Reference/Using-Ioc/index-v9.md
@@ -9,9 +9,9 @@ update-links: false
 
 # Inversion of Control / Dependency injection
 
-Umbraco 9 supports dependency injection out of the box. Umbraco uses the [ASP.NET Core built in dependency injection](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#service-lifetimes), this means that you don't have to install an external package to register and use your dependencies, and if you're familiar with APS.NET Core, the experience will be similar.
+Umbraco 9 supports dependency injection out of the box. Umbraco uses the [ASP.NET Core built-in dependency injection](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#service-lifetimes), this means that you don't have to install an external package to register and use your dependencies, and if you're familiar with APS.NET Core, the experience will be similar.
 
-`IUmbracoBuilder` is an Umbraco specific abstraction on top of the `IServiceCollection`, its purpose is to aid in adding and replacing Umbraco specific services, such notification handlers, filesystems, server role accessor, and so on. You can access the `IServiceCollection` directly to add your custom services through the `Services` property, see below for a concrete example: 
+`IUmbracoBuilder` is a Umbraco-specific abstraction on top of the `IServiceCollection`, its purpose is to aid in adding and replacing Umbraco-specific services, such as notification handlers, filesystems, server role accessor, and so on. You can access the `IServiceCollection` directly to add your custom services through the `Services` property, see below for a concrete example: 
 
 ```C#
 IUmbracoBuilder.Services
@@ -19,11 +19,11 @@ IUmbracoBuilder.Services
 
 ## Registering dependencies
 
-There's two strategies for registering your own dependencies to the container, which one you should use depends on whether you're making a package, or making custom services for your own site.
+There are two strategies for registering your own dependencies to the container, which one you should use depends on whether you're making a package, or making custom services for your own site.
 
 ### Registering dependencies for your site
 
-When working with your site, and not a package, the recommend way to registering depencies is with the `ConfigureServices` method of the `Startup` class in `Startup.cs`:
+When working with your site, and not a package, the recommended way to register dependencies is with the `ConfigureServices` method of the `Startup` class in `Startup.cs`:
 
 ```C#
 public void ConfigureServices(IServiceCollection services)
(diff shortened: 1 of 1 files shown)


Reference/Using-Ioc/index-v9.md
```diff
@@ -0,0 +1,370 @@
...
+        {
+            builder.AddNotificationHandler<ContentTypeSavedNotification, ContentTypeSavedHandler>();
+            builder.Services.AddSingleton<IFooBar, Foobar>();
+        }
+    }
+}
+```
+
+:::tip
+Remember to add `Umbraco.Cms.Core.DependencyInjection` and `Microsoft.Extensions.DependencyInjection` as 'using' statements where you register your services, to gain access to the `IUmbracoBuilder`, it's exension methods, and the Microsoft `IServiceProvider
```
jonat123, line 70:
```suggestion
Remember to add `Umbraco.Cms.Core.DependencyInjection` and `Microsoft.Extensions.DependencyInjection` as 'using' statements where you register your services, to gain access to the `IUmbracoBuilder`, its extension methods, and the Microsoft `IServiceProvider.
```

Reference/Using-Ioc/index-v9.md
```diff
@@ -0,0 +1,370 @@
...
+
+For more information, have a look at the official [Microsoft documentation](https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#service-lifetimes).
+
+## Injecting dependencies
+
+Once you have registered your services, factories, helpers or whatever you need for you application, you can go ahead and inject them where needed.
+
+### Injecting dependencies into a class
+
+If you need to inject your service intro a controller, or another service, you'll do so through the class
```
jonat123, line 179:
```suggestion
If you need to inject your service into a controller, or another service, you'll do so through the class
```

Reference/Using-Ioc/index-v9.md
```diff
@@ -0,0 +1,370 @@
...
+state: complete
+verified-against: rc-003
+update-links: false
+---
+
+# Inversion of Control / Dependency injection
+
+Umbraco 9 supports dependency injection out of the box. Umbraco uses the [ASP.NET Core built in dependency injection](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#service-lifetimes), this means that you don't have to install an external package to register and use your dependencies, and if you're familiar with APS.NET Core, the experience will be similar.
+
+`IUmbracoBuilder` is an Umbraco specific abstraction on top of the `IServiceCollection`, its purpose is to aid in adding and replacing Umbraco specific services, such notification handlers, filesystems, server role accessor, and so on. You can access the `IServiceCollection` directly to add your custom services through the `Services` property, see below for a concrete example:
```
jonat123, line 14:
```suggestion
# Inversion of Control / Dependency injection

Umbraco 9 supports dependency injection out of the box. Umbraco uses the [ASP.NET Core built-in dependency injection](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#service-lifetimes), this means that you don't have to install an external package to register and use your dependencies, and if you're familiar with APS.NET Core, the experience will be similar.

`IUmbracoBuilder` is a Umbraco-specific abstraction on top of the `IServiceCollection`, its purpose is to aid in adding and replacing Umbraco-specific services, such as notification handlers, filesystems, server role accessor, and so on. You can access the `IServiceCollection` directly to add your custom services through the `Services` property, see below for a concrete example: 
```

Reference/Using-Ioc/index-v9.md
```diff
@@ -0,0 +1,370 @@
...
+IUmbracoBuilder.Services
+```
+
+## Registering dependencies
+
+There's two strategies for registering your own dependencies to the container, which one you should use depends on whether you're making a package, or making custom services for your own site.
+
+### Registering dependencies for your site
+
+When working with your site, and not a package, the recommend way to registering depencies is with the `ConfigureServices` method of the `Startup` class in `Startup.cs`:
```
jonat123, line 26:
```suggestion
## Registering dependencies

There are two strategies for registering your own dependencies to the container, which one you should use depends on whether you're making a package, or making custom services for your own site.

### Registering dependencies for your site

When working with your site, and not a package, the recommended way to register dependencies is with the `ConfigureServices` method of the `Startup` class in `Startup.cs`:
```

Reference/Using-Ioc/index-v9.md
```diff
@@ -0,0 +1,370 @@
...
+}
+```
+
+:::tip
+Remember to add `Umbraco.Cms.Core.DependencyInjection` and `Microsoft.Extensions.DependencyInjection` as 'using' statements where you register your services, to gain access to the `IUmbracoBuilder`, it's exension methods, and the Microsoft `IServiceProvider
+:::
+
+### Builder extension methods
+
+Depending on your scenario, you may have a lot of dependencies you need to register, in this case your `Startup.cs` or Composer might become cluttered and hard to manage. A great way to manage multiple services is by creating your own custom extension methods for the `IUmbracoBuilder`, this way you can group similar dependencies in extension methods and register them all in as little as a single call:
```
jonat123, line 75:
```suggestion
### Builder extension methods

Depending on your scenario, you may have a lot of dependencies you need to register, in this case, your `Startup.cs` or Composer might become cluttered and hard to manage. A great way to manage multiple services is by creating your own custom extension methods for the `IUmbracoBuilder`, this way you can group similar dependencies in extension methods and register them all in as little as a single call:
```

Reference/Using-Ioc/index-v9.md
```diff
@@ -0,0 +1,370 @@
...
+    public class Foobar : IFooBar
+    {
+        public string Foo() => "Bar";
+    }
+}
+```
+
+### Injecting dependencies into a View or Template
+
+You might need to use services within your templates or views, fortunately you can inject services directly into your views using the `@inject` keyword. You can for example inject the `Foobar` from above into a view like so:
```
jonat123, line 219:
```suggestion
### Injecting dependencies into a View or Template

You might need to use services within your templates or views, fortunately, you can inject services directly into your views using the `@inject` keyword. You can for example inject the `Foobar` from above into a view like so: 
```

jonat123:
@nikolajlauridsen great work!!

I have once more made some suggestions to the grammar :D 

Other than that i don't have anything 🙂

nikolajlauridsen:
> I have once more made some suggestions to the grammar :D

Awesome, thanks 👍 What would the docs look like without em 🤣

PR #3469, opened 2021-09-15
Copy/paste text mistake on Cookie hijacking and protocol downgrade at…

dinc5150:
…tacks Protection page

The titles said "Click Jacking protection" instead of "Cookie hijacking and protocol downgrade attacks Protection". Titles were identical to https://our.umbraco.com/documentation/Extending/Health-Check/Guides/ClickJackingProtection so looks like a copy/paste error.

As a side note, this makes the titles rather long. Should it be shortened to "Adding Protection using NWebSec" and "Adding Protection using manual middleware" as the context of the page is only for Cookie hijacking and protocol downgrade attacks Protection, so may not be needed.


jonat123:
@dinc5150 again thanks for pointing out this 😄

Regarding your suggestion, I think it is a good point and shortening it "Adding Protection using NWebSec" and "Adding Protection using manual middleware" makes sense 😊

PR #3476, opened 2021-09-21
Add UmbracoHelper V9 docs

nikolajlauridsen:
This adds the V9 version of the `UmbracoHelper` docs.

Some of the interface has been moved to their own services, such as working with tags and members. Since I assume that people will go here to see how to do the same in V9, I've added separate articles for these things, but left in a reference to them in the `UmbracoHelper´, explaining that they no longer live on the `UmbracoHelper` and linking to the new articles I've added.

I noticed that an `IMemberManager` article has already been created, I've extended this article a bit so it now includes all the same examples that the `UmbracoHelper` docs used to have. 

I've updated all links except for one, there is a link to the examine article in the `IPublishedQuery` article, but the examine articles have not been made yet, so I was unable to update that link. 

~~I also found a bug with the `.Search(string term, int skip, int take, out long totalRecords)` method of `IPublishedQuery`, this has been fixed on the V9/dev branch, but does not work on RC-003, but since it has been fixed for RC-004 I don't thinks it's a big problem to merge regardless 😄~~ RC004 is out, so this bug is no longer a thing :D


jonat123:
@nikolajlauridsen Looks good, i don't have anything to add, so lets get it merged in :D

PR #3478, opened 2021-09-22
Spelling error

prjseal:
Changed loosing to losing.


busrasengul:
Good catch @prjseal 
Let's merge it in 💪🏼🙌

PR #3485, opened 2021-09-27
Add V9 common pitfalls

nikolajlauridsen:
This updates the common pitfalls document to V9, I've tried to use my common sense to try and make it relevant and accurate for V9, but I think it's a good idea if someone from the core team takes a look at it as well.


Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+meta.Title: "Common Pitfalls and Anti-Patterns in Umbraco"
+meta.Description: "Information on common Pitfalls and Anti-Patterns in Umbraco"
+state: complete
+verified-against: rc-004
+update-links: true
+---
+
+# Common Pitfalls & Anti-Patterns
+
+_This section is ultra important! It will describe many common pitfalls that developers fall in to. Some of the anti-patterns mentioned here can bring your site to a grinding halt, cause memory leaks, or make your site unstable or perform poorly. Make sure you read this section - it might save your site! Another great source of information is Shannon Deminick's Codegarden 2021 talk about performance tuning which can be seen [here](https://www.youtube.com/watch?v=RM2EGXm6DYU&list=PLG_nqaT-rbpx6wIDr5ufUlbHkg6qB3sxH)_
```
jonat123, line 12:
```suggestion
# Common Pitfalls & Anti-Patterns

_This section is ultra important! It will describe many common pitfalls that developers fall into. Some of the anti-patterns mentioned here can bring your site to a grinding halt, cause memory leaks, or make your site unstable or perform poorly. Make sure you read this section - it might save your site! Another great source of information is Shannon Deminick's Codegarden 2021 talk about performance tuning which can be seen [here](https://www.youtube.com/watch?v=RM2EGXm6DYU&list=PLG_nqaT-rbpx6wIDr5ufUlbHkg6qB3sxH)_
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+Depedency injection is available everywhere, and you can register your own services as well, additionally some resoureces are available through properties on certain base classes. For example, all Razor views that Umbraco creates expose an `UmbracoHelper` property you can access through `@Umbraco`, as well as a `SmidgeHelper` property. The other base classes that expose some things you might need like `UmbracoContext` are things like `SurfaceController`, but even here the services are initially gotten through DI, and you can inject further Umbraco and custom services that you might need.
+
+For more information about consuming and registering your own dependencies have a look at the [Dependency Injection](../../Reference/Using-Ioc/index-v9) documentation
```
jonat123, line 21:
```suggestion
## Usage of Singletons and Statics

Generally speaking, if you are writing software these days you should be using Dependency Injection principles.
If you do this, you probably aren't using Singletons or Statics (and for the most part you shouldn't be!). Since Umbraco 9 comes with dependency injection out of the box, there really isn't any reason to use singletons or statics. It makes your code very difficult to test but more importantly using Singletons and Statics in your code makes it very hard to manage, APIs become leaky, and ultimately you'll end up with more problems than when you started.

Dependency injection is available everywhere, and you can register your own services as well, additionally, some resources are available through properties on certain base classes. For example, all Razor views that Umbraco creates expose an `UmbracoHelper` property you can access through `@Umbraco`, as well as a `SmidgeHelper` property. The other base classes that expose some things you might need like `UmbracoContext` are things like `SurfaceController`, but even here the services are initially gotten through DI, and you can inject further Umbraco and custom services that you might need.

For more information about consuming and registering your own dependencies have a look at the [Dependency Injection](../../Reference/Using-Ioc/index) documentation
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+__Why?__
+
+It's important to understand the difference between an object that has a Request based scope/lifespan and
+an object that has an Singleton/Application based scope/lifespan ... here's the gist:
+
+* Application scope - if an object has an singleton/application scope/lifespan, that means that this single object
+instance will exist for the lifetime of the application. The single instance will be shared by every thread that
+accesses it. Static variables will always be application lifespan.
+* Request scope - The web world is made up of requests and each request has its own thread. When an object is in the scope of a Request it only survives as long as the web request survives. At the end of the web request, it may either be disposed or cleared from memory by the garbage collector. Request scoped object instances are not accessed by every other thread in the application - __unless you do something like the above!__
```
jonat123, line 94:
```suggestion
__Why?__

It's important to understand the difference between an object that has a Request based scope/lifespan and
an object that has a Singleton/Application based scope/lifespan ... here's the gist:

* Application scope - if an object has a singleton/application scope/lifespan, that means that this single object
instance will exist for the lifetime of the application. The single instance will be shared by every thread that
accesses it. Static variables will always be application lifespan.
* Request scope - The web world is made up of requests and each request has its own thread. When an object is in the scope of a Request it only survives as long as the web request survives. At the end of the web request, it may either be disposed of or cleared from memory by the garbage collector. Request scoped object instances are not accessed by every other thread in the application - __unless you do something like the above!__
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+An example of a request scoped instance is the `HttpContext`. This object exists for a single request and it definitely cannot be shared between other threads and especially not other request threads. This is because it is where the security information for a given user is stored! The `UmbracoContext` is also a request scoped object - in fact it relies directly on an instance of `HttpContext`. The `UmbracoHelper` is request scoped as well.
+
+So... in the above example, the `UmbracoHelper` which has a request scoped lifetime, will now be statically assigned to a variable, which means that this particular request scoped object is now bound to an Application scope lifetime and will never go away. This could mean that under certain circumstances that an entire Umbraco
+cache copy is stuck in memory, or that the `Security` property of the context will be accessed by multiple threads but this now contains the security information for a user for another request!
```
jonat123, line 99:
```suggestion
An example of a request-scoped instance is the `HttpContext`. This object exists for a single request and it definitely cannot be shared between other threads and especially not other request threads. This is because it is where the security information for a given user is stored! The `UmbracoContext` is also a request-scoped object - in fact, it relies directly on an instance of `HttpContext`. The `UmbracoHelper` is request-scoped as well.

So... in the above example, the `UmbracoHelper` which has a request-scoped lifetime, will now be statically assigned to a variable, which means that this particular request-scoped object is now bound to an Application scope lifetime and will never go away. This could mean that under certain circumstances that an entire Umbraco
cache copy is stuck in memory, or that the `Security` property of the context will be accessed by multiple threads but this now contains the security information for a user for another request!
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+Additionally since V9 comes with depdency injection out of the box, there's never really any reason to use static references, instead you should always use inject your required resources, and let the DI container handle the lifetimes of the objects.
```
jonat123, line 101:
```suggestion
Additionally, since V9 comes with dependency injection out of the box, there's never really any reason to use static references, instead, you should always use inject your required resources, and let the DI container handle the lifetimes of the objects.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+</ul>
+```
+
+In many cases you might know that there is only ever going to be a small number of Descendants . If so then go nuts and use Descendants or DescendantsOrSelf. It's important to be aware of the implications of what you are writing.
+
+## Too much querying (Over querying)
+
+Querying content is not Free! Anytime you make a query or resolve a property value be aware that there is overhead involved.
+You could try to think about every query you make as an SQL call - you don't want to make too many otherwise the performance of
+your website is going to suffer.
```
jonat123, line 152:
```suggestion
In many cases, you might know that there is only ever going to be a small number of Descendants. If so then go nuts and use Descendants or DescendantsOrSelf. It's important to be aware of the implications of what you are writing.

## Too much querying (Over querying)

Querying content is not Free! Anytime you make a query or resolve a property value be aware that there is overhead involved.
You could try to think about every query you make as an SQL call - you don't want to make too many otherwise the performance of
your website is going to suffer.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+Here's a common pitfall that is seen. Let's continue the menu example, in this example the menu is going to be rendered
```
jonat123, line 154:
```suggestion
Here's a common pitfall that is seen. Let's continue the menu example, in this example, the menu is going to be rendered
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+```
+
+## Using the Services layer in your views
+
+The Services layer of Umbraco is for manipulating the business logic of Umbraco directly to/from the database.
+None of these methods should be used within your views and can have a very large impact on performance and stability of
+your application.
+
+Your views should rely only on the read-only data services such as `UmbracoHelper`, `ITagQuery` and `IMemberManager` and the properties/methods that they expose. This ensures
+that the data being queried is fast (comes from cache) and that you aren't inadvertently making database changes.
```
jonat123, line 191:
```suggestion
## Using the Services layer in your views

The Services layer of Umbraco is for manipulating the business logic of Umbraco directly to/from the database.
None of these methods should be used within your views and can have a very large impact on the performance and stability of your application.

Your views should rely only on the read-only data services such as `UmbracoHelper`, `ITagQuery` and `IMemberManager` and the properties/methods that they expose. This ensures
that the data being queried is fast (comes from cache) and that you aren't inadvertently making database changes.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+Umbraco's content should not be used for volatile data, Umbraco's APIs and the way Umbraco's data is persisted
+was never designed for this. If you need to store/write/track data that changes a lot you should use a
+custom database table or another service but not Umbraco content nodes.
+
+Some examples of what not to do are:
+
+* Hit counters to track the number of times your page has been viewed - use something like Google Analytics for this or a custom database table
+* Creating new nodes for form submissions - this should be stored in a custom database table
+* Importing lots of data into Umbraco content nodes that could be stored in a custom database table (i.e. it's not going to be edited).
+In some cases this might be ok but many times we've seen bulk imports occur on an hourly/daily schedule which is generally unnecessary.
```
jonat123, line 223:
```suggestion
## Using Umbraco content items for volatile data

This is one of the worst Umbraco anti-patterns and could very well cause your site to perform ultra poorly.

Umbraco's content should not be used for volatile data, Umbraco's APIs and the way Umbraco's data is persisted
was never designed for this. If you need to store/write/track data that changes a lot you should use a
custom database table or another service but not Umbraco content nodes.

Some examples of what not to do are:

* Hit counters to track the number of times your page has been viewed - use something like Google Analytics for this or a custom database table
* Creating new nodes for form submissions - this should be stored in a custom database table
* Importing lots of data into Umbraco content nodes that could be stored in a custom database table (i.e. it's not going to be edited).
In some cases, this might be ok but many times we've seen bulk imports occur on an hourly/daily schedule which is generally unnecessary.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+synchronization of the data in the index gets better with every release. You should always ensure you are running the latest
+Umbraco and Examine versions if you are having trouble with your index data becoming out of sync with your Umbraco data.
+
+The primary reasons your data will become out of sync are:
+
+* Old version of Umbraco
+* Rebuilding indexes and restarting your app domain at the same time (try to avoid this scenario!)
+
+It is not recommended to rebuild your indexes unless you absolutely need to and if you need to do this often then it is
+advised to determine why and to try to resolve the underlying problem.
```
jonat123, line 259:
```suggestion
## Rebuilding indexes

Far too often we've seen code in people's solutions that rebuild the Examine indexes
(we've even seen this done on every request!). Rebuilding indexes can cause severe
performance penalties and is not a recommended practice. Umbraco's and Examine's index management, index stability, and
synchronization of the data in the index get better with every release. You should always ensure you are running the latest
Umbraco and Examine versions if you are having trouble with your index data becoming out of sync with your Umbraco data.

The primary reasons your data will become out of sync are:

* Old version of Umbraco
* Rebuilding indexes and restarting your app domain at the same time (try to avoid this scenario!)

It is not recommended to rebuild your indexes unless you absolutely need to and if you need to do this often then it is
advised to determine why and to try to resolve the underlying problem.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+## RenderTemplateAsync
+
+There is an API in Umbraco that should never be used unless you really know what you are doing. This API method
+is called `RenderTemplateAsync`. It allows you to be able to render a particular content item's template and get a `IHtmlEncodedString`
+in response. In some cases, this may be useful. Perhaps you want to send an email based on a content item and its template, but
+you must be very careful not to use this for purposes it is not meant to be used for.
+
+Generally speaking this method should not be used for the normal rendering of content. If abused this could cause severe
+performance problems. For normal content rendering of module type data from another content item, you should use Partial Views instead.
```
jonat123, line 282:
```suggestion
## RenderTemplateAsync

There is an API in Umbraco that should never be used unless you really know what you are doing. This API method
is called `RenderTemplateAsync`. It allows you to be able to render a particular content item's template and get a `IHtmlEncodedString`
in response. In some cases, this may be useful. Perhaps you want to send an email based on a content item and its template, but
you must be very careful not to use this for purposes it is not meant to be used for.

Generally speaking, this method should not be used for the normal rendering of content. If abused this could cause severe
performance problems. For normal content rendering of module type data from another content item, you should use Partial Views instead.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+Here's an example of how this can go wrong very quickly:
+Your tree structure is something like this:
+
+    - Root
+    -- Home
+    --- Recipes (node id = 3251, list view with 5000 nodes)
+    --- About Us
+    --- Contact Us
+
+You have a custom model that looks like:
```
jonat123, line 302:
```suggestion
## Don't put logic inside your constructors

Constructors should generally not perform any logic, they should set some parameter values, perform some null checks and perhaps validate some data but in most cases, they should not perform any logic.

There are a few reasons why this can become a huge performance problem:

* The consumer of an API doesn't expect that by creating an object that they should be worried about performance
* Creating an object can inadvertently happen a vast number of times, especially when using LINQ

Here's an example of how this can go wrong very quickly:
Your tree structure is something like this:

    - Root
    -- Home
    --- Recipes (node id = 3251, list view with 5000 nodes)
    --- About Us
    --- Contact Us

You have a custom model that looks like this:
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+```
+
+__Ouch!__ To show the top 10 voted recipe's this will end up doing the following:
+
+* This will iterate over all Recipes, create and allocate 5000 instances of `IPublishedContent`
+* This will create and allocate 5000 instances of `RecipeModel`
+* For each `RecipeModel` created, this will traverse upwards, iterate all 5000 recipes then resolve property data for 2 properties
+
+This means that there is now a minimum of __20,000__ new objects created and allocated in memory. The number of traversals/visits to each
+of these objects is now 5000 x 5000 = __25,000,000 (25 MILLION TIMES!)__
```
jonat123, line 352:
```suggestion
__Ouch!__ To show the top 10 voted recipes's this will end up doing the following:

* This will iterate over all Recipes, create and allocate 5000 instances of `IPublishedContent`
* This will create and allocate 5000 instances of `RecipeModel`
* For each `RecipeModel` created, this will traverse upwards, iterate all 5000 recipes then resolve property data for 2 properties

This means that there is now a minimum of __20,000__ new objects created and allocated in memory. The number of traversals/visits to each
of these objects is now 5000 x 5000 = __25,000,000 (25 MILLION TIMES!)__
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+_Side note: The other problem is the logic used to lookup related recipes is incredibly inefficient. Instead, each recipe should have a picker to choose its related recipe's and then each of those can be looked up by their ID.
+(There's probably a few other ways to achieve this too!)_
+
+Which leads us on to the next anti-pattern...
+
+## Don't eager load data, lazy load it instead
+
+The above example could be rewritten like this:
```
jonat123, line 361:
```suggestion
_Side note: The other problem is the logic used to lookup related recipes is incredibly inefficient. Instead, each recipe should have a picker to choose its related recipe's and then each of those can be looked up by their ID.
(There's probably a few other ways to achieve this too!)_

This leads us on to the next anti-pattern...

## Don't eager load data, lazy load it instead

The above example could be rewritten like this:
```
```suggestion
_Side note: The other problem is the logic used to lookup related recipes is incredibly inefficient. Instead, each recipe should have a picker to choose its related recipe's and then each of those can be looked up by their ID.
(There's probably a few other ways to achieve this too!)_

This leads us on to the next anti-pattern...

## Don't eager load data, lazy load it instead

The above example could be rewritten like this:
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+This is slightly better:
+
+This means that there is now a minimum of __10,000__ new objects created and allocated in memory. The number of traversals/visits to each
+of these objects is now __5000__.
+
+## Not caching epxensive lookups
+
+Based on the above 2 points, you can see that iterating content with the traversal APIs will cause new 
+instances of `IPublishedContent` to be created. When memory is used, Garbage Collection needs to occur and this
+turnover can cause performance problems. The more objects created, the more items allocated in memory, the harder the job is for the Garbage Collector == more performance problems. Even worse is when you allocate tons of items in memory and/or really large items in memory. They will remain in memory for a long time because they'll end up in something called "Generation 3" which the GC tries to ignore for as long as possible. It does so because it knows it's going to take a lot of resources to cleanup!
```
jonat123, line 423:
```suggestion
## Not caching expensive lookups

Based on the above 2 points, you can see that iterating content with the traversal APIs will cause new instances of `IPublishedContent` to be created. When memory is used, Garbage Collection needs to occur and this
turnover can cause performance problems. The more objects created, the more items allocated in memory, the harder the job is for the Garbage Collector == more performance problems. Even worse is when you allocate tons of items in memory and/or really large items in memory. They will remain in memory for a long time because they'll end up in something called "Generation 3" which the GC tries to ignore for as long as possible. It does so because it knows it's going to take a lot of resources to clean up!
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,426 @@
...
+
+
+So what can we do to mitigate this? Unfortunately there is no silver bullet that will solve all your performance issues, it will always depend on your specific scenario. One great tip though is to cache the IDs of the content you need in your critical code, and then retrieve it from the cache by ID. For instance if you need to render the same four pieces of content for your nav, then cache, or simply hardcore, the IDs of the content items and retrieve them with the ID using `Umbraco.Content`, this will always be much, much faster than trying to traverse your content tree and finding the content programatically, since it will do a direct lookup in the cache, meaning that only the `IPublishedContent` you actually need will be created.
```
jonat123, line 426:
```suggestion
So what can we do to mitigate this? Unfortunately, there is no silver bullet that will solve all your performance issues, it will always depend on your specific scenario. One great tip though is to cache the IDs of the content you need in your critical code and then retrieve it from the cache by ID. For instance, if you need to render the same four pieces of content for your nav, then cache, or simply hardcore, the IDs of the content items and retrieve them with the ID using `Umbraco.Content`, this will always be much, much faster than trying to traverse your content tree and finding the content programmatically, since it will do a direct lookup in the cache, meaning that only the `IPublishedContent` you actually need will be created.
```

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,424 @@
...
+        public IEnumerable<int> RelatedRecipes => _related ??= this.Value<IEnumerable<int>>("related").ToList();
+    }
+```
+
+This is slightly better:
+
+* This will iterate over all Recipes, create and allocate 5000 instances of `IPublishedContent`
+* This will create and allocate 5000 instances of `RecipeModel`
+
+This means that there is now a minimum of __15,000__ new objects created and allocated in memory. The number of traversals/visits to each
```
Shazwazza, line 392: As above, this will no longer allocate new instances. It will still iterate them.

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+This is slightly better:
+
+This means that there is now a minimum of __10,000__ new objects created and allocated in memory. The number of traversals/visits to each
+of these objects is now __5000__.
+
+## Not caching expensive lookups
+
+Based on the above 2 points, you can see that iterating content with the traversal APIs will cause new instances of `IPublishedContent` to be created. When memory is used, Garbage Collection needs to occur and this
```
Shazwazza, line 420: As above, this will no longer allocate new instances. But this concept still stands that the more you allocate the more the GC needs to clean up

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,424 @@
...
+
+
+So what can we do to mitigate this? Unfortunately, there is no silver bullet that will solve all your performance issues, it will always depend on your specific scenario. One great tip though is to cache the IDs of the content you need in your critical code and then retrieve it from the cache by ID. For instance, if you need to render the same four pieces of content for your nav, then cache, or simply hardcore, the IDs of the content items and retrieve them with the ID using `Umbraco.Content`, this will always be much, much faster than trying to traverse your content tree and finding the content programmatically, since it will do a direct lookup in the cache, meaning that only the `IPublishedContent` you actually need will be created.
```
Shazwazza, line 424: hardcore = hardcode

Shazwazza (changes requested):
Left comments, mostly about allocations.

Reference/Common-Pitfalls/index-v9.md
```diff
@@ -0,0 +1,424 @@
...
+		<li><a href="@recipe.Url()">@recipe.Name</a></li>
+	}	
+</ul>
+
+```
+
+__Ouch!__ To show the top 10 voted recipes's this will end up doing the following:
+
+* This will iterate over all Recipes, create and allocate 5000 instances of `IPublishedContent`
+* This will create and allocate 5000 instances of `RecipeModel`
```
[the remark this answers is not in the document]
nikolajlauridsen: Since the example does `.Select(x => new RecipeModel(x, _publishedValueFallback))`, then I suppose we'll still create an allocate 5000 instances of `RecipeModel` even though the initial content is already in the cache? Since the Linq needs to create the objects to order by votes, so I've left that in, but adjusted the rest accordingly :D

nikolajlauridsen:
I've updated the docs removing the references to allocating objects instead of the one where we explicitly create them with `Select` 😄, and fixed the examine event

jonat123:
@nikolajlauridsen Looks like there is a conflict, could you resolve it, then we can get this merged in? :D

nikolajlauridsen:
> @nikolajlauridsen Looks like there is a conflict, could you resolve it, then we can get this merged in? :D

Of course, it should be ready now 😄

jonat123:
Perfect! 

Merging it in :D

PR #3521, opened 2021-10-03
Verify members technical details and add new config options

jemayn:
Verified the details in the v8 article and added new info for the v9 config. Didn't change anything with the members service links as they should be the same?

Let me know if anything needs changing 🙂


jonat123:
@jemayn thanks for the PR :D 

We will make sure to review it asap! :D

PR #3528, opened 2021-10-05
Updated to version 9

Aaen:
Updated prepare your frontend article to v9

### DIFF
--- a/Add-ons/UmbracoForms/Developer/Prepping-Frontend/index-v9.md.md
+++ b/Add-ons/UmbracoForms/Developer/Prepping-Frontend/index-v9.md.md
@@ -1,0 +1,66 @@
+---
+versionFrom: 9.0.0
+---
+
+# Preparing your Frontend
+
+For Umbraco Forms to work correctly, you need to include some client dependencies.
+
+## Umbraco Forms version 9.0.0
+
+You can use the following Razor helper to output script tags containing the dependencies. To access this method you will need a reference to `Umbraco.Forms.Mvc`:
+
+```html
+@using Umbraco.Forms.Web
+<head>
+    @Html.RenderUmbracoFormDependencies()
+</head>
+```
+
+Alternatively, you can add the dependencies to the body tag:
+
+```html
+@using Umbraco.Forms.Web
+...
+
+<body>
+    @Html.RenderUmbracoFormDependencies()
+</body>
+```
+
+All dependencies originate from your Umbraco Forms installation, which means that no external references are needed.
+
+The easiest way to add the dependencies is to fetch them from a [CDN](https://en.wikipedia.org/wiki/Content_delivery_network). There are various CDN services you can use:
+
+- For example: [Microsoft CDN](https://docs.microsoft.com/en-us/aspnet/ajax/cdn/overview).
+- Other CDN services you might want to look at are https://www.jsdelivr.com/ and https://cdnjs.com/about, which may offer better performance and more reliable service.
+
+To add the three client dependencies, see the examples below:
+
+**Example within `head` tags.**
+
+```html
+<head>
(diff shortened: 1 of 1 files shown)


Add-ons/UmbracoForms/Developer/Prepping-Frontend/index-v9.md.md
```diff
@@ -0,0 +1,66 @@
+---
+versionFrom: 9.0.0
+---
+
+# Preparing your Frontend
+
+For Umbraco Forms to work correctly, you need to include some client dependencies.
+
+## Umbraco Forms version 9.0.0
```
AndyButland, line 9:
```suggestion
## Validation Without External References
```

Think here we don't need to have a sub-heading for the version, as this page is only for V9.  We do want a sub-heading though as per my comment below, where we need to distinguish the jquery part.

Add-ons/UmbracoForms/Developer/Prepping-Frontend/index-v9.md.md
```diff
@@ -0,0 +1,66 @@
...
+...
+
+<body>
+    @Html.RenderUmbracoFormDependencies()
+</body>
+```
+
+All dependencies originate from your Umbraco Forms installation, which means that no external references are needed.
+
+The easiest way to add the dependencies is to fetch them from a [CDN](https://en.wikipedia.org/wiki/Content_delivery_network). There are various CDN services you can use:
```
AndyButland, line 33:
```suggestion
## Validation Using jQuery

If you want to use jQuery as your validation framework for Umbraco Forms, you can manually add the following client dependencies without using the above Razor method:

- `jQuery` (JavaScript library)
- `jQuery validate` (jQuery plugin that provides client-side Form validation)
- `jQuery validate unobtrusive` (Add-on to jQuery Validation that provides unobtrusive validation via data-* attributes)

The easiest way to add the dependencies is to fetch them from a [CDN](https://en.wikipedia.org/wiki/Content_delivery_network). There are various CDN services you can use:
```

Think we need to add this sub-heading and restore the paragraph, so it's clear why we are now talking about jQuery.

Add-ons/UmbracoForms/Developer/Prepping-Frontend/index-v9.md.md
```diff
@@ -0,0 +1,66 @@
...
+versionFrom: 9.0.0
+---
+
+# Preparing your Frontend
+
+For Umbraco Forms to work correctly, you need to include some client dependencies.
+
+## Umbraco Forms version 9.0.0
+
+You can use the following Razor helper to output script tags containing the dependencies. To access this method you will need a reference to `Umbraco.Forms.Mvc`:
```
AndyButland, line 11:
```suggestion
You can use the following Razor helper to output script tags containing the dependencies. To access this method you will need a reference to `Umbraco.Forms.Web`:
```

Aaen:
Updated

Aaen:
@AndyButland @jonat123 I have updated this now

AndyButland:
I'm not seeing any updates on the PR though yet @Aaen.

And one more small thing I just saw - your new file has an extension of `md.md` but it should just be `.md`.

Aaen:
@AndyButland Better now I am a bit rusty :)

AndyButland (approved):
Looks good to me now @Aaen - thanks.

AndyButland:
Oops... completely forgotten I'd already submitted that!  I'm happy if you take this and ignore my other one (I've already looked at this one too and it's all good with me).

AndyButland:
Just realised this one may be the wrong way around @jonat123 - i.e. we're showing the V8 page by default rather than V9, and in other areas I think we changed it to be V9 by default?

PR #3536, opened 2021-10-06
Adds details of package migration configuration added in CMS and Deploy 9.0.1

AndyButland:
This should go live on Tuesday 12th October when we plan to release 9.0.1 patch versions of CMS and Deploy.


AndyButland:
Thanks - have made the updates.

PR #3554, opened 2021-10-10
Updating to remove reference to RC

crgrieve:
Umbraco 9 is now out, yay! So, removing the reference to RC and also the warning as it could be misleading.


busrasengul:
This looks great @crgrieve 
Let's merge this in! 🎉
#h5yr

PR #3570, opened 2021-10-18
Update SMTP.md for local development example

cheeseytoastie:
If you use SMTP4dev as suggested (as I do) you'll likely hit SSL handshake errors with the previous example. This example works. Also the port number in the previous example should have been an int as per intellisense / validation.


busrasengul:
Thank you @cheeseytoastie 
We'll get this reviewed soon!
🙌

marcemarc:
Think decision here is to 'close' and keep sample to the bare minimum, but reopen and shout into the wind if you disagree!

PR #3586, opened 2021-10-25
UmbracoContextAccessor doesn't contain a reference to UmbracoContext

rickbutterfield:
After trying to follow this example, I discovered that `IUmbracoContextAccessor` doesn't let you reference `UmbracoContext`, so you either have to use `_umbracoContextAccessor.GetRequiredUmbracoContext();` or `_umbracoContextAccessor.TryGetUmbracoContext()`

### DIFF
--- a/Reference/Routing/custom-routes.md
+++ b/Reference/Routing/custom-routes.md
@@ -193,32 +193,39 @@ Now that we have our dependencies, and our action methods, we're finally ready t
 ```C#
 public IPublishedContent FindContent(ActionExecutingContext actionExecutingContext)
 {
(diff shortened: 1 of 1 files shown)


marcemarc:
Thanks @rickbutterfield we've made a few tweaks but that's good to go!
I'm not sure about using the GetRequiredUmbracoContext extension method as it will throw an InvalidOperationException if no 
UmbracoContext is found (eg on a client side route like sitemap.xml) 

image

So using TryGetUmbracoContext in the example gives people a better idea of handling that in the context they are using it... 
... maybe... 

But thanks for correcting the original error!!!

PR #3599, opened 2021-10-29
Integration-Testing

Zeegaan:
As a note, this will only be possible from 9.1.0 rc and forwards

# Notes
- Added Integration-Testing
- Added Integration Testing with NotificationHandlers
- Added Integration Testing with a Schema

### DIFF
--- a/Implementation/Integration-Testing/index.md
+++ b/Implementation/Integration-Testing/index.md
@@ -1,0 +1,103 @@
+---
+versionFrom: 9.1.0
+needsV9Update: false
+meta.Title: "Integration Testing Umbraco"
+meta.Description: "A guide to getting started with integration testing in Umbraco"
+---
+
+# Integration Testing Umbraco
+
+
+These examples are for Umbraco 8+ and they rely on [NUnit](https://nunit.org/) and [Umbraco.Cms.Tests.Integration](https://github.com/Umbraco).
+
+# Using the NuGet package and attributes in your class
+
+In your NUnit class you need to use the NuGet package Umbraco.Cms.Tests.Integration, and then decorate your class with the TestFixture attribute & UmbracoTest attribute, with the UmbracoTest attributes there are multiple options:
+- None
+- NewEmptyPerFixture
+- NewEmptyPerTest
+- NewSchemaPerFixture
+- NewSchemaPerTest
+
+Besides that you also need to derive from the UmbracoIntegrationTest class, with this you are now ready to start Integration Testing.
+```csharp
+using Umbraco.Cms.Tests.Integration;
+
+[TestFixture]
+[UmbracoTest(Database = UmbracoTestOptions.Database.NewSchemaPerTest)]
+public class IntegrationTests : UmbracoIntegrationTest
+{
+
+}
+```
+
+## Testing a notification
+
+Start by making a NotificationHandler, this example will be of one cancelling overwrites on content named "Root", so if you have some content named "Root" published, you cannot change it.
+
+```csharp
+public class MyNotificationHandler : INotificationHandler<ContentSavingNotification>
+{
+    public void Handle(ContentSavingNotification notification)
+    {
+        foreach (var content in notification.SavedEntities)
+        {
+            if (content.PublishName == "Root")
+            {
+                notification.CancelOperation(new EventMessage("Cancelled", "Please do not change root content",
+                    EventMessageType.Error));
+            }
+        }
+    }
+}
+```
+Then we can make an integration tests, we do have to register our notification in the test like you would do with a composer, we can do this by overriding the `CustomTestSetupMethod` and adding the notification.
+After this we can build our ContentType and Content with their respective builders.
+When we are saving both the ContentType & Content, we need the services to do so, so we use the `GetRequiredService<IService>` method that can get the services we need.
+We can then use `Assert.Multiple()` to do multiple asserts
+
+```csharp
+
+[TestFixture]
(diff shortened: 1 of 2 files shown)


Implementation/Integration-Testing/index.md
```diff
@@ -0,0 +1,144 @@
...
+---
+
+# Integration Testing Umbraco
+
+
+These examples are for Umbraco 8+ and they rely on [NUnit](https://nunit.org/) and [Umbraco.Cms.Tests.Integration](https://github.com/Umbraco).
+
+# Using the NuGet package and attributes in your class
+
+In your NUnit class you need to use the NuGet package Umbraco.Cms.Tests.Integration, and then decorate your class with the TestFixture attribute & UmbracoTest attribute, with the UmbracoTest attributes there are multiple options:
```
jonat123, line 15:
```suggestion
In your NUnit class, you need to use the NuGet package Umbraco.Cms.Tests.Integration, and then decorate your class with the TestFixture attribute & UmbracoTest attribute, with the UmbracoTest attributes there are multiple options:
```

Implementation/Integration-Testing/index.md
```diff
@@ -0,0 +1,144 @@
...
+[UmbracoTest(Database = UmbracoTestOptions.Database.NewSchemaPerTest)]
+public class IntegrationTests : UmbracoIntegrationTest
+{
+
+}
+```
+
+## Testing a notification
+
+Start by making a NotificationHandler, this example will be of one cancelling overwrites on content named "Root", so if you have some content named "Root" published, you cannot change it.
```
jonat123, line 36:
```suggestion
Start by making a NotificationHandler, this example will be of one canceling overwrites on content named "Root", so if you have some content named "Root" published, you cannot change it.
```

Implementation/Integration-Testing/index.md
```diff
@@ -0,0 +1,144 @@
...
+                notification.CancelOperation(new EventMessage("Cancelled", "Please do not change root content",
+                    EventMessageType.Error));
+            }
+        }
+    }
+}
+```
+Then we can make an integration tests, we do have to register our notification in the test like you would do with a composer, we can do this by overriding the `CustomTestSetupMethod` and adding the notification.
+After this we can build our ContentType and Content with their respective builders.
+When we are saving both the ContentType & Content, we need the services to do so, so we use the `GetRequiredService<IService>` method that can get the services we need.
```
jonat123, line 56:
```suggestion
Then we can make an integration test, we do have to register our notification in the test like you would do with a composer, we can do this by overriding the `CustomTestSetupMethod` and adding the notification.
After this, we can build our ContentType and Content with their respective builders.
When we are saving both the ContentType & Content, we need the services to do so, so we use the `GetRequiredService<IService>` method that can get the services we need.
```

Implementation/Integration-Testing/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Testing with a schema
+
+So one of the awesome things about integration tests, is that you can setup a site, download the package for it, and we can run this state for every test.
+This means that you do not have to go through and setup your tests with data, like we do in the above example with the builder pattern.
+
+To start with we decorate our class with the `[UmbracoTest]` attribute and we again derive from `UmbracoIntegrationTest`
+Then what you wanna do is setup your umbraco site, go to the packages section and create your own package. Download the package and place the xml file next to your testing class. You want to have the build action of that xml file to be `EmbeddedRessource`
+
+Now we're almost ready to start testing! Last thing we wanna do is have a Setup method to install the package on your site.
```
jonat123, line 110:
```suggestion
So one of the awesome things about integration tests, is that you can set up a site, download the package for it, and we can run this state for every test.
This means that you do not have to go through and set up your tests with data like we do in the above example with the builder pattern.

To start with we decorate our class with the `[UmbracoTest]` attribute and we again derive from `UmbracoIntegrationTest`
Then what you wanna do is set up your Umbraco site, go to the packages section and create your own package. Download the package and place the XML file next to your testing class. You want to have the build action of that XML file to be `EmbeddedRessource`

Now we're almost ready to start testing! The last thing we wanna do is have a Setup method to install the package on your site.
```

Implementation/Integration-Testing/index.md
```diff
@@ -0,0 +1,144 @@
...
+public void MySetup()
+{
+    var xml = PackageMigrationResource.GetEmbeddedPackageDataManifest(this.GetType());
+    var packagingService = GetRequiredService<IPackagingService>();
+    packagingService.InstallCompiledPackageData(xml);
+}
+```
+
+Now you're all-set to start testing with your own site! Lets try and see how that would look! 
+Here's an example test, where we test that content is deleted, if you delete the Document Types, as you can see, this time we do not have to use builder patterns to setup our site!
```
jonat123, line 123:
```suggestion
Now you're all set to start testing with your own site! Let's try and see how that would look! 
Here's an example test, where we test that content is deleted, if you delete the Document Types, as you can see, this time we do not have to use builder patterns to set up our site!
```

jonat123:
@Zeegaan Looks really good 😄

I have made a few grammatical suggestions 😊

Let me know what you think 😃

Zeegaan:
> @Zeegaan Looks really good 😄
> 
> I have made a few grammatical suggestions 😊
> 
> Let me know what you think 😃

Looks good to me, thanks for reviewing @jonat123 ! 💪

jonat123:
@Zeegaan With the release of the RC, let's merge this PR in 😄

PR #3606, opened 2021-10-29
Backoffice search, code samples updated and verified against v9

busrasengul:


marcemarc (approved):
the original paragraph threw up a lint error, so I updated!

PR #3607, opened 2021-10-29
Information about deploying v9 to linux app service

mattou07:


busrasengul:
Thank you @mattou07 
We'll get this reviewed soon 🙌

Fundamentals/Setup/Server-Setup/azure-web-apps.md
```diff
@@ -61,3 +61,31 @@ It's important to know that Azure Web Apps may move your website between their '
 
 When your site is migrated to another worker, these variables will change.
 You cannot rely on these variables remaining static for the lifetime of your website.
+
+## Deploying to Linux App Service
+
+To avoid cross compatibility issues its recommeded to deploy to Linux App Service using a managed Dev Ops service such as Azure DevOps and configure a Linux build agent.
```
marcemarc, line 67:
```suggestion
To avoid cross compatibility issues its recommended to deploy to Linux App Service using a managed Dev Ops service such as Azure DevOps and configure a Linux build agent. 
```
typo!

mattou07: Thanks @marcemarc

sofietoft:
Hi @mattou07 

It's been a while since you made this PR - and sorry for the delay btw 🙈 

Just wondering if you know whether running Umbraco on Linux still requires this workaround?

p-m-j:
@sofietoft fixed in 9.4.0 by https://github.com/umbraco/Umbraco-CMS/pull/11961

sofietoft:
That sounds great! Thanks for pitching in @p-m-j 🙌 

Does that mean that this PR is obsoleted with the release of 9.4? (Just double-checking)

Fundamentals/Setup/Server-Setup/azure-web-apps.md
```diff
@@ -61,3 +61,31 @@ It's important to know that Azure Web Apps may move your website between their '
...
+```xml
+<ItemGroup>
+        <PackageReference Include="Clean" Version="2.0.0-beta002" />
+        <PackageReference Include="Microsoft.ICU.ICU4C.Runtime" Version="68.2.0.6" />
+        <RuntimeHostConfigurationOption Include="System.Globalization.AppLocalIcu" Value="68.2" />
+    </ItemGroup
+```
+This will cause the following error in your environment: `Process terminated. Failed to load app-local ICU: libicudata.so.68.2` and crash the SSH client in the KUDU console.
+
+Azure would attempt to start your environment and simply crash with this error in the log stream. Commenting out this ItemGroup and recreating the environment should fix this issue.
```
p-m-j, line 79:
```suggestion

```

Fundamentals/Setup/Server-Setup/azure-web-apps.md
```diff
@@ -61,3 +61,31 @@ It's important to know that Azure Web Apps may move your website between their '
...
+        <PackageReference Include="Microsoft.ICU.ICU4C.Runtime" Version="68.2.0.6" />
+        <RuntimeHostConfigurationOption Include="System.Globalization.AppLocalIcu" Value="68.2" />
+    </ItemGroup
+```
+This will cause the following error in your environment: `Process terminated. Failed to load app-local ICU: libicudata.so.68.2` and crash the SSH client in the KUDU console.
+
+Azure would attempt to start your environment and simply crash with this error in the log stream. Commenting out this ItemGroup and recreating the environment should fix this issue. 
+
+If you have slots, creating an empty slot and swapping it with the main slot is enough to recreate the environment for a future, issue free, deployment.
+
```
p-m-j, line 82: I wouldn't describe deployment slots as issue free, https://github.com/umbraco/Umbraco-CMS/issues/11935 See https://github.com/umbraco/UmbracoDocs/issues/3894

p-m-j:
> Does that mean that this PR is obsoleted with the release of 9.4? (Just double-checking)

Mostly yes, there may be utility in the logs section.

p-m-j:
@mattou07 it's probably easier to just use bash/powershell on kudu to remove the umbraco/data/TEMP folder which was likely the source of your problems, i'd be tempted to omit the suggestion of slot swapping entirely as it mandates additional required reading.

mattou07:
> @mattou07 it's probably easier to just use bash/powershell on kudu to remove the umbraco/Data/TEMP/ folder which was likely the source of your problems, i'd be tempted to omit the suggestion of slot swapping entirely as it mandates additional required reading.

My KUDU console was crashing due to the original issue mentioned around the csproj file. The app was unable to start at all and affected the web app instance on the slot.

By creating a new slot and swapping it with the bad slot you are essentially recreating the slot without touching the resource itself. Things like slot specific app settings , tags, locks etc. are maintained. 

Would this work:
## Creating a clean environment if an Azure Web App is misbehaving with slots

Lets say your Umbraco application is misbehaving and is unable to start in Azure App Service or the KUDU console itself is inaccessible. You can achieve a fresh environment by creating a new slot and swapping this slot with the misbehaving slot. Your deployment slot becomes brand new and any issues that were in the slot is moved to the copy and can be deleted or reviewed separately. You can then attempt to deploy again to see if the issue reoccurs.

nikolajlauridsen:
Just checked this again with the latest V10 version (10.1.0), and got a site up an running in a Linux AppService, with no issues, so nothing special is required for Linux. 

However I agree that the bit about logging is quite useful. In regards to the bit about slot swapping I again agree that I don't think this is something we should recommend as a solution, since it comes with its own set of issues with maindom lock, so I think we should remove that part and merge the part about logging 👍

sofietoft:
Hi @nikolajlauridsen !
Thanks a lot for your input on this one, and I think you suggestions make great sense.

@mattou07 - how does this sound to you?
Would you be up for applying the suggestions to the PR?

Mickel14 (approved):

sofietoft:
Hi all 👋 

I've taken the liberty to update this article according to the comments provided by @nikolajlauridsen 💪 
I'll merge in the PR.

Thanks again for adding to our docs @mattou07 !

PR #3611, opened 2021-10-30
Reduce wall of text to bullet points, Fundamentals/Data

mattou07:
I find it easier to absorb the edited paragraph if its listed as bullet points, rather than a wall of text.


busrasengul (approved):
Thank you @mattou07 
This looks good to me too
Let's merge this in 🙌

PR #3616, opened 2021-10-30
Updated editorService meta data

abjerner:
Spotted that the meta data for this article is wrong, so here's a PR to rectify that.

This article was originally written for Umbraco 8, and should therefore also apply to Umbraco 9.

Earlier versions of Umbraco 7 also had an `editorService`, but it was deprecated in favor of the `<umb-overlay />` directive in later versions of Umbraco 7. This article was never about any of those two.


marcemarc (approved):
thanks @abjerner !!

PR #3620, opened 2021-11-01
Update to Create Forms tutorial

OwainWilliams:
File name doesn't match the controller name in the docs.


jonat123:
@OwainWilliams Good catch!

I'll merge it in right away 🙂

PR #3633, opened 2021-11-10
Removed unnecessary usage of IUmbracoHelperAccessor

p-m-j:
Accessor should only be used to get access to scoped service in singleton contexts.


busrasengul:
Thank you @p-m-j 
We'll review this soon :)

p-m-j:
Hey @busrasengul, has there been any movement on this? Be nice to get this out asap before folks start to assume that using the accessors is the first point of call instead of last resort.

jonat123:
@p-m-j I have just taken a look at the PR and it looks good to me.

Let's merge it in right away 🙂

Issue #3628, opened 2021-11-07
IUserComposer - Is obsolete


simonantonio:
Visual Studio flags the IUserComposer interface as obsolete - this is used in the following documentation. 

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Implementation/Default-Routing/Controller-Selection/index.md***

---
Instead this seems to be the correct way to setup a Default Controller

In the **Startup.cs**
```
        /// <summary>
        /// Configures the services.
        /// </summary>
        /// <param name="services">The services.</param>
        /// <remarks>
        /// This method gets called by the runtime. Use this method to add services to the container.
        /// For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        /// </remarks>
        public void ConfigureServices(IServiceCollection services)
        {
#pragma warning disable IDE0022 // Use expression body for methods
            services.AddUmbraco(_env, _config)
                .AddBackOffice()
                .AddWebsite()
                .AddComposers()
                .Build();
#pragma warning restore IDE0022 // Use expression body for methods

            services.Configure<UmbracoRenderingDefaultsOptions>(c =>
            {
                c.DefaultControllerType = typeof(DefaultController);
            });
        }
```

p-m-j:
Configuring services in your startup class (or via convention in your container of choice) makes sense for those implementing sites.

Composers are still around for package developers (no access to startup code) that need access to the container or anyone else that wishes to use composers however IComposer should be used instead of IUserComposer.

IUserComposer which implements IComposer, used to have behavioural differences to other IComposers e.g, IUserComposer instances always ran after ICoreComposer instances. now all IComposers are treated equally.

PR #3638, opened 2021-11-12
Change the Default Controller

erikjanwestendorp:
Fixes [#3628](https://github.com/umbraco/UmbracoDocs/issues/3628)


busrasengul:
Thank you @erikjanwestendorp 
We'll review this soon!!

marcemarc (approved):
Thanks again @erikjanwestendorp yes, we are thinking any examples that use an IUserComposer that you might configure in your site should be updated to use the start.cs builder approach - only if it's specifically useful for shipping a package might it be useful to keep the composer away as an additional...

PR #3639, opened 2021-11-15
Use constant for WriterGroup

erikjanwestendorp:
Use constant for WriterGroup 😄


busrasengul:
Thank you @erikjanwestendorp 
We'll get this reviewed soon!

PR #3640, opened 2021-11-15
Content Version Cleanup

p-m-j:
Documentation for new historic content version cleanup feature.


Reference/V9-Config/ContentSettings/index.md
```diff
@@ -195,3 +200,39 @@ If you need to create a custom media document type to handle images called somet
...
+
+```json
+"ContentVersionCleanupPolicy": {
+  "EnableCleanup": false,
+  "KeepAllVersionsNewerThanDays": 7,
+  "KeepLatestVersionPerDayForDays": 90
+}
+```
+
+If you don't wish to store any historic content versions at all simply set both of the "keep" settings values to 0.
```
bergmania, line 220: This is not completely correct. This will just delete them hourly.

bergmania (approved):
Overall this looks very good to me..  💪 
I just sanity checked so someone else please check for gramma 😄

Reference/V9-Config/ContentSettings/index.md
p-m-j: :cowboy_hat_face:

Fundamentals/Data/Content-Version-Cleanup/index.md
```diff
@@ -0,0 +1,45 @@
...
+those upgrading from 9.0.0.
+
+This feature certainly isn't for everyone, some sites will have regulatory requirements to keep historic data for audit.
+
+For those sites with stricter requirements (or those who don't want the feature) it is possible to opt out both globally 
+(see [v9-Config > ContentSettings](/documentation/Reference/v9-Config/ContentSettings/index.md#contentversioncleanuppolicy)) and per document type (keep reading).
+
+Additionally it is possible to keep the feature enabled but mark specific versions to keep forever (perhaps last years awesome marketing campaign).
+
+It's worth noting that whilst we delete rows, we do not shrink database files or rebuild indexes so for upgraded sites with a lot
```
jonat123, line 25:
```suggestion
Umbraco 9.1.0 introduced a feature to clean up historic content versions (inspired by [Our.Umbraco.UnVersion](https://our.umbraco.com/packages/website-utilities/unversion/) h5yr!).

The feature is enabled by default via configuration for new installs starting from 9.1.0 but will require to opt in for 
those upgrading from 9.0.0.

This feature certainly isn't for everyone, some sites will have regulatory requirements to keep historic data for audit.

For those sites with stricter requirements (or those who don't want the feature) it is possible to opt out both globally 
(see [v9-Config > ContentSettings](/documentation/Reference/v9-Config/ContentSettings/index.md#contentversioncleanuppolicy)) and per document type (keep reading).

Additionally, it is possible to keep the feature enabled but mark specific versions to keep forever (perhaps last years awesome marketing campaign).

It's worth noting that whilst we delete rows, we do not shrink database files or rebuild indexes so for upgraded sites with a lot
```

Fundamentals/Data/Content-Version-Cleanup/index.md
```diff
@@ -0,0 +1,45 @@
...
+of history you may wish to perform these tasks (if they are not part of your regular database maintenance plan already).
+
+## Overriding global settings
+
+It is possible to override the global settings per document type in the backoffice to prevent unwanted clean up, this can be managed in the "permissions"
+content app for each document type.
```
jonat123, line 31:
```suggestion
of history you may wish to perform these tasks (if they are not part of your regular database maintenance plan already).

## Overriding global settings

It is possible to override the global settings per document type in the backoffice to prevent unwanted cleanup, this can be managed in the "permissions"
content app for each document type.
```

Reference/V9-Config/ContentSettings/index-v9.0.0.md
```diff
@@ -0,0 +1,197 @@
...
+state: complete
+verified-against: beta-3
+update-links: true
+---
+
+# Content Settings
+
+Content settings contains various settings regarding content in the CMS, such as allowed upload files, image settings, and much more. All the values in the content settings has default values, so all configuration is optional.
+
+To get an overview of the keys and values in the global section, the following snippet will contain all available options, along with their default values:
```
jonat123, line 14:
```suggestion
Content settings contain various settings regarding the content in the CMS, such as allowed upload files, image settings, and much more. All the values in the content settings have default values, so all configuration is optional.

To get an overview of the keys and values in the global section, the following snippet will contain all available options, along with their default values:
```

Reference/V9-Config/ContentSettings/index-v9.0.0.md
```diff
@@ -0,0 +1,197 @@
...
+In case of a 404 error (page not found) Umbraco can return a default page instead. This is set here. Notice you can also set a different error page, based on the current culture so a 404 page can be returned in the correct language.
+
+```json
+"Error404Collection": [{
+  "ContentId": 1,
+  "Culture": "en-US"
+}]
+```
+
+The above example shows what you need to do if you only have a single site that needs to show a custom 404 page. You specify which node that should be shown when a request for a non-existing page is being made. You can specify the node in three ways:
```
jonat123, line 67:
```suggestion
## Root-level settings

In the root level section, that is those without a separate subsection like Imaging, you can configure:

### Resolve URLs from a text string

This setting is used when you're running Umbraco in virtual directories. Setting this to true can increase render time for pages with a large number of links, however, this is required if Umbraco is running in a virtual directory.

### Error 404 collection

In case of a 404 error (page not found) Umbraco can return a default page instead. This is set here. Notice you can also set a different error page, based on the current culture so a 404 page can be returned in the correct language.

```json
"Error404Collection": [{
  "ContentId": 1,
  "Culture": "en-US"
}]
```

The above example shows what you need to do if you only have a single site that needs to show a custom 404 page. You specify which node that should be shown when a request for a non-existing page is being made. You can specify the node in three ways:
```

File: Reference/V9-Config/ContentSettings/index.md [hunk 6]
```diff
@@ -195,3 +200,41 @@ If you need to create a custom media document type to handle images called somet
...
+"ContentVersionCleanupPolicy": {
+  "EnableCleanup": false,
+  "KeepAllVersionsNewerThanDays": 7,
+  "KeepLatestVersionPerDayForDays": 90
+}
+```
+
+If you don't wish to retain any content versions except for the current draft and current published you can set both of the
+"keep" settings values to 0, after which the next time the scheduled job runs (hourly) all non current versions (except
+ those marked "prevent cleanup") will be removed.
```
jonat123, line 222:
```suggestion
If you don't wish to retain any content versions except for the current draft and currently published you can set both of the
"keep" settings values to 0, after which the next time the scheduled job runs (hourly) all non-current versions (except
 those marked "prevent cleanup") will be removed.
```

Reference/V9-Config/ContentSettings/index.md
```diff
@@ -195,3 +200,41 @@ If you need to create a custom media document type to handle images called somet
...
+
+### EnableCleanup
+
+When true a scheduled job will delete historic content versions that are not kept according to the policy every hour.
+
+When false, the scheduled job will never delete any content versions regardless of overridden settings for a document type.
+
+This defaults to false when not set in configuration which will be the case for those upgrading from v9.0.0, however the dotnet new template will supply an appsettings.json with the value set to true for all sites starting from v9.1.0.
```
jonat123, line 230:
```suggestion
### EnableCleanup

When true a scheduled job will delete historic content versions that are not kept according to the policy every hour.

When false, the scheduled job will never delete any content versions regardless of overridden settings for a document type.

This defaults to false when not set in the configuration which will be the case for those upgrading from v9.0.0, however, the dotnet new template will supply an appsettings.json with the value set to true for all sites starting from v9.1.0.
```

Reference/V9-Config/ContentSettings/index.md
```diff
@@ -195,3 +200,41 @@ If you need to create a custom media document type to handle images called somet
...
+
+### KeepAllVersionsNewerThanDays
+
+All versions that fall in this period will be kept.
+
+### KeepLatestVersionPerDayForDays
+
+For content versions that fall in this period the most recent version for each day is kept but all previous versions for that day are removed unless marked as preventCleanup.
+
+This variable is independent from KeepAllVersionsNewerThanDays, if both were set to the same value KeepLatestVersionPerDayForDays would never apply as KeepAllVersionsNewerThanDays is considered first.
```
jonat123, line 240:
```suggestion
### KeepAllVersionsNewerThanDays

All versions that fall in this period will be kept.

### KeepLatestVersionPerDayForDays

For content versions that fall in this period, the most recent version for each day is kept but all previous versions for that day are removed unless marked as preventCleanup.

This variable is independent of KeepAllVersionsNewerThanDays, if both were set to the same value KeepLatestVersionPerDayForDays would never apply as KeepAllVersionsNewerThanDays is considered first.
```

jonat123:
@p-m-j I just looked through the grammar and made some suggestions :D 

let me know what you think and once you are done lets merge it in 🙂

Reference/V9-Config/ContentSettings/index-v9.0.0.md
p-m-j: This is unchanged by my PR :p

Reference/V9-Config/ContentSettings/index-v9.0.0.md
p-m-j: Also unchanged by this PR

p-m-j:
No complaints about the other changes to the common sections, should presumably update both the 9.1 version and the 9.0 version.
However might make more sense to do it in another PR for clarity

jonat123:
@p-m-j Good points :-) 

I'm happy with the PR, let's get it merged in right away 🙂

PR #3641, opened 2021-11-16
Use constants for AdminGroupAlias

erikjanwestendorp:
I like to avoid magic strings 😄


busrasengul:
Hey @erikjanwestendorp 
Thank you for your suggestion but we'd like to keep the example as it is. Because if the user has their own custom groups it makes more sense to do it with string.
It shows how it can easily be changed :)

Maybe instead of using `admin`, we can change the example to use a custom group name e.g `marketing` or you can come up with something better 😊

PR #3647, opened 2021-11-18
Changed what mode generate button should be used

RBerntsen:
In the example it says to use the generate models in the SourceCodeAuto mode, which generates the models automatically. It should be SourceCodeManual.


marcemarc (approved):
@RBerntsen Thanks for updating!!

PR #3651, opened 2021-11-19
use Startup.cs to SetContentLastChanceFinder

erikjanwestendorp:
In line with this comment https://github.com/umbraco/UmbracoDocs/pull/3638#pullrequestreview-811068459 use the startup class instead of a I(User)Composer.


busrasengul (approved):
Thank you @erikjanwestendorp 
Let's merge this in!

PR #3652, opened 2021-11-22
Update server registration service v9

erikjanwestendorp:
Update server registration service for v9


busrasengul (approved):
Thank you @erikjanwestendorp 
Let's merge it in!

PR #3653, opened 2021-11-22
Fix small typo

erikjanwestendorp:
Media instead of macro


busrasengul:
Thank you @erikjanwestendorp 
Let's merge it!

PR #3654, opened 2021-11-22
Similar change as in the V8 doc - Added HTTP check

Ambertvu:
Added HTTP check + cleanup


busrasengul:
This looks good to me @Ambertvu 
Thank you 🙌

PR #3664, opened 2021-11-25
Use IComposer in outbound-pipeline article

erikjanwestendorp:
IUserComposer --> IComposer


busrasengul:
Thank you @erikjanwestendorp 
Let's merge this in!

PR #3666, opened 2021-11-26
Umbraco v9 configuration should be at least mentioned on the /config

dampee:
certainly because the folder is much lower in the menu. We should have created (or moved) files within the config folder. The current solution is all but handy and practical


busrasengul:
It sounds sensible to do this @dampee 
Merging right in!
Thank you 🙏

PR #3670, opened 2021-11-29
Note formatting

aguyfromdenmark:
Changes sections formatted as code inside a note to be formatted as bold. The rendering of the note looks wierd, and I suspect it's the code formatting inside the note that is the issue. I also added to the note, that you can't implement IVirtualPageController when your request is a client-side request.

What I am trying to fix:
image

### DIFF
--- a/Reference/Routing/custom-routes.md
+++ b/Reference/Routing/custom-routes.md
@@ -221,8 +221,9 @@ public IPublishedContent FindContent(ActionExecutingContextactionExecutingContex
     return productRoot;
 }
 ```
+
 :::note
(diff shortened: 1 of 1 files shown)


busrasengul:
Thank you @aguyfromdenmark 
We'll review this soon!

marcemarc (approved):

marcemarc:
@aguyfromdenmark thanks for raising this issue - there is a related issue now open on the core repo:
https://github.com/umbraco/Umbraco-CMS/issues/12083
and hopefully a fix is imminent.
This will make the UmbracoContextFactory approach work again as per the original note and how it worked in V8

So I've tweaked your original text to indicate this, hope that's ok!

For now it's broken, but at least people seeing the note will be aware of that, and then the issue will be updated when it's hopefully working again, and then we can update again!

Issue #3674, opened 2021-12-02
Example appears incomplete


warrenatatlas:
Umbraco.Core.Logging was dropped in 9 wasn't it?
"record" variable is not defined also

Thank you

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md***

PR #3675, opened 2021-12-03
Update workflow type

erikjanwestendorp:
Fixes: https://github.com/umbraco/UmbracoDocs/issues/3674 

- [x] Use `Microsoft.Extensions.Logging` instead of `Umbraco.Core.Logging;`
- [x] change record to context.Record


busrasengul:
Thank you @erikjanwestendorp 
We'll review this soon!

AndyButland (approved):
Looks good - seems we missed some updates to variables when translating this sample from V8.  The changes here are all correct.

PR #3676, opened 2021-12-06
Added meta.title and meta.description for UI Library

Adolfi:
Meta title and meta description was missing for this page.
This is especially a problem for the title in the browser, which now shows: "UI%20Library":
Screen Shot 2021-12-06 at 09 02 42


busrasengul:
Thank you @Adolfi
Good catch 🙌 
Let's merge this in!

PR #3677, opened 2021-12-06
Update azure-web-apps-v8.md

darrenferguson:
Some additional detail based on recent experience.

### DIFF
--- a/Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
+++ b/Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
@@ -32,7 +32,7 @@ You need to add these `appSettings`:
     value="Examine.LuceneEngine.Directories.SyncTempEnvDirectoryFactory, Examine" />
 ```
 
-You should also add an appsetting `WEBSITE_DISABLE_OVERLAPPED_RECYCLING` with a value of 1 to your Azure WebApp (not this needs to be added to the App configuration via the Azure portal and not via Web.config. This setting prevents Umbraco from starting on an instance before a shutdown is complete - which is important as Umbraco in Azure needs to acquire an exclusive `MainDom` lock in order to start.
+You should also add an appsetting `WEBSITE_DISABLE_OVERLAPPED_RECYCLING` with a value of 1 to your Azure WebApp. Keep in mind that this needs to be added to the App configuration via the Azure portal and not via Web.config. This setting prevents Umbraco from starting on an instance before a shutdown is complete - which is important as Umbraco in Azure needs to acquire an exclusive `MainDom` lock in order to start.
 
 You may also want add a setting `WEBSITE_ADD_SITENAME_BINDINGS_IN_APPHOST_CONFIG` with a value of 1 to your Azure WebApp. Although not required, this may help minimise app restarts during deployments. More details can be found in this post: https://ruslany.net/2019/06/azure-app-service-deployment-slots-tips-and-tricks/
 
--- a/Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
+++ b/Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
@@ -32,11 +32,15 @@ You need to add these `appSettings`:
     value="Examine.LuceneEngine.Directories.SyncTempEnvDirectoryFactory, Examine" />
 ```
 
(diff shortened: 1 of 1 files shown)


Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
```diff
@@ -32,11 +32,15 @@ You need to add these `appSettings`:
     value="Examine.LuceneEngine.Directories.SyncTempEnvDirectoryFactory, Examine" />
 ```
 
+You should also add an appsetting `WEBSITE_DISABLE_OVERLAPPED_RECYCLING` with a value of 1 to your Azure WebApp (not this needs to be added to the App configuration via the Azure portal and not via Web.config. This setting prevents Umbraco from starting on an instance before a shutdown is complete - which is important as Umbraco in Azure needs to acquire an exclusive `MainDom` lock in order to start.
```
nul800sebastiaan, line 35: This recommendation was removed for 8.6.3+ - not sure why. https://github.com/umbraco/UmbracoDocs/blob/main/Fundamentals/Setup/Server-Setup/Load-Balancing/azure-web-apps-v8.md#v860---v863

Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
```diff
@@ -32,11 +32,15 @@ You need to add these `appSettings`:
...
+
+You may also want add a setting `WEBSITE_ADD_SITENAME_BINDINGS_IN_APPHOST_CONFIG` with a value of 1 to your Azure WebApp. Although not required, this may help minimise app restarts during deployments. More details can be found in this post: https://ruslany.net/2019/06/azure-app-service-deployment-slots-tips-and-tricks/
```
nul800sebastiaan, line 37: This sounds interesting but I can't find anything about it in the context of Umbraco before. We can't recommend things without really knowing what they do or if they affect Umbraco in any way, I know it only says there's problems with WCF, but that blog also doesn't really know for sure if there may be other problems.

I guess there may actually be usages of WCF with Umbraco 8, so at least that should be warned about. Maybe best to move this to a "worth a try" section in the docs?

nul800sebastiaan (changes requested):

Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
darrenferguson: @nul800sebastiaan I am pretty certain it needs to be there still. I have observed locking issues when trying to acquire the MainDom without this.

Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
darrenferguson: @nul800sebastiaan having that setting in place is the difference between your production slot being recycled or not when swapped in. Obviously, we really don't want our live slot to be restarted. You are right, in the sense that it is nothing to do with Umbraco - but a really useful Azure tip generally. It does need to be worded better here though, you are right.

Fundamentals/Setup/Server-Setup/azure-web-apps-v8.md
```diff
@@ -32,11 +32,16 @@ You need to add these `appSettings`:
     value="Examine.LuceneEngine.Directories.SyncTempEnvDirectoryFactory, Examine" />
 ```
 
+You should also add an appsetting `WEBSITE_DISABLE_OVERLAPPED_RECYCLING` with a value of 1 to your Azure WebApp (not this needs to be added to the App configuration via the Azure portal and not via Web.config. This setting prevents Umbraco from starting on an instance before a shutdown is complete - which is important as Umbraco in Azure needs to acquire an exclusive `MainDom` lock in order to start.
```
sofietoft, line 35:
```suggestion
You should also add an appsetting `WEBSITE_DISABLE_OVERLAPPED_RECYCLING` with a value of 1 to your Azure WebApp. Keep in mind that this needs to be added to the App configuration via the Azure portal and not via Web.config. This setting prevents Umbraco from starting on an instance before a shutdown is complete - which is important as Umbraco in Azure needs to acquire an exclusive `MainDom` lock in order to start.
```

sofietoft:
Hi @darrenferguson !

Apologies for the delay on this one!
Since this issue you're referring to here has been fixed with the release Umbraco 9, we've decided to not merge the PR.

In case this is something you're still experiencing regularly, it could make sense to add the information to a "Troubleshooting" part in the article instead. That way it will not be part of the general docs, but still something you'd be able to find when running into the issue 💪 

Thanks again for helping us keep the docs up to date!

PR #3683, opened 2021-12-08
Update to Composing page

OwainWilliams:
Just missing a ';' from the end of line 46.


busrasengul:
Good catch @OwainWilliams 🙌

PR #3694, opened 2021-12-20
Add warning to chache article

erikjanwestendorp:
Add warning to v7 and v8 article as well to be more consistent


busrasengul:
Thank you @erikjanwestendorp 
Merging this in!

PR #3695, opened 2021-12-21
Remove: "(Currently only available for Umbraco 8)"

erikjanwestendorp:
Built in Property Editors articles are available for v9 right?


jonat123:
@erikjanwestendorp they totally are available for V9 🙃

Good spotted!

Merging it in right away!

PR #3696, opened 2021-12-23
Update ViewComponents-index.md

hnky:
I got an error following this tutorial, needed to use default.cshtml and not index.cshtml

### DIFF
--- a/Reference/Templating/Mvc/ViewComponents-index.md
+++ b/Reference/Templating/Mvc/ViewComponents-index.md
@@ -70,7 +70,7 @@ namespace ViewComponentSample.ViewComponents
 
 ### Create a View for ViewComponent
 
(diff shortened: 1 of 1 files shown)


jonat123:
Hello @hnky thanks for creating this PR 🙂

Hmm that does sound odd, not sure why the name that you give the file should cause issues 🤔

What was the error you got when you named the file index?

//Jonathan

hnky:
That it could not find a file and gave a list files that it was looking for.  Index was not on that list, default was.
It should be easy to reproduce.

busrasengul:
In Microsoft docs, it states as 
_If the view component is controller-specific, you can add it to the controller-specific folder (Views/ToDo/Components/PriorityList/Default.cshtml)._
If no controller-specific then
_Views/ToDo/index.cshtml_ 

Our docs we have a controller-specific example so @hnky is right! And I can also replicate this 👍 So merging this!

PR #3697, opened 2021-12-24
Rename solution to project and add note how to create a .sln file

erikjanwestendorp:
No idea if it's really a desirable addition, but I think a lot of people want to add a .sln file. I've added a note explaining how to do that.


jonat123:
@erikjanwestendorp thanks for the PR 😄

I don't see anything wrong with adding a note about how to add a solution file, we are also mentioning how to do it for Umbraco Cloud projects 🙂 

I'll merge it in right away 😁

PR #3699, opened 2021-12-30
v9: Maximum Upload Size documentation

Zeegaan:
So this comes from a number of issues raised on the CMS, where you cannot upload a file larger than 28.6MB, so I thought some documentation on how to do so would be appreciated!

This contains a how-to of the 2 most used services with .net core (kestrel & IIS) but also links to documentation on how to change this with apache and nginx 👍

### DIFF
--- a/Reference/V9-Config/MaximumUploadSizeSettings/index.md
+++ b/Reference/V9-Config/MaximumUploadSizeSettings/index.md
@@ -4,6 +4,8 @@ meta.Title: "Umbraco Maximum upload size settingsg"
 meta.Description: "Information on how to change the default cap of upload size"
 ---
 
+Umbraco does not touch the default maximum allowed content size of the different services, but you can configure this yourself.
+
 # Using IIS
 
 To configure the default 28.6mb upload limit using IIS, we have to create a web.config file in the root of the project. It should contain this:
@@ -41,3 +43,13 @@ An example of a configuration could look something like:
 ```
 
 `MaxRequestLength` is specified in kilobytes, so this configuration would limit requests, and therefore uploaded files, to 2 megabytes, and a maximum query string length of 90 characters.
+
+# Using nginx
+
+Here's the documentation link to nginx: 
+https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size
+
+# Using apache
+
+Here's the documentation link to apache:
+https://httpd.apache.org/docs/2.2/mod/core.html#limitrequestbody
--- a/Reference/V9-Config/MaximumUploadSizeSettings/index.md
+++ b/Reference/V9-Config/MaximumUploadSizeSettings/index.md
@@ -8,7 +8,7 @@ Umbraco does not touch the default maximum allowed content size of the different
 
 # Using IIS
 
-To configure the default 28.6mb upload limit using IIS, we have to create a web.config file in the root of the project. It should contain this:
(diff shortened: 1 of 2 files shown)


Reference/V9-Config/MaximumUploadSizeSettings/index.md
```diff
@@ -0,0 +1,55 @@
...
+versionFrom: 9.0.0
+meta.Title: "Umbraco Maximum upload size settingsg"
+meta.Description: "Information on how to change the default cap of upload size"
+---
+
+Umbraco does not touch the default maximum allowed content size of the different services, but you can configure this yourself.
+
+# Using IIS
+
+To configure the default 28.6mb upload limit using IIS, we have to create a web.config file in the root of the project. It should contain this:
```
jonat123, line 11:
```suggestion
Umbraco does not touch the default maximum allowed content size of the different services, but you can configure this yourself.

# Using IIS

To configure the default 28.6MB upload limit using IIS, we have to create a web.config file at the root of the project. It should contain this:
```

Reference/V9-Config/MaximumUploadSizeSettings/index.md
```diff
@@ -0,0 +1,55 @@
...
+</configuration>
+```
+
+`maxAllowedContentLength` is specified in bytes, so this configuration would limit requests, and therefore uploaded files, to 2 megabytes
+
+# Using Kestrel
+
+Runtime settings allows you to configure the `MaxRequestLength` and `MaxQueryStringLength` for kestrel. If you want to upload files larger than 28.6mb, then you have to configure these seetings. If nothing is configured reqests and query string can only be the default size and smaller.
+
+An example of a configuration could look something like:
```
jonat123, line 32:
```suggestion

`maxAllowedContentLength` is specified in bytes, so this configuration would limit requests, and therefore uploaded files, to 2 megabytes

# Using Kestrel

Runtime settings allow you to configure the `MaxRequestLength` and `MaxQueryStringLength` for kestrel. If you want to upload files larger than 28.6MB, then you have to configure these settings. If nothing is configured requests and query strings can only be the default size and smaller.

An example of a configuration could look something like this: 
```

Reference/V9-Config/MaximumUploadSizeSettings/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Using nginx
+
+Here's the documentation link to nginx: 
+https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size
+
+# Using apache
+
+Here's the documentation link to apache:
+https://httpd.apache.org/docs/2.2/mod/core.html#limitrequestbody
```
jonat123, line 55:
```suggestion
`MaxRequestLength` is specified in kilobytes, so this configuration would limit requests, and therefore uploaded files, to 2 megabytes, and a maximum query string length of 90 characters.

# Using Nginx

Here's the documentation link to Nginx: 
https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

# Using apache

Here's the documentation link to apache:
https://httpd.apache.org/docs/2.2/mod/core.html#limitrequestbody
```

jonat123:
@Zeegaan Looks good :) 

I have gone over the grammar and made some suggestions 😄

Please have a look at it 🙂

Zeegaan:
@jonat123 the grammar should now be fixed, thank you for the suggestions 👏

jonat123:
@Zeegaan awesome! 

I'll merge it in right away :D

PR #3701, opened 2021-12-30
Enhanced the description of an SMTP From address

TwoMoreThings:
To show an example of a Friendly Name and reference the relevant RFC


jonat123:
@TwoMoreThings thanks for the PR :D 

Looks like some good additions!

I'll merge it in right away 🙂

PR #3703, opened 2021-12-31
Update index.md

craigs100:
To explain what the defaults are and exactly how to use the CacheBuster in "Version" mode.  If anyone can add similar instructions for "Appdomain" and "Timestamp" that would be great too.


busrasengul:
Thank you @craigs100 
We'll review this soon!

sophie-neale (commented):
Thanks Craig! I've had a quick scan before the Docs team gets together later this week and suggested a tiny grammar tweak ☺️

PR #3707, opened 2022-01-07
Update code block language 

erikjanwestendorp:
The code samples on these pages look slightly different, probably because of the different notation?


busrasengul:
You are right @erikjanwestendorp 
I'm merging this in 🙌

PR #3716, opened 2022-01-12
Fix code view

Zeegaan:
If you take a look at https://our.umbraco.com/documentation/Reference/V9-Config/MaximumUploadSizeSettings/ 
The code view under "Using IIS" is broken, this should remedy that


marcemarc (approved):
I quite liked the black box! but agree @Zeegaan it's not very helpful.

Think you've spotted why, we'll merge and trigger the rebuild of our umbraco and hopefully the example should be back in place!!

thanks!

PR #3717, opened 2022-01-12
Add information about package telemetry properties

nikolajlauridsen:
Adds a new small section in the package manifest document, detailing the new properties added as part of the new telemetry added in 9.2.


marcemarc:
thanks @nikolajlauridsen now I need to add this to all my packages! another chart to try and get to the top of!

PR #3727, opened 2022-01-18
Fix syntax for Heartcore filtering examples

rasmusjp:


jonat123:
@rasmusjp thanks for the corrections!

I'll merge it in right away :-)

PR #3728, opened 2022-01-18
Rename alias of  WelcomeDashboard to welcomeDashboard

erikjanwestendorp:
Wouldn't it be nicer to always write an alias with a lowercase letter? This is also common in other places, such as doctypes, etc. In addition, it prevents inconsistent use within language files.


busrasengul:
It's a great idea @erikjanwestendorp 
I'm merging this in!

PR #3731, opened 2022-01-19
Update hosting requirements

nul800sebastiaan:
* Docker is not a requirement to run Umbraco
* Added the minimal server requirement (IIS 8 requires at least Windows server 2012 R2)


nul800sebastiaan:
Also added recommended hosting in addition of just the bare minimum so that we give people an idea of what the best hosting options are :)

busrasengul:
Thank you for clarifying the requirements @nul800sebastiaan 
It's clearer now!

PR #3742, opened 2022-01-25
Update content app article

erikjanwestendorp:
- [x] Update code example. Show content app for content only.
- [x] Update tip


busrasengul:
Thank you @erikjanwestendorp 
We'll get this reviewed!

PR #3746, opened 2022-01-26
Add note to creating media article

erikjanwestendorp:


busrasengul:
Thank you @erikjanwestendorp !
I'm not entirely sure about adding personal repo reference and/or uSync. We'll discuss this and get back to you :)

busrasengul (changes requested):
Hey @erikjanwestendorp 
We've come to a conclusion where we don't think it's a good idea including uSync (as it may not be everyone's case), and we thought it would be better to just link to a blog post maybe. Happy keep the rest :)

erikjanwestendorp:
@busrasengul I can imagine 😄. I just removed the uSync part.

Fundamentals/Data/Creating-Media/index.md
```diff
@@ -19,6 +19,10 @@ Media in Umbraco is handled the same way as content. Instead of defining Documen
 
 This means you don't have to define your own Media Types to start using the Media section. You have already got the tools for organizing and uploading the media.
 
+:::note
+If you have upgraded from an older version than 8.14 then the new media types are not added automatically. You can add those types manually yourselves by follwing the steps below ['Creating a new Media Type'](#creating-a-media-type). On [this page](default-media-types.md) you will find an overview of all new media types.
```
busrasengul, line 23:
```suggestion
If you have upgraded from an older version than 8.14 then the new media types are not added automatically. You can add those types manually yourselves by following the steps below ['Creating a new Media Type'](#creating-a-media-type). On [this page](default-media-types.md) you will find an overview of all new media types. 
```

PR #3750, opened 2022-01-27
Update to use Url extension method

bjarnef:
The `Url` property has been obsolete for a while in v8 and in v9 I believe it has been removed and instead `Url()` extension method should be used.


busrasengul:
Thank you @bjarnef 
Let's merge this in!

PR #3752, opened 2022-01-28
Update how to find member by Udi and Guid

bjarnef:
I noticed in Umbraco v9 we don't have `MemberShipHelper` like in Umbraco v7 and v8 and it wasn't obvious how to get a member by `Udi` or `Guid`: https://our.umbraco.com/forum/using-umbraco-and-getting-started/107936-how-to-get-member-by-udi-in-umbraco-9

It seems we need to inject `IIdKeyMap` service to resolve the id from `Udi` or `Guid`.


busrasengul:
Thank you @bjarnef 
We'll get this reviewed soon!

marcemarc (approved):

marcemarc:
@bjarnef super helpful... do you think we should add a FindMemberByUdi to the MemberManager in the core?

bjarnef:
@marcemarc that would probably be useful, but I think its a HQ decision.
Changes to the interface `IMemberManager` would be a breaking change, but `FindMemberByUdi` or just `FindMember(int id)` and `FindMember(Guid id)` ... or maybe just `GetById(int id)` / `GetById(Guid id)`.

The `Udi` could easily be resolved to a `Guid`.

marcemarc:
@bjarnef perhaps an ExtensionMethod? MemberManagerExtensions, only just poked at the code to understand your PR, seems like it's intertwined with Users, and that's why there is no getby Udi, as I don't think Users have Udis!

PR #3759, opened 2022-01-31
Remove blob migration warning

rasmusjp:
Noticed that we were still showing a blob migration warning on the Setup/Media pages, even thoug the migrations have been done for quite some time now 🙈


jonat123:
Good catch @rasmusjp! 

I'll merge this in right away.

PR #3762, opened 2022-02-01
Updated v8 example of UmbracoApiController test to v9

Adolfi:
I updated the v8 example of an UmbracoApiController unit test to a v9 example.


busrasengul:
Thank you @Adolfi 
Let's merge it in!

Adolfi:
❤️

PR #3765, opened 2022-02-01
Moved link from codeblock to normal text

Ambertvu:
Moved the link out of the codeblock as it gets wrongly interpreted (as seen in screenshot)
image


busrasengul:
Thank you @Ambertvu 
Let's merge this in right away!

PR #3766, opened 2022-02-02
Various small grammar, punctuation and readability changes.

Tariovic:
Small grammar and punctuation fixes. Redundant words removed, a few words and phrases replaced with better while retaining the original meaning, some standardization.


busrasengul:
Thank you @Tariovic 
Let's merge this in right away :)

PR #3767, opened 2022-02-02
The `Lang` directory is case-sensitive start with capital `L`

nul800sebastiaan:


nul800sebastiaan:
Unfortunately, other `lang` directories are all found by lower case, only the one in the `App_Plugins` directory needs to have an upper case `L`. I've tried to find all the references to it and added a note since this is a special case that people might easily get wrong.

busrasengul:
Thank you @nul800sebastiaan 
Merging this in!

PR #3770, opened 2022-02-03
Update inbound-pipeline-v8.md

Ambertvu:
Changed link to default page as there is no page for published-content-request-preparation for v8.


busrasengul:
This has been duplicated on #3771

PR #3784, opened 2022-02-10
Advise against Explicit Scheduling Publisher role on Azure App Service

nzdev:
Add note advising against explicit scheduling publisher on azure appservice with deployment slots.

Reason:
Explicitly registering the SchedulingPublisher Server is advised against when using the deployment slots feature on Azure Web Apps.
This is because both the staging and production slot would be scheduling publisher servers, this can result in data corruption when using scheduled publishing/unpublishing is used as both slots will try execute the scheduled publishing task.

### DIFF
--- a/Fundamentals/Setup/Server-Setup/Load-Balancing/flexible-advanced-v8.md
+++ b/Fundamentals/Setup/Server-Setup/Load-Balancing/flexible-advanced-v8.md
@@ -11,6 +11,10 @@ _This describes some more advanced techniques that you could achieve with flexib
 It is recommended to configure an explicit master scheduling server since this reduces the amount
 complexity that the [master election](flexible.md#scheduling-and-master-election) process performs.
 
+:::note
+Explicitly registering the Master Scheduling Server is advised against when using the deployment slots feature on Azure Web Apps.
+This is because both the staging and production slot would be master scheduling servers, this can result in data corruption when using scheduled publishing/unpublishing is used.
+
 The first thing to do is create a couple classes for your front-end servers and master server to use:
 
 ```csharp
--- a/Fundamentals/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
+++ b/Fundamentals/Setup/Server-Setup/Load-Balancing/flexible-advanced.md
@@ -22,6 +22,10 @@ These new terms replace 'Master and Replica', in Umbraco versions 7 and 8.
 It is recommended to configure an explicit SchedulingPublisher server since this reduces the amount
 of complexity that the election process performs.
 
+:::note
+Explicitly registering the SchedulingPublisher Server is advised against when using the deployment slots feature on Azure Web Apps.
+This is because both the staging and production slot would be scheduling publisher servers, this can result in data corruption when using scheduled publishing/unpublishing is used.
+
 The first thing to do is create a couple of small classes that implement `IServerRoleAccessor` one for each of the different server roles:
 
 ```csharp


Jeavon:
@nzdev Should use a custom ServerRoleAccessor that gets its role from config/env-var then this config should be made slot sticky so then there is only ever one SchedulingPublisher?

nzdev:
The default ServerRoleAccessor ensures only one server in the scheduling publisher role. So my recommendation is to only force the public instances to be in the replica role. Though, of course if you are changing the logic in a notification handler, and this changes between code versions, it would need to be possible to tell the prod slot to become a replica and then have the staging slot be the in the publisher role before the swap.

Jeavon:
> Hi @Jeavon, I can't remember exactly where in the Azure App Service docs I've read this, however I'm under the impression there is no guarantee on some of the Environment variables actually updating during / after the swap operation till the slot is restarted, so the app service doesn't know what slot it is in directly after a swap. If this assumption holds, both prod and staging slots need to be restarted to pick up new per slot configuration, which may result in downtime.
> 
> I'd also recommend having [umbraco/Umbraco-CMS#11972 (comment)](https://github.com/umbraco/Umbraco-CMS/pull/11972#issuecomment-1034820799) in the instructions, if approved.

When you swap slots the apps will restart and will have the correct config (at least I have never seen it fail), if you need to avoid downtime you need to use App Initiazalition to pre warm up

Jeavon:
> The default ServerRoleAccessor ensures only one server in the scheduling publisher role. So my recommendation is to only force the public instances to be in the replica role. Though, of course if you are changing the logic in a notification handler, and this changes between code versions, it would need to be possible to tell the prod slot to become a replica and then have the staging slot be the in the publisher role before the swap and retain this role after its swapped.

The default might elect a front end instance, you don't want that

nzdev:
#The default might elect a front end instance, you don't want that

That's why I suggest only forcing the front end instances to be in the replica role, and then let the edit/ backoffice slots battle it out and elect a winner using the default behaviour.

Jeavon:
This is what I've been doing, possibly we could add this to Load Balancing Azure docs
```
   public class CustomServerRoleAccessor : IServerRoleAccessor
    {
        private readonly IConfiguration _configuration;

        public CustomServerRoleAccessor(IConfiguration configuration)
        {
            _configuration = configuration;
        }
        public ServerRole CurrentServerRole
        {
            get
            {
                var serverRoleKey = Constants.Configuration.ConfigGlobalPrefix + "ServerRole";
                var serverRoleConfig = _configuration[serverRoleKey];

                var serverRole = ServerRole.Subscriber;

                if (serverRoleConfig != null)
                {
                    if (Enum<ServerRole>.TryParse(serverRoleConfig, out var configServerRole))
                    {
                        serverRole = configServerRole;
                    }
                }

                return serverRole;
            }
        }
    }
```
```
Add default
      "Global": {
        "Id": "634c07fd-99ef-4156-beb8-a9d1479bc895",
        "ServerRole" : "Subscriber", 
        "MainDomLock": "SqlMainDomLock"
      }
```
For Umbraco app (prod slot) add (Deployment slot setting TICK)
UMBRACO__CMS__GLOBAL__SERVERROLE
SchedulingPublisher

For Umbraco app (stage slot) add (Deployment slot setting TICK)
UMBRACO__CMS__GLOBAL__SERVERROLE
Subscriber

Jeavon:
> It may be that I use local cache https://docs.microsoft.com/en-us/azure/app-service/overview-local-cache

It's works really well but you have to restart after every deployment for it to pick up changes

nzdev:
We were doing the same sort of approach as your code example, with the local cache option on. I'm not sure whether we changed ServerRole per slot, which may have been our error. We removed that behaviour and have not had a issue since with scheduled publishing. Looking at the Azure Docs both slots, under the approach you suggest, at [stage 5](https://docs.microsoft.com/en-us/azure/app-service/deploy-staging-slots#swap-operation-steps) both the staging and production slots would be SchedulingPublisher as the old production site does not have the new configuration applied until after the new code is accepting requests. So there is a window where both would be Scheduling Publisher.

Jeavon:
> We were doing the same sort of approach as your code example, with the local cache option on. I'm not sure whether we changed ServerRole per slot, which may have been our error. We removed that behaviour and have not had a issue since with scheduled publishing. Looking at the Azure Docs both slots, under the approach you suggest, at [stage 5](https://docs.microsoft.com/en-us/azure/app-service/deploy-staging-slots#swap-operation-steps) both the staging and production slots would be SchedulingPublisher as the old production site does not have the new configuration applied until after the new code is accepting requests. So there is a window where both would be Scheduling Publisher.

I've not tracked the full swap process with v9 (.net5/6) but I did with v8 and the 2 apps were never running at the same time, as soon as the swap is complete the apps are booting with slot config (or warming with app initialization). Would be interesting to do again with v9...

nzdev:
When local cache is used, the last step of the slot only recycles the app pool, so the new settings / code won't apply till a restart is done.

Jeavon:
So the potential overlap time is when the slot being promoted is being warmed up whilst the existing production slot is still alive, in theory, both could attempt scheduled publishing in that period?

I was thinking about the de-promoted code in the staging slot but that gets restarted "quickly" and can be managed with a stick slot setting.

Jeavon:
v9.4 provides a new setting MainDomKeyDiscriminator which will need documenting - https://github.com/umbraco/Umbraco-CMS/pull/12037 
I'll close this PR

PR #3785, opened 2022-02-10
Custom boot failed error page

poornimanayar:
I believe that the custom boot failed error page is available for v9 as well. Hence updating the docs


busrasengul:
Thank you @poornimanayar 
We'll review this soon 🙌

jonat123:
@poornimanayar good catch, it does indeed work with v9 as well :D 

Merging it in!

PR #3786, opened 2022-02-10
Update version-specific.md

Ambertvu:
I just tried this on two separate installations (clean and an actual project), and kept running into the same issue, might be a good addition to add to the upgrade notes.


skttl:
Should be fixed in core, so you actually can upgrade from 8 to 9.3 No one wants to step over several versions in order to upgrade.

PR #3787, opened 2022-02-10
Update version-specific.md

Ambertvu:
I just tried this on two separate installations (clean and an actual project), and kept running into the same issue, might be a good addition to add to the upgrade notes.


busrasengul:
Thank you @Ambertvu 
We'll review this soon!

Ambertvu:
I might be good to add for as long as it isn't fixed in core. First people with this issue are already appearing on the forums.

PR #3792, opened 2022-02-14
Update Dashboard: Simplify code

erikjanwestendorp:
Add using `Umbraco.Cms.Core;` and simplify code a bit.


busrasengul:
This is sensible! Thanks @erikjanwestendorp 
Let's merge this in!

PR #3794, opened 2022-02-15
Fix Constructor name

erikjanwestendorp:
The name of the constructor name was different than the class name.


busrasengul:
Thank you @erikjanwestendorp 
Let's merge this in!

PR #3801, opened 2022-02-17
Changing model to Model within code snippet

OwainWilliams:
Noticed that the code sample as lower case m rather than upper case M. 
Was using the samples and noticed it didn't recognise `model` but `Model` worked fine.


busrasengul:
Thank you @OwainWilliams let's merge this in!

PR #3807, opened 2022-02-22
Added Import/export document type section

eshanrnh:


jonat123:
Looks good :-) 

I was wondering whether it maybe makes sense to add a warning that it is not possible to export it on Umbraco Cloud?

PR #3808, opened 2022-02-23
Added necessary cookies information

eshanrnh:


Reference/Security/Cookies/index.md
```diff
@@ -0,0 +1,25 @@
...
+| UMB_PREVIEW                | Allows a previewed page to act as a published page only on the browser which has initialized previewing.          | Session    |
+| UMB-WEBSITE-PREVIEW-ACCEPT | Client-side cookie that determines whether the user has accepted to be in Preview Mode when visiting the website. | Session    |
+| UMB_INSTALLID              | Used to store the Umbraco software installer id.                                                                  | Session    |
+| UMB_UPDCHK                 | Enables your system to check for the Umbraco software updates.                                                    | Session    |
+| UMB-XSRF-V                 | Used to store the backoffice antiforgery token validation value.                                                  | Session    |
+| UMB-XSRF-TOKEN             | Set for angular to pass in to the header value for "X-UMB-XSRF-TOKEN"                                             | Session    |
+| X-UMB-XSRF-TOKEN           | The header name that angular uses to pass in the token to validate the cookie                                     | Session    |
```
ronaldbarendse, line 21: This is the HTTP header name that's used together with the `UMB-XSRF-TOKEN` cookie, so we can omit it from this list.
```suggestion
```

Reference/Security/Cookies/index.md
```diff
@@ -0,0 +1,25 @@
...
+| umbraco-no-content         | Displays the route name of the page when the website has no published content.                                    | Session    |
```
ronaldbarendse, line 22: This is the route name used when no published content is available, so we can also omit this.
```suggestion
```

Reference/Security/Cookies/index.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Necessary Cookies
+
+These cookies are necessary while accessing Umbraco backoffice for the functioning of the website or to allow you to enjoy the contents and services you request.
+
+| Name                       | Purpose                                                                                                           | Expiration |
+|----------------------------|-------------------------------------------------------------------------------------------------------------------|------------|
+| UMB_PREVIEW                | Allows a previewed page to act as a published page only on the browser which has initialized previewing.          | Session    |
+| UMB-WEBSITE-PREVIEW-ACCEPT | Client-side cookie that determines whether the user has accepted to be in Preview Mode when visiting the website. | Session    |
+| UMB_INSTALLID              | Used to store the Umbraco software installer id.                                                                  | Session    |
```
ronaldbarendse, line 17: Note that cookies are case-sensitive, so this needs to be updated to:
```suggestion
| umb_installId              | Used to store the Umbraco software installer id.                                                                  | Session    |
```

Reference/Security/Cookies/index.md
```diff
@@ -0,0 +1,25 @@
+---
+versionFrom: 9.0.0
+---
+
+# Cookies
+
+We use cookies to understand the usage of the website, improve the website performance, and to increase the relevance of our communication and advertising.
```
ronaldbarendse, line 7: I don't think this applies to Umbraco and definitely not to our communication and advertising, as these cookies are set on the domain you're hosting your own Umbraco installation on.

We should primarily document what cookies are set and for what reason (basically only for accessing the back-office), so users know whether they need to include these in their own cookie policy.

ronaldbarendse (changes requested):

ronaldbarendse (approved):

PR #3816, opened 2022-02-27
generic error page section update

RyuLindow:
updated the generic error page section, as it wasn't updated for v9


jonat123:
@RyuLindow Good catch!

Merging it in right away :-)

PR #3828, opened 2022-03-07
Removed the warning

RyuLindow:
The issue is marked as solved on the tracker :)


jonat123:
@RyuLindow great catch! 

I will merge it in right away and press the button ;-)

PR #3829, opened 2022-03-08
Package good practice guide

KevinJump:
Getting this in as a draft so I don't delete it from my computer again! - 

at this stage the package team want to have a look at this before it becomes documentation.

### DIFF
--- a/Extending/Packages/Good-practice/index.md
+++ b/Extending/Packages/Good-practice/index.md
@@ -5,47 +5,49 @@ meta.Description: "Information on good practice and common defaults for Umbraco 
 ---
 
 # Good practice and defaults
-_This document provides some guides and notes on package development, and good practice that will help you maintain and support your package through multiple releases and versions of Umbraco. These good practices are not prescriptive, but mearly offer a guide as to what we have found works well, and not-so-well, when developing packages for Umbraco_
+_This document provides some guides and notes on package development, and good practice that will help you maintain and support your package through multiple releases and versions of Umbraco. These good practices are not prescriptive, but mearly offer a guide as to what we have found works well, and not-so-well, when developing packages for Umbraco._
 
 ## App_Plugins
-Files in the `App_Plugins` folder should be considered immutable, that is they are not something a user of your package should be expecting to change on their site. 
+To extend the Umbraco backoffice, a package can provide files (like a `package.manifest` and AngularJS views/controlers) that should be stored within the `App_Plugins` folder. It's recommended to put all files in a subfolder with a unique name, preferrably using the package name, e.g. `App_Plugins\MyPackage`.
 
-The default delivery method for files to the App_Plugins folder is via a `.targets` file within a package, this means when a website is built the files in this folder are copied over from the nuget cache, and any changes a user might have made will be lost. Equally if the user performs a `dotnet clean` on a solution all files in the App_Plugins folder will likely be deleted. 
+Files in the `App_Plugins` folder should be considered immutable, that is they are not something a user of your package should be expecting to change on their site.
+
+The default delivery method for files to the `App_Plugins` folder is via a `.targets` file within a package: this means when a website is built, the files in this folder are copied over from the NuGet cache and any changes a user might have made to these files will be lost. Equally, if the user performs a `dotnet clean` on a solution, all files in the `App_Plugins` folder will be deleted.
 
 If you have files that you expect users of your package to alter you should not place them in the `App_Plugins` folder.
 
 ## Views
-Views are used to render content on the front end of a website. If your plugin provides a way for the user to present the content publicly you should copy these files to the views folder. 
+Views are used to render content on the front end of a website. If your package provides a way for the user to present the content publicly, you should copy these files to the views folder.
 
-As the files will still be copied during build you should ensure your target file doesn't overwrite newer or altered files and doesn't delete files on clean. 
+As the files will still be copied during build you should ensure your target file doesn't overwrite newer or altered files and doesn't delete files on clean.
 
-_This method isn't perfect and if you do have a package that is heavy on views you might consider building a dotnet template or Razor class library to contain the views so they don't pollute your user's solutions._ 
+_This method isn't perfect and if you do have a package that is heavy on views you might consider building a dotnet template or Razor Class Library (RCL) to contain the views, so they don't pollute your user's solutions._
 
 ## License files
-Umbraco products store their licenses in `/umbraco/Licences` and it is considered a good location for third party packages that require license files to also store their licence files there. 
+Umbraco products store their licenses in `/umbraco/Licences` and it is considered a good location for third party packages that require license files to also store their licence files there.
 
-_The default .gitignore for Umbraco templates will include any files in this folder in a repo (while ignoring most of the rest of the Umbraco folder)_
+_The default .gitignore for Umbraco templates will include any files in this folder in a repo (while ignoring most of the rest of the Umbraco folder)._
 
 :::note
-The `/umbraco/Licenses` folder does not exist on a fresh installation of Umbraco so you might need to create it before you save your license file to this folder. 
+The `/umbraco/Licenses` folder does not exist on a fresh installation of Umbraco so you might need to create it before you save your license file to this folder.
 :::
 
 ## Operating system considerations
-Umbraco 9 is a .NET Core application and it can be ran on multiple operating systems (e.g Windows, Linux). When developing packages there are a few things you should be aware of so your package can run on all possible operating systems. 
+Umbraco (version 9 and up) is an ASP.NET Core application and can be ran on multiple operating systems (e.g. Windows, Linux and macOS). When developing packages there are a few things you should be aware of so your package can run on all possible operating systems.
 
 ### Case sensitivity 
-The Linux file system is case sensitive, so `App_Plugins/myPackage` is a different location to `app_plugins/MyPACKAGE`. When building your package you should ensure that you always refer to folders and paths in a consistent way. 
+The Linux and macOS file system is case sensitive (by default), so `App_Plugins/myPackage` is a different location to `app_plugins/MyPACKAGE`. When building your package you should ensure that you always refer to folders and paths in a consistent way.
 
 :::tip
-You can adjust the case sensitivity of a Windows folder by running a command against the folder. 
+You can adjust the case sensitivity of a Windows folder by running a command against the folder.
 
-``` bash
+```bash
 fsutil.exe file queryCaseSensitiveInfo <path>
 ```
 :::
 
-### Well known folders and case sensitivity  
-Some folders within Umbraco will already exist for all installations, if you access these folders you need to be aware of the case used to ensure you end up in the right place: 
+### Well known folders and case sensitivity
+Some folders within Umbraco will already exist for all installations, if you access these folders you need to be aware of the case used to ensure you end up in the right place:
 
 | Folder | Note | 
 |-|-|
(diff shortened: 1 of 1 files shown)


Rockerby (changes requested):

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,141 @@
...
+:::
+
+### Path manipulation
+Building folder path strings manually can cause problems when swapping between file systems. Windows uses the backslash character to '\' separate folders and files, Linux uses the forward slash '/'.
+
+You should use the .Net `Path` commands wherever possible when building paths to ensure that the correct path is build. 
+
+``` cs
+// WRONG: don't build paths manually.
+var myPath = myFolder + "\myfile.txt";
```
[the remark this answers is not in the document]
KevinJump: I think i need to add something about getting paths via the FileProvider or `MapPath` commands, because it should never be hardwired, but this example could then have a mapped path value for 'myFolder' and it would make more sense.

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Don't save to appsettings.json
+You should not alter `appsettings.json` in code. 
+
+Settings in aspnet core are merged from a number of different locations at runtime. You cannot guarantee that `appsettings.json` is the location that a setting is read from and your users may well not want certain settings in that file. You can read settings from the configuration but you cannot assume they have come from `appsettings.json`.
+
+### Settings locations
+There are many options for where you might save your settings and a lot will depend on the nature of your package: 
+
+#### Save to database
```
dawoe, line 142: Maybe a section about the IKeyValueService in Umbraco to store simple settings. I use this in my RedirectsViewer package : https://github.com/dawoe/umbraco-redirectsviewer/blob/develop/Source/Our.Umbraco.RedirectsViewer/Controllers/UserGroupsApiController.cs#L89

sophie-neale (commented):
Hey Kevin! 
Sophie from Docs here, thanks so much for adding this - it looks great! 
We'll be going over any outstanding PR's in our meeting next week but for now I've offered some little grammatical tweaks to help 🙂

Extending/Packages/Good-practice/index.md
KevinJump: yeah - originally i did write a load of potential implementation options into the document, it got to big 😞 - so i went for streamlined, this document is more 'pointers' than actual implementation, not sure the balance is right, but maybe there should be other pages for this type of thing ?

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
+---
+versionFrom: 9.0.0
+meta.Title: "Package development good practice"
+meta.Description: "Information on common defaults and good practice for umbraco package development"
```
ronaldbarendse, line 4:
```suggestion
meta.Description: "Information on good practice and common defaults for Umbraco package development"
```

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+---
+
+# Good practice and defaults
+_This document provides some guides and notes on package development, and good practice that will help you maintain and support your package through multiple releases and versions of Umbraco. These good practices are not prescriptive, but mearly offer a guide as to what we have found works well, and not-so-well, when developing packages for Umbraco_
+
+## App_Plugins
```
ronaldbarendse, line 10: Should this also explain what does go into the App_Plugins folder? Something like: 'Each package should create a subfolder with the package name/id (e.g. `App_Plugins\MyPackage`) to store their backoffice related code, like a `package.manifest`, the AngularJS views/controllers, CSS and language files.'

And maybe also note that these files will be served publicly (even though this folder isn't located in the `wwwroot`)?

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+The default delivery method for files to the App_Plugins folder is via a `.targets` file within a package, this means when a website is built the files in this folder are copied over from the nuget cache, and any changes a user might have made will be lost. Equally if the user performs a `dotnet clean` on a solution all files in the App_Plugins folder will likely be deleted. 
+
+If you have files that you expect users of your package to alter you should not place them in the `App_Plugins` folder.
+
+## Views
+Views are used to render content on the front end of a website. If your plugin provides a way for the user to present the content publicly you should copy these files to the views folder. 
+
+As the files will still be copied during build you should ensure your target file doesn't overwrite newer or altered files and doesn't delete files on clean. 
+
+_This method isn't perfect and if you do have a package that is heavy on views you might consider building a dotnet template or Razor class library to contain the views so they don't pollute your user's solutions._
```
ronaldbarendse, line 22: Another way is to use ASP.NET [Tag Helpers](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/intro) or [View Components](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components) instead. Or maybe even better: include the views in a Razor Class Library instead?

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+_The default .gitignore for Umbraco templates will include any files in this folder in a repo (while ignoring most of the rest of the Umbraco folder)_
+
+:::note
+The `/umbraco/Licenses` folder does not exist on a fresh installation of Umbraco so you might need to create it before you save your license file to this folder. 
+:::
+
+## Operating system considerations
+Umbraco 9 is a .NET Core application and it can be ran on multiple operating systems (e.g Windows, Linux). When developing packages there are a few things you should be aware of so your package can run on all possible operating systems. 
+
+### Case sensitivity
```
ronaldbarendse, line 36: Is using constants to keep the casing the same something we should advise?

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+### Well known folders and case sensitivity  
+Some folders within Umbraco will already exist for all installations, if you access these folders you need to be aware of the case used to ensure you end up in the right place: 
+
+| Folder | Note | 
+|-|-|
+| /App_Plugins | Uppercase `A` and `P` |
+| /App_Plugins/[Ll]ang | Uppercase `L` but in later version of Umbraco 9.3+ can be upper or lowercase `L` |
+| /Views | Uppercase `V` | 
+| /umbraco/Licenses | Lowercase `u` uppercase `L` |
+| /config | Lowercase `c` |
```
ronaldbarendse, line 56: This isn't used in v9 anymore (and the `umbraco\config\lang` folder should only include project specific/user language overrides, so shouldn't be something for package authors to deal with).

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+``` cs
+var localTempRoot = Path.GetFullPath(hostingEnvironment.LocalTempPath);
+```
+
+#### Relative paths
+If you require the path of a folder relative to the site root, you can use the `IHostingEnvironment` methods to map a path. 
+
+``` cs 
+var contentRootPath = hostingEnvironment.MapPathContentRoot("/");
+var webrootPath = hostEnvironment.MapPathWebRoot("/");
```
ronaldbarendse, line 79: This should also show how you can create an absolute path to use as URL (and maybe also explain virtual paths/application relative paths):
```suggestion
var contentRootPath = hostingEnvironment.MapPathContentRoot("~/"); // Full physical path to the content/project root
var webrootPath = hostEnvironment.MapPathWebRoot("~/"); // Full physical path to the web root (served publicly)
var absolutePath = hostEnvironment.ToAbsolute("~/"); // Absolute path to use as URL
```

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+### Folder Access 
+You should not assume things about the folder structure of a site, and ideally you should not use direct IO commands to access the file system. Within a Asp.Net Core site access to the disk is usually managed with FileProviders, you can get access to the file providers from the `IWebHostEnvironment` class.
```
ronaldbarendse, line 83:
```suggestion
You should not assume things about the folder structure of a site, and ideally you should not use direct IO commands to access the file system. Within an ASP.NET Core site access to the disk is usually managed with FileProviders, you can get access to the file providers from the `IWebHostEnvironment` class.
```

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+:::note
+This is the preferred method for file IO. because not all files served up by a site may in fact be in the wwwroot folder when you expect them to be - this is especially true if the site is using Razor class library to insert static files.
+:::
+
+### Path manipulation
+Building folder path strings manually can cause problems when swapping between file systems. Windows uses the backslash character to '\\' separate folders and files, Linux uses the forward slash '/'.
+
+You should use the .Net `Path` commands wherever possible when building paths to ensure that the correct path is built.
```
ronaldbarendse, line 116:
```suggestion
You should use the .NET `Path` commands wherever possible when building paths to ensure that the correct path is built. 
```

Extending/Packages/Good-practice/index.md
```diff
@@ -0,0 +1,161 @@
...
+## Settings
+All but the most simple package will probably require some settings to be stored for the users to control and change the behavior of how the package works on a site. Where you store these settings will depend a lot on the nature of the package. 
+
+### Property Editors
+Property Editors should store their settings as part of their datatype in Umbraco. This is the standard way property editors behavior is controlled, it is familar to users and is supported by deployment tools. 
+
+### Don't save to appsettings.json
+You should not alter `appsettings.json` in code. 
+
+Settings in aspnet core are merged from a number of different locations at runtime. You cannot guarantee that `appsettings.json` is the location that a setting is read from and your users may well not want certain settings in that file. You can read settings from the configuration but you cannot assume they have come from `appsettings.json`.
```
ronaldbarendse, line 137:
```suggestion
Settings in ASP.NET Core are merged from a number of different locations at runtime. You cannot guarantee that `appsettings.json` is the location that a setting is read from and your users may well not want certain settings in that file. You can read settings from the configuration, but you cannot assume they have come from `appsettings.json`.
```

Extending/Packages/Good-practice/index.md
KevinJump: yes this is a better way (RCL) but until v10 this isn't smooth enough - i suspect events will overtake this document 😄 and we should change it to how to do this in a RCL.

Extending/Packages/Good-practice/index.md
KevinJump: i think the caseness of these folders is worth noting, and if you want to make a package that works on all v9s you need to know that pre v9.3 the case thing was an issue. (again if we make this a v10 doc, we can simplify)

sofietoft:
Hi all 👋 

What's the state of the PR here? 
Looks like there are still some unanswered comments in the mix? Is that right?

Extending/Packages/Good-practice/index.md
ronaldbarendse: I agree that we should this document from becoming too big, but maybe we can add at least a single mention of this, e.g. 'Simple values can also be stored in the database using the `IKeyValueService`' and link that to the API docs?

Extending/Packages/Good-practice/index.md
ronaldbarendse: RCL projects have only been properly supported since v11, so it makes sense to only include additional information about how and when to use that project type in a version 11 specific documentation page...

ronaldbarendse:
@KevinJump I've gone through the document and cleaned up some paragraphs, can you please check these changes and see if the other, unresolved comments still need to be addressed (at least in the initial version of this documentation page)?

Extending/Packages/Good-practice/index.md
KevinJump: I've added a line about IKeyValueService , but i am not sure there is actually a documentation page to link to ?

Extending/Packages/Good-practice/index.md
KevinJump: Yes - added a line

sofietoft:
Hi all 👋 
I can see that this PR has been approved.
I'll give it a review as soon as possible.

Just a heads up:
I will be closing down this PR and opening a new one, as this file will not be added correctly to the new structure or fit with the format of the new docs platform.
I'll make sure to mention you in the new PR, so you can help verify that it all looks as it should.

Thanks!

PR #3833, opened 2022-03-09
Update index.md

FransLammers:
When adding rewrite rules to a Cloud project I stumbled on the issue that the XML was not deployed correctly so the site would not start up. 
See also: https://github.com/umbraco/Umbraco.Cloud.Issues/issues/558

I have added a little explanation about how to handle this.

### DIFF
--- a/Reference/Routing/IISRewriteRules/index.md
+++ b/Reference/Routing/IISRewriteRules/index.md
@@ -45,6 +45,15 @@ An example of how this can be done is
   </rules>
 </rewrite>
 ```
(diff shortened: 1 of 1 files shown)


busrasengul:
Thank you @FransLammers 
We'll review this soon!

jonat123:
Hello @FransLammers :-) 

Thank you so much for this PR :) 

However, we have actually worked internally here I'm Umbraco to update the article based on feedback from our cloud customers and our developers.

You can see the PR here: #3834

In which we have also added a note about adding the CopyToOutputDirectory in the csproj file :-) 

I will close down your PR since it is a duplicate, however, do not hesitate to create a new PR is in the future if you see some wrong info, an article that needs more details, or similar 🙂

We always love to look at PRs from the community 😄

Have a great day Frans 😄

PR #3834, opened 2022-03-10
Url Rewrites in V9 update

jonat123:
Some changes to the article based on feedback from customers and internally at Umbraco HQ


ronaldbarendse:
I've improved the texts to explain the changes a little bit better and have updated (and locally tested) the examples. Although using the Wildcard syntax is a lot easier to read then the Regex patterns, this [isn't supported by the URL Rewriting Middlware](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/url-rewriting?view=aspnetcore-5.0#unsupported-features-1).

jonat123:
@ronaldbarendse Thanks for taking a look at it! 

The changes looks great 😁

I'll merge the PR in right away :D

PR #3850, opened 2022-03-15
Rename folders in reference security

erikjanwestendorp:
Rename folders (Reference --> Security) so they are capitalized consistently. 🙈 
image


busrasengul:
Thank you @erikjanwestendorp 
I'm merging this right in!

PR #3853, opened 2022-03-16
Updated this article due to content node limits

Aaen:
We have removed the content node limits so I have made this PR


jonat123:
@Aaen thanks for the PR :D 

Merging it in right away!

PR #3854, opened 2022-03-17
Update index.md

jacklawry:
It doesn't explain that in XPath errorPages is the DocTypeAlias for the error page.

### DIFF
--- a/Tutorials/Custom-Error-Pages/index.md
+++ b/Tutorials/Custom-Error-Pages/index.md
@@ -88,6 +88,10 @@ XPath example:
 }
 ```
 
+:::note
+In the above XPath example `//errorPages` is the DocTypeAlias
+:::
+
 ## Errors with booting a project
 
 Sometimes you might experience issues with booting up your Umbraco project. This could be a brand new project, or it could be an existing project after an upgrade.


greystate:
I do think it would be a good idea to use a better example now that this is being updated - the existing one assumes the Document Type has been named **Error Pages** in Umbraco (which would generate the alias `errorPages` - a much more likely example would be a singular name here, e.g. `errorPage`.

busrasengul:
Thank you @jacklawry 
We'll look into this asap!

sofietoft:
Hi @jacklawry !
Thanks for the PR - and your first contribution to the UmbracoDocs 🥳 

I think it's a great idea to add a little note about the name of the Document Type, as that might not be entirely clear from the code sample! 💪 

@greystate I hear what you're saying about updating the entire sample, and I think it's a good idea!
It does create a bigger task though, so perhaps we should add that as a task on the Issue Tracker instead? 🤔 
Might even be something @jacklawry, would be up for doing? 😄

sofietoft:
Hi @jacklawry and @greystate !

I'll merge this PR in as it is now.
Thank you both for pitching in 😄 

As this is your first ever PR to the Umbraco Docs @jacklawry - I'll make sure to give you a C-Trib badge on your Our profile 🥳
Looking forward to seeing a lot more contributions from your site 😉

PR #3857, opened 2022-03-19
Small updates to Deploy and Forms information

AndyButland:
This PR is just correcting on error on Deploy configuration I noticed, and added some detail about Forms "magic strings" that looked to be missing.


jonat123:
@AndyButland Looks good,

merging it in right away :-)

PR #3865, opened 2022-03-24
Added a note about read-only file systems

harry-gordon:
Ran into this issue today and spent a while scratching my head, so I think it's definitely worth mentioning.

One thing I haven't included was that I inadvertently ran into this issue using Azure Devops (see my comments on the issue) but this feels overly specific so I haven't included this in the documentation.

Here's the relevant issue: https://github.com/umbraco/Umbraco-CMS/issues/12043

### DIFF
--- a/Fundamentals/Setup/Server-Setup/azure-web-apps.md
+++ b/Fundamentals/Setup/Server-Setup/azure-web-apps.md
@@ -47,6 +47,14 @@ __If you are load balancing or require the scaling ("scale out") ability of Azur
 
 It is important to know that Azure Web Apps uses a remote file share to host the files to run your website (i.e. the files running your website do not exist on the machine running your website). In many cases this isn't an issue but it can become one if you have a large amount of IO operations running over remote file share.
 
+## Issues with read-only filesystems
+
+Although Umbraco can be configured to use environmental storage it still requires its working-directory to be writable. If Umbraco is deployed to a read-only file system it will [fail to boot](https://github.com/umbraco/Umbraco-CMS/issues/12043).
+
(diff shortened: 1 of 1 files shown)


Jeavon:
@harry-gordon I think it might be good to actually like to that issue?

Fundamentals/Setup/Server-Setup/azure-web-apps.md
```diff
@@ -47,6 +47,12 @@ __If you are load balancing or require the scaling ("scale out") ability of Azur
 
 It is important to know that Azure Web Apps uses a remote file share to host the files to run your website (i.e. the files running your website do not exist on the machine running your website). In many cases this isn't an issue but it can become one if you have a large amount of IO operations running over remote file share.
 
+## Issues with read-only filesystems
+
+Although Umbraco can be configured to use environmental storage it still requires its working-directory to be writable. If Umbraco is deployed to a read-only file system it will [fail to boot](https://github.com/umbraco/Umbraco-CMS/issues/12043).
```
harry-gordon, line 52: @Jeavon I linked to the issue here but only in passing, does that seem ok or would you rather it was more explicit?

Jeavon: All good, wondering if we should add a link from https://github.com/umbraco/UmbracoDocs/blob/main/Fundamentals/Setup/Server-Setup/Load-Balancing/azure-web-apps.md#deployment-considerations to your new section also

harry-gordon: Possibly! If so, I think it might be a footnote under temporary files?

I hope though that someone reading the load balancing notes has read the more general Azure guidance (and/or successfully deployed Umbraco on a single instance), but I defer to you on this 👍.

Jeavon: It's a good point, I've added a link to it at the beginning of the load balancing document, thanks!

PR #3871, opened 2022-03-25
Updates package.manifest example with missing info about bundleOptions

warrenbuckley:
We had missing info about a property available in our package.manifest
https://github.com/umbraco/Umbraco-CMS/pull/10691

### DIFF
--- a/Extending/Property-Editors/package-manifest.md
+++ b/Extending/Property-Editors/package-manifest.md
@@ -238,7 +238,7 @@ The parameter editors array follows the same format as the property editors desc
 
 * `Default` - The default bundling behavior for assets in the package folder where the assets will be bundled with the typical packages bundle.
 * `None` - The assets in the package will not be processed at all and will all be requested as individual assets and will essentially be a bundle that has composite processing turned off for both debug and production
-* `Independent` - The packages assets will be processed as it's own separate bundle. (in debug, files will not be processed)
+* `Independent` - The packages assets will be processed as its own separate bundle. (In debug, files will not be processed)
 
 ## JSON Schema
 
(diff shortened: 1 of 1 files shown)


Extending/Property-Editors/package-manifest.md
```diff
@@ -228,6 +233,13 @@ The parameter editors array follows the same format as the property editors desc
 }
 ```
 
+## Bundling
+`bundleOptions` is an enum that expects one of the following values:
+
+* `Default` - The default bundling behavior for assets in the package folder where the assets will be bundled with the typical packages bundle.
+* `None` - The assets in the package will not be processed at all and will all be requested as individual assets and will essentially be a bundle that has composite processing turned off for both debug and production
+* `Independent` - The packages assets will be processed as it's own separate bundle. (in debug, files will not be processed)
```
busrasengul, line 241:
```suggestion
* `Independent` - The packages assets will be processed as its own separate bundle. (In debug, files will not be processed)
```

busrasengul (commented):
Thank you @warrenbuckley 
Added tiny tweaks. If you can apply them it's good to merge!

Extending/Property-Editors/package-manifest.md
```diff
@@ -228,6 +233,13 @@ The parameter editors array follows the same format as the property editors desc
 }
 ```
 
+## Bundling
+`bundleOptions` is an enum that expects one of the following values:
```
sofietoft, line 237: I don't know the suggested changes here makes sense, but `enum` threw me off a bit 🤔 
Perhaps we could loose the ´type´?

```suggestion
`bundleOptions` is an enumerable type that expects one of the following values:
```

sofietoft (commented):
Great additions 💪 

I've added a single suggestion - let me know if it makes sense!

warrenbuckley:
@sofietoft & @busrasengul

sofietoft (approved):

sofietoft:
Looks great @warrenbuckley !

Merging in 💪

Issue #3884, opened 2022-03-29
Add UmbracoApplicationLifetime notifications


ronaldbarendse:
## What type of issue is it?
Missing documentation: the `UmbracoApplicationStartingNotification`, `UmbracoApplicationStartedNotification`, `UmbracoApplicationStoppingNotification` and `UmbracoApplicationStoppedNotification` aren't documented

## What article/section is this about?
[Notifications](https://our.umbraco.com/documentation/Reference/Notifications/) and subpages

## Describe the issue
Umbraco 9.3.0 includes 2 new notifications (see PR https://github.com/umbraco/Umbraco-CMS/pull/11857 and https://github.com/umbraco/Umbraco-CMS/pull/11883), but I also can't find any documentation on the other 2 existing notifications (besides some references).

The referenced PRs include most information about when and why to use these notifications, but let me know if more information is needed to write the documentation.

I also noticed that not all documented notifications below [Notifications](https://our.umbraco.com/documentation/Reference/Notifications/) are added in the navigation (currently only [EditorModel Notifications](https://our.umbraco.com/documentation/Reference/Notifications/EditorModel-Notifications/) and [MemberService Notifications](https://our.umbraco.com/documentation/Reference/Notifications/MemberService-Notifications/) are shown).

sofietoft:
Hi @ronaldbarendse 

I've had a look through our docs, and you're right that we do not have those 4 notifications documented.
Some of them are referenced in various articles, but that's it.

I'll see about having them documented as soon as possible. We will reach out if we need anything further in order to write the articles 💪 

So, the reason that most of the documented notifications aren't shown in the navigation, is because the navigation only includes folders. And because only the EditorModel Notifications and MemberService Notifications articles are added into folders, they are the only ones shown there 😅 
Yes, Our is a _magical_ place, if you didn't know already! 🤪

It would make sense to add folders for the rest, so they will be shown in the navigation as well 💪

eshanrnh:
These are now documented and can be found at https://our.umbraco.com/documentation/Reference/Notifications/UmbracoApplicationLifetime-Notifications/

Closing the issue 🙂

PR #3890, opened 2022-04-01
Create folders for notifications

erikjanwestendorp:
Move the notification articles to folders  so that they are displayed in the menu.

Belongs to:  https://github.com/umbraco/UmbracoDocs/issues/3884


eshanrnh (approved):
Looks good 💪

eshanrnh:
Thanks for the PR @erikjanwestendorp. Merging it in 🙂

dampee:
@eshanrnh just checking, have you updated the REDIRECTS before merging?

sofietoft:
Good point about the rewrites @dampee !

We've done some testing here, and it does not seem like we actually need to make the rewrites here.
As the new folder names uses the same naming as the files did before, I think we should be OK.
It looks like Our automatically "ignores" the file extension, so the path will actually be the same 🤞🏻

So for now, we're not gonna add the redirects. Let's keep an eye on it - we can do it later, if it turns out we run into some broken links here and there.

PR #3891, opened 2022-04-01
Update Smidge Config to reflect the UmbracoSmidgeConfigCacheBuster

marcemarc:


marcemarc:
hi @p-m-j I've just been puzzling over why the docs setting of smidge version doesn't result in a v12345 on the assets, and saw your update to the UmbracoSmidgeConfigCacheBuster - I've updated the explanation here to hopefully reflect your changes, be ace if you had a moment to cast your eye over it!

Reference/V9-Config/RuntimeMinificationSettings/index.md
```diff
@@ -15,45 +15,49 @@ This section allows you to configure the runtime minifications (defaults shown),
 "Umbraco": {
   "CMS": {
     "RuntimeMinification": {
-      "UseInMemoryCache": false,
-      "CacheBuster": "Version"
+      "UseInMemoryCache": true,
+      "CacheBuster": "Timestamp"
```
p-m-j, line 19: This section is intended to show the defaults so it needs to be reverted. See https://github.com/umbraco/Umbraco-CMS/blob/79fdb9a876a822f8662679711f5f801f3159ff42/src/Umbraco.Core/Configuration/Models/RuntimeMinificationSettings.cs for ref

Reference/V9-Config/RuntimeMinificationSettings/index.md
```diff
@@ -15,45 +15,49 @@ This section allows you to configure the runtime minifications (defaults shown),
...
 
 ## Cache buster
 
-Specifies what type of cache buster to use, the options are:
+This is the setting that determines how Smidge should break it's cache.
+
+The options are:
 
-* Version - Caches will be busted every time the version string changes.
+* Version - Caches will be busted every time the version string changes in your appsettings.json file
```
p-m-j, line 34: appsettings is only one way to manage configuration, possibly beter to mention "configuration" with a link to [Managing configuration](https://our.umbraco.com/documentation/Reference/V9-Config/#managing-configuration)

Reference/V9-Config/RuntimeMinificationSettings/index.md
```diff
@@ -15,45 +15,49 @@ This section allows you to configure the runtime minifications (defaults shown),
...
 
-## Use in memory cache
-
-Specifies if Smidge should use an in memory cache or not.
+This setting determines whether Smidge should save it's cached output in memory, or in a file on disk. If set to false, then the folder will be created at the wwwroot of your Umbraco site in a folder called 'Smidge'/
 
 ## Cache buster
 
-Specifies what type of cache buster to use, the options are:
+This is the setting that determines how Smidge should break it's cache.
```
p-m-j, line 30:
```suggestion
Specifies mechanism for cache invalidation.
```

Reference/V9-Config/RuntimeMinificationSettings/index.md
p-m-j, line 19:
```suggestion
      "UseInMemoryCache": false,
      "CacheBuster": "Version"
```

This section is intended to show the defaults so it needs to be reverted. See [RuntimeMinificationSettings.cs](https://github.com/umbraco/Umbraco-CMS/blob/79fdb9a876a822f8662679711f5f801f3159ff42/src/Umbraco.Core/Configuration/Models/RuntimeMinificationSettings.cs) for ref

p-m-j:
Hey @marcemarc, I added some comments with my $0.02, the docs team are much better at writing style than I am so probably better to wait for their input as opposed to mine.

Looks good though thanks for the PR

Reference/V9-Config/RuntimeMinificationSettings/index.md
p-m-j, line 34:
```suggestion
* Version - Caches will be busted when your assembly version changes, Umbraco version changes & when version string in configuration changes.
```

p-m-j:
Might be good to avoid explicit mention of smidge as its use is just an implementation detail of the umbraco RuntimeMinification abstractions.

marcemarc:
> Hey @marcemarc, I added some comments with my $0.02, the docs team are much better at writing style than I am so probably better to wait for their input as opposed to mine.
> 
> Looks good though thanks for the PR

Thanks @p-m-j! - will revert to the docs team! they do an awesome job...

Reference/V9-Config/RuntimeMinificationSettings/index.md
```diff
@@ -21,39 +21,43 @@ This section allows you to configure the runtime minifications (defaults shown),
...
+      "UseInMemoryCache": true,
       "CacheBuster": "Version",
       "Version": "1234"
     }
   }
 }
 ```
+The actual 'Version' number will not be visible in the url of the assets, this is because it is combined, along with the Umbraco Version from configuration and the your project assembly dll, and then once combined a 'hash' is generated to obscure these details.
+
+in the HTML link thus: ```<link href='/sc/69a3dbf6.1cf661e7.css.d=efb7329470a16d7020272a294742726ebe1da5f1' rel='stylesheet' type='text/css'/>```
```
Jeavon, line 55: Using `Version` cache buster the output in v9.4.1 is like this

`<link href='/sb/umbraco-backoffice-init-css.css.v7a71f91360259c5f7c3337f152b0df01eeee36f0' rel='stylesheet' type='text/css'/>`

marcemarc: done

p-m-j: This changes based on the value of [Hosting:Debug](https://our.umbraco.com/documentation/Reference/V9-Config/HostingSettings/#debug).

I'm not sure the emphasis on the hash appearing is useful as long as users are aware of when then cache busting takes place.

However I do wonder if we should change the value to be {{hash}.{configured explicit version value} so the value from config shows in the URL to avoid confusion.

Jeavon:
@p-m-j whilst figuring this out with @marcemarc it has also occurred to me that I think Umbraco should be setting the Smidge dataFolder to the Umbraco Temp or at least providing the option to set it, what do you think?

p-m-j:
@Jeavon I think that's already covered although perhaps it could be clearer, see final paragraph on [RuntimeMinificationSettings](https://our.umbraco.com/documentation/Reference/V9-Config/RuntimeMinificationSettings/)

Jeavon:
@p-m-j but setting dataFolder to a path isn't going to observe the LocalTempStorageLocation setting which I think you should be able to do. Perhaps can be done in code...

p-m-j:
Ahh I see what you mean, sounds like a good idea, also useful for working towards supporting [WEBSITE_RUN_FROM_PACKAGE ](https://github.com/umbraco/Umbraco-CMS/issues/12043)

Jeavon:
> Ahh I see what you mean, sounds like a good idea, also useful for working towards supporting [WEBSITE_RUN_FROM_PACKAGE ](https://github.com/umbraco/Umbraco-CMS/issues/12043)

Very true! I had a quick try at setting it in code but couldn't get it swapped in time so I think should be added into Umbraco

Reference/V9-Config/RuntimeMinificationSettings/index.md
Jeavon: > However I do wonder if we should change the value to be {{hash}}.{{configured explicit version value}} so the value from config shows in the URL to avoid confusion.

I think this makes sense!

marcemarc: > However I do wonder if we should change the value to be {{hash}}.{{configured explicit version value}} so the value from config shows in the URL to avoid confusion.

Yes, great observation! - it's the 'setting' of an explicit version in the appsettings configuration that then doesn't translate to what's seen on the asset URLs, that causes confusion!

Reference/V9-Config/RuntimeMinificationSettings/index.md
```diff
@@ -21,39 +21,43 @@ This section allows you to configure the runtime minifications (defaults shown),
...
+      "UseInMemoryCache": true,
       "CacheBuster": "Version",
       "Version": "1234"
     }
   }
 }
 ```
+The actual 'Version' number will not be visible in the url of the assets, this is because it is combined, along with the Umbraco Version from configuration and the your project assembly dll, and then once combined a 'hash' is generated to obscure these details.
+
+in the HTML link thus: ```<link href='/sb/umbraco-backoffice-init-css.css.v7a71f91360259c5f7c3337f152b0df01eeee36f0' rel='stylesheet' type='text/css'/>```
```
Jeavon, line 55:
```suggestion
in the HTML link thus: ```<link href='/sb/umbraco-backoffice-init-css.css.v7a71f91360259c5f7c3337f152b0df01eeee36f0' rel='stylesheet' type='text/css'/>``` (when [Umbraco:CMS:Hosting:Debug:false](../HostingSettings/index.md))
```

sofietoft:
Hi @Jeavon and @marcemarc ..

What's the status on this PR?
Ready to merge, or are there still some pending questions/comments, etc?

Jeavon (approved):

PR #3895, opened 2022-04-04
Adds another e-mail service

aguyfromdenmark:
Adds Rapidmail, a 100% GDPR compliant service based in the EU. 
No affiliation, just this it's a good service to mention, since Umbraco is used a lot in the EU, and I don't really know any alternatives to it. 
I have tested the SMTP side of things out on Umbraco Cloud, and it works as expected.


eshanrnh (approved):
Looks good. Merging it in 💪

PR #3898, opened 2022-04-05
Update external services with port 25 specifics

sajumb:
It has now been confirmed by Microsoft that port 25 will continue to be blocked in our static outbound ip setup.


eshanrnh:
Thanks, @sajumb for the PR.

Looks good. Merging it in 💪

PR #3921, opened 2022-04-13
Removed tip as links are already removed

busrasengul:


eshanrnh:
Thanks for the PR, @busrasengul 💪

Looks good. Merging it in 😀

PR #3939, opened 2022-04-26
Eye Dropper Color Picker value type in v9

bjarnef:
In v9 we updated the value type of Eye Dropper Color Picker to return a `string` instead of a generic `object`, which is also the default generated value type using in ModelsBuilder if no value converter is found.
https://github.com/umbraco/Umbraco-CMS/pull/10853


eshanrnh (approved):
Looks good, @bjarnef 💪 Merging it in 🙂

PR #3960, opened 2022-05-11
Conditions for Date field

eshanrnh:
1. Added Conditions for Dates section in Conditional Logic article. 
2. Noticed "Preparing your Frontend" article in https://our.umbraco.com/documentation/Add-ons/UmbracoForms/Developer/ still links to version 7 of the article. Changed it to v9.


sofietoft (commented):
Looks great @eshanrnh ! 💪 

One quick note:
Since you've made two versions of the "Preparing the frontend" doc, does it make sense to still mention the version in the titles inside the document? Seems a bit redundant to me.

What do you think?

eshanrnh:
Oh yes, Good catch. I removed it from the v9 article however I left it for the v7 article since the dependencies are different for 8.6.0+ and below

PR #3963, opened 2022-05-11
Update to JSON Security Value for appSettings.json

OwainWilliams:
It seems the "umbracoSecurity" is now just "Security"


eshanrnh (approved):
Thanks for the PR, @OwainWilliams 💪

Merging it in 🙂

PR #3966, opened 2022-05-13
SmidgeHelper ignores Umbraco:CMS:Hosting:Debug

drpeck:
I think this edit makes it clear that SmidgeHelper needs some help in knowing whether to bundle or not.


sofietoft:
Hi @drpeck !

Thanks for the PR 💪 
All looks good!

We'll merge it right in.

PR #3972, opened 2022-05-19
Added a note in an attempt to avoid confusion

sofietoft:
When visiting this article right now, it looks like it should be valid for Umbraco 9 as well.

However, in the YAML tags is says `versionRemoved: 9.0.0`.
I've added a note to the article, to make it clear that this article is not valid for v9 or above.
Also added a `versionTo` tag, to make it a little clearer.

### DIFF
--- a/Reference/Querying/MemberShipHelper/index.md
+++ b/Reference/Querying/MemberShipHelper/index.md
@@ -1,7 +1,14 @@
 ---
 versionFrom: 8.0.0
+versionTo: 8.0.0
 versionRemoved: 9.0.0
 ---
+
+:::warning
+The topic covered in this article is only relevant to you if you are using Umbraco 8 or below.
+
+This article is not relevant to you if you are using Umbraco 9 or above.
+:::
 
 # Membershiphelper
 
--- a/Reference/Querying/MemberShipHelper/index.md
+++ b/Reference/Querying/MemberShipHelper/index.md
@@ -7,7 +7,7 @@ versionRemoved: 9.0.0
 :::warning
 The topic covered in this article is only relevant to you if you are using Umbraco 8 or below.
 
-This article is not relevant to you if you are using Umbraco 9 or above.
+If you are using Umbraco 9 or above, see the [IMemberManager](../IMemberManager/index.md) article.
 :::
 
 # Membershiphelper
--- a/Reference/Querying/MemberShipHelper/index.md
+++ b/Reference/Querying/MemberShipHelper/index.md
@@ -1,6 +1,6 @@
 ---
 versionFrom: 8.0.0
-versionTo: 8.0.0
+versionTo: 8.18.0
 versionRemoved: 9.0.0
 ---
 


sofietoft:
Another thing: We do we link to both v7 and v8 API docs from both versions of this article? 🤔 
That seems a bit reduntant.

Reference/Querying/MemberShipHelper/index.md
```diff
@@ -1,8 +1,15 @@
 ---
 versionFrom: 8.0.0
+versionTo: 8.0.0
```
eshanrnh, line 3: I am just wondering if we add this then the version box would show as 8.0.0 - 8.0.0 (current) which will make it question if the article is not valid for 8+versions.

Maybe add versionTo: 8.18.4?

eshanrnh, line 3: I am just wondering if we add this then the version box would show as 8.0.0 - 8.0.0 (current) which will make it question if the article is not valid for 8+versions.

Maybe add versionTo: 8.18.4?

Reference/Querying/MemberShipHelper/index.md
```diff
@@ -1,8 +1,15 @@
...
 versionRemoved: 9.0.0
 ---
 
+:::warning
+The topic covered in this article is only relevant to you if you are using Umbraco 8 or below.
+
+This article is not relevant to you if you are using Umbraco 9 or above.
```
eshanrnh, line 10:
```suggestion
If you are using Umbraco 9 or above, see the [IMemberManager](../IMemberManager/index.md) article.
```

eshanrnh, line 10:
```suggestion
If you are using Umbraco 9 or above, see the [IMemberManager](../IMemberManager/index.md) article.
```
```suggestion
This article is not relevant to you if you are using Umbraco 9 or above.
```

eshanrnh (commented):
Thanks for the Pr, @sofietoft.💪 I've added couple of suggestions. 🙂

eshanrnh:
> Another thing: We do we link to both v7 and v8 API docs from both versions of this article? 🤔 That seems a bit reduntant.

That's a good catch 🙂Possibly something we missed on earlier. We should link the relevant API doc from the article. Are you up for cleaning that bit?🙈

Reference/Querying/MemberShipHelper/index.md
sofietoft: That's also an option.
Then perhaps let's do 8.18.0 then?
That way, if we release patches then we won't have to go in an update this one every time? 🙈

eshanrnh: Yup, sounds good 💪

sofietoft:
> > Another thing: We do we link to both v7 and v8 API docs from both versions of this article? 🤔 That seems a bit reduntant.
> 
> That's a good catch 🙂Probably something we missed earlier. We should link the relevant API doc from the article. Are you up for cleaning that bit?🙈

I'll make a new PR with a link-cleanup 💪

eshanrnh (approved):
Looks good 💪 Merging it in.

PR #3989, opened 2022-05-25
Update version on deploy ci/cd doc

AndyButland:
This just updates the versioning on a documentation page for Deploy, that only applies to V8.  We'll need to create a V9 version.

### DIFF
--- a/Add-ons/Umbraco-Deploy/Installing-Deploy/CICD-Pipeline/ci-cd-azure-dev-ops.md
+++ b/Add-ons/Umbraco-Deploy/Installing-Deploy/CICD-Pipeline/ci-cd-azure-dev-ops.md
@@ -1,13 +1,13 @@
 ---
 versionFrom: 8.0.0
-versionTo: 9.0.0
+versionTo: 8.0.0
 meta.Title: "Setting up a CI/CD Build and Deployment Pipeline Using Azure DevOps"
 meta.Description: "Steps and examples on how to setup a build and deployment pipeline for Umbraco Deploy using Azure DevOps"
 ---
 
 # Setting up CI/CD build server with Azure DevOps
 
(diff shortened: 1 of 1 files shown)


Add-ons/Umbraco-Deploy/Installing-Deploy/CICD-Pipeline/ci-cd-azure-dev-ops.md
```diff
@@ -1,13 +1,13 @@
 ---
 versionFrom: 8.0.0
-versionTo: 9.0.0
+versionTo: 8.0.0
 meta.Title: "Setting up a CI/CD Build and Deployment Pipeline Using Azure DevOps"
 meta.Description: "Steps and examples on how to setup a build and deployment pipeline for Umbraco Deploy using Azure DevOps"
 ---
 
 # Setting up CI/CD build server with Azure DevOps
```
sofietoft, line 9: How about adding a note here at the top as well, to ensure people see it before they dive into the article?

```suggestion

:::note
Please be aware that this document is **only relevant to you, if you are using Umbraco 8**.
:::

```

AndyButland: I think that's perhaps a bit strong... the information is still useful to understand the steps but you can't take it verbatim for Umbraco 9,  I've changed the lead paragraph to reference Umbraco 8 so there's some indication, but given we don't currently have an Umbraco 9 version of these instructions, it's probably still of some use to people looking to setup a pipeline using that,

sofietoft: Yeah, true. And perhaps it is enough that is only says Umbraco 8.0.0 in the versioning box!

sofietoft (approved):

PR #4006, opened 2022-05-30
Verified Tracking article for v10

eshanrnh:
There is one difference between v10 and v9 when testing this article.
For v10, VS suggests adding a ? in the C# sample:
```
public bool IsForEditor(IDataEditor? dataEditor) => dataEditor?.Alias == Constants.PropertyEditors.Aliases.Label;

public IEnumerable<UmbracoEntityReference> GetReferences(object? value)
```


eshanrnh:
Added the code under two headings to accommodate the nullable reference check for v10

PR #4008, opened 2022-05-30
Update system requirements for 10.x

p-m-j:
Bump .Net requirements from net5.0 -> net6.0

The existing article can be modified slightly to fit both 9 and 10 however it doesn't look quite right in the same way that a code example change would.


eshanrnh (approved):
Looks good 💪 Merging it in 🙂

PR #4013, opened 2022-05-30
Add missing documentation about the UmbracoContext

dampee:
Not sure if I put in in the right folder. What do you think?

### DIFF
--- a/Reference/Querying/UmbracoContext/index.md
+++ b/Reference/Querying/UmbracoContext/index.md
@@ -67,4 +67,4 @@ public class PeopleController : UmbracoApiController
 }
 ```
 
-UmbracoContext is registered with a scoped lifetime (see [Microsoft documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#lifetime-and-registration-options) for more information), as a service scope is created for each request you can resolve an instance directly in a controller.
+UmbracoContext is registered with a scoped lifetime. See the [Microsoft documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#lifetime-and-registration-options) for more information. A service scope is created for each request, which means you can resolve an instance directly in a controller.
--- a/Reference/Querying/UmbracoContext/index.md
+++ b/Reference/Querying/UmbracoContext/index.md
@@ -8,7 +8,7 @@ meta.Description: "The UmbracoContext is a helpful service provided on each requ
 
 The UmbracoContext is the simplified way to work with the current request on your website.
 
-You can use the UmbracoContext to access the content & media cache. Other useful properties are the the original & cleaned url's of the current request. And last, but not least, you can also check if the current request is running in "preview" mode.
+You can use UmbracoContext to access the content and media cache. Other useful properties are the original and cleaned URLs of the current request. You can also check if the current request is running in "preview" mode.
 
 ## How to reference UmbracoContext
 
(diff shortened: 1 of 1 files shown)


Reference/Querying/UmbracoContext/index.md
```diff
@@ -0,0 +1,70 @@
...
+versionFrom: 9.0.0
+meta.Title: "Umbraco Request Context"
+meta.Description: "The UmbracoContext is a helpful service provided on each request to the website"
+---
+
+# UmbracoContext helper
+
+The UmbracoContext is the simplified way to work with the current request on your website.
+
+You can use the UmbracoContext to access the content & media cache. Other useful properties are the the original & cleaned url's of the current request. And last, but not least, you can also check if the current request is running in "preview" mode.
```
sofietoft, line 11:
```suggestion
You can use UmbracoContext to access the content and media cache. Other useful properties are the original and cleaned URLs of the current request. You can also check if the current request is running in "preview" mode.
```

Reference/Querying/UmbracoContext/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+## How to reference UmbracoContext
+
+If you are using Views or Partial View Macros you can reference UmbracoHelper with the syntax: `@UmbracoContext`
```
sofietoft, line 15: Shouldn't `UmbracoHelper` here be `UmbracoContext` instead? 🤔 

```suggestion
If you are using Views or Partial View Macros you can reference the UmbracoContext with the syntax: `@UmbracoContext`
```

Reference/Querying/UmbracoContext/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+If you need an `UmbracoContext` in your own controllers, you need to inject an `IUmbracoContextAccessor`.
+
+Example of getting `UmbracoContext` in a controller:
```
sofietoft, line 19:
```suggestion
The following is an example of how to get access to the `UmbracoContext` in a controller:
```

Reference/Querying/UmbracoContext/index.md
```diff
@@ -0,0 +1,70 @@
...
+                .Children<Person>() ?? Array.Empty<Person>())
+            .Select(p => p.Name);
+        return personContentType == null
+            ? this.Problem("Person Content Type is null")
+            : Ok(personNodes);
+    }
+}
+```
+
+UmbracoContext is registered with a scoped lifetime (see [Microsoft documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#lifetime-and-registration-options) for more information), as a service scope is created for each request you can resolve an instance directly in a controller.
```
sofietoft, line 70:
```suggestion
UmbracoContext is registered with a scoped lifetime. See the [Microsoft documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#lifetime-and-registration-options) for more information. A service scope is created for each request, which means you can resolve an instance directly in a controller.
```

sofietoft (changes requested):
Looks great @dampee !

And I think the placement of the doc is correct. Would it make sense to add a link to it from here as well: https://our.umbraco.com/documentation/Reference/Querying/ ?

I made a few suggestions as well.
Let me know if something doesn't make sense!

dampee:
Accepted all your changes. They made sense indeed. Thanks for reviewing!

> And I think the placement of the doc is correct. Would it make sense to add a link to it from here as well: [https://our.umbraco.com](https://our.umbraco.com/documentation/Reference/Querying/)[/documentation/Reference/Querying/](https://our.umbraco.com/documentation/Reference/Querying/) ?

To me that makes sense indeed as you would use it in most lot of cases to query for content.

PR #4015, opened 2022-05-30
Document approaches to migrate from SQL CE

p-m-j:
Couple of questions here.

1) I'm not sure if this advice belongs on this page, on `general.md`, or on both.
2) @nul800sebastiaan you have a recommended tool for SQL CE -> SQL Server, is that the one used in the article https://www.saotn.org/convert-sqlce-database-to-sql-server/


sofietoft (approved):
Looks great @p-m-j !

I think the information is placed where it should. This is exactly the kind of information this "Version specific" article is for.

You mention uSync - will it also be possible to migrate using Umbraco Deploy? Because I think that should be one of our recommended approaches then 😁

p-m-j:
Oops yes will do shortly

p-m-j:
> Are you still waiting for a response from Seb, or are we good to merge this one?

Seb has used this tool with success so I think good 2 go

PR #4017, opened 2022-05-31
Updated 2FA article for v10

eshanrnh:
The only difference between v10 and v9 in `UmbracoUserAppAuthenticator.cs` code:

```
public string? QrCodeSetupImageUrl { get; set; }
public string? Secret { get; set; }
```


eshanrnh:
Added the code under 2 headings to accommodate the nullable reference check for v10.

PR #4031, opened 2022-06-01
Updated "Starter Kit" tutorial for v10

eshanrnh:
- [x] Step 3 of the Open Graph lesson does not render the composition as expected.


sofietoft (approved):
Looks great 💪 

Regarding the Open Graph lessons - let's leave that for now. It only covers 0.1% of our yearly views.
We need some developer-help in order to figure it out.

I will add a note to all the pages of the Open Graph lesson, on how it's not updated for the latest version of Umbraco.

PR #4041, opened 2022-06-02
Added note about "deprecating" the Models docs.

sofietoft:


Reference/Management/Models/index.md
```diff
@@ -5,6 +5,12 @@ versionTo: 9.0.0
 
 # Persistence Models reference
 
+:::note
+As of the release of Umbraco 10 we will no longer be updating the articles in this section.
```
eshanrnh, line 9:
```suggestion
As of Umbraco 10 release, we will no longer be updating the articles in this section.
```

eshanrnh (commented):
Just added a small suggestion 😃

eshanrnh (approved):

Reference/Management/Models/index.md
sofietoft: Perhaps instead we could do: 
Since the release of Umbraco 10, we will no longer ... 

Would that be better?

eshanrnh: Yes that sounds better 🙌

eshanrnh (approved):

PR #4054, opened 2022-06-02
Updated backoffice user manager docs

bergmania:
The page is still valid for v9 and since 9.5 it make sense to change `true` to `false ` in the example.


eshanrnh (approved):
Looks great 👏Merging it in 😃

PR #4057, opened 2022-06-03
Verified Managing Hostnames article for v10

eshanrnh:


Umbraco-Cloud/Set-Up/Manage-Hostnames/Rewrites-on-Cloud/index.md
```diff
@@ -8,12 +9,11 @@ To make rewrite rules on Umbraco Cloud as seamless as possible, we've installed
...
 The rewrite rules should be added to the `<system.webServer><rewrite>` module in your projects `Web.config` file.
 
-```xml
-<!--
-If you wish to use IIS rewrite rules, see the documentation here:
-https://our.umbraco.com/documentation/Reference/Routing/IISRewriteRules
--->
+:::note
+If you are running Umbraco 9 and above on IIS, you can still add a `web.config` file to configure IIS features such as URL rewrites. If you wish to use IIS rewrite rules, see the [URL Rewrites in Umbraco 9](https://our.umbraco.com/documentation/Reference/Routing/IISRewriteRules) article.
+:::
```
sofietoft, line 14: I think we need to confirm with the Cloud team, whether we recommend people adding a `web.config` file, or if we recommend using the [URL Rewriting Middleware in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/url-rewriting?view=aspnetcore-5.0), as mentioned in [the article for the CMS](https://our.umbraco.com/documentation/Reference/Routing/IISRewriteRules/).

sofietoft (commented):
All good.
Except the part about the rewrites. We need to figure out what we should recommend on Cloud first.

Umbraco-Cloud/Set-Up/Manage-Hostnames/Rewrites-on-Cloud/index.md
eshanrnh: Okay, Before I added the note, the second para in the article mentions _The rewrite rules should be added to the <system.webServer><rewrite> module in your projects Web.config file_. and the link to the URL Rewrites in Umbraco 9 (CMS) article in the code sample. 

So I tried to make it simple by adding the same note from the CMS article and moving the link to the CMS article from the code snippet below to the note.

Since it's already mentioned in the article do we still need to confirm?

sofietoft: It's what we recommend for the CMS, but it might not be what we recommend for Cloud, right?

It think it would be good to have it confirmed by the Cloud team! Just in case.

sofietoft:
Let's merge this one in.

I'll add a task to our Trello boards, for having this double-checked with the Cloud team.

Issue #4039, opened 2022-06-02
v10: Verify Key Vault docs


sofietoft:
Hi @erikjanwestendorp !😄 

Was wondering if you could find some time to help us update this article for Umbraco 10? 😇 
You wrote it, so thought I'd try asking you, before we dive into it ourselves.

Let me know!

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Extending/Key-Vault/index.md***

sofietoft:
That's just wonderful, thanks 🙌 

No worries about getting to it today - just as soon as you have some spare time! 😄

erikjanwestendorp:
@sofietoft The Program.cs class looks slightly different in v10. I have created a PR with a v10 version of the article.

PR #4064, opened 2022-06-03
Update KeyVault article for v10

erikjanwestendorp:
Belongs to: https://github.com/umbraco/UmbracoDocs/issues/4039

### DIFF
--- a/Extending/Key-Vault/index-v10.md
+++ b/Extending/Key-Vault/index-v10.md
@@ -1,0 +1,74 @@
+---
+versionFrom: 10.0.0
+meta.Title: "Azure Key Vault"
+meta.Description: "A guide configuring an Azure Key Vault"
+---
+
+# Configuring an Azure Key Vault
+
+From a security perspective, it's always a good solution to store your application secrets (such as a connection string and others keys) in an Azure Key Vault. This article tells you how to configure your application so that it's ready to use a Key Vault
+
+## Installing the package
+
+Before you begin, you need to install the `Azure.Extensions.AspNetCore.Configuration.Secrets` and the `Azure.Identity` NuGet packages. There are two approaches to installing the packages:
+
+1. Use your favorite IDE and open up the NuGet Package Manager to search and install the packages
+1. Use the command line to install the package
+
+## Installing through command line
+
+Navigate to your project folder, which is the folder that contains your .csproj file. Now use the following 'dotnet add package' command to install the package:
+
+```
+dotnet add package Azure.Extensions.AspNetCore.Configuration.Secrets
+dotnet add package Azure.Identity
+```
+
+## Configuration
+The next step is to add the Azure Key Vault endpoint to the 'appsettings.json' file. 
+
+```json
+{
+  "AzureKeyVaultEndpoint": "https://{your-key-vault-name}.vault.azure.net",
+}
+```
+
+After adding the Key Vault endpoint you have to update the `CreateHostBuilder` method which you can find in the `Program.cs` class. 
+
+```csharp
+public static IHostBuilder CreateHostBuilder(string[] args) =>
+    Host.CreateDefaultBuilder(args)
+        .ConfigureLogging(x => x.ClearProviders())
+        .ConfigureAppConfiguration((context, config) =>
+        {
+            var settings = config.Build();
+            var keyVaultEndpoint = settings[Constants.Azure.AzureKeyVaultEndpoint];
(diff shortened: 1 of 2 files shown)


eshanrnh:
Thanks for the PR, @erikjanwestendorp 👏 We will review this soon 🙂

Extending/Key-Vault/index-v10.md
```diff
@@ -0,0 +1,74 @@
...
+From a security perspective, it's always a good solution to store your application secrets (such as a connection string and others keys) in an Azure Key Vault. This article tells you how to configure your application so that it's ready to use a Key Vault
+
+## Installing the package
+
+Before you begin, you need to install the `Azure.Extensions.AspNetCore.Configuration.Secrets` and the `Azure.Identity` NuGet packages. There are two approaches to installing the packages:
+
+1. Use your favorite IDE and open up the NuGet Package Manager to search and install the packages
+1. Use the command line to install the package
+
+## Installing through command line
```
sofietoft, line 18: Looks to me like a subheading to the "Installing the package" section? I could be wrong though. What do you think?

```suggestion
### Installing through command line
```

Extending/Key-Vault/index-v10.md
```diff
@@ -0,0 +1,74 @@
...
+
+Navigate to your project folder, which is the folder that contains your .csproj file. Now use the following 'dotnet add package' command to install the package:
```
sofietoft, line 20:
```suggestion
Navigate to your project folder, which is the folder that contains your `.csproj` file. Now use the following `dotnet add package` command to install the package:
```

Extending/Key-Vault/index-v10.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+dotnet add package Azure.Extensions.AspNetCore.Configuration.Secrets
+dotnet add package Azure.Identity
+```
+
+## Configuration
+The next step is to add the Azure Key Vault endpoint to the 'appsettings.json' file.
```
sofietoft, line 28:
```suggestion
The next step is to add the Azure Key Vault endpoint to the `appsettings.json` file. 
```

Extending/Key-Vault/index-v10.md
```diff
@@ -0,0 +1,74 @@
...
+## Configuration
+The next step is to add the Azure Key Vault endpoint to the 'appsettings.json' file. 
+
+```json
+{
+  "AzureKeyVaultEndpoint": "https://{your-key-vault-name}.vault.azure.net",
+}
+```
+
+After adding the Key Vault endpoint you have to update the `CreateHostBuilder` method which you can find in the `Program.cs` class.
```
sofietoft, line 36: I realize that it is a "class", but I think in the context here, it makes more sense to use "file", as it's a file you need to open and edit, right? 🤔 

```suggestion
After adding the Key Vault endpoint you have to update the `CreateHostBuilder` method which you can find in the `Program.cs` file. 
```

Extending/Key-Vault/index-v10.md
```diff
@@ -0,0 +1,74 @@
...
+            {
+                config.AddAzureKeyVault(validUri, new DefaultAzureCredential());
+            }
+        })
+        .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>());
+```
+
+## Authentication 
+
+There are several ways to access the Azure Key Vault. It is important that the user you are logging in with has access to the Key Vault. You can assign roles using the Azure Portal.
```
sofietoft, line 56:
```suggestion
There are different ways to access the Azure Key Vault. It is important that the user you are logging in with has access to the Key Vault. You can assign roles using the Azure Portal. 
```

sofietoft (changes requested):
Thanks for verifying this for v10 👏 

I've made a few suggestions and added some comments.
Hope it all makes sense 🤞

erikjanwestendorp:
@sofietoft Thanks for you review! 😄. 

I just updated the PR with the suggestions. I assume these changes should also be made for the v9 article?

sofietoft:
Yeah, that's a good point! 😅 
If the text is a one-to-one, then they should also be made on the other version.

You up for it? 😄 
Perhaps even add it to this one, so we can "smack two flies with one smack" (is that a danish saying? 🤔 😂 )!

erikjanwestendorp:
@sofietoft Yes the text is one-to-one there is only a small difference in the code.  Just updated the v9 article. 👍. 

And "smack two flies with one smack" is a dutch saying also! 🤣

sofietoft (approved):
This is great @erikjanwestendorp ! 🙌 

Thanks again for updating the articles.
I'll get merge it right in!

Hope to see you next week for Codegarden 🍻 🎉

PR #4067, opened 2022-06-07
Updated Reference Routing Authorized to v10

bergmania:
Opdated this page to v10. The only changes between v9 and v10 is the nullable reference types, so I still thing it is better to share the page for both v9 and v10, compared to copy paste every thing.


eshanrnh (approved):
Looks great, @bergmania 💪 Merging it in.

PR #4070, opened 2022-06-07
Updated the install options for Umbraco 10

jonat123:
The install articles has been updated for V10.

We should wait with merging it in till the release date :)


sofietoft (commented):
Looks good 💪 

What are the reasons for not merging this in now?
Apart from the file names.

PR #4077, opened 2022-06-08
v10: Update articles with references to embedded resources

sofietoft:
Updated the Language Files article.

* [ ] Where do the `[LANG].user.xml` files need to be added in v10? (lines 65-81)

Updated the Backoffice Tours articles

* [ ] Where do the custom tour files need to be placed in v10? (lines 16-22)

Other related articles:
[Backoffice - Login](https://our.umbraco.com/documentation/Fundamentals/Backoffice/Login/)
[Custom Error Pages](https://our.umbraco.com/documentation/Tutorials/Custom-Error-Pages/#errors-with-booting-a-project)

### DIFF
--- a/Extending/Backoffice-Tours/index.md
+++ b/Extending/Backoffice-Tours/index.md
@@ -1,5 +1,6 @@
 ---
 versionFrom: 9.0.0
+versionTo: 10.0.0
 meta.Title: "Backoffice Tours"
 meta.Description: "A guide configuring backoffice tours in Umbraco"
 ---
@@ -16,7 +17,9 @@ The tour functionality will load information from multiple locations.
 
 - **Core tours and custom tours**
 
-    The tour files that ship with Umbraco are stored in `/Config/BackofficeTours`. This is also the recommended location for storing your own tour files.
+    The tour that ship with Umbraco are embedded into the CMS.
+
+    When you want to implement custom tours into your website, we recommend that you create
 
 - **Plugin tours**
 
--- a/Extending/Backoffice-Tours/index.md
+++ b/Extending/Backoffice-Tours/index.md
@@ -17,9 +17,7 @@ The tour functionality will load information from multiple locations.
 
 - **Core tours and custom tours**
 
-    The tour that ship with Umbraco are embedded into the CMS.
-
-    When you want to implement custom tours into your website, we recommend that you create
+    The tour that ship with Umbraco are embedded into the CMS assemblies.
 
 - **Plugin tours**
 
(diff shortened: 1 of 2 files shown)


Extending/Language-Files/index.md
```diff
@@ -60,17 +65,19 @@ The `App_Plugins` version of the `Lang` directory is case sensitive on Linux sys
...
+```xml
+/config/lang/{language}.user.xml
+```
 
 By default, these files are empty but you can add any new keys you want or override existing ones with your own translations. The nice part about the user files is that they will not get overwritten by the installer when you upgrade your Umbraco versions.
 
 In order for these files to deploy when you do a `dotnet publish`, you need to add the following to your `.csproj` file:
 
-```xml
+```html
```
dampee, line 76: I think this needs to be XML instead of html.

```suggestion
```xml
```

Extending/Backoffice-Tours/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 versionFrom: 9.0.0
+versionTo: 10.0.0
```
dampee, line 3: I guess this needs to be removed again if we update this article to make it work for 10.
As I guess the backoffice tours are not meant to leave?

dampee (changes requested):

dampee:
As for your question "Where do the custom tour files need to be placed in v10?" it is in the plugin folder. :-)

bergmania:
> Where do the custom tour files need to be placed in v10? (lines 16-22)

This is unchanged between v9 and v10, but for both, it is not really supported. You will need to add them like plugin tours in the app_plugins folder.


>Where do the [LANG].user.xml files need to be added in v10? (lines 65-81)

Same location as in 8 and 9. Still `/config/lang/{language}.user.xml`

Extending/Backoffice-Tours/index.md
```diff
@@ -16,7 +17,9 @@ The tour functionality will load information from multiple locations.
 
 - **Core tours and custom tours**
 
-    The tour files that ship with Umbraco are stored in `/Config/BackofficeTours`. This is also the recommended location for storing your own tour files.
+    The tour that ship with Umbraco are embedded into the CMS.
+
+    When you want to implement custom tours into your website, we recommend that you create
```
bergmania, line 22:
```suggestion
    The tour that ship with Umbraco are embedded into the CMS assemblies.
```

It's not supported. People needs to do like for plugins

bergmania (approved):

Extending/Backoffice-Tours/index.md
sofietoft: I will merge this PR in now as is.

I have a task on my board, on running some more tests with these tours - in relation to the tests we ran at our previous meeting.

PR #4086, opened 2022-06-08
v10: Verified User Picker article for v10

sofietoft:
The samples didn't work for v9 either, so I updated the article to cover both.


eshanrnh (approved):
Looks great 👏

PR #4089, opened 2022-06-09
Cleaned up product dependencies on cloud

jonat123:
Removed V7 and courier and added v10


sofietoft (approved):
Looks great 💪 
About time we got Courier out of there 🙈

PR #4092, opened 2022-06-09
Remove forms folder added

thejokerd3:


sofietoft:
Thanks for the PR @thejokerd3 !

I'm not entirely sure I understand the change you've made.
Would you mind elaborating a bit?

eshanrnh:
@sofietoft, as per [slack conversation](https://umbraco.slack.com/archives/CSQMK997E/p1654771901927539), Muslim has created the PR. 😀

sofietoft:
Ah! Makes sense then 💪 

@thejokerd3 - for future reference, please make sure you add a few words to your PRs, so it's easy for everyone involved to understand why the change is being made! 😄 

Seems that deleting this Forms folder is only relevant if you're actually using Umbraco Forms on your website, right?
I think that should be mentioned in this article as well then.

PR #4114, opened 2022-06-18
Updated to use correct usings in v10

Ambertvu:
Not sure about the usage for the versionFrom and versionTo fields. Couldn't really find anything on it, or not searching correctly :-)

Added some V10 versions of the files as in V10 Scoping is moved to Infrastructure


eshanrnh:
Thanks for the PR, @Ambertvu 💪
We will review this ASAP 😃

sofietoft (commented):
Thanks again for the PR @Ambertvu !

I just made a few suggestions regarding the versioning.
It's not super transparent how it works, but in general when the article covers the latest version of Umbraco, we keep it "open-ended" - so no need to add eg. versionTo: 11.0.0 just yet 😄 

As for the rest of the PR, I'll make sure we give it a proper review as soon as possible!

Ambertvu:
@sofietoft Thanks for your suggestions, I'll keep it in mind in the future!

bergmania:
Instead of using `IScopeProvider` from the `Infrastructure` namespace I suggest using `ICoreScopeProvider` from the `Core` namespace.  The difference is whether you need direct access to the DB or not. This should never (In a perfect world) be the case from services or other business logic.

Ambertvu:
@bergmania Thanks for the comments, learning something new everyday ;-)
I used my own project as reference, but I was adding items to the database with direct queries, so hence I used the `IScopeProvider` 
Updated the files.

sofietoft:
Thanks for amending the files here @Ambertvu ! 💪 

I noticed that a few places in the 10 version of the Database article, it mentions Umbraco 9 and 8, and a few notes about when migrating from 8 to 9.
Those should be removed from the new article, right? 🤔

Ambertvu:
@sofietoft Silly me, I only looked at the code changes itself instead of the actual information... Thanks for the sharp eye!

I'm about to do a new commit that removes the first section (IComposer) entirely, as the new approach is favored (please correct me if i'm wrong about this). Also did some more cleanup.

Ambertvu:
@sofietoft I think some things got overwritten, not entirely sure, I'll thoroughly dive into this tonight!

sofietoft:
Ah shoot! Sorry about that.
Thanks for looking into it 🙌

Ambertvu:
@sofietoft Not sure if it was the merge, or if I overlooked a few things (not a github master), but I think all is fixed now in my last commit. Also rewrote a few small parts because of the Vale Linter warnings. Cheers!

sofietoft:
Fantastic @Ambertvu !
Thanks for all the work you've put in here 👏 It's much appreciated 🙏 

Let's get this merged in then 😄 🎉

Issue #3994, opened 2022-05-25
External login providers - Microsoft example


erikjanwestendorp:
Is it an idea to extend the article about [external login providers](https://our.umbraco.com/documentation/reference/security/external-login-providers/) with code examples of a Microsoft implementation?

erikjanwestendorp:
@eshanrnh Yes of course I want to give it a try. 😄. 

Would it be a good idea to move the article to a folder so that it appears in the menu?

eshanrnh:
That's amazing, @erikjanwestendorp 💪

You can definitely move the article to a folder. It would be good to have it listed in the menu. 😀

PR #4120, opened 2022-06-20
Authenticate with active directory 9 and 10

erikjanwestendorp:
Added code samples for how to configure AD sign in. 

Fixes: https://github.com/umbraco/UmbracoDocs/issues/3994


eshanrnh:
Thanks for the PR, @erikjanwestendorp 💪 We will review this soon 😀

eshanrnh:
D-Team will review it in the next sprint.

Reference/Security/Authenticate-with-Active-Directory/index-vrpe9.md
```diff
@@ -0,0 +1,88 @@
...
+
+Finally, to use your `UmbracoStandardOwinStartup` class during startup, update this setting to Web.config:
+
+```xml
+<appSettings>
+    <add key="owin:appStartup" value="MyOwinStartup" />
+</appSettings>
+```
+
+If the active directory setup uses usernames instead of emails for authentication this will need configuring against the Umbraco user. This can be done in Umbraco backoffice under a specific user in user management by setting the name and Username to be the active directory username. Making Username visible for editing requires `usernameIsEmail` in umbracoSettings.config to be set to false:
```
sophie-neale, line 80:
```suggestion
If the active directory setup uses usernames instead of emails for authentication this will need configuring against the Umbraco user. This can be done in Umbraco backoffice under a specific user in user management by setting the name and username to be the active directory username. Making username visible for editing requires `usernameIsEmail` in umbracoSettings.config to be set to false:
```

Reference/Security/Authenticate-with-Active-Directory/index-vrpe9.md
```diff
@@ -0,0 +1,88 @@
...
+
+```xml
+<usernameIsEmail>false</usernameIsEmail>
+```
+
+:::note
+if the username entered in the login screen does not already exist in Umbraco then `ActiveDirectoryBackOfficeUserPasswordChecker()` does not run.  Umbraco will fall back to the default authentication.
```
sophie-neale, line 87:
```suggestion
If the username entered in the login screen does not already exist in Umbraco then `ActiveDirectoryBackOfficeUserPasswordChecker()` does not run.  Umbraco will fall back to the default authentication.
```

sophie-neale (commented):
Hey @erikjanwestendorp! Thanks so much for adding this! When I meet with the Docs Curator team on Monday we'll go through it 👍🏼 but for now, I've added some very small suggestions 🙂

erikjanwestendorp:
Hi @sophie-neale Thanks for your your review 😄. I just updated the PR with your suggestions.

bergmania:
Hi @erikjanwestendorp .. Thanks for the update. 

Overall it looks correct, but it is also very similar to the current docs for [External login providers](https://our.umbraco.com/Documentation/Reference/Security/External-login-providers/).

I'm not sure if it brings value to showcase explicitly for `Microsoft.AspNetCore.Authentication.MicrosoftAccount` like this PR does. - I'll let the Docs Curator Team decide, but from a technical perspective this PR is correct :)

erikjanwestendorp:
Hi @bergmania Thanks for your review. 😄. 

I get your point, it is very similar to [External login providers](https://our.umbraco.com/Documentation/Reference/Security/External-login-providers/) indeed. The initial idea was to just add a small note that's explains the differences between a Google and a Microsoft implementation. 

@eshanrnh You suggested to update [Authenticating on the Umbraco backoffice with Active Directory credentials](https://our.umbraco.com/documentation/Reference/Security/Authenticate-with-Active-Directory/) for v9+. Is this still the idea?

eshanrnh:
Yes that's right, we definitely wanted to update this article to the latest versions. It also helps answer the queries we receive on the [Umbraco Learning Base channel](https://www.youtube.com/c/UmbracoLearningBase/videos) 🙂

Yilmas:
@erikjanwestendorp Nothing major, but there is a "}" too many in `{your_client_secret}}`

erikjanwestendorp:
@Yilmas Great Good point, just updated the PR. 😄

sofietoft:
Hi all 👋 

Looks like we're almost ready to merge this one in 🙌 
One little thing though:
The file name of the v8 article is `index-vrpe9.md` - would you mind updating that to `index-v8.md` instead?
And if you're up for a little more, perhaps also help us get rid of the `index-v800.md` file, which covers only 3 patch versions of Umbraco 8, and that does not really seem worth it.

Finally, we should add a link to this article from the "External Login Providers" article.
Would you be up for doing that?
Otherwise, we'll add it to our backlog 💪 Let me know!

sofietoft:
Awesome, thanks @erikjanwestendorp ! Looks great 🙌 
And thanks for helping out with the other article 💪 

Looks like we're ready to get this one merged in! 😄 
I'll do that right away!

PR #4126, opened 2022-06-22
Update blob version number in Cloud 9 to 10 guide

skttl:
When I build I get the message that Umbraco.Cloud.StorageProviders.AzureBlob version 5.0.0 doesn't exist, and it will use 10.0.0 instead.


eshanrnh (approved):
Looks great, @skttl 🙌 Merging it in 🙂

PR #4129, opened 2022-06-22
Updated the partial views with the new method signature

jonat123:
for v9 and 10 and moved the note a bit down to reflect that it is only the three code examples that has not been verified yet.


eshanrnh (approved):
Looks good, @jonat123 💪 Merging it in 🙂

Issue #3879, opened 2022-03-29
Issues with Handler code and fixes


thenexus00:
Inside GetPostFieldValue
When this is called - var field = GetPostField(e, key);
If the field does not exist it can create and instances of an object error.

Checking if field is not null fixes the issue.

In the main validation if you do say..

 var msgField = GetPostFieldValue(e, "message");
                    if (String.IsNullOrEmpty(msgField) == false)
                    {

For example you can not avoid the error.
I would update the code example so GetPostFieldValue has the null check for field.

***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Add-ons/UmbracoForms/Developer/Extending/Adding-an-Event-Handler-v8.md***

sofietoft:
Hi @thenexus00 

Thanks for reporting this!
Seems like there should definitely be a null check in the sample 🙈 Great find!

Actually, it looks like it has been added in the [sample for Umbraco 9](https://our.umbraco.com/Documentation/Add-ons/UmbracoForms/Developer/Extending/Adding-an-Event-Handler), which is great!

Would you be up for adding the null check to the Umbraco 8 sample? 🤞 
Otherwise I'll add it to our to-do 💪 Let me know!

PR #4138, opened 2022-06-23
Add null check

erikjanwestendorp:
Add missing null check as suggested in #3879. 

Fixes: #3879


eshanrnh (approved):
Thanks for the PR, @erikjanwestendorp 🙌

As always looks great. Merging it in 😀

PR #4113, opened 2022-06-18
Create V10 version of Azure Blob Storage documentation

Ambertvu:
As section about 'UseAzureBlobMediaFileSystem()' is no longer needed in v10.

PR #4142, opened 2022-06-24
Changes for v10

ault:
Middleware is no longer necessary... see: https://github.com/umbraco/Umbraco.StorageProviders/releases

### DIFF
--- a/Extending/FileSystemProviders/Azure-Blob-Storage/index.md
+++ b/Extending/FileSystemProviders/Azure-Blob-Storage/index.md
@@ -82,7 +82,36 @@ Invoke the `.AddAzureBlobMediaFileSystem()` extention method in the `ConfigureSe
         }
 ```
 
+## Setting the middleware !!V9 Only!!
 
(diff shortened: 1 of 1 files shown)


eshanrnh:
Thanks for the PR, @ault 💪 We will review this soon 🙂

Extending/FileSystemProviders/Azure-Blob-Storage/index.md
```diff
@@ -82,32 +82,7 @@ Invoke the `.AddAzureBlobMediaFileSystem()` extention method in the `ConfigureSe
         }
 ```
 
-Next invoke `UseAzureBlobMediaFileSystem();` in the `.WithMiddleware` call, like so:
```
ronaldbarendse, line 85: This article targets both v9 and v10, so instead of removing all code, we should add a note saying this is not required in v10 anymore.

ronaldbarendse (changes requested):
The Umbraco.StorageProviders package for v10 indeed doesn't require custom middleware anymore, so we should either create a separate page for v9/v10 or add a note.

sofietoft:
Hi all 👋 

It seems that there was already an approved PR for updating this article: #4113.
As that PR adds a separate version for v10, we decided to go with that instead.

Thanks again for contributing and helping us keeping the docs up to date 💪

PR #4155, opened 2022-06-29
Update storage providers docs for v10

crgrieve:
Updating for v10 based on changes to readme it looks like this section isn't needed now: https://github.com/umbraco/Umbraco.StorageProviders#umbracostorageproviders

Do you also want me to create a v9 specific page for the v9 way to work with this? or a note on this same page?

Thanks,
Carole


sofietoft:
Hi @crgrieve  👋

I have just merged a PR for this article with similar changes: https://github.com/umbraco/UmbracoDocs/pull/4113.
As that PR adds a separate version for v10, we decided to go with that instead.

Thanks again for contributing and helping us keep the UmbracoDocs up to speed 😄 

Thanks again for contributing and helping us keeping the docs up to date 💪

PR #4161, opened 2022-06-30
Media Picker should be used instead of File Upload

sofietoft:
I've made a few different things in this PR:

- Verified and updated the C# code sample for the Media Picker. Replaced the existing with the one from the legacy Media Picker article, as that worked for this Media picker as well. Not sure why the other sample was added to the Media Picker article - but I might be missing something.
- Noted that we're not going to update the C# sample in the File Upload article as we want people to be using the Media Picker for uploading media and files to their projects instead.


eshanrnh (approved):
Looks great, @sofietoft 🙌

The Note (have not and will not) is very aggressive and funny though 😃

sofietoft:
Yes, you're right about that sentence being a bit too aggressive 🙈 
I've updated it now - let me know what you think!

eshanrnh:
Looks perfect 💪 Merging it in 😃

PR #4175, opened 2022-07-03
Configuring the route to not be considered as a client side request h…

marcemarc:
…as non c# code samples and so important bit ```<UmbracoRequestOptions>``` is stripped

Just adding the ```C# bit in to make it render the correct information!


eshanrnh (approved):
Looks good, @marcemarc 💪 Merging it in 😃

PR #4176, opened 2022-07-04
missing "public access" under the v9-10

alina-tincas:
the public access feature is missing under the page for the version 9-10 in comparison to v8 and below.

### DIFF
--- a/Umbraco-Cloud/Set-Up/Project-settings/index.md
+++ b/Umbraco-Cloud/Set-Up/Project-settings/index.md
@@ -35,6 +35,20 @@ Manage transport security settings for your project. You can configure certain t
 
 Used to upload and bind your own custom security certificate to your hostnames instead of using the TLS (HTTPS) certificates provided by the Umbraco Cloud service.
 
+## Public access
+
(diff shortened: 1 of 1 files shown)


jonat123:
@alina-tincas I have talked with Søren about the basic auth feature, which has now become available.

However, there have been some changes to how it works.

First, by default basic auth is disabled on all environments not enabled.

For trials, it is not enabled as well and it cannot be enabled.

Also, basic auth is not available on V9 projects.

And lastly, the allow list is not working at the moment, unfortunately...

If you want and would be up for it, could you update the PR with the above information? 😊

jonat123:
@alina-tincas If you saw my last message, just disregard it :) 

The plan from D-team is that they are gonna make sure that everything is working as described in your PR in their next cycle :)

Until then we will update the text with a minimal description, and once the feature is fully complete we can merge in your PR 😊

jonat123:
@alina-tincas Again, thanks for your PR, 
Unfortunately public access won't work quite the same as it used to do before hand, so the information you are adding back, is out-dated.

I will therefore close down this PR.

PR #4182, opened 2022-07-05
Add comments about changes regarding Umbraco Forms

bjarnef:
... when upgrading Umbraco 9 to 10.

I had fixed these locally, but these was nothing to commit, because this was included in gitignore in Umbraco 9. The gitignore at same directory as "src" folder.

So it Umbraco Forms views in "repository" folder on Development environment it still used `schema` and `Html.SetFormThemeCssFile(Model, "~/App_Plugins/UmbracoForms/Assets/themes/default/style.css");` which caused the build to fail.

chrome_kcMeITizRJ

```
##
## Umbraco CMS .NETCore
##

# Dont commit Umbraco TEMP folder containing Examine Indexes, NuCache etc
**/umbraco/Data/TEMP/

# Umbraco log files
**/umbraco/Logs/

# Umbraco backoffice language files
# Nuget package Umbraco.Cms.StaticAssets will copy them in during dotnet build
# Customize langguage files in /config/lang/{language}.user.xml
**/umbraco/config/lang/

# JSON Schema file for appsettings
# This is auto generated from the build
**/umbraco/config/appsettings-schema.json

# This is the no-nodes, installer & upgrader pages from Umbraco
# Nuget package Umbraco.Cms.StaticAssets will copy them in during dotnet build
**/umbraco/UmbracoWebsite/
**/umbraco/UmbracoInstall/
**/umbraco/UmbracoBackOffice/

# Comment out the line below if you wish to change or add any new templates to PartialView Macros
**/umbraco/PartialViewMacros/Templates

# Umbraco Static Assets of Backoffice
# Nuget package Umbraco.Cms.StaticAssets will copy them in during dotnet build
**/wwwroot/umbraco/assets/
**/wwwroot/umbraco/js/
**/wwwroot/umbraco/lib/
**/wwwroot/umbraco/views/

# Umbraco Media Files
# Media is transfered via Deploy as content
**/wwwroot/media

# Umbraco Models Builder out of date flag file
**/umbraco/models/ood.flag

# Umbraco HQ Package - UmbracoID Login
**/App_Plugins/UmbracoId/

# Umbraco HQ Package - Deploy and its marker files on disk
# Ensure you can ALWAYS commit the UDA files in umbraco/Deploy/Revision folder
**/App_Plugins/Deploy/
**/umbraco/Deploy/deploy*

# Umbraco HQ Package - Umbraco Forms
**/App_Plugins/UmbracoForms/
**/umbraco/Data/UmbracoForms/
**/Views/MacroPartials/InsertUmbracoFormWithTheme.cshtml
**/Views/MacroPartials/RenderUmbracoFormScripts.cshtml
**/Views/Partials/Forms/Emails/Example-Template.cshtml
**/Views/Partials/Forms/Export/excel.cshtml
**/Views/Partials/Forms/Themes/bootstrap3-horizontal/
**/Views/Partials/Forms/Themes/default/
```

I updated these manually in "respository" folder via Kudo. Afterward I made a small commit from local to development, which triggered a new build and then I got assemblies in the .net6.0 folder (which also cleared the existing .net5.0 folder) and shortly after Umbraco and Forms were upgraded unattended.

image

### DIFF
--- a/Umbraco-Cloud/Upgrades/Upgrading-from-9-to-10/index.md
+++ b/Umbraco-Cloud/Upgrades/Upgrading-from-9-to-10/index.md
@@ -146,6 +146,10 @@ Remove the following files and folders *manually* from your local project:
 * `/umbraco/config/lang`
 * `/App_Plugins/UmbracoForms`
 
+Furthermore make the following change in these two files in Umbraco Forms:
+* Change `schema` to `Schema` in `/Views/Partials/Forms/Export/excel.cshtml` (and any custom exports).
+* Change `Html.SetFormThemeCssFile(Model, "~/App_Plugins/UmbracoForms/Assets/themes/default/style.css")` to `Html.SetFormThemeCssFile("~/App_Plugins/UmbracoForms/Assets/themes/default/style.css")` in `/Views/Partials/Forms/Themes/default/Form.cshtml` (any custom themes overriding default form view).
+
 By default, Umbraco Deploy will create an SQLite database. If you want to re-use the existing LocalDB database, configure the [ConnectionStrings](https://our.umbraco.com/documentation/Add-ons/Umbraco-Deploy/Upgrades/version-specific#database-initialization) or use the [`PreferLocalDbConnectionString` setting](https://our.umbraco.com/documentation/Add-ons/Umbraco-Deploy/Deploy-Settings/#preferlocaldbconnectionstring). 
 
 Build and run your project locally to verify the Umbraco 10 upgrade.
@@ -157,8 +161,9 @@ Build and run your project locally to verify the Umbraco 10 upgrade.
 Once the Umbraco 10 project runs locally without any errors, the next step is to deploy and test on the Cloud Development environment.
 
 1. Remove the folders mentioned above on the **Development** environment using [KUDU](../../Set-Up/Power-Tools/index.md) from the `repository` and `wwwroot` folders.
-2. Push the changes to the **Development** environment. See the [Deploying from local to your environments](../../Deployment/Local-to-Cloud/index.md) article.
-3. Test **everything** on the **Development** environment.
+2. In `repository` folder also make the two changes regarding Umbraco Forms in `excel.cshtml` and `Form.cshtml` mentioned previously.
+3. Push the changes to the **Development** environment. See the [Deploying from local to your environments](../../Deployment/Local-to-Cloud/index.md) article.
+4. Test **everything** on the **Development** environment.
 
 We highly recommend that you go through everything on your Development environment. This can help you identify any potential errors after the upgrade, and ensure that you are not deploying any issues on to your Live environment.
 


bjarnef:
@AndyButland it seems these changes was made in Umbraco Forms v10 right?

AndyButland:
These changes were made in Forms 10 @bjarnef yes, but I'm not sure what you've documented here is quite correct.  These files are now distributed as part of a Razor class library, and hence the better approach is to remove these files rather than amend them (assuming you are using the default theme).  Then the versions from the RCL will be used.

We documented this [here](https://our.umbraco.com/documentation/Add-ons/UmbracoForms/Installation/Version-Specific#views-and-client-side-files), but not very clearly I realise as we've only talked about client-side files.  So I think it's probably better I make an update to that rather than including the changes you've provided here.

AndyButland:
Suggested updates are [here](https://github.com/umbraco/UmbracoDocs/pull/4183).

bjarnef:
@AndyButland okay, I have deleted the `App_Plugins/UmbracoForms` folder, but regarding these it was an issue with the views.

So we should basically remove the files in `Views/Partials/Forms` except if we have a custom theme with files, e.g. a "standard" theme. In that case this should probably not be ignores in gitignore file.

AndyButland:
Yes, that's right.  Which is what I've tried to make clear now in the PR I've linked above.

bjarnef:
Yes, thanks for make it clearer :)

bjarnef:
@AndyButland I removed the Forms macro partial views and see the default selected view.

But when I remove this and select a macro partial view, the overlay is empty.
Is that expected?

image

AndyButland:
No, that's a bug - this dialog isn't reading the files provided in the Razor class library.  I can fix it though via a notification handler, so will do that for the next Forms release.

eshanrnh:
Just checking in, is this PR still relevant?

bjarnef:
@eshanrnh the changes in the PR @AndyButland submitted did the trick, so we don't need the changes in this PR.
However the migration guide here https://our.umbraco.com/documentation/Umbraco-Cloud/Upgrades/Upgrading-from-9-to-10/ should link to the Umbraco Forms specific docs.

bjarnef:
@AndyButland maybe related to the bug you mentioned, but I also noticed when saving the macro "Insert Form with Theme" Deploy removes the dependency to the macro partial view:

image

AndyButland:
I hadn't seen that, but actually I think it's OK.  This view isn't found on disk, and so it's actually not something Umbraco Deploy needs to track to ensure that the dependency is available in the target environment.  It's coming from the RCL and hence will be available (at least as long as Forms is installed in the target environment).

bjarnef:
@AndyButland okay thanks. What if a custom macro partial view was chosen for this macro (or any other macro)? Would it then include the physical file as a dependency?

AndyButland:
Yes, that should happen.  It would find the file on disk and added it as a dependency.

PR #4185, opened 2022-07-06
Delete code block and show add package command inline.

erikjanwestendorp:
A code block in a note looks a bit weird (in my opinion), so maybe show the command inline?  Like `dotnet add package Umbraco.TheStarterKit`

image

### DIFF
--- a/Reference/Searching/Examine/Quick-Start/index.md
+++ b/Reference/Searching/Examine/Quick-Start/index.md
@@ -9,8 +9,11 @@ _This guide will help you get set up quickly using Examine with minimal configur
 
 ## Performing a search
 
-:::note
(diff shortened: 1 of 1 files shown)


eshanrnh (commented):
That's right, @erikjanwestendorp 🙌 Thanks for creating a PR to fix this. 

Can I suggest removing the Note tag as it doesn't add any specific informative value in this case? It's just a simple paragraph so we can remove the Note tag and rather use the code block to display the code? 

It might be easier to miss out on the code if it's not in a code block 😀

erikjanwestendorp:
Thanks @eshanrnh, the note doesn't add specific information indeed. Just updated the PR 😄.

eshanrnh (approved):
Looks great 🙌 Merging it in.

PR #4189, opened 2022-07-06
Added component registration

AaronSadlerUK:
Updated the documentation to show how to register the component with a composer

### DIFF
--- a/Fundamentals/Design/Stylesheets-Javascript/index.md
+++ b/Fundamentals/Design/Stylesheets-Javascript/index.md
@@ -118,6 +118,23 @@ namespace MyNamespace
 }
 ```
 
+You will then need to register your component:
+```csharp
+using Umbraco.Cms.Core.Composing;
+using Umbraco.Cms.Core.DependencyInjection;
+
+namespace MyNamespace
+{
+    public class MyComposer : IComposer
+    {
+        public void Compose(IUmbracoBuilder builder)
+        {
+            builder.AddComponent<MyComponent>();
+        }
+    }
+}
+```
+
 Then, you can render the bundles by the bundle name in a view template file:
 
 ```csharp


eshanrnh:
Thanks for the PR, @AaronSadlerUK 💪 We will review this soon.

Jeavon:
@AaronSadlerUK The existing sample contains a `ComponentComposer` so it shouldn't be necessary to register the component in a composer? 
It looks like this:
```
    public class MyComposer : ComponentComposer<MyComponent>
    { }
```

AaronSadlerUK:
@Jeavon I missed that line in the file... 

However isn't it good practice to have the composers separate to the component?

Jeavon:
@AaronSadlerUK ComponentComposer exists for this type of usage to avoid needing a composer class. However, in v9/10/+ Components and Composers are mainly supported for packages so for this sample it would be a good idea to move it to a notification handler instead? I have created the sample in our samples project https://github.com/umbraco/Umbraco.Docs.Samples/blob/develop/Umbraco.Docs.Samples.Web/Stylesheets-Javascript/CreateBundlesNotificationHandler.cs 
This handler would then be registered in startup.cs like this `.AddNotificationHandler<UmbracoApplicationStartingNotification, CreateBundlesNotificationHandler>()`
Additionally, we should note debug mode when rendering using the standard (tag) method in the same way as the `SmidgeHelper ` In the samples project I've injected the Umbraco Debug setting to both options, maybe update the docs to use that? https://github.com/umbraco/Umbraco.Docs.Samples/blob/develop/Umbraco.Docs.Samples.Web/Views/Master.cshtml#L39

Let me know if you fancy updating with these suggestions yourself otherwise I would be happy to collaborate on this PR

Jeavon:
This is superseded by https://github.com/umbraco/UmbracoDocs/pull/4382

PR #4190, opened 2022-07-07
Minor updates to the Packages docs

sofietoft:
- Added more details about package migrations to the "Creating a package" article.
- Updated version numbers on the UmbPack article.
- Added note to UmbPack article about how to use NuGet instead.

### DIFF
--- a/Extending/Packages/Creating-a-Package/index.md
+++ b/Extending/Packages/Creating-a-Package/index.md
@@ -220,6 +220,10 @@ namespace CustomWelcomeDashboardProject.Migrations
 
 ![Automatic package migration](images/embeded-resource.png)
 
+:::note
+Whenever the embedded package.xml file changes, the automatic package migration plan is executed again. This is due to the fact that the migration state is based on the file hash. Existing schema or content will not be overwritten in this process.
+:::
+
 #### Custom Package Migration
 
 Instead of creating an automatic package migration plan, we will inherit from the `PackageMigrationPlan` and again specify the name of the package in the base constructor. Further on, we will define the plan using a unique GUID - in the example below we have a single migration called `MyCustomMigration`.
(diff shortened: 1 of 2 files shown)


Extending/Packages/Creating-a-Package/index.md
```diff
@@ -296,6 +300,8 @@ Whichever migration plan you choose to create, you will be able to see that your
 
 ![Installed package](images/installed-package.png)
 
+When using a custom package migration plan, any changes to the plan will trigger all migrations again even those that have already run previously. This is due to the `IgnoreCurrentState` being set to `true` when changes are made to the plan. You can override this property and set it to `false` again to make it behave like regular migration plans and only run the migrations that have not yet been executed on the current environment.
```
ronaldbarendse, line 303:
```suggestion
When using a custom package migration plan, the current state is ignored by default, causing it to execute all migrations again whenever this isn't the same as the final state of the plan (e.g. if you added a new migration). This is due to the `IgnoreCurrentState` being set to `true` in the `PackageMigrationPlan` base class. You can override this property and set it to `false` again to make it behave like regular migration plans and only run the migrations that have not yet been executed on the current environment.
```

Extending/Packages/UmbPack/index.md
```diff
@@ -10,11 +11,17 @@ meta.Description: "How to use the UmbPack tool to deploy package versions to Our
 UmbPack is an open source NuGet tool that can deploy packages to Our, and also help you pack your files into a package.
 
 :::note
-You still need to create the package on Our before the tool can be used. Once it has been set up you will then be able to add new versions of the package zip files using UmbPack
+**Are you using Umbraco 9+?**
+
+As of Umbraco 9 all packages needs to be installed through NuGet, and it is no longer to install packages from .zip files.
```
ronaldbarendse, line 16: Should this also reference the blogpost (https://umbraco.com/blog/packages-in-umbraco-9-via-nuget/) explaining this change?

```suggestion
As of Umbraco 9, packages can only be installed using NuGet and it is no longer possible to install packages from package zip files.
```

ronaldbarendse (changes requested):
Hi @sofietoft! I've read through the changes and only have two minor suggestions... Great additions and will hopefully safe people some time figuring out why their migrations are executed more than once👍🏻

sofietoft:
Thanks for the suggestions @ronaldbarendse ! 💪 
I've applied them both - but did make a small adjustment to shorten the sentence length a bit 😄 

From:
When using a custom package migration plan, **the current state is ignored by default, causing it to execute all migrations** again whenever this isn't the same as the final state of the plan (e.g. if you added a new migration).
To:
When using a custom package migration plan, **the current state is ignored by default. This causes it to execute all migrations** again whenever this isn't the same as the final state of the plan (e.g. if you added a new migration).

PR #4193, opened 2022-07-07
Various fixes

eshanrnh:
Added Note as per Trello task.
Clean up umbracoSettings article, deleted all the pre/post files and merged its content in index-v7/index.md file.


Reference/Configuration-for-Umbraco-7-and-8/umbracoSettings/index-v7.md
```diff
@@ -166,6 +166,29 @@ On new installs this set to true. When you are upgrading from a lower version th
...
+
+You can specify your own background image for the login screen here. The image will automatically get an overlay to match backoffice colors. This path is relative to the ~/umbraco path. The default location is: /umbraco/assets/img/installer.jpg
+
+```xml
+<loginBackgroundImage>../App_Plugins/Backgrounds/login.png</loginBackgroundImage>
+```
+
+## `<AllowedUploadFiles>` (Available on Umbraco versions 7.6.2 and above)
+
+If greater control is required than available from the above, this setting can be used to store a "white list" of file extensions.  If provided, only files with these extensions can be uploaded via the backoffice.
```
sofietoft, line 179:
```suggestion
If greater control is required than available from the above, this setting can be used to store an "allow list" of file extensions.  If provided, only files with these extensions can be uploaded via the backoffice.
```

sofietoft (commented):
Looks great @eshanrnh !

Just a single suggestion based on a vale complaint 💪

PR #4218, opened 2022-07-14
-all changed to --list

craigs100:
-all doesn't work, --list does


eshanrnh (approved):
Thanks for the PR, @craigs100 🙌 Looks good 🙂

Merging it in.

PR #4219, opened 2022-07-15
Removed the moving from courier to deploy article

jonat123:
As there should be no more courier projects on Umbraco Cloud.


eshanrnh (approved):
Looks great, @jonat123 💪 Merging it in

PR #4220, opened 2022-07-15
Renamed the manual upgrade deploy folders, so it matches versions

jonat123:
Hopefully this will make the v9 version show up as well as it was not showing up


eshanrnh (approved):
Looks good, @jonat123 🙌 Merging it in.

PR #4225, opened 2022-07-18
Added clarity for deploy setting

Frost117:
Ran into a confusion scenario in the upgrade process.

Adding this into the documentation that we link to will give a direct overview of what you need in case you want to keep your old database when going from v9 -> v10


eshanrnh (approved):
Thanks for the PR, @Frost117 🙌 The PR looks good. 💪

Merging it in.

PR #4230, opened 2022-07-20
fix: typo where publishes should be published

MMasey:


eshanrnh (approved):
Thanks for the Pr, @MMasey 💪 Looks great 🙌

PR #4233, opened 2022-07-21
Removed the courier section from troubleshooting

jonat123:
As there should be no sites with courier anymore on cloud.


eshanrnh (approved):
Looks good, @jonat123 🙌 Merging it in.

PR #4244, opened 2022-07-25
Minor changes and removed the note about the version

jonat123:
All project on the new infra should be on UmbracoID therefore the note is not necessary any more


eshanrnh (approved):
Looks great, @jonat123 💪 Merging it in.

PR #4259, opened 2022-08-02
New version of dictionary items article including import/export feature

sofietoft:


Fundamentals/Data/Dictionary-Items/index-v8.md
```diff
@@ -0,0 +1,69 @@
...
+meta.Description: "Creating Dictionary Items in Umbraco"
+versionFrom: 8.0.0
+versionTo: 10.0.0
+---
+
+# Dictionary Items
+
+Depending on how your site is set up, not all content is edited through the **Content** section. There might be some text in your templates or macros that needs translation. Using Dictionary Items, you can store a value for each language. Dictionary Items have a unique key which is used to fetch the value of the Dictionary Item.
+
+Dictionary Items can be managed from the **Translation** section. Let's take a look at an example. In this example, we will translate "Welcome to Umbraco" from within the template and add it to the dictionary:
```
jonat123, line 12:
```suggestion
Depending on how your site is set up, not all content is edited through the **Content** section. There might be some text in your templates or macros that needs translation. Using Dictionary Items, you can store a value for each language. Dictionary Items have a unique key that is used to fetch the value of the Dictionary Item.

Dictionary Items can be managed from the **Translation** section. Let's take a look at an example. In this example, we will translate "Welcome to Umbraco" from within the template and add it to the dictionary:
```

jonat123 (changes requested):
Looks good!¨

Saw one small thing, other than that i don't have anything else :)

PR #4260, opened 2022-08-02
Removed the Manual-Upgrade-UaaS-Tool article

jonat123:
The article is old and should not be relevant anymore as it is not possible to run older versions of Umbraco 8 on the new hosting platform.

The lowest v8 version able to run on cloud is 8.6.7 according to the note.

If anyone is against removing the article, please let me know


sofietoft:
Makes sense to remove the article.

Before removing it, we need to do two things:
* [ ] Verify that there are no other articles linking to it
* [ ] Add a rewrite to Our, just in case someone has the link or finds it on Google. I think sending them to the general Cloud Upgrade docs would be fine.

jonat123:
@sofietoft No links to the article in the docs.

Will look into the rewrite tomorrow :)

jonat123:
PR for rewrites created here: https://github.com/umbraco/OurUmbraco/pull/788

jonat123:
Pr has been merged in, so merging this one in.

Issue #4274, opened 2022-08-09
Verify Unit Testing docs for v10


sofietoft:
Hi @Adolfi ,

Hope you've enjoyed your summer! ☀️ 
We have yet to get the Unit Testing docs verified against Umbraco 10, which was released a few months back.
Is there any chance you have some time to spare in the near future, to help us with the task? 🤞 😄 
As you're the original author of these docs, I thought it appropriate to loop you in here!

Adolfi:
Hi @sofietoft.
I will have a look at verify it against v10 to see if there’s any changes needed to be made.
Thanks for the heads up.
Cheers!

PR #4276, opened 2022-08-11
Verified Unit Testing docs for v10

Adolfi:
@sofietoft Turns out nothing new in v10 breaking these examples, everything works just as it did in v9.
So just needed to update the **versionTo** header.

Cheers!


sofietoft:
Amazing! Thanks a lot for checking @Adolfi !

Much appreciated!

Fixes #4274

PR #4277, opened 2022-08-11
Corrected description of health check settings first run time.

AndyButland:
I came across this when looking to re-use this setting pattern for a feature in Umbraco Forms, and realised that it had been described incorrectly.

@nikolajlauridsen - would you mind double-checking that you agree with my change please? Thanks.


nikolajlauridsen (approved):
LGTM 👍 

and https://crontab.guru/ agrees too 👍

PR #4279, opened 2022-08-12
Updating Callbackpath

thejokerd3:
Callbackpath cannot be the same as the default /umbraco backoffice route.


sofietoft:
Looks great @thejokerd3 !

PR #4282, opened 2022-08-15
Update to Magic String information for record.id

OwainWilliams:
Making a note that Record.Id is only accessible on workflows trigged "on approve". 

This update was made after raising an issue - https://github.com/umbraco/Umbraco.Forms.Issues/issues/835


eshanrnh (approved):
Thanks for the PR, @OwainWilliams. It looks good 🙌

PR #4289, opened 2022-08-17
Change 'versionTo' from 9 to 10

erikjanwestendorp:
All built in property editors are verified against Umbraco 10.


eshanrnh (approved):
Ah, Good catch, @erikjanwestendorp 👏 I'll merge it right away 😃

PR #4293, opened 2022-08-18
Use hardcoded language

dampee:
While it makes sense to get this from the CurrentCulture, this IS NOT related to the culture which is linked to the variant. Even more, your backoffice user can work in dk while the website is only available in en-US...  which would never show this link.

I think it makes more sense to use a hardcoded string to make sure people will change it before raising support requests.

Or maybe it should be something like `notification.UmbracoContext.PublishedRequest.Culture`

### DIFF
--- a/Reference/Notifications/EditorModel-Notifications/Customizing-the-links-box.md
+++ b/Reference/Notifications/EditorModel-Notifications/Customizing-the-links-box.md
@@ -16,7 +16,7 @@ public void Handle(SendingContentNotification notification)
 {
     notification.Content.Urls = new[]
     {
-        new UrlInfo($"/products/?id={notification.Content.Id}", true, CultureInfo.CurrentCulture.Name)
+        new UrlInfo($"/products/?id={notification.Content.Id}", true, "en-US")
     };
 }
 ```


eshanrnh:
Thanks for the PR, @dampee 💪 We will review this soon.

kjac:
@dampee thank you for this 😄 

The `culture` parameter doesn't actually do a whole lot other than displaying a badge next to the link. 

This code for example:

image

Yields:

image

So... perhaps it would make more sense to leave the `culture` parameter out entirely in the first example, and add an example with different cultures and hostnames? Something like

```csharp
public void Handle(SendingContentNotification notification)
{
    notification.Content.Urls = new[]
    {
        new UrlInfo($"/products/?id={notification.Content.Id}", true, null)
    };
}
```

and 

```csharp
public void Handle(SendingContentNotification notification)
{
    notification.Content.Urls = new[]
    {
        new UrlInfo($"https://mysite.com/products/?id={notification.Content.Id}", true, "en-US"),
        new UrlInfo($"https://mysite.dk/produkter/?id={notification.Content.Id}", true, "da-DK")
    };
}
```

eshanrnh:
@dampee, Just checking in - Are you up for incorporating the changes suggested by @kjac? 🤔

sofietoft:
Hi all 👋 

I migrated this PR: https://github.com/umbraco/UmbracoDocs/pull/4719 
Please take a look when you have a minute 🙏

PR #4295, opened 2022-08-19
Release notes for 2022-08-19 Umbraco Portal release

sajumb:


eshanrnh:
Thanks for the PR, @sajumb 💪 We will review this soon.

Umbraco-Cloud/Release-Notes/2022-08-19-ReleaseNotes.md
```diff
@@ -0,0 +1,30 @@
...
+
+You can access Dedicated Resources from the Settings menu on your Cloud project. Here you’ll find 3 different dedicated options (CPU and memory configurations) to choose from bringing flexibility in how requirements are met, and how much extra compute power is added. As soon as you select and confirm an option the process of migrating to dedicated resources begins. If the extra resources are no longer needed, the project can be moved back to shared resources in the same way.
+
+![ProjectsDashboard](images/DedicatedResourceMvp.gif)
+
+Dedicated Resources is currntly available for projects on **Professional** and **Enterprise** plans. More options will be added later in the year to make the feature available for Starter and Standard plans as well. 
+
+## Projects Dashboard (beta)
+
+We are currently working on multiple initiatives to give Umbraco Cloud users a fantastic **developer-friendly experience**. One of these is the Projects Dashboard (as it’s known internally). Together with partners and Cloud users, we have worked to identify what would be the most useful features and tools to provide an overview of all your projects.
```
sofietoft, line 22: > One of these is the Projects Dashboard (as it’s known internally).

Is the "as it's known internally" necessary?
Is it because the name might change later, because then perhaps we could say something like "this is what it's currently called".

What do you think?

sofietoft (commented):
Looks great! 💪 
Just a single comment added.

Also, do we want to create a new article for the Projects Dashboard?
Or will it eventually replace the Projects Overview, and therefor also the related article?

Umbraco-Cloud/Release-Notes/2022-08-19-ReleaseNotes.md
sajumb: Yup, your suggestion would be better. You could also skip the content in the parenthesis. That would be find with me.

sofietoft: Let's skip it then.
I'll amend and merge 💪

sajumb:
> Looks great! 💪 Just a single comment added.
> 
> Also, do we want to create a new article for the Projects Dashboard? Or will it eventually replace the Projects Overview, and therefor also the related article?

Yes, the Projects Dashboard is currently just a beta, where the user can have a sneak peek. When version 1.0 of this page is launched (in many 3-4 weeks) from now, we will write a new Our page and introduce it properly. I expect that the users (at least for a longer period) can choose between the old and the new one.

PR #4302, opened 2022-08-22
Added Note and fixed markdown errors

eshanrnh:
- Added Notes referring to v9 and above article
- Fixed markdown errors.


sofietoft (commented):
Looks great @eshanrnh !

Just a small note. I'm not sure `cs` is a valid format for the code snippets.
Perhaps `xml` or `none` would be better? 🤔

eshanrnh:
Ah ok. I'll change it to `none `🙂

PR #4313, opened 2022-08-25
Updated the code snippet for config transforms

Jay-umbr:
Some xdt magic that don't work no more on v10. Not sure when it changed but I updated the snippet and tested it on Cloud.


eshanrnh (approved):
Looks good to me, @Jay-umbr 🙌

PR #4326, opened 2022-08-31
Update IPublishedContentQuery to use proper standards

Zeegaan:
# Notes
- Changed code examples from being directly in view to being in a service

### DIFF
--- a/Reference/Querying/IPublishedContentQuery/index.md
+++ b/Reference/Querying/IPublishedContentQuery/index.md
@@ -9,13 +9,25 @@ meta.Description: "Querying in views with IPublishedContentQuery in Umbraco"
 
 The `IPublishedContentQuery` interface contains various query methods for accessing strongly typed content in templates.
 
-## How to reference IPublishedContentQuery
+## How to inject IPublishedContentQuery
 
-In order to reference the `IPublishedContentQuery` in your views, you must add a using statement for `Umbraco.Core` and inject the service using the `@inject` keyword.
+In order to inject the `IPublishedContentQuery` into your services, you must add a using statement for `Umbraco.Cms.Core` and inject the service using the constructor.
 
 ```C#
-@using Umbraco.Cms.Core
-@inject IPublishedContentQuery _publishedContentQuery;
+using Umbraco.Cms.Core;
+
+namespace MySite.Services
+{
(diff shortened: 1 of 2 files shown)


Zeegaan:
This will again need a internal review from the D-team 👍

sofietoft:
I have a quick question on this one, based on a conversation with the community docs team earlier today.

The first code sample you've updated - that still goes into a template/view, right?
Just wondering, as the initial intro to the article mentions the templates - also, where we link to this article also mentions strongly types and templates. If we're changing this, we need to change it the right way 💪

Zeegaan:
Hey @sofietoft, no this is all supposed to be backend code, but I do see we still mention templates etc. 🤔 
I will have a look at cleaning this up so its clear 👍

sofietoft:
Thanks for clearing that up @Zeegaan ! (and Happy Birthday btw 🥳 )

Does the service still need to be injected in a template/view somewhere?
If it does - should we add a sample for that as well?

Also, perhaps this is a bigger discussion around how we handle these best practice code samples in our docs in general 😄

Zeegaan:
@sofietoft No, I have refactored our Quick start: https://our.umbraco.com/documentation/Reference/Searching/Examine/quick-start/ 
This now uses a custom view model to use in the view instead, this article then builds on how you'd use the IPublishedContentQuery in the SearhService. 
Now that i say that, I see that is not mentioned at all, so I will add that in aswell 🙈

sofietoft:
Ah! Missed that one.

Cool, thanks! 😄

Reference/Querying/IPublishedContentQuery/index.md
```diff
@@ -9,13 +9,25 @@ meta.Description: "Querying in views with IPublishedContentQuery in Umbraco"
 
 The `IPublishedContentQuery` interface contains various query methods for accessing strongly typed content in templates.
```
kjac, line 10:
```suggestion
The `IPublishedContentQuery` interface contains various query methods for accessing strongly typed content in services etc.
```

sofietoft:
Looks good 💪 

I just updated the text on the page linking to the "IPublishedContentQuery" article, using the same text as in the intro of the article.

PR #4335, opened 2022-09-05
v10.2: New AllowEditInvariantFromNonDefault setting

sofietoft:
Added details about the `AllowEditInvariantFromNonDefault` security setting added in 10.2.

### DIFF
--- a/Reference/Configuration/SecuritySettings/index.md
+++ b/Reference/Configuration/SecuritySettings/index.md
@@ -20,6 +20,7 @@ A full configuration with all default values can be seen here:
       "AllowPasswordReset": true,
       "AuthCookieName": "UMB_UCONTEXT",
       "AuthCookieDomain": "",
+      "AllowEditInvariantFromNonDefault": true,
       "UsernameIsEmail": true,
       "UserPassword": {
         "RequiredLength": 10,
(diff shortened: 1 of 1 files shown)


Reference/Configuration/SecuritySettings/index.md
```diff
@@ -69,6 +70,14 @@ The authentication cookie which is set in the browser when a backoffice user log
 
 The authentication cookie which is set in the browser when a backoffice user logs in is automatically set to the current domain.
 
+### Allow Edit Invariant From Non-Default
+
+Invariant properties are properties on a multilingual site that are not varied by culture, which means it shared the same value across all languages added to the website.
+
+When the setting is set to `false` the invariant properties that are shared between all languages can be edited by all no matter what language you have permission to manage.
```
Zeegaan, line 77:
```suggestion
When the setting is set to `false` the invariant properties that are shared between all languages can only be edited from the default language. This means you need access to the default language, in order to edit the property.
```

Zeegaan (changes requested):
Found a small correction 👍

sofietoft:
Great! Thanks for the suggestion @Zeegaan !
I had a hard time finding my way around the `false` / `true` difference, so I might have confused myself a bit there.

So, this mean we're ready to merge? 😄

PR #4359, opened 2022-09-07
added "for"

Jette94:
What may be 'easy' for you, might not be easy another user reading the article. > What may be 'easy' for you, might not be easy for another user reading the article.


eshanrnh (approved):
Looks great, @Jette94 🙌

PR #4362, opened 2022-09-08
Fixed incorrect configuration key

AndyButland:
We spotted this error today where the docs and code don't match up, hence this PR to resolve.

It can be published as soon as it is reviewed and approved.


sofietoft:
Thanks for updating this @AndyButland ! 💪 

Merging in right away.

PR #4368, opened 2022-09-09
Added comma, dot and deleted "so"

Jette94:
Old sentence: "Next we will create a HTML file inside this folder called 'WelcomeDashboard.html' the html file will contain a fragment of a html document and so does not need <html><head><body> entities."
New sentence: "Next, we will create a HTML file inside this folder called 'WelcomeDashboard.html'. The html file will contain a fragment of a html document and does not need &lt;html&gt;&lt;head&gt;&lt;body&gt; entities."


eshanrnh (approved):
Looks great, @Jette94 🙌 Merging it in,.

PR #4370, opened 2022-09-12
Added note about how and when to use global settings

sofietoft:


kjsguld (commented):
Looks great

sofietoft:
Thanks for the review @kjsguld ! 💪 

Merging it in.

PR #4378, opened 2022-09-13
Update to 'Adding a type to the provider model'

hfloyd:
Adding a note about additional NuGet package.


eshanrnh (approved):
Looks good, @hfloyd 🙌 Merging it in.

PR #4386, opened 2022-09-19
Add note about CopyRazorGenerateFilesToPublishDirectory

ronaldbarendse:
The current documentation isn't very clear on the requirement of having `<CopyRazorGenerateFilesToPublishDirectory>true</CopyRazorGenerateFilesToPublishDirectory>` set in your project file when using the new runtime mode together with build/publish time Razor compilation.

This property is already set in the default [dotnet new template (alongside a descriptive comment)](https://github.com/umbraco/Umbraco-CMS/blob/c5654f031195544d6bbb9d8b27c21458ec692890/templates/UmbracoProject/UmbracoProject.csproj#L19-L22), but isn't mentioned in our docs and resulted in https://github.com/umbraco/Umbraco.Forms.Issues/issues/877 being reported on the Forms issue tracker.


Fundamentals/Setup/Server-Setup/runtime-modes.md
```diff
@@ -55,6 +55,10 @@ Ensure your models are generated by running Umbraco and navigating to **Settings
 
 Fix any compilation errors you might get after this, e.g. if you accidentally referenced deleted models or properties. Running the application will still show the rendered content and you're now ready to optionally enable `Production` mode on your production environment.
 
+:::note
+Ensure you have the `<CopyRazorGenerateFilesToPublishDirectory>true</CopyRazorGenerateFilesToPublishDirectory>` property in your `csproj` project file, so Razor views are always copied to the publish directory. This is required by the CMS to diplay the contents in the backoffice (as read-only when running in Production runtime mode) and for Deploy to be able to compare schemas (otherwise you'll get schema mismatches).
```
ronaldbarendse, line 59: Should this also mention the property is already set when using the dotnet new template?

ronaldbarendse, line 59: And we should probably also include the requirement for Forms, as it otherwise can't find custom theme partial views.

ronaldbarendse: This property was already included since 9.0.1, so let's not make this note bigger than it already is 😄

ronaldbarendse: I've updated the note to mention Forms too, so should be good to merge now! 🚀

eshanrnh (approved):
Looks good, @ronaldbarendse 🙌 Merging it in.

PR #4390, opened 2022-09-20
Removed version 9-10 version of the Media Picker article

sofietoft:
Rearranged the somewhat confusing documentation from the legacy Media Picker since Umbraco 8.14.
When using Umbraco 8.14 or later versions, we should recommend that people use the other Media Picker instead - Media Picker 3 (which we shouldn't call it 😅 ).

I hope the changes I've made, makes sense.
Otherwise, let me know 🤞


eshanrnh (approved):
The changes look good to me. Since, we don't have the latest versions (v9+) it won't be confusing and the note makes sense.

Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker/index.md
```diff
@@ -1,20 +1,24 @@
 ---
-versionFrom: 9.0.0
-versionTo: 10.0.0
+versionFrom: 8.1.0
+versionTo: 8.14.0
```
eshanrnh, line 3: Should this be 8.13.0? Since the note below mentions 8.14 and links to the new media picker? It can be a little confusing if the versionTo has 8.14.0 cz it means the article is valid for it but then the note says otherwise.

eshanrnh (commented):
The changes look good to me. Since, we don't have the latest versions (v9+) it won't be confusing and the note makes sense. However, I have a single comment to the versionTo tag.

Fundamentals/Backoffice/Property-Editors/Built-in-Property-Editors/Media-Picker/index.md
sofietoft: Yep. That's a great point! I'll update it to 8.13.0 💪

eshanrnh (approved):
Looks good, @sofietoft 🙌 Merging it in.

PR #4393, opened 2022-09-20
Correct typo in MaximumUploadSizeSettings meta title

callumbwhyte:
I found this typo when Googling for some docs, so thought I would correct it!


dampee:
ACE @callumbwhyte

dampee:
Your change will not be deployed right away as the deploy proces is still a manual intervention.  But I am sure that it will happen tomorrow.  Thanks a lot for the effort :-)

PR #4400, opened 2022-09-22
Avoid using EnsureUmbracoContext in a loop

callumbwhyte:
The docs for using UmbracoContext via the `UmbracoContextFactory` in C# include a code sample that recommends creating a new UmbracoContext for each item in a collection.

This is okay and will work fine, but definitely won't be as efficient as it should be (especially for larger collections).

This PR changes this code sample to first create an UmbracoContext (once), then loop over the items in the collection.


eshanrnh:
Thanks for the PR, @callumbwhyte 💪 We will review this soon.

nul800sebastiaan (approved):

Issue #4391, opened 2022-09-20
When setting up SignalR routes, you need to add the routes to `ReservedPaths`


nul800sebastiaan:
## What type of issue is it? (Choose one - delete the others)
Amending existing docs

## What article/section is this about?
[ReservedPaths](https://github.com/umbraco/UmbracoDocs/blob/main/Implementation/Custom-Routing/signalR.md)

## Describe the issue
As I've just learned from https://github.com/umbraco/Umbraco-CMS/issues/13005#issuecomment-1252282141 - when you add a SignalR route, Umbraco will try to find the page and end up in either custom 404 handlers or execute Umbraco's default `ContentFinderByConfigured404`. This is unnecessary and gives quite a bit of overhead. Instead we should add the SignalR route to the `ReservedPaths` config, like so:

```

  "Umbraco": {
    "CMS": {
      "Global": {
        "ReservedPaths": "~/app_plugins/,~/install/,~/mini-profiler-resources/,~/umbraco/,~/umbraco/testhub/"
        }
      }
    }
  }

Note that you need to provide the default reserved paths (`~/app_plugins,~/install/` etc), else you'll run into the issue described in https://github.com/umbraco/Umbraco-CMS/issues/12965

eshanrnh:
Thanks, @nul800sebastiaan. We will add the info in the docs soon 🙂

eshanrnh:
You can find this documentation at https://our.umbraco.com/Documentation/Implementation/Custom-Routing/signalR#add-the-route-in-appsettingsjson-file

PR #4403, opened 2022-09-23
Fixes Issue 4391

eshanrnh:
#4391


jonat123:
@eshanrnh Looks good to me :) 

Only thing is i remember @sofietoft and i talked about the article not being up to the standard of articles that we wanted in our docs, however i can see that changes has been made to it so i guess it has been reworked and is okay now? 

If that is the case maybe we should also link it somewhere so people can actually find it :)

eshanrnh:
Hey @jonat123 Yup, the article has been reworked and I've added a link from the Custom Routing article so it's easier to find.

jonat123 (approved):
Great! looks good to me :)

PR #4406, opened 2022-09-26
Small changes to the prevalues article

jonat123:
Added an image of how it looks in the backoffice with the prevalues and the captions.

As well as a minor change to the text.


eshanrnh (approved):
Looks good to me, @jonat123 🙌 Merging it in.

PR #4412, opened 2022-09-27
Added rule that catches whether an acronym has been defined

sofietoft:
All acronyms should be defined before continuous use in an article.
This means that when an acronym is used in an article, the first use case needs to be accompanied by a definition.

Examples have been added to the PR.

Also, a long list of excluded acronyms have been added to the rule.
This is in order to accommodate the API docs, markup languages and other commonly used acronyms like `IIS` and `CMS`.

### DIFF
--- a/.github/styles/UmbracoDocs/Acronyms.yml
+++ b/.github/styles/UmbracoDocs/Acronyms.yml
@@ -22,6 +22,7 @@ exceptions:
   - ASP
   - REST
   - HTTP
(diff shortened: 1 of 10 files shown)


eshanrnh (approved):
Looks Amazing 🙌🎉

sofietoft:
I'll merge this one tomorrow, once I've also added info about the rule to the Style Guide contribution article.

PR #4435, opened 2022-10-07
Updated the Custom Indexing code to match to Samples repo

busrasengul:


Reference/Searching/Examine/Indexing/index.md
```diff
@@ -164,24 +164,17 @@ To create this index we need five things:
...
-            ILoggerFactory loggerFactory,
-            string name,
-            IOptionsMonitor<LuceneDirectoryIndexOptions> indexOptions,
-            IHostingEnvironment hostingEnvironment,
-            IRuntimeState runtimeState) : base(loggerFactory,
-            name,
-            indexOptions,
-            hostingEnvironment,
-            runtimeState)
+            ILoggerFactory loggerFactory, string name, IOptionsMonitor<LuceneDirectoryIndexOptions> indexOptions, IHostingEnvironment hostingEnvironment, IRuntimeState runtimeState)
```
sofietoft, line 176: Isn't it a shame, that the "list overview" of the items here is lost? 🤔 
I think the "list overview" is easier to read - might just be me!

Reference/Searching/Examine/Indexing/index.md
```diff
@@ -323,16 +322,17 @@ using Umbraco.Cms.Core.Composing;
...
 
-namespace MySite.MyCustomIndex
+namespace Umbraco.Docs.Samples.Web.CustomIndexing
 {
     public class ExamineComposer : IComposer
     {
         public void Compose(IUmbracoBuilder builder)
         {
             builder.Services.AddExamineLuceneIndex<ProductIndex, ConfigurationEnabledDirectoryFactory>("ProductIndex");
-            builder.Services.ConfigureOptions<ConfigureCustomIndexOptions>();
```
sofietoft, line 333: Should this be deleted? 🤔 
There's still a section at the top of the article called "Creating a ConfigureOptions class".

sofietoft (commented):
Thanks for updating the samples in this article @busrasengul 💪 

I've added a few comments. Take a look and let me know what you think!

Reference/Searching/Examine/Indexing/index.md
busrasengul: In this tutorial, we're not actually adding any `ConfigureCustomIndexOptions` so there is no need to add that line. And on line 46 we're already explaining if we add 'ConfigureNamedOptions' then the next code block shows how to register it :)

sofietoft: Oh, alright 🤔 
Then I wonder why it was added in the first place!

busrasengul: It was I who added that while trying the tutorial then I realized we don't need that so I removed :)

PR #4441, opened 2022-10-10
Macro Parameter Editors has a broken link

OwainWilliams:
Just a small edit to fix a broken link that I found when reading the documentation for Macros. 

The link was pointing to a now removed page for V8, I've change the link to point to the main Property Editors page.

### DIFF
--- a/Extending/Macro-Parameter-Editors/index-v8.md
+++ b/Extending/Macro-Parameter-Editors/index-v8.md
@@ -25,7 +25,7 @@ It is possible to create custom macro parameter types.
 
 ### isParameterEditor
 
-To create a custom Macro Parameter Type, first create a custom 'Property Editor' (or copy one from the core). See [Property Editors documentation](../../Extending/Property-Editors/) and in the corresponding [Package Manifest file](../../Extending/Property-Editors/Package-Manifest/index.md) for the editor, set the `isParameterEditor` property to be true.
+To create a custom Macro Parameter Type, first create a custom 'Property Editor' (or copy one from the core). See [Property Editors documentation](../../Extending/Property-Editors/index-v7-8) and in the corresponding [Package Manifest file](../../Extending/Property-Editors/Package-Manifest/index.md) for the editor, set the `isParameterEditor` property to be true.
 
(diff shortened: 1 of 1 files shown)


Extending/Macro-Parameter-Editors/index-v8.md
```diff
@@ -25,7 +25,7 @@ It is possible to create custom macro parameter types.
 
 ### isParameterEditor
 
-To create a custom Macro Parameter Type, first create a custom 'Property Editor' (or copy one from the core). See [Property Editors documentation](../../Extending/Property-Editors/index-v8.md) and in the corresponding [Package Manifest file](../../Extending/Property-Editors/Package-Manifest/index.md) for the editor, set the `isParameterEditor` property to be true.
+To create a custom Macro Parameter Type, first create a custom 'Property Editor' (or copy one from the core). See [Property Editors documentation](../../Extending/Property-Editors/) and in the corresponding [Package Manifest file](../../Extending/Property-Editors/Package-Manifest/index.md) for the editor, set the `isParameterEditor` property to be true.
```
sofietoft, line 28: Shouldn't this instead link to the [Umbraco 7-8 version of the "Property Editor documentation"](https://our.umbraco.com/Documentation/Extending/Property-Editors/index-v7-8)?
Since this article is marked for Umbraco 8, I think it should also link to Umbraco 8 valid docs, right? 🤔

sofietoft (commented):
Thanks for taking a look at this broken link @OwainWilliams ! 💪 

Looks like the article wasn't removed, but renamed - which would result in the same behavior, really.
I've added a single suggestion - let me know what you think!

OwainWilliams:
Thanks @sofietoft  - I thought it might have been moved but just couldn't find what the new home would be. 
Fixed the link :)

sofietoft:
Thanks for updating the link @OwainWilliams !

I do realize that the versioning system isn't always easy to figure out 🙈 Hopefully that will soon be a thing of the past!

Oh, and congratz on being one step closer to completing the Hacktoberfest challenge 😎

PR #4442, opened 2022-10-10
Added element type section

eshanrnh:
Belongs to airtable task: https://airtable.com/appAelapv34OT1avk/tbliWYWb9SC0z4k3I/viwIrOG4IpVmsxFOD/recpOt2LzJYZ3smtu?blocks=hide

**Note:** We might have to skip _SHIFT_ and _ALT_ (Keyboard keys) from the acronyms check list.

- [x] Add Element type section in v8 and v9 articles.
- [x] Add references to Element Type section from Block List and Block Grid Editor articles
- [x] Add Creating an Element Type section.


eshanrnh:
The line throwing a warning (write shorter sentences) contains 24 words. Not sure why it's still throwing a warning message.

Fundamentals/Data/Defining-content/index-v8.md
```diff
@@ -9,27 +9,27 @@ versionTo: 9.2.0
...
 
 ### Properties
 
-Each field on a Document Type is called a property. A property is given a name, an alias (used to output the properties contained in a template), and an editor. The editor determines what type of data the property will store and the input method. There is a wide range of editors available out of the box (textstring, Rich text, media picker, and so forth) and you can customize and add additional editors.
+Each field on a Document Type is called a property. A property is given a name, an alias (used to output the properties contained in a template), and an editor. The editor determines what type of data the property will store and the input method. There is a wide range of property editors available out-of-the box (Textstring, Richtext, Media Picker, and so forth) and you can customize additional editors.
```
sofietoft, line 24: Strange that it gives the error here. Could be because of "out-of-the-box" 🤔 Let's try something like this:

```suggestion
Each field on a Document Type is called a property. The property is given a name, an alias (used to output the properties contained in a template), and an editor. The editor determines what type of data the property will store and the input method. There is a wide range of default property editors available (Textstring, Richtext, Media Picker, and so forth) and you can customize additional editors.
```

Fundamentals/Data/Defining-content/index.md
```diff
@@ -9,27 +9,27 @@ versionTo: 10.0.0
...
 
 ### Properties
 
-Each field on a Document Type is called a property. A property is given a name, an alias (used to output the properties contained in a template), and an editor. The editor determines what type of data the property will store and the input method. There is a wide range of editors available out of the box (textstring, Rich text, media picker, and so forth) and you can customize and add additional editors.
+Each field on a Document Type is called a property. A property is given a name, an alias (used to output the properties contained in a template), and an editor. The editor determines what type of data the property will store and the input method. There is a wide range of property editors available out-of-the box (Textstring, Richtext, Media Picker, and so forth) and you can customize additional editors.
```
sofietoft, line 24:
```suggestion
Each field on a Document Type is called a property. The property is given a name, an alias (used to output the properties contained in a template), and an editor. The editor determines what type of data the property will store and the input method. There is a wide range of default property editors available (Textstring, Richtext, Media Picker, and so forth) and you can customize additional editors.
```

Fundamentals/Data/Defining-content/index.md
```diff
@@ -208,15 +214,30 @@ and this time we'll add two properties. First, make a property called Summary us
...
+Go to the __Content__ section and click the menu icon (•••) next to the *Home* node and select the Text page Document Type. We'll name the page "About us". We now have a basic content structure.
 
 ![Basic content structure](images/v8Screenshots/createAboutUs.png)
 
-Document Types are very flexible and can be used in a myriad of ways from defining a piece of reusable content or an entire page, to acting as a container or repository.
+Document Types are flexible and can be used for defining pieces of reusable content or an entire page, to acting as a container or repository.
+
+## What is an Element Type?
+
+An Element Type is a Document Type without a template containing smaller schema configurations for repeating a set of properties. These can be used for defining schema in the Block List Editor, Nested Content, Block Grid Editor, or other Element Type based editors. Element Types cannot be used to create content that resides in the Content tree.
```
sofietoft, line 229:
```suggestion
An Element Type is a Document Type without a template containing schema configurations for repeating a set of properties. These are for defining schema in the Block List Editor, Nested Content, Block Grid Editor, or other Element Type based editors. Element Types cannot be used to create content that resides in the Content tree.
```

Fundamentals/Data/Defining-content/index.md
```diff
@@ -208,15 +214,30 @@ and this time we'll add two properties. First, make a property called Summary us
...
+When you create an Element type, it automatically sets the **Is Element Typw** flag to **True** on the **Permissions** tab.
```
sofietoft, line 230:
```suggestion
When you create an Element type, it automatically sets the **Is Element Type** flag to **True** on the **Permissions** tab.
```

Fundamentals/Data/Defining-content/index.md
```diff
@@ -208,15 +214,30 @@ and this time we'll add two properties. First, make a property called Summary us
...
+
+![Element Type](images/Element-Type.png)
+
+## Creating an Element Type
+
+An Element Type is created under Document Types in the __Settings__ section.
+
+Go to the __Settings__ section in the backoffice. On the __Document Types__ node click the menu icon (•••) to bring up the context menu. Here, choose __Element Type__. This will create an Element Type.
+
+![Creating an Element Type](images/Element-Type-create.png)
```
sofietoft, line 240: I don't think it's necessary to add steps on how to create an Element Type.
The entire article is about how to create a Document Type, which is pretty much the same workflow.

Perhaps instead, something like this could be added:
"Element Types are created using the same workflow as regular Document Types, but usually contain fewer properties. You can also create Element Types as part of configuring a Block Grid or Block List Data Type."

What do you think? 🤔

sofietoft (changes requested):
Looks great @eshanrnh ! 💪 
And thanks for cleaning up all the vale warnings as well.

I've added some comments here and there - but only made suggestions for the Element Type section on one of the article. Changes should of course be applied to both 😅 

Finally, I think the section should be placed further up, along with where the other options for Document Types are explained - around lines 36-37.

Fundamentals/Data/Defining-content/index.md
eshanrnh: Yup, makes sense. I'll remove the section and add the suggested line instead 💪

eshanrnh:
@sofietoft Thanks for the feedback. I've applied the suggestions and moved the Element Type section below the Creating a Doc Type section so as to not break that section.

Fundamentals/Data/Defining-content/index-v8.md
```diff
@@ -39,19 +39,28 @@ You can also use __Composition__ to create a new Document Type. Compositions all
 
 ![Creating a Compositions](images/v8Screenshots/compositions.png)
 
+## What is an Element Type?
```
sofietoft, line 42:
```suggestion
Finally, you can create a Document Type as an __Element Type__, which can be used to build Block Grid and Block List editors. Learn more about Element Types below.

### What is an Element Type?
```

sofietoft (commented):
Thanks for applying the suggestions.

I've just added a little note about how "Element Types" is also an option when creating a Document Type. I think that's more in-line with the rest of the article.

Let me know what you think!

sofietoft:
And regarding the vale errors on the "acronyms" - it would be great if you could add them directly to the [exceptions-list in the YAML file](https://github.com/umbraco/UmbracoDocs/blob/main/.github/styles/UmbracoDocs/Acronyms.yml) 💪

eshanrnh:
I think this should be good to go now 🤞

sofietoft (approved):
Looks great 💪 

Let's get it merged in.

PR #4446, opened 2022-10-11
Notifications matching to the Samples site

busrasengul:


sofietoft:
Related Sample Project PR: https://github.com/umbraco/Umbraco.Docs.Samples/pull/10 
(would be great if you could link between the PRs - no need to review the code twice 😄 )

sofietoft:
I see that the referenced PR on the Sample Project was closed.
What should happen to this one? 😄

busrasengul:
We already have the codes in the Samples site but haven't updated the docs. So If you're happy with it we can merge :)

sofietoft:
Ah, cool!
I'll merge it in then 💪

Issue #4416, opened 2022-09-28
Improve readability of the docs


sofietoft:
**Hacktoberfest** is upon us, and on the UmbracoDocs we have a very important task that we need **your** help with:

Improving readability!
===================================

Some time back, we implemented a style guide on the documentation - a series of "rules" to ensure consistency in the docs and articles that are easy to read.

The style guide is implemented using Vale, and every time a commit is made to the UmbracoDocs, it will be checked against all of our rules. The [rules can be found right here on the repository](https://github.com/umbraco/UmbracoDocs/tree/main/.github/styles/UmbracoDocs), and they include rules that recommends against the use of certain words, a rule that ensures that acronyms are defined and rules that ensures consistent spelling of certain terms.

How do I help out?
-----------------------------------

In order to make the task a little more edible, we've divided the docs into some sections.

**Write a comment** to this Issue, telling us which section you'd like to check, and make sure to reference this issue when you make the PR.
That way we can keep track of the progress, and other contributors will know which sections are already done or being worked on 💪 

- [x] Getting Started
- [ ] Tutorials (all!)
- [ ] Umbraco Forms
- [ ] Umbraco Deploy
- [ ] Umbraco Plumber

**Fundamentals**
- [ ] Setup
- [x] Backoffice
- [ ] Data
- [ ] Design
- [ ] Code

**Implementation**
- [x] Default and Custom routing
- [x] Controllers and Data Persistence
- [x] Composing and Services
- [ ] Integration Testing, Unit Testing and Nullable Reference Types

**Extending**
- [ ] Dashboard, Section Trees and Property Editors
- [ ] Health Checks and Language Files
- [ ] Backoffice Search, Tours and API
- [ ] Content App and Database
- [ ] Embedded Media Provider and FileSystemProviders
- [ ] Packages

**Reference**
- [ ] Configuration
- [ ] Templating
- [ ] Querying
- [ ] Routing
- [ ] Searching
- [ ] Management
- [ ] Plugins and Cache
- [ ] Debugging and Language Variation
- [ ] Notifications
- [ ] Scheduling and Using IoC

**Umbraco Cloud**
- [ ] Getting Started
- [ ] Setup
- [ ] Deployment
- [ ] Databases, Upgrades and Security
- [ ] Troubleshooting and Frequently Asked Questions

erikjanwestendorp:
I would like to work on the Fundamentals part. 😄

sofietoft:
[...]

PR #4450, opened 2022-10-11
Some readability improvements

erikjanwestendorp:
Some readability improvements. Ok I think I have to do this in a more structured way.. 😛

Belongs to: https://github.com/umbraco/UmbracoDocs/issues/4416


sofietoft (commented):

Phew those long sentences can be tough to figure out - I think I did an OK job though 🙈 
Could you take an extra look, to make sure I didn't loose the meaning of the sections with my edits?

Also, just a quick little side-note:
We're actively looking into migrating the documentation to a new platform. In this process, we will be leaving docs for 6-8 behind on Our. So, in case you're gonna look more into readability, I just wanted to highlight that you can skip articles for these versions if you want 😄

erikjanwestendorp:
@sofietoft Thanks for your review. Just committed your suggestions, they seems good to me!

I wanted to look at some readability so good to know! 😄

sofietoft:
That sounds wonderful @erikjanwestendorp ! 😄 Thanks for helping out - as always!

Let's get this one merged in then! 💪

PR #4452, opened 2022-10-11
Readability on Where can I get help.

busrasengul:
The previous version implied contacting directly to the Team but the link was for the Docs issue tracker.  I thought it would be better when consistent with the CMS issue tracker.


sofietoft:
Thanks for adding these changes @busrasengul !
I think it makes perfect sense 💪

PR #4453, opened 2022-10-11
Update HSTS health check guide based on ASP.NET docs

JasonElkin:
While looking at https://github.com/umbraco/Umbraco-CMS/pull/13166 I noticed that this guide does not use the recommended approach to enabling HSTS in ASP.NET Core.

I've updated the guide to use the recommended extension method built into ASP.NET Core.


sofietoft:
Hi @JasonElkin !

Thanks a bunch for putting in the work and adding this PR 💪 
I'll run it by our developers for a check-up!

nikolajlauridsen (approved):
Looks fine to me 👍 I mean it's what Microsoft recommends, so makes sense that's what we recommend too 😄

sofietoft:
Thanks for the review @nikolajlauridsen ! 🙌 

And thanks again @JasonElkin , for adding this to the docs - Much appreciated.
Let's get this merged in then, yeah? 😄

PR #4455, opened 2022-10-11
Update block grid docs

sussexrick:
'wether' is a misspelling of 'whether'

'neglect' didn't make sense in this context - to neglect to do something is a less deliberate action than Umbraco absolutely refusing to move a block to an incompatible area


sofietoft:
Hi @sussexrick ! Thanks a bunch for the PR! 🥳 

These are some great additions - I completely agree with the use of "neglect". Definitely an oversight from us, so thanks for catching this one 😄 

I'll merge this in right away!
And congratz on being one step closer to completing the Hacktoberfest challenge 😎

PR #4462, opened 2022-10-12
Updated details about fixed typo in the forms AllowEditableFormSubmissions configuration setting.

AndyButland:


sofietoft (approved):
Thanks for updating this so quickly @AndyButland ! 💪 🙌 

I just made a tiny change to use our note format instead.
Will merge it right in!

PR #4463, opened 2022-10-12
Add V8 version of Nucache settings

### LINKED ISSUE #4456: Can we add instructions to disable the NuCache file generation on Umbraco 8?
***This is the page with issues: https://github.com/umbraco/UmbracoDocs/edit/main/Reference/Configuration/NuCacheSettings/index.md***

The sample code to disable on Umbraco 8 is as follows:

```csharp
    public class DisableNuCacheDatabaseComposer : IComposer
    {
        public void Compose(Composition composition)
        {
            var settings = new Umbraco.Web.PublishedCache.NuCache.PublishedSnapshotServiceOptions
            {
                IgnoreLocalDb = true
            };
            composition.Register(factory => settings, Lifetime.Singleton);
        }
    }
```

marcemarc:
We mention how to set ignorelocaldb for V8 in load balancing docs but now we have a specific NuCache settings page for V9+ makes sense to do a V8 version here?

Fixes #4456

### DIFF
--- a/Reference/Configuration-for-Umbraco-7-and-8/nucache-settings.md
+++ b/Reference/Configuration-for-Umbraco-7-and-8/nucache-settings.md
@@ -7,25 +7,23 @@ meta.Description: "Information on the NuCache settings section"
 
 # NuCache Settings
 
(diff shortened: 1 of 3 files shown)


sofietoft:
Thanks a bunch for the PR and for your input on this @marcemarc ! 💪 

I do not think this should be added to the "Reference/Configuration" section as this is for docs on configuration for v9+.
Instead, the article should be added somewhere in the "Reference/Configuration-for-7-and-8" section.

There should be links between the articles for 9/10 and 8 of course, but that should be easy enough to solve.

Would you be up for moving the article?
I'd be happy to do it as well - just let me know 😄

PR #4464, opened 2022-10-12
Fix link to Image Cropper v7 page

greystate:
I was on the "Built-in Property Editors" page for v7 and clicked the Image Cropper link, and noticed I ended up on the "v10 (current)" version, so did some digging to see if there was a version for v7. Lo and behold, I found it, so decided to fix the link 😀


sofietoft:
Good morning @greystate 🌞 

Thanks a bunch for the PR! Good to have those broken links eliminated 💪

PR #4467, opened 2022-10-13
Update Fundamentals -- Login -- index.md

erikjanwestendorp:
Update Fundamentals -- Login to improve readability. 

Belongs to:

https://github.com/umbraco/UmbracoDocs/issues/4416 

- [x] Add SMTP to acronyms exceptions?
- [x] Write shorter sentence
- [x] changes css to CSS

### DIFF
--- a/Fundamentals/Backoffice/Login/index.md
+++ b/Fundamentals/Backoffice/Login/index.md
@@ -49,7 +49,7 @@ The **Forgotten password?** link allows your backoffice users to reset their pas
 
 Set it to `true` to enable the password reset feature, and `false` to disable the feature.
 
-You will also need to configure an SMTP server in your `appsettings.json` file. When you get a successful result on the SMTP configuration when running a health check in the backoffice, you are good to go!
+You will also need to configure an SMTP: Simple Mail Transfer Protocol, server in your `appsettings.json` file. When you get a successful result on the SMTP configuration when running a health check in the backoffice, you are good to go!
 
 An example:
 
--- a/Fundamentals/Backoffice/Login/index.md
+++ b/Fundamentals/Backoffice/Login/index.md
@@ -31,7 +31,7 @@ The login screen features a greeting which you can personalize by overriding the
     <key alias="greeting6">Happy Caturday</key>
 </area>
 ```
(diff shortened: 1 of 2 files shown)


Fundamentals/Backoffice/Login/index.md
```diff
@@ -31,7 +31,7 @@ The login screen features a greeting which you can personalize by overriding the
     <key alias="greeting6">Happy Caturday</key>
 </area>
 ```
-You can customize other text in the login screen as well, grab the default values and keys from `~/umbraco/config/lang/en.xml` and copy the ones you want to translate into `~/config/lang/en_us.user.xml` file. Note: the new /config/ folder needs to be created at the site root.
+You can customize other text in the login screen as well, first grab the default values and keys from `~/umbraco/config/lang/en.xml`. Thereafter copy the ones you want to translate into `~/config/lang/en_us.user.xml` file. Note: the new /config/ folder needs to be created at the site root.
```
sofietoft, line 34:
```suggestion
You can customize other text on the login screen as well. First, grab the default values and keys from `~/umbraco/config/lang/en.xml`. Thereafter copy the ones you want to translate into `~/config/lang/en_us.user.xml` file. Note: the new /config/ folder needs to be created at the site root.
```

sofietoft (commented):

I've just made a small edit to the sentence and shortened it even more. Let me know what you think!

And then, about "SMTP"... I think that would actually be one of the acronyms we should define in the docs! It's often mentioned in articles with a target audience that are not necessarily developers and might not know this acronym 🤔 
Would you be up for defining the acronym, instead af adding it to the exceptions list? 😁 
We have some docs on the formats to use in the [Contribution guidelines](https://our.umbraco.com/documentation/Contribute/Style-Guide/#defined-acronyms).

erikjanwestendorp:
@erikjanwestendorp Thanks for your review. Just updated the PR, but now there is a warning about the word Simple 🙈

sofietoft:
Thanks for updating @erikjanwestendorp ! 💪 

We can ignore this warning, as "Simple" is actually a necessary word in this case 🙈 
I mean, the rules aren't perfect 😄 🤪 

I'll get this merged in 💪

PR #4471, opened 2022-10-14
Added video link...

eshanrnh:
...  and fixed markdown errors.

The video is currently private and this PR can be merged in on Thursday, 20/10/22.

### DIFF
--- a/Add-ons/UmbracoForms/Editor/Attaching-Workflows/index.md
+++ b/Add-ons/UmbracoForms/Editor/Attaching-Workflows/index.md
@@ -43,7 +43,7 @@ A new workflow can be of different types and Umbraco Forms ships with a few defa
 
 ### Update Type-specific Settings
 
-Once the Workflow Type has been selected, you will need to configure the workflow. There are settings depending on the type that has been selected.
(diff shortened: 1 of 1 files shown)


Add-ons/UmbracoForms/Editor/Attaching-Workflows/index.md
```diff
@@ -39,7 +43,7 @@ A new workflow can be of different types and Umbraco Forms ships with a few defa
 
 ### Update Type-specific Settings
 
-Once the Workflow Type has been selected, you will need to configure the workflow. There are various settings depending on the type that has been selected.
+Once the Workflow Type has been selected, you will need to configure the workflow. There are settings depending on the type that has been selected.
```
sofietoft, line 46: I think that removing "various" in this case actually takes away some of the meaning. So how about adding "different" instead?

```suggestion
Once the Workflow Type has been selected, you will need to configure the workflow. There are different settings depending on the type that has been selected.
```

sofietoft (commented):
Looks great 💪 
Just added a single suggestion!

PR #4474, opened 2022-10-14
Updated Examine docs for V8

poornimanayar:
Notied that there is no mention of ValueSetBuilder and IndexPopulator in V* docs. So added a link to the v9 example with update to the V8 composer on how to register the implementations

### DIFF
--- a/Reference/Searching/Examine/Indexing/index-v8.md
+++ b/Reference/Searching/Examine/Indexing/index-v8.md
@@ -298,8 +298,8 @@ public class ProductIndexCreator : LuceneIndexCreator, IUmbracoIndexesCreator
 }
 ```
 
-You can create an `IValueSetBuilder` implementation that builds the value sets for the index following the example [here](index.md/productindexvaluesetbuilder).
-You can also create an `IndexPopulator` implementation that populates the index with the value sets following the example [here](index.md/productindexpopulator).
+You can create an `IValueSetBuilder` implementation that builds the value sets for the index following the [example for Umbraco 9](index.md/#productindexvaluesetbuilder).
+You can also create an `IndexPopulator` implementation that populates the index with the value sets following the [example for Umbraco 9](index.md/#productindexpopulator).
 
 ### ProductComponent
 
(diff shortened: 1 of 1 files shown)


Reference/Searching/Examine/Indexing/index-v8.md
```diff
@@ -298,6 +298,9 @@ public class ProductIndexCreator : LuceneIndexCreator, IUmbracoIndexesCreator
 }
 ```
 
+You can create an `IValueSetBuilder` implementation that builds the value sets for the index following the example [here](index.md/productindexvaluesetbuilder).
+You can also create an `IndexPopulator` implementation that populates the index with the value sets following the example [here](index.md/productindexpopulator).
```
sofietoft, line 302: We prefer to not use "here" as link texts, so I've tried to amend your suggestions a bit.
Let me know what you think!

```suggestion
You can create an `IValueSetBuilder` implementation that builds the value sets for the index following the [example for Umbraco 9](index.md/#productindexvaluesetbuilder).
You can also create an `IndexPopulator` implementation that populates the index with the value sets following the [example for Umbraco 9](index.md/#productindexpopulator).
```

sofietoft (commented):
Thanks for the PR @poornimanayar ! 💪 

So, just to confirm - these things you're adding to the Umbraco 8 article.. is that in fact something that one can do on Umbraco 8?
If it is indeed possible in both versions, why is it not just added to both version of the article as well? 🤔 

I hope it makes sense!

poornimanayar:
Yes, this is something achievable in V8. I wasn't sure a repetition of code is welcome. So I took this approach, I can copy the code to V8 as well. I agree, it should be there for V8.

sofietoft:
The code should be there in both versions then, I think!

We will soon be moving all docs for 9+ to a new platform, and I think that makes it even more important that the code is added to both versions this article!

poornimanayar:
Let me add it then @sofietoft. More PRs incoming other than this.

sofietoft:
Thanks for updating the Umbraco 8 article @poornimanayar ! 💪 
Much appreciated.

I'll merge it in!

PR #4477, opened 2022-10-15
Add instructions on migrating members from Umbraco 8 to 10

sussexrick:
When I migrated my site from Umbraco 8 to 9 I followed this guide, but I found that migration of members was not covered. I asked Cloud support and they were only able to tell me there was no tool to support it, and I would have to do it on my own with database queries. It seemed as if I was the first one to ask, yet this must be a common requirement. This documents the process I ended up using, which worked for my site.

### DIFF
--- a/Umbraco-Cloud/Upgrades/Migrating-from-8-to-10/index.md
+++ b/Umbraco-Cloud/Upgrades/Migrating-from-8-to-10/index.md
@@ -144,36 +144,47 @@ If you have Members in your Umbraco database these were not transferred when you
    SELECT MAX(id) FROM umbracoContentVersion
    SELECT MAX(id) FROM umbracoPropertyData
    ```
-* Use the SQL Server Import and Export Wizard to query data from your Umbraco 8 database and insert it into your Umbraco 10 database. You can start the wizard by right-clicking any database and selecting Tasks > Import Data... Use the 'Microsoft OLE DB Driver for SQL Server' to connect first to your Umbraco 8 database as the data source, then your Umbraco 10 database.
+* Use the SQL Server Import and Export Wizard to query data from your Umbraco 8 database and insert it into your Umbraco 10 database. You can start the wizard by right-clicking any database and selecting Tasks > Import Data... 
+ 
+    Use the 'Microsoft OLE DB Driver for SQL Server' to connect first to your Umbraco 8 database as the data source, then your Umbraco 10 database as the destination. Next, select 'Write a query to specify the data to transfer', and use the following queries. You will need to run this wizard again from the start for each query. 
 
-You will need to run this wizard once for each of the following queries. 
-
-Replace `{umbracoNode}` and `{umbracoContentVersion}` with the results from your previous query.
+    In this query replace `{umbracoNode}` and `{umbracoContentVersion}` with the results from your previous query. After you enter your query in the wizard you'll see the 'Select Source Tables and Views' page of the wizard. Change the destination from `[dbo].[Query]` to `[dbo].[umbracoNode]`. Click 'Edit Mappings...` and tick 'Enable identity insert'. You're now ready to complete the wizard.
 
    ```
-   -- Insert the results of this query on your Umbraco 8 database into the umbracoNode table on your Umbraco 10 database. This requires you to enable Identity Insert in the wizard options.
    SELECT id + {umbracoNode} AS id, uniqueid, parentid, level, path, sortOrder, trashed, -1 AS nodeUser, text, nodeObjectType, createDate 
    FROM umbracoNode 
    WHERE nodeObjectType IN ('39EB0F98-B348-42A1-8662-E7EB18487560', '366E63B9-880F-4E13-A61C-98069B029728')
+   ```	
+   
+* Run the wizard again and select your Umbraco 8 database as the data source, then your Umbraco 10 database as the destination. Enter the query below. Replace `{umbracoNode}` with the result from your earlier query. Change the destination to `[dbo].[umbracoContent]`, and this time you don't need to select 'Enable identity insert'.
+ 
+    ```
+    SELECT nodeId + {umbracoNode} AS nodeId, contentTypeId 
+    FROM umbracoContent 
+    WHERE nodeId IN (SELECT nodeId FROM cmsMember)
+	```
+    
+* Run the wizard again and select your Umbraco 8 database as the data source, then your Umbraco 10 database as the destination. Enter the query below. Replace `{umbracoNode}` with the result from your earlier query. Change the destination to `[dbo].[cmsMember]`, and this time you don't need to select 'Enable identity insert'.
+
+    ```
+    SELECT nodeId + {umbracoNode} AS nodeId, Email, LoginName, Password 
+    FROM cmsMember
+    ```
 	
-   -- Run the wizard again and insert the results of this query on your Umbraco 8 database into the umbracoContent table on your Umbraco 10 database.
-   SELECT nodeId + {umbracoNode} AS nodeId, contentTypeId 
-   FROM umbracoContent 
-   WHERE nodeId IN (SELECT nodeId FROM cmsMember)
+* Run the wizard again and select your Umbraco 8 database as the data source, then your Umbraco 10 database as the destination. Enter the query below. Replace `{umbracoContentVersion}` and `{umbracoNode}` with the results from your earlier queries. Change the destination to `[dbo].[umbracoContentVersion]`. This time you **do** need to select 'Enable identity insert'.
+
+    ```
+    SELECT id + {umbracoContentVersion} AS id, nodeId + {umbracoNode} AS nodeId, versionDate, userId, [current], [text] 
+    FROM umbracoContentVersion 
+    WHERE nodeId IN (SELECT nodeId FROM cmsMember) 
+    ```
 	
-   -- Run the wizard again and insert the results of this query on your Umbraco 8 database into the cmsMember table on your Umbraco 10 database.
-   SELECT nodeId + {umbracoNode} AS nodeId, Email, LoginName, Password 
-   FROM cmsMember
-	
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks a bunch for the PR @sussexrick ! 💪 

We do write in the "Requirements" part of the article, that all Member data needs to be migrated manually after completing the migration process:

> - Contains no member data
>    - If you do have member data, these will need to be imported manually after the migration"

But it would be super nice, to actually show a way to do this!

Perhaps we should add your steps here to a separate article instead? 🤔 
The migration article is quite long already 🙈 

I'm sending this to our Cloud expert to have it tested etc 💪 
Just to clarify - this can be done after the rest of the migration is done, right?

jonat123:
Hey @sussexrick I have been working on testing out your article and so far I have made most of it work, however, I have hit a roadblock with the last part:

 _Now you need to transfer the properties that belong to the Members. Run the following query on both your Umbraco 8 database and your Umbraco 10 database and note the results, which will probably be different between the two:
    ```
    SELECT id, Alias
    FROM cmsPropertyType
    WHERE contentTypeId = (SELECT nodeId FROM cmsContentType WHERE alias = 'Member')  
    AND Alias NOT LIKE 'umbracoMember%' 
    ORDER BY Alias
    ```_

I've run the query on both my v8 and v10 projects database, however, once the query finishes nothing shows up 🤔

Is there something I have missed?

sussexrick:
@jonat123 I get a result like this. 
image

jonat123:
@sussexrick Thanks!

I'll do some more testing and see if i can verify the steps :)

jonat123:
@sussexrick Okay, I have tested a bit more and I am still unable to see anything when running that query.

When that is said, I followed the guide up to that step and when I look in the backoffice I can see that the members and member groups have been migrated successfully to my environment.

I am therefore wondering if the last is necessary?

I also tested whether it is possible to log in with the member and that works as well using the partial view snippets for login.

jonat123:
@sussexrick I've been looking a bit more into it and is it correct to understand that it is if you have made custom changes to the member types, then the query is supposed to find the alias for those?

Is that correctly understood?

sussexrick:
@jonat123 Yes, you're right, those are all my custom properties.

jonat123:
@sussexrick Great!

Thanks for letting me know :D 

I'll get back to testing :)

Umbraco-Cloud/Upgrades/Migrating-from-8-to-10/index.md
```diff
@@ -120,7 +120,113 @@ Read more about these changes in the [IPublishedContent section of the Documenta
...
+	DELETE FROM umbracoNode where nodeObjectType IN ('39EB0F98-B348-42A1-8662-E7EB18487560', '366E63B9-880F-4E13-A61C-98069B029728')
+    ```
+
+* In your Umbraco 10 database, run the following queries and make a note of the results as you will need them for later queries:
+
+   ```
+   SELECT MAX(id) FROM umbracoNode
+   SELECT MAX(id) FROM umbracoContentVersion
+   SELECT MAX(id) FROM umbracoPropertyData
+   ```
```
jonat123, line 146: This query should be run on the local database that we have taken from the v8 project and used on our v10 project right?

Otherwise two out of the three values are just gonna be NULL as our v10 DB is empty on the cloud environment.

jonat123:
Okay, @sussexrick I have done some more testing and unfortunately, I am unable to fully verify the guide that you have written.


Most of the article works as I am able to migrate my members and member groups from my v8 project to V10 and I can see them in the backoffice and they are able to log in, which is great!

However, the last part where we transfer the custom properties that belong to the Members does not seem to work on my end.

First of all the first query:

   ```
 SELECT id, Alias
    FROM cmsPropertyType
    WHERE contentTypeId = (SELECT nodeId FROM cmsContentType WHERE alias = 'Member')  
    AND Alias NOT LIKE 'umbracoMember%' 
    ORDER BY Alias
```

When running on the v 10 database is not showing anything.

This does make sense to me as the properties that we are looking for are from the Member Types, which can be edited in the settings section, they need to be pushed up to the cloud environment before being able to see them

Once that has been pushed up I am able to see the ID's for the properties.


Then when trying to run the next query:

```
SELECT id + {umbracoPropertyData} AS id, versionid + 29 AS versionid,  
	CASE WHEN propertyTypeId = {umbraco 8 id for alias approvalToken} THEN {umbraco 10 id for alias approvalToken} 
		 WHEN propertyTypeId = {umbraco 8 id for alias approvalTokenExpires} THEN {umbraco 10 id for alias approvalTokenExpires} 
		 WHEN propertyTypeId = {umbraco 8 id for alias blockLogin} THEN {umbraco 10 id for alias blockLogin} 
                 WHEN propertyTypeId = {umbraco 8 id for alias migratedMemberId} THEN {umbraco 10 id for alias migratedMemberId} 
		 WHEN propertyTypeId = {umbraco 8 id for alias passwordResetToken} THEN {umbraco 10 id for alias passwordResetToken} 
		 WHEN propertyTypeId = {umbraco 8 id for alias passwordResetTokenExpires} THEN {umbraco 10 id for alias passwordResetTokenExpires} 
		 WHEN propertyTypeId = {umbraco 8 id for alias requestedEmail} THEN {umbraco 10 id for alias requestedEmail} 
		 WHEN propertyTypeId = {umbraco 8 id for alias requestedEmailToken} THEN {umbraco 10 id for alias requestedEmailToken} 
		 WHEN propertyTypeId = {umbraco 8 id for alias requestedEmailTokenExpires} THEN {umbraco 10 id for alias requestedEmailTokenExpires} 
		 WHEN propertyTypeId = {umbraco 8 id for alias totalLogins} THEN {umbraco 10 id for alias totalLogins}
  ELSE propertyTypeId END AS propertyTypeId, 
  languageId, segment, intValue, decimalValue, dateValue, varcharValue, textValue
  FROM umbracoPropertyData
  WHERE versionid IN (SELECT id FROM umbracoContentVersion WHERE nodeId IN (SELECT nodeId FROM cmsMember))
```


It fails with the error: "The INSERT statement conflicted with the FOREIGN KEY constraint `"FK_umbracoPropertyData_cmsPropertyType_id"."`

Unfortunately, I am not sure where to go from there, do you maybe have an idea to what the issue might be?


We also talked a bit about the placement of this guide and where we think it would fit best.

Since everyone might not be using members, so having it as a specific step in the guide does not make sense to us, instead, it should be a separate article under the migration guide, and then make a note in the main article that refers to the member's migration guide.

We also feel like migrating members is something that should be done as one of the last things, when most of the things have been migrated over then they can look into migrating the members and groups if they are using it.

**Comment by sussexrick** [comment]
@jonat123 Let me do some more testing today and I'll get back to you. Some discrepancies might be because the notes I based this on were from upgrading to v9, not v10. I still have a local copy of the v8 database I started from.

File: Umbraco-Cloud/Upgrades/Migrating-from-8-to-10/index.md (hunk 1 above)
Comment (sussexrick) [reply Umbraco-Cloud/Upgrades/Migrating-from-8-to-10/index.md:146]: Good point, it does need to be clearer. Reading the whole guide there are at least three databases in play - v8 on Cloud, v8 upgraded to 10 locally, and v10 on Cloud. 

The local one may or may not have members already depending on whether it was created from a backup or by cloning down content (Step 1, first bullet), so actually the right source database is v8 on Cloud, the right target database is v10 on Cloud, and the right time is within Step 5, after "Transfer Content and Media from the local clone to the Development environment." and before "Test everything in the Development environment." Do you agree?

**Comment by sussexrick** [comment]
@jonat123 I'm happy to move it to a separate page, but let's get your testing done first. 

I replicated your error. It indicates that you got one of the `{umbraco 10 id for alias xxx}` values wrong. Get the values right and it does work.

I've updated that last part about transferring custom properties so that it handles custom properties for any member type, not just 'Member', and so that it describes how to build up a query for whatever custom properties are there rather than hard-coding mine.

**Comment by jonat123** [comment]
@sussexrick Great!

Will test it out again sometime in our current sprint (This week or next) and let you know the result 😊

**Comment by jonat123** [comment]
@sussexrick Okay, so I've tested out the guide again with the updated steps.

Unfortunately, I am still unable to run the last query for the custom data for member types... It is failing
 ```
with The INSERT statement conflicted with the FOREIGN KEY constraint "FK_umbracoPropertyData_cmsPropertyType_id". The conflict occurred in database "risoizzlruf", table "dbo.cmsPropertyType", column 'id'.".
 (SQL Server Import and Export Wizard)
```

Just so we are on the same page, i connected to both my v8 and v10 cloud databases, ran the query to get the different id's, which looks like this:

V8:

image

and the V10:

image

I then created the new query based on your example, which looks like this:

```
SELECT id + 2 AS id, versionid + 2 AS versionid,  
CASE 
WHEN propertyTypeId = 53 THEN 54 
WHEN propertyTypeId = 65 THEN 60 
WHEN propertyTypeId = 64 THEN 61  
WHEN propertyTypeId = 63 THEN 62
WHEN propertyTypeId = 66 THEN 63 
ELSE propertyTypeId END AS propertyTypeId, 
languageId, segment, intValue, decimalValue, dateValue, varcharValue, textValue
FROM umbracoPropertyData
WHERE versionid IN (SELECT id FROM umbracoContentVersion WHERE nodeId IN (SELECT nodeId FROM cmsMember))
```

To me this all looks correct, however, the query still fails 😕

Unfortunately, I don't think that I will get much further at this point, and think we will see if we can get one of our developers to try and take a look at it to see if they can help us get this verified :)

jonat123:
@sussexrick I Just want to let you know that we recently updated the documentation with the deploy setting for AllowMembersDeploymentOperations and TransferMemberGroupsAsContent which lets you deploy members between environments  😄

This makes this PR redundant and I will close it down :)

jonat123:
And the documentation for those settings can be found here:
- [AllowMembersDeploymentOperations and TransferMemberGroupsAsContent](https://docs.umbraco.com/umbraco-deploy/deploy-settings#allowmembersdeploymentoperations-and-transfermembergroupsascontent)

PR #4478, opened 2022-10-16
Configuring Azure Key Vault via Azure Web App

mattou07:
Another method to incorporate Key Vault into an Umbraco application running on an Azure Web App

PR #4489, opened 2022-10-18
Fix the KeyVault code sample

Jeavon:
CreateHostBuilder updated to v10 (it's currently the v9 version) and the Constants.Azure.AzureKeyVaultEndpoint doesn't exist so it should be a string in the sample.

### DIFF
--- a/Extending/Key-Vault/index.md
+++ b/Extending/Key-Vault/index.md
@@ -25,7 +25,7 @@ dotnet add package Azure.Identity
 ```
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


sofietoft (approved):
Looks great @Jeavon ! 💪 

This will cause some merge conflicts on this one though: https://github.com/umbraco/UmbracoDocs/pull/4478 🙈

Jeavon:
I don't think it will @sofietoft 🙏

PR #4491, opened 2022-10-18
cleaned up code, removed li tag

abhamla:
li tag not needed when not looping through links (max == 1)


sofietoft:
Thanks for the PR @abhamla ! 💪 

Makes sense to remove the `<li>` tags here, as there's also no wrapping `<ul>` tag!
Great catch! 😄

PR #4493, opened 2022-10-19
Examine Searchers for v10

iOvergaard:
Update screenshot and description about examine searchers for v10+.


iOvergaard:
@sofietoft > I think this article should actually cover from v9 to v10+

There's another article [examine-management-v7.md](https://github.com/umbraco/UmbracoDocs/blob/main/Reference/Searching/Examine/examine-management-v7.md) which probably should cover both v7 and v8 in that case.

Any thoughts on that? :)

sofietoft:
Thanks for the PR @iOvergaard ! 💪 

I agree that the two versions of the articles should cover 9-10 and 7-8 🤔 
I don't believe that there are that many changes between the versions though, but it still would be good to get cleared up!

iOvergaard:
I tried to add the version ranges, @sofietoft. Hope I did it correctly!

PR #4494, opened 2022-10-19
Fix custom error page config formatting

callumbwhyte:
The format of the configuration examples on the "Custom Error Page" docs is a bit out of whack... I've fixed it!

Screenshot 2022-10-19 at 15 28 32


eshanrnh (approved):
Thanks @callumbwhyte for fixing the code block 💪

PR #4498, opened 2022-10-20
Update config sample

sussexrick:
The provided config sample was out-of-date. It doesn't match what you get in Umbraco 10.3 and it doesn't match the description in the second paragraph after the "Specifying minimum log level" heading.


eshanrnh (approved):
Thanks for updating the config sample, @sussexrick 💪

PR #4517, opened 2022-10-24
Added some bold to the fields

Jette94:


sofietoft (approved):
Looks great @Jette94 ! 💪 
Thanks for improving the consistency of the uses of italics and bolds in this article - much appreciated!

PR #4519, opened 2022-10-24
V11: Update LogViewer docs 

nikolajlauridsen:
This PR updates the LogViewer docs to use the newer version of the Azure Table Storage client, since the old one is obsolete, and also cleans up/improves the code a bit. 

I've tested this on both V10 and V11 and both work fine now.


Fundamentals/Backoffice/LogViewer/index.md
```diff
@@ -35,124 +35,94 @@ With the flexibility of Umbraco, we give you the power to implement your own `IL
 
 ### Create your own implementation
 To do this we can implement a base class `SerilogLogViewerSourceBase` from `Umbraco.Cms.Core.Logging.Viewer` like so.
-*Note:* This uses the `WindowsAzure.Storage` nuget package
+*Note:* This uses the `Azure.Data.Tables` nuget package
```
sofietoft, line 38:
```suggestion
:::note
This uses the `Azure.Data.Tables` NuGet package.
:::
```

Fundamentals/Backoffice/LogViewer/index.md
```diff
@@ -35,124 +35,94 @@ With the flexibility of Umbraco, we give you the power to implement your own `IL
...
+
+public class AzureTableLogEntity : LogEventEntity, ITableEntity
+{
+    public DateTimeOffset? Timestamp { get; set; }
+
+    public ETag ETag { get; set; }
+}
 ```
 
+Do note that we have to implement our own version of a `LogEventEntity` this is beacuse the `TableClient` needs whatever it's fetching to implement the `ITableEntity` interface.
```
sofietoft, line 110:
```suggestion
Keep in mind that we have to implement our own version of a `LogEventEntity`. This is because the `TableClient` needs whatever it is fetching to implement the `ITableEntity` interface.  
```

Fundamentals/Backoffice/LogViewer/index.md
```diff
@@ -35,124 +35,94 @@ With the flexibility of Umbraco, we give you the power to implement your own `IL
...
+    public DateTimeOffset? Timestamp { get; set; }
+
+    public ETag ETag { get; set; }
+}
 ```
 
+Do note that we have to implement our own version of a `LogEventEntity` this is beacuse the `TableClient` needs whatever it's fetching to implement the `ITableEntity` interface.  
+
 ### Register implementation
 Umbraco needs to be made aware that there is a new implementation of an `ILogViewer` to register and replace the default JSON LogViewer that we ship in the core of Umbraco.
```
sofietoft, line 113:
```suggestion
Umbraco needs to be made aware that there is a new implementation of an `ILogViewer` to register. We also need to replace the default JSON LogViewer that we ship in the core of Umbraco.
```

sofietoft (commented):
Thanks for verifying this article @nikolajlauridsen ! 💪 

I've added a couple of suggestions.
Could you take a look at some of the sentences I've tried to shorten? Just to ensure they still hold the same meaning 😅 

Also, there are a few "TODO"s in the article 🤔 I'm wondering if they can be taken out?

nikolajlauridsen:
Suggestions looked good to me, reads a bit easier but still says the same 😄 

In regards to the TODOs, I think it makes sense that they're still there since these methods aren't implemented to use Azure Table (They're all for saved searches), however, saved searches are by default stored in the database, and honestly, I'm not entirely sure why you would want them in table storage, but I'm sure there's a use case.

But I can see that it might be a bit confusing that they're called TODO, so it might make sense to instead have them say something like "If you want to store saved storage in Azure Table Storage, implement these methods"

Alternatively, we could just implement these methods in the example, but I'm a bit afraid that this might lead people to think they must implement these, which are not the case.

sofietoft:
Aha. I see! So it's not something that needs to be added but something that is more "optional" 🤔 
How about we replace the TODOs with these instead:

```
//Optional: If you want to store saved searches in the Azure Table Storage, implement here a method to fetch from the Azure Table.

//Optional: If you want to store saved searches in the Azure Table Storage, implement here a method to add to the Azure Table.

//Optional: If you want to store saved searches in the Azure Table Storage, implement here a method to remove from the Azure Table
```

(I took what you wrote, but replaced "saved storage" with "saved searches" which is what you wrote in the section above that. Did I do it right, or should it actually be "saved storages"? 🙈 )

nikolajlauridsen:
Yes exactly, and you got it exactly right 😄 writing "saved storage" was just an oopsie from my side, table storage got into my head haha

Fundamentals/Backoffice/LogViewer/index.md
```diff
@@ -42,125 +42,97 @@ With the flexibility of Umbraco, we give you the power to implement your own `IL
...
+            .Skip(skip)
+            .Take(take)
+            .Select(x => LogEventReader.ReadFromString(x.Data))
+            .Where(filter.TakeLogEvent)
+            .ToList();
+    }
+
+    public override IReadOnlyList<SavedLogSearch>? GetSavedSearches()
+    {
+        //TODO: Fetch from Azure Table
```
sofietoft, line 94:
```suggestion
        // Optional: If you want to store saved searches in the Azure Table Storage, implement here a method to fetch from the Azure Table.
```

Fundamentals/Backoffice/LogViewer/index.md
```diff
@@ -42,125 +42,97 @@ With the flexibility of Umbraco, we give you the power to implement your own `IL
...
-            //TODO: Add to Azure Table
-            return base.AddSavedSearch(name, query);
-        }
-
-        public override IReadOnlyList<SavedLogSearch> DeleteSavedSearch(string name, string query)
-        {
-            //TODO: Remove from Azure Table
-            return base.DeleteSavedSearch(name, query);
-        }
+        //TODO: Add to Azure Table
```
sofietoft, line 100:
```suggestion
        //Optional: If you want to store saved searches in the Azure Table Storage, implement here a method to add to the Azure Table.
```

Fundamentals/Backoffice/LogViewer/index.md
```diff
@@ -42,125 +42,97 @@ With the flexibility of Umbraco, we give you the power to implement your own `IL
...
+        return base.AddSavedSearch(name, query);
     }
+
+    public override IReadOnlyList<SavedLogSearch>? DeleteSavedSearch(string? name, string? query)
+    {
+        //TODO: Remove from Azure Table
```
sofietoft, line 106: //Optional: If you want to store saved searches in the Azure Table Storage, implement here a method to remove from the Azure Table.

sofietoft (commented):
Final suggestions @nikolajlauridsen ! 🤞 😄

nikolajlauridsen:
Amazing, looks good to me 👍

sofietoft:
Grrrrreat 😄 Thanks again 🙌 

I'll merge it in right away!

PR #4521, opened 2022-10-25
Make Zapier Tutorial more friendly

OwainWilliams:
- There are a couple of sentences I just couldn't find a way to get in to the 25 word limit. I think if they were shortened at all they would lose their meaning.

### DIFF
--- a/Tutorials/Connecting-Umbraco-Forms-and-Zapier/index.md
+++ b/Tutorials/Connecting-Umbraco-Forms-and-Zapier/index.md
@@ -133,7 +133,10 @@ Now you can choose which Google Drive to use, find the spreadsheet and choose th
 
 ![Zapier Google Sheets Possible Fields](images/zapierGoogleSheetsPossibleFields.png)
 
-The fields showing are all columns in your spreadsheet that have a name in row 1. To map the input data to the different fields in the spreadsheet, you click on the “Type or insert…” field and choose which data to put in that field.
+The fields showing are all columns in your spreadsheet that have a name in row 1. To map the input data to the different fields in the spreadsheet follow these steps:
+
+* Select the “Type or insert…” field.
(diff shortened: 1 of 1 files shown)


Tutorials/Connecting-Umbraco-Forms-and-Zapier/index.md
```diff
@@ -133,7 +133,7 @@ Now you can choose which Google Drive to use, find the spreadsheet and choose th
 
 ![Zapier Google Sheets Possible Fields](images/zapierGoogleSheetsPossibleFields.png)
 
-The fields showing are all columns in your spreadsheet that have a name in row 1. To map the input data to the different fields in the spreadsheet you now click on the “Type or insert…” field and choose which data to put in that field.
+The fields showing are all columns in your spreadsheet that have a name in row 1. To map the input data to the different fields in the spreadsheet, you click on the “Type or insert…” field and choose which data to put in that field.
```
sofietoft, line 136: If we put the steps into a list like this, we avoid the too long sentence. What do you think?

```suggestion
The fields showing are all columns in your spreadsheet that have a name in row 1. To map the input data to the different fields in the spreadsheet follow these steps:

* Select the “Type or insert…” field.
* Choose which data to put in the field.
```

sofietoft (commented):
Thanks for improving the readability of this article @OwainWilliams ! 💪 😄 

I've taken a shot at shortening the sentence you refer to.
Let me know what you think 😄

OwainWilliams:
Like the option of creating steps rather than a sentence. Thanks for looking at this so quickly Sofie :)

sofietoft:
Great! 😄 
Yeah, sometimes it's necessary to have to think "out side the box" in order to resolve the long sentences 😄 

Let's merge it in!

PR #4522, opened 2022-10-25
V11: Remove reference to UmbracoApplicationMainDomAcquiredNotification in common pitfalls

nikolajlauridsen:
Removes the reference to the `UmbracoApplicationMainDomAcquiredNotification` in the common pitfalls document, since this notification no longer exists in V11 😄 

I'm still a bit shaky about the whole multiple versions thing, but I've copied the old one and named it `index-v9.md` even though it covers both v9 and v10 and then made index.md be the v11 one. Feel free to change this if I've done it wrong 😄

### DIFF
--- a/Reference/Common-Pitfalls/index-v11.md
+++ b/Reference/Common-Pitfalls/index-v11.md
@@ -1,6 +1,5 @@
 ---
-versionFrom: 9.0.0
-versionTo: 10.0.0
+versionFrom: 11.0.0
 meta.Title: "Common Pitfalls and Anti-Patterns in Umbraco"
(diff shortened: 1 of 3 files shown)


sofietoft:
Thanks for verifying the article @nikolajlauridsen ! 💪 

I get that it's tricky with the versioning - it sometimes is for me as well 🤷 😄 
But good news: Soon we don't have to deal with it anymore 😎 

Anywho!
I've renamed the files - having index-v11 files for the new stuff, makes it a little easier for us, when we start the migration 😁 

Also, let's ignore all those vale warnings this time 🙈

nikolajlauridsen:
Awesome, thanks 👍 I'll keep that in mind for the next time 😄

PR #4524, opened 2022-10-25
V11: Update Custom Package Migration examples

nikolajlauridsen:
Updates the Custom Package Migration example to use the new constructor 😄


sofietoft:
Thanks for verifying and updating the article @nikolajlauridsen ! 😄 

I've just renamed the files again, to ensure we catch the Umbraco 11 stuff when we migrate the docs to the new platform 😄 
(again, let's just ignore those vale warnings 🙈 )

PR #4527, opened 2022-10-25
V11: Fix RecurringHostedServiceBase example

nikolajlauridsen:
Updates the `RecurringHostedServiceBase` the be up to date for V11.

The major thing was that it used the obsolete `IScopeProvider`, I've updated this to use the appropriate `ICoreScope`.

### DIFF
--- a/Reference/Scheduling/index-v11.md
+++ b/Reference/Scheduling/index-v11.md
@@ -1,0 +1,180 @@
+---
+versionFrom: 11.0.0
+meta.title: "Scheduling with hosted services in Umbraco"
+meta.Description: "Use hosted services to run a background task"
+---
+
+# Scheduling with RecurringHostedServiceBase
+
+In Umbraco 9+ it is possible to run recurring code using a hosted service.
+Below is a complete example showing how to create and register a hosted service that will regularly empty out the recycle bin every five minutes. 
+
+:::warning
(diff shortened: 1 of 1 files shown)


Reference/Scheduling/index-v11.md
```diff
@@ -0,0 +1,180 @@
...
+
+```C#
+public CleanUpYourRoom(
+    ...)
+    : base(HowOftenWeRepeat, DelayBeforeWeStart)
+{
+    ...
+}
+```
+:::
```
nikolajlauridsen, line 99:
```suggestion
```

We might want to remove this since this doc is for v11 specifically. 

Also this example looks a bit odd [on the live site](https://our.umbraco.com/documentation/Reference/Scheduling/#recurringhostedservice-example) even though the markdown is correct

nikolajlauridsen (commented):
I noticed this minor thing when reviewing this doc too

Reference/Scheduling/index-v11.md
sofietoft: I agree that this should be removed from the v11 version of the article 💪 
Have no idea why the formatting of this code sample is funky on the frontend - but I've seen it before, and it seems related to the fact that the code sample is inside a note 🤷‍♀️ 😅

PR #4530, opened 2022-10-25
V11: Fix publishing notifications

nikolajlauridsen:
Adds a new v11 document and fixed the code example to work with v11.

I also noticed that the V10 version is wrong (missing logger, and not properly using ICoreScopeProvider) so I've update the V10 version as well to fix this 😄


sofietoft (commented):
Thanks for verifying the article @nikolajlauridsen ! 💪 
And for fixing the v10 version as well 😄 

I'll get it merged.

PR #4533, opened 2022-10-25
Update IContentFinder.md

kmeilander:
Adding using System.Threading.Tasks; to example code since it's needed for Task<bool>


sofietoft:
Great catch @kmeilander ! 💪 and thanks for making a PR to correct the code sample!
Much appreciated 😄

PR #4534, opened 2022-10-25
Added SendingChildrenNotification to list of editormodel notifications

patrickdemooij9:
This functionality has been present in the codebase for some time now, but I hadn't gotten around to adding the documentation for it.

### DIFF
--- a/Reference/Notifications/EditorModel-Notifications/index.md
+++ b/Reference/Notifications/EditorModel-Notifications/index.md
@@ -206,7 +206,7 @@ namespace Umbraco.Docs.Samples.Web.Notifications
       </ul>
     </td>
     <td>
-    Published right before the allowed children of a node are sent back. (whenever you click create on a node)<br>
+    Published right before the allowed children of the selected Content Type are sent back during content creation in the Content Section.<br>
     NOTE: Children is a collection of ContentTypeBasic, each object gives you access to Alias, Description, Thumbnail and more. You can remove or add new children to the list in the notification.
     </td>
   </tr>
--- a/Reference/Notifications/EditorModel-Notifications/index.md
+++ b/Reference/Notifications/EditorModel-Notifications/index.md
@@ -196,6 +196,20 @@ namespace Umbraco.Docs.Samples.Web.Notifications
     NOTE: Dashboards is a collection of IDashboardSlim, each object gives you access to Label, Alias, Properties, whether it's expanded, and whether it IsActive.
     </td>
   </tr>
+  
+  <tr>
+    <td>SendingAllowedChildrenNotification</td>
+    <td>
+      <ul>
+        <li>IEnumerable&ltContentTypeBasic&gt Children</li>
+        <li>IUmbracoContext UmbracoContext</li>
+      </ul>
+    </td>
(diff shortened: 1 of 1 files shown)


Reference/Notifications/EditorModel-Notifications/index.md
```diff
@@ -196,6 +196,20 @@ namespace Umbraco.Docs.Samples.Web.Notifications
...
+    <td>SendingAllowedChildrenNotification</td>
+    <td>
+      <ul>
+        <li>IEnumerable&ltContentTypeBasic&gt Children</li>
+        <li>IUmbracoContext UmbracoContext</li>
+      </ul>
+    </td>
+    <td>
+    Published right before the allowed children of a node are sent back. (whenever you click create on a node)<br>
+    NOTE: Children is a collection of ContentTypeBasic, each object gives you access to Alias, Description, Thumbnail and more. You can remove or add new children to the list in the notification.
```
sofietoft, line 210: Instead of the sentence in the parameter, would it make sense to combine the first two sentences?
I'm not entirely sure I understand "whenever you click create on a node" - what is it that happens whenever you click create on a node? 🤔 

```suggestion
    Published right before the allowed children of a node are sent back. (whenever you click create on a node)<br>
    NOTE: Children is a collection of ContentTypeBasic. Each object gives you access to Alias, Description, Thumbnail, and more. You can remove or add new children to the list in the notification.
```

sofietoft (commented):
Thanks for the PR @patrickdemooij9 ! 💪 
Much appreciated!

I've added a single suggestion, and then I'll also have it by our developers for a check-up!

Reference/Notifications/EditorModel-Notifications/index.md
patrickdemooij9: I thought the sentence `Published right before the allowed children of a node are sent back` doesn't really make it clear when this is actually called. Whenever you want to create a new node, it'll list you the available node types that you can pick from according to the node that you clicked create on. That is when this notification is sent out. So I was hoping to add some more context to it like that.

Maybe something along the lines of: `Published right before the allowed childrens of a node are sent back. These are the node types that are allowed under a given a node in your content section.`

sofietoft: Ah, I see! 🤔 I think I was thrown by the "sent back", not really sure what that meant. But it makes sense now!

I like your second suggestion here, but how about we merge it into just one sentence instead?
Something like:
"Published right before the allowed children of the selected Content Type are sent back during content creation in the Content Section."

patrickdemooij9: Good suggestion, I've added that sentence now instead of the other one.

nikolajlauridsen (approved):
Looks good to me 👍

sofietoft:
Thanks for the review @nikolajlauridsen 🙌 
And thanks again for adding this to our docs @patrickdemooij9 - MUCH appreciated!

I see this is also you're first-ever contribution to the Umbraco docs - that's umbazing! 🥳 #H5YR
Hope to see you back here with even more great stuff 👏

PR #4537, opened 2022-10-26
Update and rename/move static asset cache tutorial to "Reference/Response Caching"

Jeavon:
The current tutorial name of "Cache static assets" is not accurate. This tutorial does not cache anything it allows the setting of a response header that is used to determine how long a file should be cached for by the browser or a proxy server.

This PR updates the title for static assets cache control tutorial, updates code samples from Samples project, and swaps from IComposer to Startup.cs registration

### DIFF
--- a/Reference/Response-Caching/index.md
+++ b/Reference/Response-Caching/index.md
@@ -1,8 +1,11 @@
 ---
 versionFrom: 9.3.0
 ---
+# Response Caching
 
-# Modify the `Cache-Control` header for Static Files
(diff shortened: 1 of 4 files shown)


sofietoft:
Thanks for correcting this article @Jeavon ! 💪 

We actually have a task on our list here at HQ, to move this article to the Caching section in Reference instead.
It does not really read as a tutorial, as it's more or less a code sample that can be added.
What do you think about moving this to the Reference/Caching section instead?

Jeavon:
@sofietoft agreed! I've moved the article to "Reference/Response Caching". Additionally, I've added an example of how to add the `cache-control` header for content rendering and updated indexes accordingly. 😎

sofietoft:
Looks great @Jeavon ! 💪 Thanks for taking the article out of the Tutorials section.

I'm happy with the changes - let's merge it in 😄

PR #4567, opened 2022-11-02
Fixes incorrect reference to Forms configuration key.

AndyButland:
This was reported via [this functional issue](https://github.com/umbraco/Umbraco.Forms.Issues/issues/907), that looked to be due to an incorrect value in the documentation.


sofietoft (approved):
Thanks for updating the docs @AndyButland ! 🙌

PR #4591, opened 2022-11-04
Fix casing for block grid components path

kjac:
The docs have a bad case of casing going on 😄 this will cause issues for Linux users.

The correctly cased full path is: `/Views/Partials/blockgrid/Components/`


sofietoft:
Thanks for the PR @kjac !

Just curious.. Can we say anything more general about how to do casing - perhaps, especially in a Linux setting?
Do they need to be changed because of the names of the directories in the path? 🤔

kjac:
@sofietoft it's very much on a case-by-case basis. Umbraco was not originally built for case sensitive systems, so consistent was never top of mind.

In this case the "blockgrid" directory is lower-cased while the rest of the path is pascal-cased. There is a historical reason for this that ties to the block list, so it's not entirely random, even if it probably seems that way.

sofietoft:
Thanks for explaining 😄 
Figured it might be something like that - just wanted to check, as it looked a little strange to me 🙈 

But nothing to do about that - from here at least!

PR #4592, opened 2022-11-04
Cleanup Fundamentals section Final Part

eshanrnh:


sofietoft:
Looks great @eshanrnh !

The links on the Fundamentals landing page are not updated though.
Perhaps we should get rid of it all together? What do you think?
We could let GitBook generate the overview for us.

eshanrnh:
Oh yes, I missed updated the links in that page since the table is generated using <html> tags 🙈. I can delete that page and let Gitbook generate an overview for us, if that's possible.

sofietoft:
Yes, let's do that I think!
We can always add a landing page at a later point 💪

PR #4595, opened 2022-11-04
added null check

abhamla:


sofietoft:
Thanks for the PR @abhamla ! 
Totally makes sense to add a null check to this code 💪 

I'll merge it right in!

PR #4601, opened 2022-11-07
Improve readability - Fundamentals Code Creating-Forms

erikjanwestendorp:
Belongs to: https://github.com/umbraco/UmbracoDocs/issues/4416

- [x] write shorter sentences
- [x] use 'for example' instead of 'e.g.'
- [x] define an acronym, but there is still an issue?

### DIFF
--- a/Fundamentals/Code/Creating-Forms/index.md
+++ b/Fundamentals/Code/Creating-Forms/index.md
@@ -18,7 +18,7 @@ In this example we'll create a basic contact form containing a name, email and m
 
 ### Creating the view model
 
-First, we're going to create the model for the contact form by adding a new class to the `/Models` folder (if the folder doesn't already exist, create it at the root of your website). Let's call it `ContactFormViewModel.cs`
+First, we're going to create the model for the contact form by adding a new class to the `/Models` folder. (If the folder doesn't already exist, create it at the root of your website). Let's call it `ContactFormViewModel.cs`
 
 ```csharp
 namespace MyFirstForm.Models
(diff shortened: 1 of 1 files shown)


Fundamentals/Code/Creating-Forms/index.md
```diff
@@ -65,13 +65,13 @@ The view can be built with standard MVC helpers:
 ```
 
 ### Adding the controller
-Finally, we're going to add the controller. Create a new empty class in the `/Controllers` folder (if the folder doesn't already exist, create it at the root of the website). Name it `ContactFormController` and make it inherit from `SurfaceController`. Inheriting from `SurfaceController` requires that you call its base constructor, most IDE's can do this automatically for you.
+Finally, we're going to add the controller. Create a new empty class in the `/Controllers` folder (if the folder doesn't already exist, create it at the root of the website). Name it `ContactFormController` and make it inherit from `SurfaceController`. Inheriting from `SurfaceController` requires that you call its base constructor, most IDE's: Integrated Development Environments, can do this automatically for you.
```
sofietoft, line 68: I think the issue might be that the acronyms has `'s` after it, so the formatting rule can't really figure it out 😅 

How about rephrasing the sentence to:
> Inheriting from `SurfaceController` requires that you call its base constructor. If you are using an IDE: Integrated Development Environments, this can be done automatically.

sofietoft (commented):
Thanks for the PR @erikjanwestendorp ! 🙌 

I've made a suggestion to the acronym situation. I think I know why it doesn't "approve" your solution!

Fundamentals/Code/Creating-Forms/index.md
erikjanwestendorp: @sofietoft I think so too 😅. And rephrasing the sentence is a good solution 😄 !

erikjanwestendorp:
@sofietoft Just updated the PR with your suggestion! 😄

sofietoft:

I'll merge it right in then 😄

PR #4603, opened 2022-11-08
Reference script instead of html view

erikjanwestendorp:
Reference script instead of html view


sofietoft (approved):
Good catch @erikjanwestendorp ! 😄 

I'll merge it right in.

PR #4616, opened 2022-11-10
Example of how to create async RenderController

enkelmedia:
It was hard to find any documented way to use an async Index()-method but this approach works. I think that official docs should have an example for people looking to use a async implementation in the controller. 

Without the sync Index() with the [NoAction]-attribute the call would be ambiguous between the async version on the ProductPageController and the sync Index()-method on RenderController.

As far as I can tell this is the best approach at the moment.

### DIFF
--- a/Reference/Routing/Custom-Controllers/index.md
+++ b/Reference/Routing/Custom-Controllers/index.md
@@ -66,7 +66,7 @@ namespace My.Website
 
 ALL requests to ANY 'product' pages in the site will be'hijacked' and routed through the custom ProductPageController.
 
-If you perfer to use a async controller your need to overide both the sync and the async Index()-methods to disable the default behavior from the base controller.
+If you prefer to use an async controller your need to override both the sync and the async Index()-methods. This is done to disable the default behavior from the base controller.
 
 ```csharp
 public class ProductPageController : RenderController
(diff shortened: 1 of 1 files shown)


Reference/Routing/Custom-Controllers/index.md
```diff
@@ -66,6 +66,24 @@ namespace My.Website
 
 ALL requests to ANY 'product' pages in the site will be'hijacked' and routed through the custom ProductPageController.
 
+If you perfer to use a async controller your need to overide both the sync and the async Index()-methods to disable the default behavior from the base controller.
```
sofietoft, line 69:
```suggestion
If you prefer to use an async controller your need to override both the sync and the async Index()-methods. This is done to disable the default behavior from the base controller.
```

sofietoft (commented):
Thanks for the PR @enkelmedia ! 💪 
I've made a single suggestion to fix a long sentence.

Also, I'll have it by our developers for a double-check on the code!

enkelmedia:
Great! Just feel free to change and adjust =D

Zeegaan (approved):
LGTM 👍

sofietoft:
Excellent! Thanks for the review @Zeegaan ! 🙌 

Looks like all lights are green here @enkelmedia ! I'll merge it in 😄 
Thanks again for helping us improve the docs!

PR #4617, opened 2022-11-10
Added info about custom IContentLastChanceFinder

enkelmedia:


sofietoft (commented):
Thanks for the PR @enkelmedia ! 💪 

I think this is very valuable information.
Perhaps it would be better to have it in a note below the list? 
That would keep the list "clean", but still provide the information in a highlighted way.
What do you think? 🤔

enkelmedia:
Feel free to change it and adjust =D I just wanted to "raise the issue", spent about 2 hours trying to get the 404Errors configured in "appSettings" to work but found that it did not because a packaged used the LastChangeContentFinder. So I figured its good to have this in the docs =D

sofietoft:
Hi @enkelmedia !
I've tried adding the info to a warning below the list instead.

Take a look, and let me now if it still makes sense or if it should be rewritten.

enkelmedia:
Looks great! =D

sofietoft:
Excellent @enkelmedia ! 😄 
I'll merge it in.

Thanks again for adding this info to the docs 💪

Issue #4426, opened 2022-10-03
Creating a custom index does not work


AaronSadlerUK:
## What type of issue is it? (Choose one - delete the others)
Wrong documentation

## What article/section is this about?
https://our.umbraco.com/documentation/reference/searching/examine/indexing/

## Describe the issue
Following the example to create a custom index for a single document type does not work.

First the composer for the content populator is registered incorrectly and should be:

`builder.Services.AddSingleton<IIndexPopulator, ProductIndexPopulator>();`

Secondly even when everything is registered correctly, the index remains empty and does not get populated with a "product" item.

eshanrnh:
Thanks for flagging this issue, @AaronSadlerUK 🙌 I will pass it by the D-team to hear their thoughts 🙂

Wiggee11:
I have logged the same issue here https://github.com/umbraco/Umbraco-CMS/issues/13070 and can confirm the documentation is wrong an out of date.

nul800sebastiaan:
Specifically, this line is wrong:

`yield return new ValueSet(content.Id.ToString(), IndexTypes.Content, indexValues);` 

The third argument should be the doctype alias!

So we end up with:

`yield return new ValueSet(content.Id.ToString(), IndexTypes.Content, content.ContentType.Alias, indexValues);`


Additionally, my composer looks like the following. The order is very important, I found..

```
    public class MyIndexComposer : IComposer
    {
        public void Compose(IUmbracoBuilder builder)
        {
            builder.Services.ConfigureOptions<ConfigureMyIndexOptions>();
            builder.Services.AddSingleton<MyIndexValueSetBuilder>();
            builder.Services.AddSingleton<IIndexPopulator, MyIndexPopulator>();
            builder.Services.AddExamineLuceneIndex<MyIndex, ConfigurationEnabledDirectoryFactory>(IndexConstants.IndexName);
        }
    }
```

(replace `My` with `Product` and `home` with `product` to match the docs examples). The `IndexName` comes from: 

```csharp
    private static class IndexConstants
    {
        public static string IndexName => "MyCustomIndex";
    }
``` 

Full working example (do test before copy/pasting!): https://gist.github.com/nul800sebastiaan/b9a6730a6ce1bf9f8d00fcd82e087f1f

[...]

PR #4625, opened 2022-11-15
Update code examples

Zeegaan:
Fixes https://github.com/umbraco/UmbracoDocs/issues/4426

There were a few bits missing from the code examples, mainly the `ConfigureOptions` from the ExamineComposer 👍 

This would be really great if it was tested, so we actually know if this resolves the issue 🙏


sofietoft:
I've closed the Issue related to this PR, as it was recently tested and found to be working 💪 

I'll see about adding the commits from this PR to the new docs - this was made for Our, it looks like 😅

PR #4627, opened 2022-11-15
Update index.md

biapar:
Change Google library import.

### DIFF
--- a/Reference/Security/Two-factor-authentication/index.md
+++ b/Reference/Security/Two-factor-authentication/index.md
@@ -22,7 +22,7 @@ Despite the name, this package works for both Google and Microsoft authenticator
 ```Csharp
 using System;
 using System.Threading.Tasks;
-using Google.Authenticator;
+using GoogleAuthenticatorService.Core;
 using Umbraco.Cms.Core.Security;
 using Umbraco.Cms.Core.Services;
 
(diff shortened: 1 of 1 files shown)


Reference/Security/Two-factor-authentication/index.md
```diff
@@ -83,7 +83,7 @@ namespace Umbraco9
             var member = _memberService.GetByKey(userOrMemberKey);
 
             var twoFactorAuthenticator = new TwoFactorAuthenticator();
-            SetupCode setupInfo = twoFactorAuthenticator.GenerateSetupCode("My application name", member.Username, secret, false);
+            SetupCode setupInfo = twoFactorAuthenticator.GenerateSetupCode("My application name", secret, 200,200  );
```
sofietoft, line 86: What is the reason for removing the "member.Username"?
What does "200,200" mean? Would be great with a code-comment here, to explain the numbers 💪 

Also, if `member` should not be used, can line 83 be omitted? 🤔

sofietoft (commented):
Thanks for the PR @biapar ! 💪 

I've added a comment in the code.
I'll make sure we test this as soon as possible 💪

Reference/Security/Two-factor-authentication/index.md
biapar: Hi, I use another library. In your example, you use this https://github.com/BrandonPotter/GoogleAuthenticator. I see that the version 3 support .NET Core 6.
Sorry for my mistake. You can reject the changes.

sofietoft: Ah, I see!
So just to make sure I understand completely - is it the entire PR is can disregard, or only this part? 🤔

sofietoft:
Hi @biapar !

I haven't heard back from you on this one, so I've decide to close it.
Since we've migrated the docs to a new platform, we wouldn't be able to merge this one in directly anyways.

Please feel free to open a new PR, if you still feel the article needs to be updated 💪

PR #4641, opened 2022-11-17
Update contentsettings documentation

Zeegaan:
With https://github.com/umbraco/Umbraco-CMS merged, we will have 2 new settings in the ContentSettings 👍 
I have made a note for the v10 docs, that from 10.4 an onwards you can use the new settings.
I have also created v11 docs and removed mention of the old settings, and replaced them with the new ones 😁


sofietoft:
Thanks for the PR @Zeegaan !
I've made vale happy (well, ish!), so I'll merge it in 🙌

PR #4642, opened 2022-11-17
Link to Azure Key Vault page added to Web App page

mattou07:


Fundamentals/Setup/Server-Setup/azure-web-apps.md
```diff
@@ -77,3 +77,7 @@ The quickest way to get to your logs is using the following URL template and rep
 `https://{app}.scm.azurewebsites.net/api/logstream`
 
 You can also find this in the KUDU console by clicking **Advanced Tools** > **Log Stream** on the Web App in the Azure Portal.
+
+### Web App secret management
```
sofietoft, line 81: How about changing the heading to H2 instead?
Doesn't sound like something that's related to "Web worker migrations" - but I might be wrong? 😄 

```suggestion
## Web App secret management
```

sofietoft (commented):
Thanks for the PR @mattou07 ! 💪 

Makes a lot of sense adding this link here!
I've added a single suggestion.

mattou07:
> Thanks for the PR @mattou07 ! 💪
> 
> Makes a lot of sense adding this link here! I've added a single suggestion.

It's not. As always thanks @sofietoft !

sofietoft:
Brilliant! Well, let's get it merged in then 😄 

Thanks again for helping out @mattou07 ! 💪

PR #4643, opened 2022-11-17
Updates to the Umbraco Forms developer documentation

AndyBoot:
- Updated - Umbraco Forms > Developer > Themes doc - Updated URL to the downloadable 'default' theme for Umbraco Forms 10.2.
- Updated - Umbraco Forms > Developer > Rendering Scripts doc - Somewhat overhauled the page with the intention for it to be more informative by mentioning the Scripts.cshtml file, what it contains and what it does.
- Updated - Umbraco Forms > Developer > Extending > Adding a Workflow Type doc - Added some additional code to the example snippet to demo how to alter the form submission data.

### DIFF
--- a/Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md
+++ b/Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md
@@ -50,6 +50,20 @@ namespace MyFormsExtensions
 
                 // or get it as a string
                 rf.ValuesAsString();
+
+                // set or replace values against a field in the current form submission
+                if(rf.Alias.Equals("myField", StringComparison.InvariantCulture))
+                {
+                    rf.Values.Clear();
+                    rf.Values.Add("My new value");
+                }
+
+                // Update the form submission via the UpdateRecord method, but only when the Workflow is in an Approval chain. If not in an Approval chain, the values are still saved against the RecordField.
+                if (context.State.Equals(FormState.Approved))
+                {
+                    _recordStorage.UpdateRecord(context.Record, context.Form);
+                }
+
             }
 
             //Change the state
--- a/Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
+++ b/Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
@@ -5,25 +5,58 @@ versionTo: 10.0.0
 
 # Rendering Forms Scripts
 
-Forms output some JavaScript which is by default rendered right below the markup.
-
-In many cases, you might prefer rendering your scripts at the bottom of the page, e.g. before the closing `</body>` tag, as this generally improves site performance.
-
-In order to render your scripts where you want, you need to add the following snippet to your template. Make sure you add it below your scripts, right before the closing `</body>` tag:
+Script delivery is done automatically using [Themes](/documentation/Add-ons/UmbracoForms/Developer/Themes/) without the need to include any additional code into your own views. The 'default' theme (bundled with Umbraco Forms) comes with the file `Scripts.cshtml` containing all the relevant logic & data required by the Umbraco Forms JavaScript. For reference, the file looks like this *(as per Umbraco Forms 10.2)*:
 
 ```csharp
-@using Umbraco.Forms.Web.Extensions;
+@model Umbraco.Forms.Web.Models.FormViewModel
 
-@if (TempData["UmbracoForms"] != null)
-{
-    foreach (var form in TempData.Get<List<Guid>>("UmbracoForms"))
-    {
-        @await Component.InvokeAsync("RenderFormScripts", new {formId = form, theme = "bootstrap3-horizontal"})
-    }
+@using Newtonsoft.Json
+@using Umbraco.Forms.Core
+@using Umbraco.Forms.Web
+
+@{
+    Html.AddFormThemeScriptFile("~/App_Plugins/UmbracoForms/Assets/themes/default/umbracoforms.js");
 }
+
+<div class="umbraco-forms-form-config"
+     style="display: none"
+     data-id="@Model.FormClientId"
+     data-serialized-page-button-conditions="@JsonConvert.SerializeObject(Model.PageButtonConditions, FormsJsonSerializerSettings.Default)"
+     data-serialized-fieldset-conditions="@JsonConvert.SerializeObject(Model.FieldsetConditions, FormsJsonSerializerSettings.Default)"
+     data-serialized-field-conditions="@JsonConvert.SerializeObject(Model.FieldConditions, FormsJsonSerializerSettings.Default)"
+     data-serialized-fields-not-displayed="@JsonConvert.SerializeObject(Model.GetFieldsNotDisplayed(), FormsJsonSerializerSettings.Default)"
+     data-trigger-conditions-check-on="@Model.TriggerConditionsCheckOn"
+     data-form-element-html-id-prefix="@Model.FormElementHtmlIdPrefix"></div>
+
+@* Only render out scripts on the page if the form has not been submitted yet *@
+@if (Model.SubmitHandled == false)
+{
+    @*
+        If the current page of the form has any Partial view files attached to render
+        Likely used by events and third parties adding tracking or other 3rd party functionality to a form
+    *@
+    if (Model.CurrentPage.PartialViewFiles.Any())
+    {
+        foreach (var partial in Model.CurrentPage.PartialViewFiles)
+        {
+            @await Html.PartialAsync(partial.Value)
+        }
+    }
+
+    @* Render references to javascript files needed by fields on the current page*@
+    @Html.RenderFormsScripts(Url, Model, Model.JavaScriptTagAttributes)
+    @Html.RenderFormsStylesheets(Url, Model)
+}
+
 ```
 
+:::note
+If you are upgrading from an older version of Umbraco Forms and using a variant of `Scripts.cshtml` within your own project you may be missing some required and fundamental parts of this file. As an example, the `<div class="umbraco-forms-form-config" />` element introduced the `data-form-element-html-id-prefix` attribute in Umbraco Forms 10.2. Failing to update this within your own version of this file may lead to a loss of functionality and/or JavaScript errors.
+:::
+
 ## Enabling `ExcludeScripts`
+
+Forms output some JavaScript which is by default rendered right below the markup.
 
 If you do not want to render the associated scripts with a Form, you need to explicitly say so. You need to make sure `ExcludeScripts` is checked/enabled, whether you are inserting your Form using a macro or adding it directly in your template.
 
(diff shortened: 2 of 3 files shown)


Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
```diff
@@ -5,26 +5,59 @@ versionTo: 10.0.0
...
+
+    @* Render references to javascript files needed by fields on the current page*@
+    @Html.RenderFormsScripts(Url, Model, Model.JavaScriptTagAttributes)
+    @Html.RenderFormsStylesheets(Url, Model)
 }
+
 ```
 
+:::note
+If you are upgrading from an older version of Umbraco Forms and using a variant of `Scripts.cshtml` within your own project you may be missing some required and fundamental parts of this file. As an example, the `<div class="umbraco-forms-form-config" />` element introduced the `data-form-element-html-id-prefix` attribute in Umbraco Forms 10.2. Failing to update this within your own version of this file may lead to a loss of functionality and/or JavaScript errors.
```
sofietoft, line 54:
```suggestion
When upgrading from an older version of Umbraco Forms and using a variant of `Scripts.cshtml`, you may be missing required parts of this file. As an example, the `<div class="umbraco-forms-form-config" />` element introduced the `data-form-element-html-id-prefix` attribute in Umbraco Forms 10.2. Failing to update this within your own version of this file may lead to a loss of functionality and/or JavaScript errors.
```

Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
```diff
@@ -43,6 +76,11 @@ To enable `ExcludeScripts`:
 `ExcludeScripts = "1"` prevents the associated scripts from being rendered. Any other value, an empty value, or if the parameter is excluded, will render the scripts on the Form.
 :::
 
+The `Scripts.cshtml` file (from the current or default theme) will not be initiated and won't be included as part of the current form. You may however use parts (or all) of the code from the snippet of `Scripts.cshtml` within your own template or partial views.
```
sofietoft, line 79:
```suggestion
The `Scripts.cshtml` file (from the current theme) will not be initiated and therefore not included as part of the current form. You may however use parts (or all) of the code from the snippet of `Scripts.cshtml` within your own template or partial views.
```

Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
```diff
@@ -43,6 +76,11 @@ To enable `ExcludeScripts`:
...
+
+
+In many cases, you might prefer rendering your scripts at the bottom of the page, e.g. before the closing `</body>` tag, as this generally improves site performance.
```
sofietoft, line 82:
```suggestion
In many cases, you might prefer rendering your scripts at the bottom of the page, for example before the closing `</body>` tag. This will generally improve site performance.
```

Add-ons/UmbracoForms/Developer/Themes/index.md
```diff
@@ -15,7 +15,7 @@ To create a theme, you need to create a folder at `/Views/Partials/Forms/Themes/
 
 Copy the explicit files you wish to override in your theme, it may be a single file or all files from the default theme folder. Make the necessary changes you desire to CSS class names, markup etc.
 
-For Umbraco 9 and previous, it's straightforward to simply copy the files you need from the default theme folder.  Umbraco 10 distributes these files as part of a Razor class library, so you won't find them on disk. The easiest way to obtain copies of the files included in the default theme is to [download this zip file](https://our.umbraco.com/FileDownload?id=23911) and extract the ones you need.
+For Umbraco 9 and previous, it's straightforward to simply copy the files you need from the default theme folder.  Umbraco 10 distributes these files as part of a Razor class library, so you won't find them on disk. The easiest way to obtain copies of the files included in the default theme is to [download this zip file](https://our.umbraco.com/FileDownload?id=24022) and extract the ones you need.
```
sofietoft, line 18:
```suggestion
For Umbraco 9 and previous, it's straightforward to copy the files you need from the default theme folder. Umbraco 10 distributes these files as part of a Razor class library, so you won't find them on disk. The easiest way to obtain copies of the files included in the default theme is to [download this zip file](https://our.umbraco.com/FileDownload?id=24022). From there you can extract the ones you need.
```

sofietoft (commented):
Thanks a bunch for the PR @AndyBoot ! 💪 

I've made a couple of suggestions to the grammar and there were some long sentences in there as well 😄 Take a look at let me know what you think!

I'll be sending the PR by our Forms expert as well, as I'll need some extra eyes on the technical part of this PR!

Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md
```diff
@@ -50,6 +50,20 @@ namespace MyFormsExtensions
 
                 // or get it as a string
                 rf.ValuesAsString();
+
+                // set or replace values against a field in the current form submission
```
AndyButland, line 54: This is useful to add for reference, so happy to include this.

Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md
```diff
@@ -50,6 +50,20 @@ namespace MyFormsExtensions
...
+                if(rf.Alias.Equals("myField", StringComparison.InvariantCulture))
+                {
+                    rf.Values.Clear();
+                    rf.Values.Add("My new value");
+                }
+
+                // Update the form submission via the UpdateRecord method, but only when the Workflow is in an Approval chain. If not in an Approval chain, the values are still saved against the RecordField.
```
AndyButland, line 61: This though I'm not so sure about - i.e. updating a record within a loop that iterates through the fields.  Maybe it works OK in the context it was used in, but we have to be 100% sure it's a good idea in the general case before using it in the docs.  So I think we should remove this part, or at least move it outside of the loop.

Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
```diff
@@ -5,26 +5,59 @@ versionTo: 10.0.0
 
 # Rendering Forms Scripts
 
-Forms output some JavaScript which is by default rendered right below the markup.
+Script delivery is done automatically using [Themes](/documentation/Add-ons/UmbracoForms/Developer/Themes/) without the need to include any additional code into your own views. The 'default' theme (bundled with Umbraco Forms) comes with the file `Scripts.cshtml` containing all the relevant logic & data required by the Umbraco Forms JavaScript. For reference, the file looks like this *(as per Umbraco Forms 10.2)*:
```
AndyButland, line 8: This is great to add, but I think we should remove the "For reference... " sentence, and the code block.  It'll only get out of date here.  I'd rather direct people to where they can download the theme files.

Add-ons/UmbracoForms/Developer/Rendering-Scripts/index.md
AndyButland, line 54: This is fair point, but very specific about a particular version and issue - and really the point is general.  If you're using a custom theme, you will want to review when the default one is updated to make sure you apply to your custom theme what changes may be necessary

Rather than adding this, probably something to that effect can be usefully added [here](https://our.umbraco.com/documentation/Add-ons/UmbracoForms/Developer/Themes/#creating-a-theme).

Add-ons/UmbracoForms/Developer/Themes/index.md
AndyButland, line 18: Thanks - we forgot to update this link clearly after the last release, and this is the latest one at the time or writing.

AndyButland (commented):
Thanks for this @AndyBoot - looks good, but I've left a few comments, mostly just to make sure we don't discuss very specific issues that are better presented in a more general way.  Nor make a maintenance headache by copying too much into the docs that is available elsewhere.

AndyBoot:
Thank you for your rapid response to this PR @AndyButland & @sofietoft! I will take a look through each of your comments and respond where appropriate. 👍

Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md
AndyBoot: To be honest, I only added the example in the context of the already existing snippet / loop. On my own project, I've actually done it slightly differently (outside of a loop). 

```
RecordField? myField = context.Record.RecordFields.Values.FirstOrDefault(x => x.Alias.Equals("myFieldAlias", StringComparison.InvariantCulture)) ?? throw new Exception($"Could not find field 'myFieldAlias'");
myField.Values.Clear();
myField.Values.Add("My new value");
```

This could sit either side of the loop if you're happy with it?

I personally feel this information would be super useful to anybody needing to perform a similar task inside their own custom workflows.

Add-ons/UmbracoForms/Developer/Themes/index.md
AndyBoot: This may mean more work for you each time a new version is released, but is it worth having this link in a TinyURL type service? I seem to recall (unless I dreamt it) that Umbraco has its own similar service which would be an idea candidate for this? What would then get used for the documentation instead is the static URL, but when a version is released the URL is simply updated with the new target.

Add-ons/UmbracoForms/Developer/Extending/Adding-a-Workflowtype.md
sofietoft: It sounds like we should try to put the code outside the loop here.
Would you be up for making the changes @AndyBoot ?

Add-ons/UmbracoForms/Developer/Themes/index.md
AndyButland: Yes, it sounds like a good idea, and I have found that we do have such a service.  Will see if I can get access to it and set it up for this link.

AndyButland: I've set one up, so if we amend this to `https://umbra.co/umbraco-forms-default-theme` that will point to the latest download.  And I can update it if and when we have an updated link.

sofietoft: I updated my suggestion here, to include the new URL 💪

sofietoft:
Hi @AndyBoot !
Just wanted to ping you on this one - have you had a chance to look at the suggestions? 😄 

Also, since we migrated the docs to a new platform, I'm afraid it will no longer be possible to merge in this PR directly.
Instead, we'll have to create a new PR that targets the articles on the new platform: 

- https://docs.umbraco.com/umbraco-forms/developer/extending/adding-a-workflowtype 
- https://docs.umbraco.com/umbraco-forms/developer/rendering-scripts
- https://docs.umbraco.com/umbraco-forms/developer/themes

Would you be up for making the new PR?
I'd be happy to do it - just let me know 😄

PR #4649, opened 2022-11-18
Update documentation to use private cache instead of shared

Zeegaan:
We have discussed this on the D-team and we will no longer be recommending shared cache mode, therefore the documentation that mentions shared cache for Sqlite needs to be changed to private cache, if there's anywhere else this is mentioned please let me know 🤔


sofietoft:
Thanks for the PR @Zeegaan !
I've made some minor changes - just updating the link ref and shortened the sentence 😄

Zeegaan:
@sofietoft Thanks for the assist 👊

PR #4640, opened 2022-11-17
Fix images debugging article v11

erikjanwestendorp:
In addition to: https://github.com/umbraco/UmbracoDocs/pull/4634 

Rename images for v11

PR #4650, opened 2022-11-18
Fix images debugging article for v11

erikjanwestendorp:
In addition to:  https://github.com/umbraco/UmbracoDocs/pull/4640


sofietoft:
Excellent @erikjanwestendorp !
Let's try this again 💪 🤞

sofietoft:
Alright! Looks like it didn't work.

Thanks for putting in all the work to try to figure this one out 🙏 Really appreciate it a lot! 😄 

I'll take it from here.
Will see if I can fix it directly on the GitBook platform 🤞 (we have a UI, where I might be able to fetch the images from and update the article there!)

PR #4652, opened 2022-11-21
Fixing broken references in the cloud section

jonat123:
Fixed the broken references in the Cloud section.


jonat123:
Cannot fix the broken links for some of the images so will add them directly...

Will merge this in as it has some other changes that should be fixed.

PR #4664, opened 2022-11-23
Renamed meta.Description to description

mastrup:
GitBook uses the description as the meta description.
This PR will make the old descriptions visible below the title and also fill out the meta tag in GitBook.

### DIFF
--- a/Add-ons/Umbraco-Deploy/Deploy-Settings/index-v8.md
+++ b/Add-ons/Umbraco-Deploy/Deploy-Settings/index-v8.md
@@ -1,7 +1,7 @@
 ---
 versionFrom: 8.0.0
 meta.Title: "Umbraco Deploy settings"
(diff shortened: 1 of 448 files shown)


sofietoft:
Wow! That's a big PR 😅 
Thanks @mastrup , for putting in the efforts to do this 💪 

However, I'm afraid I'm gonna have to tell you, that your PR is targeting the wrong branch 🙈 
Currently, the migrated docs live on the `gb/10` and `11.x` branches.
We will be merging 11.x into main when we officially go live with the new platform, but I'm afraid that will just completely overwrite all the changes you've made here.

It's totally on me for not having communicat this more clearly - just really wasn't expecting a PR of this magnitude 😄 

Is there any chance I can convince you to do this on the `gb/10` and `11.x` branches instead? 🙈 🤞

PR #4670, opened 2022-11-24
added null checker for single item mode

abhamla:
error thrown if item doesn't exist so added null check


eshanrnh:
Thanks for the PR, @abhamla 💪 We will review this soon.

sofietoft:
Hi @abhamla !

It definitely makes sense to have a null check here 🙌 
Thanks a bunch for adding it!

I see the changes you made here, are for the Umbraco 8 version though - would you be up for taking a look at the newer versions as well? 🤞

abhamla:
Hi @sofietoft,

Yeah sure, I can take a look! 😄

PR #4673, opened 2022-11-25
Added nullcheckers to V9/10 Nested Content docs

abhamla:


sofietoft (approved):
Awesome @abhamla ! 🙌 
Looks great - thanks for adding the additional null-checks here 😄 

I'll merge it right in!

PR #4674, opened 2022-11-25
Another Nested Content nullchecker for V9/10 docs

abhamla:
Missed one from the previous merge!


sofietoft:
Ah! Good catch 😄 Looks like I was a bit too quick on the trigger there as well 💪

PR #4700, opened 2022-11-29
Azure Blob Storage: Add note about private blob containers

georg-jung:
The need and how to configure this setting seems to be unclear to parts of the community, see <https://our.umbraco.com/forum/using-umbraco-and-getting-started/100593-log-flooding-with-imageprocessor-errors-using-blobstorage>. It took me some time to figure this out too, as basic image serving still works if this is misconfigured, but the log is full of related error messages though. Thus maybe it's worth to put a note in the corresponding docs.

### DIFF
--- a/Extending/FileSystemProviders/Azure-Blob-Storage/index-v8.md
+++ b/Extending/FileSystemProviders/Azure-Blob-Storage/index-v8.md
@@ -41,7 +41,7 @@ The following six keys will have been added to the `<appSettings>` in your `web.
 </appSettings>
 ```
 
-Make sure to update this configuration to match your own setup. E.g. note that `AzureBlobFileSystem.UsePrivateContainer:media` needs to be set to `true` if your blob container is not publicly accessible.
+Make sure to update this configuration to match your own setup. For example `AzureBlobFileSystem.UsePrivateContainer:media` needs to be set to `true` if your blob container is not publicly accessible. `ImageProcessingException`s will be thrown when images are served and this is not configured properly.
 
 When you're installing the package from the **Package** section of the Umbraco Backoffice, you'll be able to fill in the configuration directly from the backoffice:
 
--- a/Extending/FileSystemProviders/Azure-Blob-Storage/index-v8.md
+++ b/Extending/FileSystemProviders/Azure-Blob-Storage/index-v8.md
@@ -144,7 +144,7 @@ Any new media files you upload to the site, will automatically be added to the B
 
 If you don't want to use a publicly accessible container for media storage, you need to follow these additional steps:
 
-1. In your `Web.config`, `AzureBlobFileSystem.UsePrivateContainer:media` needs to be set to `true`.
+1. In your `Web.config`, `AzureBlobFileSystem.UsePrivateContainer:media` needs to be set to `true`: `<add key="AzureBlobFileSystem.UsePrivateContainer:media" value="true"/>`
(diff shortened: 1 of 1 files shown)


Jeavon:
Hi @georg-jung If you are using a private container then you will also need to swap ImageProcessor from using the CloudImageService to the AzureImageService so that probably needs to go into your note?

georg-jung:
I see, you're probably right. I didn't find any good docs for this myself, how does it work? I guess I need to install `ImageProcessor.Web.Plugins.AzureBlobCache` via NuGet and replace

    <service prefix="media/" name="CloudImageService" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
      <settings>
        <setting key="Container" value="media"/>
        <setting key="MaxBytes" value="8194304"/>
        <setting key="Timeout" value="30000"/>
        <setting key="Host" value="https://mygreatproject.blob.core.windows.net/"/>
      </settings>
    </service>

with ...?

I only found [this post](https://our.umbraco.com/forum/using-umbraco-and-getting-started/96120-imageprocessor-azureblobcache-with-azure-storage-emulatorazurite#comment-303993) suggesting

    <service prefix="media/" name="AzureImageService" type="ImageProcessor.Web.Plugins.AzureBlobCache.AzureImageService, ImageProcessor.Web.Plugins.AzureBlobCache">
      <settings>
        <setting key="StorageAccount" value="UseDevelopmentStorage=true" />
        <setting key="Container" value="dev" />
        <setting key="AccessType" value="Blob" />
      </settings>
    </service>

but `UseDevelopmentStorage=true` doesn't feel right to me.

Jeavon:
@georg-jung that's correct. You need to replace `UseDevelopmentStorage=true` with your own storage account connection string and also update the container. They should match the ones you use for the blob provider.

georg-jung:
Ok, thanks! Please see my recent changes to this PR, where I tried to incorporate that info.

Jeavon:
Great job! The only thing I'd mention is that when you install `ImageProcessor.Web.Plugins.AzureBlobCache` it modifies `cache.config` which you may not want if you are only installing for the AzureImageService....

georg-jung:
How about the changes to this repo in general? Is this PR still the way to modify v8 docs?

georg-jung:
Added a note about `cache.config`. Didn't happen to me because I use an SDK-style csproj, I guess.

Jeavon (approved):

sofietoft:
Hi @georg-jung ! Thanks for making this PR 💪 

As you've probably noticed, we migrated our docs to a new platform since you initially created this PR.
I'm afraid that means that it will not be possible to merge this one in directly.

As the file you made changes to here, `index-v8.md` doesn't exist under that name anymore, we need to create a new PR with the changes here.
The new PR should be made to the `legacy-docs` branch. for this article: https://github.com/umbraco/UmbracoDocs/blob/legacy-docs/Extending/FileSystemProviders/Azure-Blob-Storage/index.md

Apologies for the inconvenience 🙈 
Are you up for making the new PR? 🤞

sofietoft:
Hi @georg-jung !
Since I haven't heard back from you regarding this one, I've taken the liberty to add the changes to the new structure myself: https://github.com/umbraco/UmbracoDocs/pull/4964

That also means I'll close down this PR.
Thanks again for putting in the effort to help us update the docs and keep it up to speed.
MUCH appreciated 😄 🙌

georg-jung:
Oh, thanks for adapting this to the new structure. I'm afraid I lost track of that a bit, sorry.

PR #4704, opened 2022-11-29
Updated with the media folder step

Aaen:
I have added that you need to restore the media files on your Umbraco 8 project because in step 2 you will need to copy over the media folder to the new v10 project

### DIFF
--- a/umbraco-cloud/upgrades/migrating-from-8-10.md
+++ b/umbraco-cloud/upgrades/migrating-from-8-10.md
@@ -32,6 +32,8 @@ Video example.
 
 * A backup of your Umbraco 8 project database.
 
+* Restore media on your Umbraco 8 project 
+
 {% hint style="info" %}
 We strongly recommend having at least **2 environments** on the Umbraco 10 project.
 


sofietoft:
Thanks for the PR @Aaen ! 💪 

What you've added - isn't that already covered in this step: 
> * Create a backup of the database from your Umbraco 8 project using the [database backup guide](../databases/README.md) *OR* clone down the V8 project and take a backup of the local Database. Make sure to restore the content from your cloud environment. ?

Aaen:
I rhink we should do it like this 

Create a backup of the database from your Umbraco 8 project using the [database backup guide](https://github.com/umbraco/UmbracoDocs/databases/README.md) OR clone down the V8 project and take a backup of the local Database. Make sure to restore the content and media from your cloud environment. 

So we say explicit that you also should restore media because if the media are not referenced in the content then you dont get it down if I remember correct?

@sofietoft

sofietoft:
That makes sense @Aaen ! 💪 

I've made some changes to the PR - let me know what you think!

PR #4749, opened 2022-12-09
Fixed broken image Reference/Templating/Macros

wwrobel:
The image path on page https://docs.umbraco.com/umbraco-cms/reference/templating/macros is broken


sofietoft:
Good catch @wwrobel ! 💪 
Thanks for helping out with this!

PR #4752, opened 2022-12-12
Indicate that Grid Layout is obsolete

Wiggee11:
Suggest marking Grid Layout and Nested Content as obsolete and directing users to replacement Block Grid and Block List respectively.

### DIFF
--- a/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/grid-layout/README.md
+++ b/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/grid-layout/README.md
@@ -1,5 +1,9 @@
 # Grid Layout
 
+{% hint style="info" %} We highly recommend that you use the [Block Grid](../../block-editor/block-grid-editor.md) instead.
+
+Grid Layout has been marked as obsolete and will be removed from future versions of Umbraco. {% endhint %}
+
 `Returns: JSON`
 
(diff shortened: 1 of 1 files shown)


umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/grid-layout/README.md
```diff
@@ -1,4 +1,8 @@
-# Grid Layout
+# Grid Layout (Legacy)
+
+{% hint style="info" %} We highly recommend that you use the [Block Grid](../../block-editor/block-grid-editor.md) instead.
+
+Grid Layout has been marked as obsolete and will be removed from future versions of Umbraco. {% endhint %}
```
sofietoft, line 5: As we can't give a specific version number as to when we will remove the property, how about we write something like this instead:

```suggestion
Grid Layout has been marked as obsolete and development on the property editor has been discontinued. {% endhint %}
```

sofietoft (commented):
This is a great addition @Wiggee11 !
Makes sense to direct people to the new editors instead 💪 

I've added a single suggestion - let me know what you think!

Wiggee11:
Sensbile suggestion. I didn't really know how to phrase it!

sofietoft:
Yeah, it's definitely a tricky one to get "just right" 😅 
Glad you agreed with my suggestions.

Let's get this one merged in 💪

PR #4754, opened 2022-12-12
Update install-umbraco-with-templates.md

rhealefever:
From the .NET CLI:
"Warning: use of 'dotnet new --install' is deprecated. Use 'dotnet new install' instead." Have replaced dotnet new -i with dotnet new install


sofietoft:
Good catch @rhealefever ! 💪 

I'll merge it in right away!

PR #4759, opened 2022-12-13
Appsettings not correct format

jkulker:
Changed "Umbraco" to "Heartcore"


jonat123 (approved):
@jkulker Good spotted that the format is wrong.

Merging in this PR right away :D

PR #4762, opened 2022-12-15
Update mvc-samples.md

jkulker:
UmbracoController doesn't exist in Heartcore


sofietoft:
Thanks for the PR @jkulker ! 💪 

I'll make sure we review this as soon as possible!

Issue #4772, opened 2022-12-20
fix .Net SDK version for Umbraco Cloud local development


idseefeld:
## What type of issue is it? 

Wrong documentation

## What article/section is this about?

[Working with a Local Clone](https://docs.umbraco.com/v/10.x-lts/umbraco-cloud/set-up/working-locally)

## Describe the issue
Section "Running the site Locally" references the latest .Net SDK as requirement to run the cloned Umbraco Cloud project locally. But actually you have to look into file `global.json` to figure out the correct SDK version.

eshanrnh:
Thanks for reporting this issue and raising a PR, @idseefeld 🙌 We will review this soon.

jonat123:
@idseefeld Some time ago the issue that was the reason for us adding the global.json file to projects has been fixed and the global.json file has been removed from all Cloud projects.

I will therefore close this issue.

PR #4773, opened 2022-12-20
fix .Net SDK version for Umbraco Cloud local development

idseefeld:
This is a PR for issue #4772


idseefeld:
Is it me, or is there an setup issue with this vale linter check?

idseefeld:
Finally I figured out why vale linter has thrown an error. It was the acronym SDK without a definition. I have also open a discussion issue #4774 
😀

sofietoft:
Thanks for the PR @idseefeld ! 💪 

We've actually added a note about this in the v10 version of the article - as I believe this is only necessary for 10 projects? 🤔 

In any case, instead of adding a new step to the guide, what we did in the v10 article was adding a bullet to the "Tools" section that lists what you need in order to work locally:
https://docs.umbraco.com/v/10.x-lts/umbraco-cloud/set-up/working-locally#tools

I have a task on my desk to also add a note about this to the v11 article.
What do you think about doing something like that instead?

idseefeld:
Hey @sofietoft,
I think the tools list would be a good place for this detail. 
As far as I know, this is a quite new change because of recent cloud issues. 
And I guess this will be true for all future versions. 
Currently I have not tried version 11 in Umbraco Cloud.

Btw. I am impressed how quick you and your team respond to new issues and PRs!
Merry Xmas & Happy New Year!

sofietoft:
You're right there was an issue regarding this 💪 so what you're suggesting adding here is totally legit!
I have actually just been made aware that the Cloud team is rolling out a fix for this today: https://status.umbraco.io/ 
This will take out the `global.json` altogether, so my colleague is taking out the references to this from the docs as we speak.

With that, I'll close this PR for now!
Thanks again for the efforts behind this PR 🙌 
Hope you'll be up for reporting and suggesting again in the future, should you see something amiss in the Docs!

PR #4775, opened 2022-12-21
add error 500 handling example

idseefeld:
I have copied from a former PR, because there are merge conflicts (cause by the new structure?)
Hopefully this helps you😊

I close the _old_ PR

### DIFF
--- a/umbraco-cms/tutorials/custom-error-page.md
+++ b/umbraco-cms/tutorials/custom-error-page.md
@@ -9,6 +9,7 @@ Custom error handling might make your site look more on-brand and minimize the i
 This article contains guides on how to create custom error pages for the following types of errors:
 
 - [404 errors ("Page not found")](#404-errors)
+- [505 errors ("Internal Server Error")](#500-errors)
 
 ## In-code error page handling
 
--- a/umbraco-cms/tutorials/custom-error-page.md
+++ b/umbraco-cms/tutorials/custom-error-page.md
@@ -9,7 +9,7 @@ Custom error handling might make your site look more on-brand and minimize the i
 This article contains guides on how to create custom error pages for the following types of errors:
 
 - [404 errors ("Page not found")](#404-errors)
-- [505 errors ("Internal Server Error")](#500-errors)
+- [500 errors ("Internal Server Error")](#500-errors)
 
 ## In-code error page handling
 
--- a/umbraco-cms/tutorials/custom-error-page.md
+++ b/umbraco-cms/tutorials/custom-error-page.md
@@ -188,3 +188,5 @@ The following steps guides you through setting up a page for internal server err
     ...
     }
     ```
+
+    For local testing in Visual Studio replace `app.UseDeveloperExceptionPage();` by `app.UseExceptionHandler("/error");`. Otherwise you will get the default error page with stack trace etc.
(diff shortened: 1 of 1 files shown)


umbraco-cms/tutorials/custom-error-page.md
```diff
@@ -128,3 +128,65 @@ If your code or any packacges configures a custom `IContentLastChanceFinder`, th
...
+                    else if (Response.StatusCode != StatusCodes.Status200OK)
+                    {
+                        return Redirect("/statuscodes");
+                    }
+                    return Redirect("/");
+                }
+            }
+        }
+        ```
+
```
sofietoft, line 164:
```suggestion
1. Create a `~/controllers` folder in your Umbraco web project.
2. Create a file in this folder, called `ErrorController.cs`.
3. Add the following code to the file:

        ```csharp
        using Microsoft.AspNetCore.Mvc;
        namespace UmbracoProject.Web.Controllers
        {
            public class ErrorController : Controller
            {
                [Route("Error")]
                public IActionResult Index()
                {
                    if (Response.StatusCode == StatusCodes.Status500InternalServerError)
                    {
                        return Redirect("/statuscodes/500");
                    }
                    else if (Response.StatusCode != StatusCodes.Status200OK)
                    {
                        return Redirect("/statuscodes");
                    }
                    return Redirect("/");
                }
            }
        }
        ```

```

sofietoft (commented):
Thanks for recreating this PR @idseefeld ! 💪 
Much appreciated.

Only a couple of things I noticed here:
1. Instead of having a step called "First steps", would it make sense to add the three steps as part of the main ordered list? (see suggestion!)
2. Is there a way to test whether or not the error page actually works, once it's been implemented? Just wondering, as I have no idea, but thought it would be a nice thing to be able to do as a final thing 🤔 
3. Finally, we should add a link to this part of the article in the top, right? 😁

idseefeld:
Regarding testing: A simple test could be to force an error with status code 500 by adding 

```
@{
    var x = 0;
    var y = 1 / x;
}
```
to template. This will throw such an error and instead of the default you should see the configured error page. I am actually using this in my current projects, but cannot easily share these.

idseefeld:
The top link is a good idea. I like this new feature of on page navigation / linking.

idseefeld:
Now I have create a simple demo: https://github.com/idseefeld/CustomErrorPage

sofietoft:
Hi @idseefeld !
Happy New Year 🎆 

Thanks a bunch for providing the sample for testing.
I had a colleague test it out, and it works like a charm 👏 

Regarding the top link - it needs to be added manually.
GitBook automatically generated the deep-links displayed in the right-hand side of the article.
I still think, as this article now holds two separate tutorials, that it makes sense to add links to both of them at the top of the article 😄

idseefeld:
Happy New Year!🥳
thank you for your kind words!
Can you please add the top link?
I am currently not sure how to do this.🙄

Dirk

sofietoft:
Sure thing @idseefeld !

I've made some minor changes to the article.
Changed it to unordered lists for the steps, as it would otherwise be split into two ordered lists, which looked a little weird!

I'll merge in the tutorial!
Thanks again for adding this to the Umbraco Docs 🙌 
Much appreciated!

PR #4778, opened 2022-12-21
add hint for acronyms

idseefeld:
I spent quite a bit of time yesterday figuring out how to use acronyms.
What do you think?


idseefeld:
Just found acronyms mentioned in the style guide

PR #4785, opened 2023-01-02
added block grid docs to 10.4 rc

jonat123:


jonat123:
merging it in as it looks alright when previewing

PR #4787, opened 2023-01-02
changed pricing images to not have the real price

jonat123:

PR #4788, opened 2023-01-02
added the image

jonat123:
Changed the image so it does not hold any real prices


sofietoft (commented):
Same as: https://github.com/umbraco/UmbracoDocs/pull/4787#pullrequestreview-1234170206

jonat123:
Yeah same as in #4787 :-)

sofietoft (approved):

PR #4789, opened 2023-01-02
Article explaining the versioning strategy of the docs

sofietoft:


other-resources/documentation-and-versions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Major vs minor versions
+
+The Umbraco Documentation covers all supported versions of Umbraco CMS. A version exists for each supported **major** version of Umbraco CMS.
+
+When a Release Candidate (RC) for a new **minor** version of Umbraco CMS is released, a new version of the documentation will be available. When the minor version is released, the documentation for it will be merged into the related **major** version.
+
+The documentation for each major version will always document the latest version of that major version of Umbraco CMS.
+
+When an RC for a new **major** version of Umbraco CMS is released, a new version of the documentation will be available. When the major version is released, the documentation for it will be the new **default version of the Umbraco Documentation**.
```
eshanrnh, line 33:
```suggestion
When a RC for a new **major** version of Umbraco CMS is released, a new version of the documentation will be available. When the major version is released, the documentation for it will be the new **default version of the Umbraco Documentation**.
```

eshanrnh (approved):
Looks good to me. I think it covers all the required points related to the versioning strategy of our docs 🙌

other-resources/documentation-and-versions.md
sofietoft: I think let's stick with `an` here.
It should be `a` when reading RC as "Release Candidate", but since it's the acronym that written here, it should be `an` 😄 
At least according to Grammarly 😬

PR #4792, opened 2023-01-03
Fixed incorrect configuration key.

AndyButland:
This came up in a support request where we realised there was a misalignment between the key used in the code and in the docs, for Forms 8.

Can go live at any time.


eshanrnh (approved):
Looks good to me 🙌

PR #4797, opened 2023-01-03
Use extension method

erikjanwestendorp:
Wouldn't it be nicer to check if the user is an admin using the extension?


nikolajlauridsen (approved):
Looks great to me, it's the same code in the extension method, so this is just a cleaner way to do the same 👍

sofietoft:
Thanks for the review @nikolajlauridsen 💪 
And thanks again @erikjanwestendorp for simplifying the code in this article 😄 

I'll merge it in 🙌

PR #4800, opened 2023-01-05
Add null check in ContentService Notifications

erikjanwestendorp:


sofietoft:
Definitely makes sense to add a null check to this one.

I'll merge it in!

PR #4804, opened 2023-01-05
Prevents error "cannot convert from 'void' to 'object?'"

craigs100:
This might just be in Rider, but I suspect not.  If the RenderSiteMapUrlEntry(siteHomePage) is inline then there is an error. If not, it runs ok.


sofietoft:
Thanks for the PR @craigs100 !

I agree - it would be strange if this was related to only Rider.
I'll make sure we get this tested using another IDE as soon as possible 💪

sofietoft:
Thanks again for the PR @craigs100 !

I've just run through the Sitemap tutorial, and found that this change is indeed necessary also when using simply the backoffice editor 😅

PR #4806, opened 2023-01-06
fixed links

skartknet:
Links that use uppercase don't work. Github urls are case-sensitive.


sofietoft:
Thanks a bunch for the PR @skartknet 💪 
Great to have these broken links fixed!

PR #4810, opened 2023-01-07
Fixing a typo in the code example

readingdancer:
Removed an S from the Service(s) class name as it should have been Service.


sofietoft:
Great catch @readingdancer ! 💪 Thanks for fixing this!

I'll merge it right in!

PR #4812, opened 2023-01-09
Document asynchronous surface controller actions

kjac:
Adds documentation for how to use `Async` action name suffixes.

Originates from https://github.com/umbraco/Umbraco-CMS/issues/12191

### DIFF
--- a/umbraco-cms/reference/routing/surface-controllers/README.md
+++ b/umbraco-cms/reference/routing/surface-controllers/README.md
@@ -232,6 +232,49 @@ $.ajax({
 
 For more information, see the [Antiforgery in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/anti-request-forgery?view=aspnetcore-6.0#antiforgery-in-aspnet-core-1) article.
 
+### Asynchronous surface controller actions
+
+Surface controller actions can be asynchronous. However, the common naming convention for asynchronous methods (using an `Async` suffix for the action name) will not work by default, due to the inner workings of ASP.NET Core MVC. 
+
+Consider the following asynchronous surface controller action:
+
+```csharp
+public class MySurfaceController : SurfaceController
+{
+    // ...
+
+    [HttpPost]
+    [ValidateAntiForgeryToken]
+    public async Task<IActionResult> SomeMethodAsync()
+    {
+        // do some async work here
+        return CurrentUmbracoPage();
+    }
+}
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @kjac ! 💪 

A couple of questions:
- First of all, the CMS Issue you're referring to is specifying Umbraco 9.3.1 - I just want to make sure this is indeed an issue in v10 and 11 as well before adding it to the main docs 🤞 
- As I read what you're adding here, using Aync methods requires a workaround. Do we recommend using Async methods, if a workaround like you're describing is required? 🤔 
- I did a quick search in our docs to see if we use/recommend Async methods in our docs - and we do in this one: https://docs.umbraco.com/umbraco-cms/tutorials/members-registration-and-login#assigning-new-members-to-groups-automatically - however, it does not include the workaround 🤔 

These questions might be a result of my not understanding the technical parts here, but I just want to make sure we're consistent with our recommendations 💪

kjac:
Hi @sofietoft 👋 

- This is indeed an issue for all versions of Umbraco, as it is tied to the underlying ASP.NET Core MVC framework.
- It depends. Async methods are highly recommended for performing async work (like calling external services or performing database work). If the action does not perform any async work there is very little gain in using async methods.
- The workaround is *only* necessary when using the `Async` suffix for action names (which is a commonly used naming convention, but not at all required). In the linked code, the action name is `HandleRegisterMember` and thus it should just work.

sofietoft:
Thanks for the reply @kjac ! 💪 

Sounds like it was the "actions" part that confused me here, so thanks for helping me understand that a little better.
We have an article specific for "[Surface Controller actions](https://docs.umbraco.com/umbraco-cms/reference/routing/surface-controllers/surface-controllers-actions)" - does it make sense to move this bit into that article instead? 
Looks like it currently only holds some examples and not much else, so not sure 🤔

kjac:
Anytime @sofietoft 😄 

I did consider the "Surface controller actions" article, but it only concerns itself with return values from actions. It might possibly be badly named 😆 but in any case, this PR does not belong in that article.

sofietoft:
Great 💪 
Thanks for double-checking.

Sounds like that Surface Controller Actions article is very simple at the moment then 😅 
Any ideas for a new name for the article?

I'll merge this one in, in the mean time!
Thanks again for adding this to the docs 💪 #H5YR

PR #4830, opened 2023-01-12
Elaborate Linq extensions

greystate:
Hi there 👋 

I'm back ith a couple of updates to the **Filtering using Linq** section, to make the extensions as self-explanatory as possible (I hope - open to critique, of course 😁)

### DIFF
--- a/umbraco-11/filtering-using-linq.md
+++ b/umbraco-11/filtering-using-linq.md
@@ -7,7 +7,7 @@
 .GroupBy(p => p.Category)
 .Take(3)
 .Skip(3)
-.InGroupsOf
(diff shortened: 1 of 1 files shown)


sofietoft:
This is great @greystate ! 💪 
Let me run it by the Curators - and then I'll see if I can make it fit on the sheet without having to make major changes 🤪 😄

sofietoft:
I was sure you added an extra line to the sheet here, but you totally didn't!
Fitting this in the sheet was no problem 😎 

Thanks again for helping us make this cheatsheet even better!
MUCH appreciated 😄

PR #4834, opened 2023-01-15
Update port reference in SMTP settings

nathanwoulfe:
Value for port is an integer not a string, per appsettings-schema.json 😄


sofietoft:
Good catch @nathanwoulfe ! 😄 
Thanks for updating this.

I'll merge it right in 💪

PR #4842, opened 2023-01-17
Check that the cloud project works before pushing

HamDerAndrew:
Adding an extra check to make sure that people don't push changes to the v11 project on Cloud that is in a bad state :D  Feel free to re-word it 😁

The idea came from a recent case I had.


sofietoft:
Thanks for the PR @HamDerAndrew !

Shouldn't we be able to assume that when you create a brand new project on Umbraco Cloud, it works? 🤞 
I know there has been some issues recently, but I'm a little worried about mentioning something like this in our docs.

jonat123:
@HamDerAndrew First of all thanks for the PR! 

I agree with Sofie since it is a clean project with nothing on it, I think it should be safe to assume it is working.

Also, I'm not too fond of having a note insinuating that a clean project on Umbraco Cloud can break without anything happening on it.

So all in all, I don't think we should add the note so I will close this PR for now.

PR #4849, opened 2023-01-20
removed the migration article from the summary

jonat123:
Removed the migration to v11 article from the summary file, so it won't be displayed in the docs as the issue preventing the upgrade should be fixed now.

Gonna keep the article for a while just in case it somehow might be needed again.


sofietoft:
LGTM 💪

sofietoft (approved):

PR #4853, opened 2023-01-23
Updates to Major Upgrade workflow

sofietoft:
As it wasn't possible to upgrade to the latest major version by going through each package indidually, I've updated the guide to update all packages at the same time.

**This needs to be tested by someone else before it's merged**


jonat123 (approved):
Looks good to me, I tested out the upgrade steps on a project, and it was failing when trying to upgrade the packages one by one.

Upgrading them all at once worked!

PR #4863, opened 2023-01-27
Readability improvements - datatypes.md

erikjanwestendorp:
Use Data Type instead of data type


sofietoft:
Thanks for helping keep the docs consistent 🙌

PR #4867, opened 2023-01-27
Readability improvements - udi-identifiers.md

erikjanwestendorp:
Remove several


sofietoft:

Now that we're looking at this article -
I can see that the GUID and UDI reference links are directed at 9 and 10, and not 11 which they really should be in this article 😅 
Would you be up for updating those links as well?

sofietoft:
Thanks @erikjanwestendorp ! 💪 

I'll merge it in!

PR #4868, opened 2023-01-27
Update create-content-programmatically.md

jkulker:
Always look for the code line I'm missing when creating content programmatically. With an Umbraco instance with multiple cultures

### DIFF
--- a/umbraco-cms/reference/management/services/contentservice/create-content-programmatically.md
+++ b/umbraco-cms/reference/management/services/contentservice/create-content-programmatically.md
@@ -23,3 +23,7 @@ person.SetValue("isAuthor", false);
 // Save the child item
 contentService.Save(person);
 ```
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @jkulker ! 💪 

Definitely makes sense to add the culture aspect here!
I'll make sure we run some tests!

Also, I was a little confused about the fact that the Document Types used as examples in this article are called "Person" and "People" 😅 
If I were to change these to, say, "Article" and "News area" or something the like... Do you think that would do the article here any good?

jkulker:
I think the relation between an Article and a News area is very clear. So yes, I would support the change to the Article and news area.

sofietoft:
Excellent! I'll do that while testing the code here as well then 💪

jkulker:
Check and let me know if you need something else :) If I need to change the line of code to fit yours

PR #4869, opened 2023-01-27
Remove duplicated description

wpplumber:
I think this is obvious a duplication.

### DIFF
--- a/umbraco-cms/extending/language-files.md
+++ b/umbraco-cms/extending/language-files.md
@@ -4,7 +4,7 @@ description: "Language files are used to translate the Umbraco backoffice user i
 
 # Language Files & Localization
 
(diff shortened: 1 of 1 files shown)


sofietoft:
You are right that this is a duplication for sure 😅 

How about using the text in the article instead, and then update the description to something that captures the essence of the entire article? 🤔 
I'd be happy to write something up - let me know what you think!

wpplumber:
I don't see my skills suitable to participate in that but anything usefull for **UX** and **SEO**.

sofietoft:
That's completely fair @wpplumber ! 💪 

I'll merge this one in for now, and see about updating the description a bit later on.
Removing the duplicate is definitely a step in the right direction - so thanks for that 😄

PR #4875, opened 2023-01-31
Added the UK region

Jette94:
Added the UK region as baselines are not available here


sofietoft:
Good catch @Jette94 !
Thanks for adding this to the docs 💪

PR #4876, opened 2023-01-31
Registering with C# an Umbraco section hardcoded

wpplumber:
While trying to develop an application as custom section for Umbraco, the solution on docs seems a hardcoded way when asking to register it via Startup.cs


sofietoft:

I am not entirely sure I understand why you're suggesting removing the sample for adding a section through C#.
Did you test it out and it didn't work?
It would be great if you could elaborate a bit on this!

wpplumber:
>I am not entirely sure I understand why you're suggesting removing the sample for adding a section through C#.

Yes I suggest that because the idea of a plugin is its simplicity to integrate easily without manual changes (hardcoded) but for this example it asks for modification on` Startup.cs`.

>Did you test it out and it didn't work?
It would be great if you could elaborate a bit on this!

I refer you to a conflict thread on our [forum](https://our.umbraco.com/forum/using-umbraco-and-getting-started/110726-umbraco-8-custom-section#comment-110726).

sofietoft:
Thanks a bunch for elaborating @wpplumber ! 💪 

I see your point about how these custom sections are usually added as plugins, in which case it really shouldn't be necessary to modify the `Startup.cs` file.

There might still be cases where someone would need to register the sections via C# though 🤔 
Perhaps some rephrasing would be better instead - like, use the recommended approach instead, but this is also an option if it's needed.

I'll run it by our developers - they might have some input as to what it is we want to promote as best practice around this 😄

wpplumber:
>There might still be cases where someone would need to register the sections via C# though

I think for doing some hacks or tests at staging but as Umbraco a professional CMS I think better to go best practices and I knew from discord there is a package validator now.

Anyway I think the idea received, thank you too.

nikolajlauridsen:
Heya.

While I agree that generally speaking it is preferable to add the section in a `package.manifest`, especially since the ability to do it from C# ***may*** be removed in v13. I do think that entirely removing the documentation might be a bit of an overreaction, the feature is still there, and it still works.

In regards to

> Yes I suggest that because the idea of a plugin is its simplicity to integrate easily without manual changes (hardcoded) but for this example it asks for modification on Startup.cs.

Indeed! Packages should be easy to install and use, and not necessarily require you to edit Statup.cs. Fortunately, this is still completely possible, by using a composer, maybe it would be better if we update the example in question to something like this maybe: 

## Suggestion

For your C# type to be discovered by Umbraco at application start up, it needs to be appended to the `SectionCollectionBuilder`.

You can achieve this by updating the `ConfigureServices` method in the `Startup.cs` class, however, since this will often be done from a package, it may be desirable to do this from a composer instead:


```csharp
using Umbraco.Cms.Core.Composing;  
using Umbraco.Cms.Core.Sections;  
  
namespace Umbraco.Cms.Web.UI;  
  
  
public class MySectionComposerComposer : IComposer  
{  
    public void Compose(IUmbracoBuilder builder) => builder.AddSection<MyFacouriteSection>();  
}
```

PR #4877, opened 2023-02-01
Fix title in dashboard tutorial

lassefredslund:
In the creating custom dashboard tutorial, the HTML example was titled with the XML-file. This PR fixes that.


sofietoft:
Great catch @lassefredslund ! 💪 
Thanks a bunch for correcting this!

PR #4880, opened 2023-02-02
More readability improvements

erikjanwestendorp:
- [x] Add PDF to the exceptions list
- [x] Define SSL
- [x] Define FIPS
- [x] Write shorter sentence

### DIFF
--- a/.github/styles/UmbracoDocs/Acronyms.yml
+++ b/.github/styles/UmbracoDocs/Acronyms.yml
@@ -50,3 +50,4 @@ exceptions:
   - YAML
   - PNG
   - ISO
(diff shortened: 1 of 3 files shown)


sofietoft:
Looks good @erikjanwestendorp ! 💪 

Instead of defining PDF - how about we add it to the exceptions list?
I think it's safe to assume that most people know what that is.. what do you think? 🤔

erikjanwestendorp:
@sofietoft Totally agree with that, just updated the PR.

sofietoft:
Super 😄 Thanks!

I'll merge it in 💪

PR #4885, opened 2023-02-03
Fixed auto-generated links...

eshanrnh:
- Added the `needsV8Update: "true"` tag where the "We have moved" box has to be removed since there is no higher version for that article.
- Removed links to v9+ articles since folders are not available in the current branch and throws broken link error on Our.
- Removed warning msg with links to new doc site since the "We have moved" box has the correct link.

### DIFF
--- a/Fundamentals/Code/Subscribing-To-Events/index.md
+++ b/Fundamentals/Code/Subscribing-To-Events/index.md
@@ -3,7 +3,7 @@ meta.title: "Subscribing to events"
 meta.description: "Subscribing to events allows you to execute custom code on a number of events both before and after the event occurs"
 versionFrom: 8.0.0
 versionTo: 8.18.0
-needsV8Update: "true"
+meta.RedirectLink: "/umbraco-cms/fundamentals/code/subscribing-to-notifications"
 ---
 
 # Subscribing to events
--- a/Fundamentals/Code/Subscribing-To-Events/index.md
+++ b/Fundamentals/Code/Subscribing-To-Events/index.md
@@ -3,6 +3,7 @@ meta.title: "Subscribing to events"
 meta.description: "Subscribing to events allows you to execute custom code on a number of events both before and after the event occurs"
 versionFrom: 8.0.0
 versionTo: 8.18.0
+needsV8Update: "true"
 ---
 
 # Subscribing to events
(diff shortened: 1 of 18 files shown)


Fundamentals/Code/Subscribing-To-Events/index.md
```diff
@@ -3,6 +3,7 @@ meta.title: "Subscribing to events"
 meta.description: "Subscribing to events allows you to execute custom code on a number of events both before and after the event occurs"
 versionFrom: 8.0.0
 versionTo: 8.18.0
+needsV8Update: "true"
```
sofietoft, line 6: Would it make sense to link to the "Subscribing the Notifications" here instead? 🤔 
Notification is the equivalent to Events in legacy Umbraco 💪

Fundamentals/index.md
```diff
@@ -62,10 +62,6 @@ Learn more about the features in Umbraco v8.
                 <h5><a href="Setup/Upgrading/">Upgrading</a></h5>
                 <small>How to upgrade your copy of Umbraco to a newer version?</small>
             </div>
-            <div class="col-sm-6">
-                <h5><a href="Setup/Config/">Configuration</a></h5>
```
sofietoft, line 66: How about linking this to the Configuration section in Reference instead?

Reference/Management/Services/SectionService/Index.md
```diff
@@ -1,9 +1,10 @@
 ---
 versionFrom: 8.0.0
+needsV8Update: "true"
```
sofietoft, line 3: I think this should be `versionRemoved: 8.0.0` instead, is it sounds like it has been replaced by something else 🤔

sofietoft (commented):
Added a couple of comments!

Fundamentals/Code/Subscribing-To-Events/index.md
eshanrnh: Makes sense. Will link it to subscribing to notifications article.

Fundamentals/index.md
eshanrnh: I can do that but it anyways is not part of the Fundamentals section and the Reference section landing page has Configuration mentioned it. Let me know if I should go ahead with linking to the Reference section.

Reference/Management/Services/SectionService/Index.md
eshanrnh: Yup, it no longer exists and is redundant. Will add the `versionRemoved` tag instead.

Fundamentals/index.md
sofietoft: Yeah, that's also a good point.
I think it makes sense to just take it out like you've already done 💪

sofietoft (approved):

PR #4892, opened 2023-02-06
Fixed auto-gen links

eshanrnh:
For the https://our.umbraco.com/Documentation/Fundamentals/Code/Source-Control/index-v7, I've removed the `needsV8Update: "true"` tag since we have a v8 article for it.


Fundamentals/Setup/Server-Setup/Load-Balancing/flexible.md
```diff
@@ -1,5 +1,6 @@
 ---
 versionFrom: 8.0.0
+meta.RedirectLink: "/umbraco-cms/fundamentals/setup/server-setup/load-balancing/file-system-replication"
```
sofietoft, line 3: Add "versionRemoved" here instead, as it was removed and replaced in Umbraco 8.

sofietoft (approved):

PR #4894, opened 2023-02-06
Update document-types.md

jacksorjacksor:
Removed a bunch of trailing "\\" - no idea if they actually mean anything for markdown but they're showing up in the docs and don't think that they should be!


sofietoft:
Thanks for the PR @jacksorjacksor ! 💪 
I've tried to figure out why these slashes have been added - I think it's something that the platform added, but can't figure out why.
It doesn't appear to make a difference to the content though, so let's merge it in!

jacksorjacksor:
@sofietoft you 👑

PR #4901, opened 2023-02-08
Small fixes for tree article

erikjanwestendorp:
- [x] Change return type of CreateRootNode method
- [x] Change several to multiple
- [x] Change `angularJS` to `AngularJS`
- [x] Write shorter sentences

### DIFF
--- a/umbraco-cms/extending/section-trees/trees/README.md
+++ b/umbraco-cms/extending/section-trees/trees/README.md
@@ -8,7 +8,7 @@ This section describes how to work with and create trees with Umbraco APIs.
 
 ## Creating trees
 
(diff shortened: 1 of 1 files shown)


sofietoft:

Before doing anything here, I want to check with you... Are you working on shortening up some sentence? 😄 
The PR passes all checks, but that might be because the long sentences aren't part of the changes you've made, and thus not picked up by the checker..

erikjanwestendorp:
@sofietoft Thanks for your review. I was working on updating the sentences. I just commited the latest version. Do you want to check is the sentences are still correct and make sense? 😄

sofietoft:
Hi @erikjanwestendorp 👋 
Apologies for the delay!
The sentences looks great 💪 
I'll merge it in!

PR #4906, opened 2023-02-10
Readability improvements - Update backoffice-search.md

erikjanwestendorp:
- [x] use 'for example' instead of 'eg'


sofietoft (approved):
Thanks for updating this @erikjanwestendorp ! 💪 

I'll merge it right in.

PR #4918, opened 2023-02-14
Duplicate section "Scaling Blocks"

wpplumber:
image


sofietoft:

Just to make sure I understand this one - it's a rebase of this one, right? : https://github.com/umbraco/UmbracoDocs/pull/4876 

Also, did you consider the comments left by @nikolajlauridsen , about registering in a composer?

wpplumber:
I only want the #4918 to be merged, @sofietoft.

Seems like the #4876 passed and I don't want that which is clear for me that the decision is up to the team. (I mentioned my idea)

sofietoft:
I am not entirely sure I understand what you're trying to say here @wpplumber ! 🤔 

This PR still removes the section about registering a section through C#.
As I understood from the comments on [the other PR](https://github.com/umbraco/UmbracoDocs/pull/4876), it should be kept there, but have some different options instead, right?

wpplumber:
>_it should be kept there, but have some different options instead, right?_

Yes

>clear for me that the decision is up to the team. (I mentioned my idea)

>Seems like the https://github.com/umbraco/UmbracoDocs/pull/4876 passed and I don't want that

I mean I didn't want to include the C# removed code commit in this PR, I let you cherry-pick the right commit. (#4918 )

sofietoft:
Thanks for explaining - I think I understand!
I've added back the C# code so that only the updates to the Block Grid Editor remains.

I see that you're also removing some bits and pieces from the v10 version of the article - Could you elaborate a bit more on these changes?

wpplumber:
Others than Duplicate section **"Scaling Blocks"** just avoid them all.

sofietoft:
Alrighty!
I'll close down this PR and create a new PR that only removed the "Scaling Blocks" section.
It would be too much work for me to sit an pick/revert some of the changes in your PR here.

I hope you understand!
And thanks again for helping us keep our docs up to speed 😄

PR #4931, opened 2023-02-16
Readability improvements language files for packages

erikjanwestendorp:
- [x] write shorter sentences

### DIFF
--- a/10/umbraco-cms/extending/packages/language-files-for-packages.md
+++ b/10/umbraco-cms/extending/packages/language-files-for-packages.md
@@ -7,7 +7,7 @@ description: "Information on how to use language files to make your Umbraco pack
 
 # Language file for packages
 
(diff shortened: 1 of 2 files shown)


sofietoft:
I'll merge it in 👏 

Quick question, if you don't mind:
Would you say that it is easier to add changes to more than one version, now that it's all in the same branch? 🤔 
Just curious to how to the new structure looks from the contributor side of things!

erikjanwestendorp:
@sofietoft No of course I don't mind! 😄 I think this is easier. Now it is possible to make changes in one PR for several versions. 👍

sofietoft:
That's great to hear 💪 
I'm also hoping it will actually be easier to find the correct version to add the changes to.

We'll have to see!
Thanks as always for all the work you put into the docs ⭐

PR #4939, opened 2023-02-17
Renamed folders to change duplicate group/folder name in URL

eshanrnh:


11/umbraco-workflow/.gitbook.yaml
```diff
@@ -8,4 +8,6 @@ redirects:
     configuration: getting-started/configuration.md
     dashboards-and-buttons: getting-started/dashboards-and-buttons.md
     notifications: getting-started/notifications.md
-    workflow-content-app: getting-started/workflow-content-app.md
\ No newline at end of file
+    workflow-content-app: getting-started/workflow-content-app.md
+    upgrading-workflow/README: upgrading-workflow/upgrading.md
```
sofietoft, line 12: You need to use the URL path here and not the file name for the old path.

sofietoft (commented):
Instead of "Workflow section overview" - would "Workflow Section Dashboard" be better? 🤔 
Isn't that what the article is showing?

sofietoft (approved):
LGTM 💪

PR #4955, opened 2023-02-22
Update rewrite rules docs

RobertCopilau:
We need to enrich the www redirect rule to contain ".*azurewebsites.net*" pattern, otherwise we will have issues deploying as the rule will redirect to www.{stie}.azurewebsites.net which will break the deployment process.

Fixing this in the site extension is a no-go, as we don't want to hijack the customer's rewrite rules.

DISCUSSION:
Do we want to include a warning hint mentioning this? 🤔  
I've added it, but can be removed or updated if it does not make sense.

### DIFF
--- a/11/umbraco-cms/reference/routing/iisrewriterules.md
+++ b/11/umbraco-cms/reference/routing/iisrewriterules.md
@@ -108,5 +108,5 @@ Another example would be to redirect from non-www to www (except for the Umbraco
 
 {% hint style="warning" %}
 Adding wwww redirect rule requires <add input="{HTTP_HOST}" pattern=".*azurewebsites.net*" negate="true" ignoreCase="true" /> condition to be added as well in order for Umbraco to function correctly.
-Failing to do so will cause the deployment service to fail and content transfer between environments will fail.
+This step is required for the deployment service and the content transfer between environments to function properly.
 {% endhint %}
--- a/11/umbraco-cms/reference/routing/iisrewriterules.md
+++ b/11/umbraco-cms/reference/routing/iisrewriterules.md
@@ -99,9 +99,15 @@ Another example would be to redirect from non-www to www (except for the Umbraco
   <match url=".*" />
   <conditions>
     <add input="{HTTP_HOST}" pattern="^www\." negate="true" />
+    <add input="{HTTP_HOST}" pattern=".*azurewebsites.net*" negate="true" ignoreCase="true" />
     <add input="{HTTP_HOST}" pattern="^localhost(:[0-9]+)?$" negate="true" />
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks for the PR @RobertCopilau ! 😄 

Before moving forward with this, I just need to verify one thing:
Is this related to Umbraco Cloud setups, or is this a general CMS thing? 🤔

RobertCopilau:
Hey @sofietoft, it's an Umbraco Cloud thing :).

RobertCopilau:
Ah, now I see. I'm in the Umbraco CMS docs 😅.
Should the warning mention that is required only if the Umbraco installation is hosted on Umbraco Cloud then?

sofietoft:
Aha! 😄 

Actually, in the text just above the rewrites you're editing it mentions that it doesn't apply to Cloud hostnames, right? 😅 

Would it make sense to enrich one or more of the rewrite rules documented here instead: https://docs.umbraco.com/umbraco-cloud/setup/project-settings/manage-hostnames/rewrites-on-cloud 🤔

RobertCopilau:
You are correct, Sofie, I will take my business to the other documentation page 😄.
I am wandering, however, is it possible that some users might use this example to add rewrite rules to their cloud projects?
Would it make sense to keep the warning (with the text updated slightly)?

11/umbraco-cms/reference/routing/iisrewriterules.md
```diff
@@ -105,3 +105,8 @@ Another example would be to redirect from non-www to www (except for the Umbraco
   <action type="Redirect" url="https://www.{HTTP_HOST}/{R:0}" />
 </rule>
 ```
+
+{% hint style="warning" %}
+Adding wwww redirect rule requires <add input="{HTTP_HOST}" pattern=".*azurewebsites.net*" negate="true" ignoreCase="true" /> condition to be added as well in order for Umbraco to function correctly.
+This step is required for the deployment service and the content transfer between environments to function properly.
```
sofietoft, line 111:
```suggestion
Adding the www redirect rule requires a `<add input="{HTTP_HOST}" pattern=".*azurewebsites.net*" negate="true" ignoreCase="true" />` condition to be added in order for Umbraco Cloud to function correctly.

This step is required for the deployment service and the content transfer between environments to continue to function.
```

sofietoft (commented):
Yes, let's keep the warning here 💪 
I think that's a great idea.

I've made some minor suggestions to the text. Let me know what you think 😁

RobertCopilau:
I think it looks great! 😄

RobertCopilau:
FYI, I've added the warning for the v10 docs as well.

PR #4958, opened 2023-02-22
Enriches the rewrite documentation.

RobertCopilau:
Adds www redirect rule example as well as a warning hint mentioning the  `".*azurewebsites.net*"` pattern requirement


sofietoft:
Looks great @RobertCopilau ! 💪 

I'll merge it right away!

PR #4959, opened 2023-02-22
Updated creating a basic website tutorial

eshanrnh:
- Updated Getting Started article
- Updated Images in some of the articles
- Cleaned up articles to remove `<figure>` tag and add markdown image syntax
- Fixed markdown errors


sofietoft (commented):
Looks great 💪 

You're replacing some of the images - could you make sure that the old images are then deleted from the repo?
Did you try separating the images from the ordered lists to see how that would look?

eshanrnh:
Ah, good point. I went through each image and deleted all the old and unused images 💪

No, I did not try separating the images from the ordered lists since I made these changes using VS code and not GitBook. It was faster to do it through VS code 🙈

sofietoft:
Super 💪 

You can also test this with the ordered list locally: Move the images to a separate line, instead of indent it like the list items.
And I think it might even be easier to do locally, as you will not have to amend the list numbers at all, as they should stay the same! 😄 

But this can always be tested at a later point of course!

PR #4966, opened 2023-02-23
Update to IMemberManager

OwainWilliams:
Adding a missing namespace to allow for IMemberManager to work correctly.

If using Rider or Visual Studio this using may automatically appear but not always. Useful to have it in the docs just in case it doesn't appear automatically via intellisense.

### DIFF
--- a/10/umbraco-cms/reference/querying/imembermanager.md
+++ b/10/umbraco-cms/reference/querying/imembermanager.md
@@ -18,6 +18,7 @@ There are different ways to reference MembershipHelper:
 While working with templates, the methods are available when you inject `@IMemberManager` to access member data:
 
 ```csharp
+@using Umbraco.Cms.Core.Services;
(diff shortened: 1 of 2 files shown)


sofietoft:
Looks great @OwainWilliams ! 💪 
Makes a lot of sense to add the namespaces to the sample.

You up for verifying whether this should also be updated in the v10 version?

OwainWilliams:
I'll update the V10 version too, as that is the version I actually had the issue with but wasn't sure if this would be back ported to the V10 docs. I'll update this PR.

OwainWilliams:
V10 docs now updated.

sofietoft:
Thanks @OwainWilliams ! 😄 

I'm afraid we're not so fancy here in the docs repo with automation and such 🙈 
So for now, updating both versions is a manual process!

I'll get this one merged in!

PR #4969, opened 2023-02-24
Fixes broken links

abjerner:
As pointed out by @hfloyd elsewhere, the links behind the cards are broken. The destination in two of the links is literally `broken-reference/`.

I'm not entirely sure how the README.md is converted and shown on GitBook, but I've tried fixing the links.


sofietoft:
Thanks for the PR @abjerner !
Had to link the cards directly to the first article under both the Tutorial and the Fundamentals section in order to get these to link correctly 💪 

I'll merge it right in!

PR #4974, opened 2023-02-27
fixing typo

alina-tincas:


sofietoft:
Good catch @alina-tincas !
I'll merge it in 💪

PR #4982, opened 2023-03-01
Fix typo

greystate:
"Rendering From Scripts" => "Rendering Forms Scripts"


sofietoft (approved):
Good catch @greystate !
I'll merge it right in.

PR #4987, opened 2023-03-02
Update docs on usage of language-files

wvhooijdonk:
Language files are not only used in backoffice, but also for member identity errors in a website. This is missing in the docs.

### DIFF
--- a/umbraco-cms/extending/language-files.md
+++ b/umbraco-cms/extending/language-files.md
@@ -10,9 +10,8 @@ description: >-
 # Language Files & Localization
 
 Language files are used for translating:
-- The Umbraco backoffice user interface so that end users can use Umbraco in their native language.
-  This is particularly important for content editors who do not speak English.
-- The member identity errors in a Umbraco website so that end users can use Umbraco in the website language.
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks a bunch for the PR @wvhooijdonk ! 💪 
I've made some tiny adjustments, and will make sure we get this reviewed as soon as possible!

sofietoft:
Oh! I just realized that this PR is targeting a branch that we currently do not use.
So sorry for the confusion!

This should be added to the v10 docs, right?
The file you need is located on the **main** branch, in the 10/umbraco-cms/extending folder.
If it's also valid for v11, then the 11/umbraco-cms/extending as well 💪 

You can also use the "Edit on GitHub" button, when viewing the article on the docs site.

wvhooijdonk:
oh, bummer. I'll do that.
But the guidelines should be changed I think, because that's where I got the gb/10 hint from.
See https://docs.umbraco.com/welcome/contribute/pull-request#option-1-creating-a-pr-directly-on-github

sofietoft:
Yes, you're absolutely right!
It's been on my to-do for a couple of weeks - I'll make sure it's updated!

PR #4992, opened 2023-03-02
Deleted nested-contend.md

C0derPr0:
Nested Content was removed in v11 and should be removed from the documentation (or I could write something to that effect).


sofietoft:
Thanks for the PR @C0derPr0 !

You're absolutely right - the Nested Content property is no longer in Umbraco 11.
The article should definitely be deleted, but there are a few things we also need to consider:
- Links to this article will be broken when we remove the article
- We need to redirect this path to something else - perhaps the Block List article? 🤔
- We should remove the link to this article from the list of all the available property editors...
- The article should be removed from the 11/umbraco-cms/SUMMARY.md file which lists all articles on the site.

I can make the redirect - are you up for removing the link from the property editors' article? 😄

C0derPr0:
Hiya There, @sofietoft,

Sure. I can help with that. Should I also update the Block List article so users aren't confused about where the nested content control went?

sofietoft:
Ah, yes. That's a good idea!

Perhaps a note in the top about how the Block List has replaced the Nested Content property.

C0derPr0:
@sofietoft 

Alrighty... I made all of the requested changes and updated Block List to include a note about nested content. 

I noticed a note at the top: "A default Umbraco CMS installation does not ship with a defined Data Type using the Block List editor. In order to start using the property, follow the steps outlined below." This was not true for a couple of versions, so I replaced that text with my note.

C0derPr0:
Argh! I pushed directly to the main branch. Should I roll back those changes and push to this patch? I didn't think I'd have permission to do that. 😮 Sorry!

sofietoft:
I don't see any commits being pushed directly to the main branch, and you also should not have permission to do that 😅 

But, just to comment on the "defined Data Types" for the Block List - I don't see any in my Umbraco 11.1 installation, so I'm pretty sure this is still true. It's just not used in a default Data Type..

C0derPr0:
It's rather early here.... I pushed to the c0derPr0/UmbracoDocs main branch and freaked out. I deleted the comment, but unfortunately couldn't delete it from your email! 😂 I moved my changes into the correct branch. You should see those now.

sofietoft:
Excellent! 😄 

There are some conflicts currently - suspect it's due to some commits made by GitBook.
We're looking into resolving them so we can get this one merged in.

Also, did you see my comment about removing the Data Types note at the top of the Block List article?
"But, just to comment on the "defined Data Types" for the Block List - I don't see any in my Umbraco 11.1 installation, so I'm pretty sure this is still true. It's just not used in a default Data Type.."

sofietoft:
Hi @C0derPr0 ! 👋 

It's been a while here, and since there are quite a few conflicts on your PR, I'll be closing it down for now.

From the changes you were adding here, the following articles still need to be updated:
- [Data Types](https://docs.umbraco.com/umbraco-cms/v/11.latest/fundamentals/data/data-types)
- [View Components](https://docs.umbraco.com/umbraco-cms/v/11.latest/reference/templating/mvc/viewcomponents)
- [Install Deploy](https://docs.umbraco.com/umbraco-deploy/getting-started/install-configure)

I'll add a task internally for this - are you still up for it, please don't hold back submitting another PR.
I promise we'll get that solved in a timely!

Let me know if you have any questions!

PR #5005, opened 2023-03-07
readability improvements - contentsniffingprotection

erikjanwestendorp:
- [x] define MIME


sofietoft:

I know the rule we've setup says to define the acronym on the first occurrence on a page. I think though, that in this circumstance, it would be OK to define it in the line below the title?
The title will also be the title used in the navigation, so we wanna keep these titles somewhat short as well 😅 

What do you think?

sofietoft:
Thanks @erikjanwestendorp ! 💪 
Made one tiny adjustment - looks like a word has gone missing 🤪 

I'll merge it in!

PR #4766, opened 2022-12-19
First draft of Azure Pipelines for the DevOps docs

mattou07:
Simply a first draft for a new DevOps focused section in the Umbraco Docs. To enable Umbracians to have somewhere to grab YAML to easily build an Umbraco solution.

PR #5006, opened 2023-03-07
Create a basic build pipeline in Azure DevOps

mattou07:


sofietoft:
Thanks for the PR @mattou07 ! 💪 
I'll make sure we review it as soon as possible.

A few comments about the structure you've created:
- Tutorials is a great fit here!
- I don't think there's a need to create two separate folders here. Placing the tutorial directly in the `azure-devops` folder would be fine, right?

So the structure would be this instead:
- Tutorials
- - Azure Devops
- - - Getting started
- - - Creating [...]

Perhaps even consider whether the "Getting Started" page is necessary. The two could be combined? 🤔 

Finally, all new articles need to be added to the `SUMMARY.md` file for that product.
You can find it in the root of the `/10/umbraco-cms` directory.

sofietoft:
Hi again @mattou07 !

As I send this to our developers for review, I was reminded that we have already discussed adding this tutorial earlier this year. We came to the conclusion that it is not something we can add to our docs (https://github.com/umbraco/UmbracoDocs/pull/4766#issuecomment-1370710171).

Apologies for the confusion and the mixed messages here. This must have completely slipped my mind.
I'm afraid we will not be accepting this PR.

I remember that you had some other great material listed in your plans for this section - like, the articles about logging.
Those would still be great, as they tie more directly into Umbraco and to something that has already been configured elsewhere.

I hope you understand, and do let me know if you have any questions.

PR #5010, opened 2023-03-08
Added a link to Examine events

drpeck:
Our Umbraco docs used to cover the Examine events `TransformingIndexValues` with the view to allowing developers to customise the External index. This has been dropped because we're not seeking to document Examine, but it leaves developers a little unsure how to achieve customisation of indexes.

I've sort a minimal update to avoid this confusion by adding a link to the Examine events documentation, where they are briefly mentioned.

### DIFF
--- a/10/umbraco-cms/reference/searching/examine/indexing.md
+++ b/10/umbraco-cms/reference/searching/examine/indexing.md
@@ -9,7 +9,7 @@ versionTo: 10.0.0
 
 You can modify the built-in indexes in the following ways:
 
-* Events - giving you control over exactly what data goes into them and how the fields are configured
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks for bunch for making the PR @drpeck !

I just this morning saw on Twitter that Shannon released new docs for Examine - unfortunately that breaks the link you've put in here 🙈
I've browsed through and found this instead: https://shazwazza.github.io/Examine/articles/indexing.html#events
Seems like the link we're after here, right? 🤞

drpeck:
Good spot. Updated

sofietoft:
Thanks for correcting the link @drpeck !

I'll make sure we merged this in!

PR #5017, opened 2023-03-09
Proposing change to include legacy version(s)

jesperlange:
As there are still plenty of people working on v8 who might want to migrate to cloud, I believe it's still worth to have these merge steps included, as they are not identical to the steps on v9+.

Here's just an idea on how to include it.


sofietoft:
This should definitely be added back into the docs, yes 💪 

How about using tabs for it instead? 🤔 
See how they are used here as an example: https://docs.umbraco.com/umbraco-workflow/upgrading-workflow/upgrading/migrating-workflow 

That way we avoid people starting to follow the steps, before they notice the option for 7/8 as well.

jesperlange:
That seems like a really good idea to have it in tabs instead.
Just so I'm sure I'm on the same page as you.
Whole step 6 is in tabs, default tab is new versions, other tab is legacy versions.
Am I understanding that correctly?

jesperlange:
Good point, I've removed it 🙈

sofietoft:
I think we're ready to merge.
Normally we can preview the changes before merging, but for some reason that doesn't work here.

So, I guess we'll just have to try it out 😅

PR #5019, opened 2023-03-09
Licensing article for Forms/Deploy

eshanrnh:
A licensing article for each of the products...


10/umbraco-deploy/the-licensing-model.md
```diff
@@ -92,15 +57,13 @@ Once you've configured your license with the correct domains, you are ready to i
 1. Download your license from your Umbraco.com account - this will give you a `.lic` file
 2. Place the file in the `/umbraco/Licenses` directory in your Umbraco installation
 
-The `.lic` file must be placed in the `/umbraco/Licenses` directory in order to be registered by Umbraco Deploy or Umbraco Forms. If the file isn't placed correctly, the application will automatically switch to trial mode.
-
-![Installing Umbraco Forms license](../images/install-forms-license-v9.gif)
+The `.lic` file must be placed in the `/umbraco/Licenses` directory in order to be registered by Umbraco Deploy. If the file isn't placed correctly, the application will automatically switch to trial mode.
```
sofietoft, line 60: I don't think it will switch to trial mode, when it's deploy, does it? 🤔 
In the beginning of the article it mentions that there isn't at least... 

Or maybe it should just be called something different - not sure 😅

PR #5026, opened 2023-03-14
Point to newest major version in API Doc

jemayn:


sofietoft:
Good catch @jemayn !
Thanks for updating this 🙌

PR #5031, opened 2023-03-15
Updated Deploy release notes.

AndyButland:
To include the updates for the patch release going out on 21st March.

These will also be added in the old location on our, where people will be directed to to view for now.


jonat123 (approved):
@AndyButland Looks good to me :) 

Should I wait with merging it until the 21st or should it just be merged right away? :)

AndyButland:
No, right away is fine, thanks.  I currently publish these at the point I prepare the release and are confident on the date it'll go out, which is usually in the week before we actually release the update to NuGet and/or Cloud.

So there may be a bit more short-notice on these than usual docs updates I'm afraid.  Previously I was in control of pushing these live when they lived on our.umbraco.com, but now of course I need to - and it's best to - go through you and the docs team.

jonat123:
Merging it in then 😄

PR #5035, opened 2023-03-17
Adds release notes for Forms 10.3.0 and 11.1.0.

AndyButland:
These are related to a release going out on March 21st, but I've got them on our.umbraco.com too so they can go live whenever you prefer.

### DIFF
--- a/10/umbraco-forms/release-notes.md
+++ b/10/umbraco-forms/release-notes.md
@@ -18,6 +18,14 @@ In this section, you can find the release notes for each version of Umbraco Form
 
 <summary>Version 10</summary>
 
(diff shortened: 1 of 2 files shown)


AndyButland:
Updates all look, have accepted.  But I'm not quite clear on what you are suggesting with the RC.  If you remove the "10.3.0-rc001" heading, then the individual updates under there will appear like they are in the section above.  Perhaps if you do that you also need to move all the individual lines under the "10.3.0" release?

AndyButland:
Ah, I see.  Yes - sounds good to me.

eshanrnh (approved):
Looks good to me 💪 Merging it in.

PR #5043, opened 2023-03-20
Remove a line from the article

sergiutesu:
Removed the following line: 
"By default, we add Basic Authentication to all trials and non-live environments."

This is not correct I checked with Søren and v9+ projects and all trials don't have the Basic Authentication enabled by default. The Public access option will be enabled on all the trials at some point in the next sprints.


jonat123 (approved):
LGTM 👍

PR #5044, opened 2023-03-20
Add OnParseCommandsAsync example

ronaldbarendse:
This adds some context and example code on how to add your custom ImageSharp command parsing logic without overriding the build-in maximum width/height validation.


ronaldbarendse:
> Does this apply to v10 as well?

Yes, this applies to v9 and up (v8 uses ImageProcessor, instead of ImageSharp) 👍🏻

sofietoft:
Thanks. Tried to add it to the branch here, but it didn't get added to the PR for some reason.

Would you be up for adding it to the v10 version of this article? 🤞

ronaldbarendse:
Looks like you already managed to add it to the v10 page (just without the empty lines) 😄

PR #5045, opened 2023-03-21
Added a note to the Depth parameter.

RyuLindow:


RyuLindow:
The Depth parameter only applies to the REST and Preview APIs.

sofietoft:
Super! And that also means that this limit (0 - 5) applies to all, right? 🤞🏻

RyuLindow:
Yes, that's correct.

PR #5050, opened 2023-03-24
readability improvements - contentapps

erikjanwestendorp:
- [x] back-office -> backoffice


sofietoft:
Good catch @erikjanwestendorp ! 💪 
I'll merge it right in.

PR #5053, opened 2023-03-27
It said "hasOnce" - changed to Once only

Jette94:


sofietoft:
Great catch @Jette94 ! 💪 
I'll merge it right in!

PR #5058, opened 2023-03-28
Fixed links

eshanrnh:
Changed from external to internal links


sofietoft:
The links are working as internal links now.
I fixed it directly on GitBook yesterday, as I tried the same like in the PR, but to no avail.

PR #5062, opened 2023-03-29
removed Visual Studio

alina-tincas:
we have had cases where clients try to pull and push changes to and from Cloud and confirmed it is slower than it should be and are asking us to help with this, which is not something we can do. We always recommend clients to use a git client such as Fork, SourceTree, Gitkraken etc as they are faster and work as expected.

### DIFF
--- a/umbraco-cloud/set-up/working-locally.md
+++ b/umbraco-cloud/set-up/working-locally.md
@@ -18,7 +18,7 @@ We recommend using the following tools to work with a local clone of your Umbrac
 
 * Git needs to be installed on your computer
   * Example: [Fork](https://git-fork.com/), [SourceTree](https://www.sourcetreeapp.com/), or [GitKraken](https://www.gitkraken.com/)
-* Visual Studio
+
 
 {% hint style="info" %}
 In the root of your local project, there is a README file containing information about the project structure and the build process on Umbraco Cloud.&#x20;


alina-tincas:
yes, we do recommed VS to work with Umbraco, but right now it is under Tools sections, and that section is mostly referring to how you can clone down from Umbraco Cloud....

alina-tincas:
or maybe an idea would to elaborate a bit more on what you need VS for so clients don´t confuse that we recommend them to clone the project with it?

sofietoft:
The article is about "**Working** locally with Umbraco Cloud" and not only cloning...
For this you need several different tools - a Git client AND Visual Studio.

I do not agree that we should take out Visual Studio from this list, as it is a prerequisite to working with a Cloud project locally.

If it is not clear enough at the moment, we can update the text where is mentions the cloning part, and ensure it refers to the Git tool at that point.

Would that be a way to solve this perhaps?

sofietoft:
Feel free to give it a go.

I'll add it to our task list to get it updated.

alina-tincas:
Sure, am thinking something like this:
- Git needs to be installed on your computer - in order to clone down the project and push your changes from local
- Visual Studio - which is to used to work with your code locally

sofietoft:
Yeah, I think that makes sense 💪 
We should still recommend a couple of different Git Clients for readers who've never worked with git before!

I can see we currently only mention that git needs to be installed, and then we give some examples.
Perhaps this needs to be three bullets instead: Git installed, Git client, and Visual Studio.

umbraco-cloud/set-up/working-locally.md
```diff
@@ -18,7 +18,7 @@ We recommend using the following tools to work with a local clone of your Umbrac
 
 * Git needs to be installed on your computer
   * Example: [Fork](https://git-fork.com/), [SourceTree](https://www.sourcetreeapp.com/), or [GitKraken](https://www.gitkraken.com/)
-* Visual Studio
+
```
jonat123, line 21:
```suggestion
* Visual Studio - for running the project on your local machine.
```

umbraco-cloud/set-up/working-locally.md
```diff
@@ -18,7 +18,7 @@ We recommend using the following tools to work with a local clone of your Umbrac
 
 * Git needs to be installed on your computer
   * Example: [Fork](https://git-fork.com/), [SourceTree](https://www.sourcetreeapp.com/), or [GitKraken](https://www.gitkraken.com/)
```
jonat123, line 20:
```suggestion
  * We recommend using one of the following git-clients if you are new to git:: [Fork](https://git-fork.com/), [SourceTree](https://www.sourcetreeapp.com/), or [GitKraken](https://www.gitkraken.com/)
```

umbraco-cloud/set-up/working-locally.md
```diff
@@ -18,7 +18,7 @@ We recommend using the following tools to work with a local clone of your Umbrac
 
 * Git needs to be installed on your computer
```
jonat123, line 19:
```suggestion
* Git needs to be installed on your computer in order to clone down the project and push your changes up to Cloud.
```

jonat123:
@alina-tincas @sofietoft I've made some suggestions to this PR based on your previous comment 😄

Please let me know what you think!

sofietoft (approved):

PR #5066, opened 2023-03-29
Release details for Deploy 4.8/9.6/10.2/11.1 

AndyButland:
This PR contains updated relevant for the next Deploy release due April 11th.  Ideally these updates would be published first thing on that day.


ronaldbarendse (dismissed):
The CMS documentation should be updated to 'Distributed Locks', because using the database for these locks is an implementation detail, see [relevant settings](https://docs.umbraco.com/umbraco-cms/reference/configuration/globalsettings#distributed-locking-mechanism).

Should we also add more information about the difference between read and write locks, timeouts, lazy (default `ReadLock()` and `WriteLock()`) and eager locks (`EagerReadLock()` and `EagerWriteLock()`), the fact that they are released when the scope completes/rolls back and the different IDs can be used to 'scope' the lock to a certain operation/entity/database table.

If we want to go into the implementation details, both `SqlServerDistributedLockingMechanism` and `SqliteDistributedLockingMechanism` do indeed use the `umbracoLock` database table. However, the `value` column doesn't represent whether the lock is acquired or not: it's the fact that the value can be updated to a different value. SQLite depends on the fact there can only be a single database writer and SQL Server uses row locks using the `REPEATABLEREAD` isolation level.

jonat123 (approved):
@AndyButland I don't have anything to add to the PR so I'll approve it from my end.

@ronaldbarendse are you planning on adding more information to this PR or are we just going to leave it as it is now?

ronaldbarendse:
@jonat123 These changes look good and can be extended with further details in another PR, so you can merge this as-is 👍🏻

PR #5074, opened 2023-03-31
Added details on use of telemetry data on Umbraco marketplace

AndyButland:


marketplace-and-integrations/marketplace/package-telemetry-data.md
```diff
@@ -0,0 +1,91 @@
...
+        public void Filter(List<PackageManifest> manifests) =>
+            manifests.Add(new PackageManifest
+            {
+                PackageName = "MyPackage",
+            });
+    }
+}
+```
+
+The class needs to be registered as part of application startup, via an extension method (as shown here), or a composer:
```
sofietoft, line 74: The example below is an extension method, right?
If so, I think we should update this text a bit:

```suggestion
The class needs to be registered as part of the application startup, via a composer or an extension method as shown here:
```

sofietoft (commented):
Thanks for the PR @AndyButland ! 💪 

I've added a single comment.
Also, the new article needs to be added to the SUMMARY file as well, in order to be added to the GitBook site.

AndyButland:
Thanks @sofietoft - have accepted your improved wording and added the link on the summary page.

Issue #5073, opened 2023-03-31
Custom UrlProviders should not return base.GetUrl


simonech:
## What type of issue is it? 
Wrong documentation

## What article/section is this about?
https://docs.umbraco.com/umbraco-cms/reference/routing/request-pipeline/outbound-pipeline#custom-url-provider

## Describe the issue
In the article, it is explained to return `base.GetUrl(content, mode, culture, current);` if the doctype doesn't match the one for which you want to create the custom URL.

This is fine if you have just one custom URL provider, but if you want to have multiple URL providers, returning the default URL will stop Umbraco to cycle through all the other providers, so only the first will be executed.

The code sample should return "null" if the condition is unmet.

simonech:
I just noticed it is a duplicate of #4265 
I'll see if I can update the doc myself

sofietoft:
Thanks for reporting the issue and creating the PR.

alina-tincas:
Hi @simonech it seems that this PR #4265 is currently merged therefore I am closing this issue 😊

Issue #4265, opened 2022-08-05
Custom Url Provider Example Circumvents other Url Providers


eden-jh:
## What type of issue is it? (Choose one - delete the others)
Other (write it here)

## What article/section is this about?
[Outbound Request Pipeline - Custom Url Provider Example](https://our.umbraco.com/documentation/reference/routing/request-pipeline/outbound-pipeline#example-1)

## Describe the issue
In the example given, the custom Url Provider is written to apply only to a specific content type, but it's actually written in a way that prevents other Url Providers from being used. 

For content of different content types, it [returns the base GetUrl result](https://github.com/umbraco/UmbracoDocs/blob/70aace625e0550037d48cb594b972c68b74e70c9/Reference/Routing/Request-Pipeline/outbound-pipeline.md?plain=1#L282-L306). 

```
if (content.ContentType.Alias == "productPage")
{
   ...
}
// Otherwise return the base GetUrl result:
return base.GetUrl(content, mode, culture, current);
```

Because the [`GetUrl` method of `UrlProvider`](https://apidocs.umbraco.com/v9/csharp/api/Umbraco.Cms.Core.Routing.UrlProvider.html#Umbraco_Cms_Core_Routing_UrlProvider_GetUrl_Umbraco_Cms_Core_Models_PublishedContent_IPublishedContent_Umbraco_Cms_Core_Models_PublishedContent_UrlMode_System_String_Uri_) looks for the first non-null Url in the IEnumerable of available Url Providers, returning the base GetUrl result for other content types prevents any subsequent Url Providers from ever being called. Neither the text of the section nor the comments in the example tell the reader that it will do this.

If the example was written this way intentionally, it would be great for the content to say so, and describe what reason(s) one might have for using this approach. If not, the code example should be updated to return null for content types that this custom Url Provider should not apply to.

sofietoft:
Hi @eden-jh !
Thanks for reporting this issue!
We agree that the example in the article here might be a little too simplistic, and doesn't consider the bigger picture where you might have multiple UrlProviders.

How about we add a note about that to the article, or make the sample return null?
What do you think would work best for explaining the concept of UrlProviders?

eden-jh:
I think making the sample return null when that UrlProvider does not apply makes sense, because that seems like the most likely use case.

Then the description could explain how the order and the GetUrl methods of the registered UrlProviders determine which one will be used for a content item:
- GetUrl should return null if that UrlProvider should not apply to the content item
[...]

PR #5075, opened 2023-03-31
Added explanation on return value of GetUrl

simonech:
Made the code sample return `null` and added an explanation on the sequence of execution.

Solves tickets #4265 and #5073

### DIFF
--- a/11/umbraco-cms/reference/routing/request-pipeline/outbound-pipeline.md
+++ b/11/umbraco-cms/reference/routing/request-pipeline/outbound-pipeline.md
@@ -282,8 +282,8 @@ public class ProductPageUrlProvider : DefaultUrlProvider
                     return new UrlInfo(customUrl, true, defaultUrlInfo.Culture);
                 }
             }
-            // Otherwise return the base GetUrl result:
-            return base.GetUrl(content, mode, culture, current);
+            // Otherwise return null
+            return null;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @simonech ! 💪 

I'll make sure we review the PR as soon as possible.

Zeegaan:
LGTM 👍

Zeegaan (approved):

Zeegaan:
Looks good from a code perspective 👍 I'll pass it back to the docs team  @sofietoft 💪

11/umbraco-cms/reference/routing/request-pipeline/outbound-pipeline.md
```diff
@@ -307,6 +307,10 @@ namespace RoutingDocs.UrlProviders
 }
 ```
 
+{% hint style="info" %}
+If you want to have multiple UrlProvider, you can add them one after the other with multiple `Insert` methods. Umbraco will cycle though all the providers registered until it finds one that doesn't return `null`. If all custom url providers return `null` it will fall back to the default url provider. Note that the last added with `Insert` is the first being executed.
```
sofietoft, line 311:
```suggestion
If you want to have multiple URL providers, you can add them one after the other with multiple `Insert` methods. Umbraco will cycle through all the providers registered until it finds one that doesn't return `null`. If all custom URL providers return `null` it will fall back to the default URL provider. The last added with `Insert` is the first that will be executed. 
```

sofietoft (commented):
Just a couple of suggestions from me @simonech !
Let me know if you think it makes sense.

PR #5080, opened 2023-04-04
Add culture to the "Creating content programmatically" article

eshanrnh:
I've updated the sample to set the localized language and provided a reference to retrieve multilingual languages.

Further updated the sample to reflect "Catalogue"/"Product" instead of "People"/"Person"

Let me know if anything else needs to be added.


jonat123 (approved):
LGTM 💪

PR #5092, opened 2023-04-10
readability improvements - ajaxforms

erikjanwestendorp:
- [x] Remove the word easily

### DIFF
--- a/10/umbraco-forms/developer/ajaxforms.md
+++ b/10/umbraco-forms/developer/ajaxforms.md
@@ -409,7 +409,7 @@ Firstly, with server-to-server integrations you will want to disable the antifor
 
 This is done by setting the `Umbraco:Forms:Security:EnableAntiForgeryTokenForFormsApi` configuration key to a value of `false`.
 
-You should then configure an API key `Umbraco:Forms:Security:FormsApiKey`.  This can be any string value, but shouldn't be easily guessable by a brute force attack.
+You should then configure an API key `Umbraco:Forms:Security:FormsApiKey`.  The string value can be any, but it should not be susceptible to a brute force attack.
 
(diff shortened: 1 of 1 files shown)


10/umbraco-forms/developer/ajaxforms.md
```diff
@@ -409,7 +409,7 @@ Firstly, with server-to-server integrations you will want to disable the antifor
 
 This is done by setting the `Umbraco:Forms:Security:EnableAntiForgeryTokenForFormsApi` configuration key to a value of `false`.
 
-You should then configure an API key `Umbraco:Forms:Security:FormsApiKey`.  This can be any string value, but shouldn't be easily guessable by a brute force attack.
+You should then configure an API key `Umbraco:Forms:Security:FormsApiKey`.  The string value can be any, but it should not be susceptible to a brute force attack.
```
sofietoft, line 412: Not sure if this should be "any value" or "anything", like I've suggestion.
But I felt something was missing from the sentence. What do you think?

```suggestion
You should then configure an API key `Umbraco:Forms:Security:FormsApiKey`. The string value can be anything, but it should not be susceptible to a brute force attack.
```

sofietoft (commented):

I've added a single comment.

10/umbraco-forms/developer/ajaxforms.md
erikjanwestendorp: @sofietoft I had exactly the same doubt 🙈. Just updated the PR with your suggestion 😄

erikjanwestendorp:
@sofietoft  I had exactly the same doubt 🙈. Just updated the PR with your suggestion 😄

sofietoft:
Great! Let's go with that then 😄 

I'll merge it in.

PR #5107, opened 2023-04-16
Workflow 10.1.2 updates

nathanwoulfe:
Adds doco for new configuration options in 10.1.2 and 11.1.2
Adds release notes for versions 10.1.0 and 11.1.0 through to current

**Release:** April 18th

### DIFF
--- a/10/umbraco-workflow/SUMMARY.md
+++ b/10/umbraco-workflow/SUMMARY.md
@@ -3,6 +3,7 @@
 * [Umbraco Workflow Documentation](README.md)
 * [Installation](installing-workflow.md)
 * [Licensing](licensing.md)
+* [Release notes](release-notes.md)
 
 ## Getting Started
 
--- a/10/umbraco-workflow/getting-started/configuration.md
+++ b/10/umbraco-workflow/getting-started/configuration.md
@@ -100,6 +100,7 @@ All available `SettingsCustomization` options are illustrated below along with t
       “General”: {
         “FlowType”: 0|1|2 matching the FlowType enum values,
         “LockIfActive”: bool,
+        "MandatoryComments": bool,
         “AllowAttachments”: bool,
         “AllowScheduling”: bool,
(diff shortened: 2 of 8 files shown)


sofietoft:
Thanks for the PR @nathanwoulfe !
Let me know when it's ready for review.

Also, could you add the release notes to the 11/umbraco-workflow directory as well?
And the 11+ notes can be omitted from the 10/umbraco-workflow directory 💪

nathanwoulfe:
Hey @sofietoft, plan is to release the patch tomorrow, I'll publish this PR once the patch is out. 

I'll also add the docs to the v11 folder, forgot that we're fancy and have version-specific pages :)

sofietoft:
Oh! Alright.

Please make sure that the Release notes articles are added to the SUMMARY.md file in both directories as well.
Otherwise, it will not show up on the platform.

In the Umbraco Forms and Deploy docs, the release notes are added to the root of the content.
Should we do the same for the Workflow release notes? Just to follow a similar structure for each of the products.

Finally, if want this merged tomorrow - could you mark it as ready for review, so that I can review it properly? 😄

nathanwoulfe:
Hey @sofietoft I've updated for 11 too - I've stuck to the release notes structure from Forms, where we list all version info in both the v10 and v11 docs. If the preference is to keep just v10 for v10 and 11 for 11, I can change to suit.

No rush with reviewing and publishing either, release is a small patch with a couple of configuration options, nothing urgently needing docs to explain

sofietoft:
I've fixed a heading that was broken on GitBook - missing a blank space 😅 

Also, on both the Forms and Deploy release notes, we do not have 11 mentioned in the 10 versions.
Let's keep it consistent and take out 11 from the 10 version here as well 💪 

Finally, I like how you're highlighting the new features in the release notes!
Would it make sense to also link to the articles detailing these new features? I think that would be a nice touch!
I'm happy to add the links. Let me know what you think!

PR #5108, opened 2023-04-17
Ignore Umbraco case when used in a URL

nathanwoulfe:
Vale should only warn when `umbraco` is not used in a URL - changes here add a regular expression to only match `umbraco` when it is NOT preceded by a `.` or `/` character.


sofietoft:
This is great @nathanwoulfe !
Thanks 💪

PR #5113, opened 2023-04-18
Add clarifty on what PublishedValueFallback is

Matthew-Wise:
In the Fallback section all the examples have the first parameter of PublishedValueFallback  but dont explain what it is or how to access it.

Hopefully this adds the extra bit of information for people.

I do wonder if we also need to highlight its optional, as there is a "friendly" overload.

### DIFF
--- a/11/umbraco-cms/reference/querying/ipublishedcontent/properties.md
+++ b/11/umbraco-cms/reference/querying/ipublishedcontent/properties.md
@@ -230,6 +230,11 @@ var mediaItems = Model.Value<IEnumerable<IPublishedContent>>(PublishedValueFallb
 
 If the current content item doesn't have the requested value, use an alternative 'fallback' value in its place.
 
+Each of the examples below make use of an injected PublishedValueFallback. This is achieved by adding the following at the top of your Razor file:
+
+@inject IPublishedValueFallback PublishedValueFallback
+This parameter is optional, but can make unit testing easier.
+
 ### Fallback to Default Value
 
 If a content page has a 'title' property, to fallback to use the 'Name' of the content item if the 'title' is not populated. Set the Fallback type to be Fallback.ToDefaultValue, and set the DefaultValue accordingly:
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks for the PR @Matthew-Wise ! 💪 
I've fixed a few grammar issues..
Also, following along on Discord - sounds like this injection is optional? Should we add a note about that as well? Or perhaps not a note - but mention it at least?
What do you think?

Matthew-Wise:
"I've fixed a few grammar issues.." Story of my life!

I also noticed that there is a v11 version so I have updated that one as well :)

11/umbraco-cms/reference/querying/ipublishedcontent/properties.md
```diff
@@ -230,6 +230,11 @@ var mediaItems = Model.Value<IEnumerable<IPublishedContent>>(PublishedValueFallb
 
 If the current content item doesn't have the requested value, use an alternative 'fallback' value in its place.
 
+Each of the examples below make use of an injected PublishedValueFallback. This is achieved by adding the following at the top of your Razor file:
+
+@inject IPublishedValueFallback PublishedValueFallback
```
sofietoft, line 235:
```suggestion
```csharp
@inject IPublishedValueFallback PublishedValueFallback
```
```

sofietoft (commented):
Great catch with the 11 version as well 💪 

Just one final suggestion, and I think we're golden ⭐

11/umbraco-cms/reference/querying/ipublishedcontent/properties.md
sofietoft: Not sure why it doesn't add the end tags for this codeblock element.
Might clash with the "suggestions" formats here on GitHub 😅

Matthew-Wise:
Accepted the suggestion then added the code end block :D

sofietoft:
Excellent - thanks @Matthew-Wise ! 💪 

Time to get this merged in then 🎉

PR #5121, opened 2023-04-18
Readability improvements - rendering scripts

erikjanwestendorp:
- [x] write shorter sentences
- [x] use for example instead of e.g.


sofietoft (commented):
Looks great @erikjanwestendorp ! Thanks for fixing these up.

I'll merge it right in!

PR #5124, opened 2023-04-19
umbraco-cloud/security: Tried to make it more clear what solution you need for your project

Ambertvu:
I ran into this this week. I was rather new to cloud and didn't know Cloudflare was used. I read the headlines and thought, oh, I dont have cloudflare, I need the second solution.

Oh boy, I was wrong :-)

After debugging, rechecking, I finally got notified on  the above part by Sebastiaan. I totally overlooked it.

I'm hoping it's a bit clearer for newer people like this.

### DIFF
--- a/umbraco-cloud/security.md
+++ b/umbraco-cloud/security.md
@@ -157,7 +157,7 @@ In the first rule we exclude the Umbraco Deploy endpoints, so that all deploymen
 
 **Projects from before December 8th, 2020 (Non-Reverse Proxy, eg. non-Cloudflare)**
 
-When your Cloud project is not using Cloudflare, your site needs to get the Remote IP address of the website visitor. In this case, you should use the second variation as shown below, when restricting access to your backoffice.
+When your Cloud project is not using Cloudflare, your site needs to use the Remote IP address of the website visitor. In this case, you should use the second variation as shown below, when restricting access to your backoffice.
 
 ```xml
 <rule name="Excluding Umbraco Deploy" enabled="true" stopProcessing="true">
--- a/umbraco-cloud/security.md
+++ b/umbraco-cloud/security.md
@@ -156,7 +156,11 @@ You can read more about the HTTP request headers coming from Cloudflare in the [
 {% tab title="Legacy (non-reverse proxy)" %}
 **Projects from before December 8th, 2020 (Non-Reverse Proxy, eg. non-Cloudflare)**
 
-When your Cloud project is not using Cloudflare, your site needs to use the Remote IP address of the website visitor. Use the example below to restrict access to your backoffice using IP filtering.
+If you created the Cloud project earlier than December 8th, 2020, Cloudflare is not used on your project.
+
+When your Cloud project is not using Cloudflare, your site needs to use the Remote IP address of the website visitor.
+
+Use the example below to restrict access to your backoffice using IP filtering:
 
 ```xml
 <rule name="Excluding Umbraco Deploy" enabled="true" stopProcessing="true">
--- a/umbraco-cloud/security.md
+++ b/umbraco-cloud/security.md
@@ -129,7 +129,7 @@ If you are unsure whether your Cloud project uses Cloudflare or not, get in touc
 {% tab title="Default (reverse proxy)" %}
 **Default setup (reverse proxy version, eg. Cloudflare)**
 
-For all cloud projects since December 8th, 2020, Cloudflare is used as a reverse proxy service for your project. This means that you'll need to get the IPs from the `X-Forwarded-For` header. Use the example below to restrict access to your backoffice using IP filtering.
+For all cloud projects since December 8th, 2020, Cloudflare is used as a reverse proxy service for your project. This means that you'll need to get the IPs from the `CF-Connecting-IP` header. Use the example below to restrict access to your backoffice using IP filtering.
 
 You can read more about the HTTP request headers coming from Cloudflare in the [Cloudflare Documentation.](https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/).
 
--- a/umbraco-cloud/security.md
+++ b/umbraco-cloud/security.md
@@ -129,7 +129,7 @@ If you are unsure whether your Cloud project uses Cloudflare or not, get in touc
 {% tab title="Default (reverse proxy)" %}
 **Reverse proxy version (default)**
 
-For all cloud projects since December 8th, 2020, Cloudflare is used as a reverse proxy service for your project. This means that you'll need to get the IPs from the `CF-Connecting-IP` header. Use the example below to restrict access to your backoffice using IP filtering.
+Cloudflare is used as a reverse proxy service for your Cloud project. This means that you'll need to get the IPs from the `CF-Connecting-IP` header. Use the example below to restrict access to your backoffice using IP filtering.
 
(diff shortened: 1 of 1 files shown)


umbraco-cloud/security.md
```diff
@@ -125,9 +125,9 @@ Since December 8th, 2020 all Umbraco Cloud sites uses Cloudflare for DNS, so new
 If you are unsure whether your Cloud project uses Cloudflare or not, get in touch with the friendly support team, and they will help you out.
 {% endhint %}
 
-**Reverse Proxy version (eg. Cloudflare)**
+**Default setup (reverse Proxy version, eg. Cloudflare)**
 
-When using Cloudflare, which is the default setup for all Cloud projects, the project will from behind a reverse proxy get the IPs from the `CF-Connecting-IP` header. In this case, which is most cases, use the first variation here to restrict access to your backoffice using IP filtering.
+For all cloud projects since December 8th, 2020, Cloudflare is used as a reverse proxy service for your project. This means that you'll need to get the IPs from the `X-Forwarded-For` header. Use the first variation here to restrict access to your backoffice using IP filtering.
```
sofietoft, line 130: "Use the first variation **here** to ..."

What is the "here" referring to?
I realize this is not something you've added specifically - but if it's referring to the code snippet a few lines below, should we ensure that the code snippet comes directly after this line? 🤔

sofietoft (commented):
Thanks a bunch for this PR @Ambertvu !
Sure seems like we should clean up the wording here!

I had a thought while reviewing your PR.
Would it make sense to use tabs for this section?
We could give it a heading like "Proxy Version" or something.
And then below it, two tabs: "Reverse Proxy (Cloudflare)" and "Non-Reverse Proxy".
Within each of these tabs, we could then make it clear for which projects they're relevant and add the code snippet that fits.

Might also be an option to use the "Before/After December 2020" creating date as the tab names - not sure what would be best here 🤔 

Any thoughts on a setup like this?
You can see some examples of tabs in this article: https://docs.umbraco.com/umbraco-cloud/getting-started/git-repository-in-a-cloud-project

Ambertvu:
Those tabs seem like a great idea for this as well, I'll put some work into it tonight/tomorrow to convert it in a nice way. Thanks for the suggestion!

sofietoft:
Sounds great @Ambertvu ! 😄 
Looking forward to see what you come up with 😎 

Doesn't look like GitBook has any good docs on how to work with tabs in the markdown files, but look at some other of our articles where we use them - like the one I linked you already, and we also have some here: https://docs.umbraco.com/umbraco-cms/v/10.latest-lts/extending/section-trees/searchable-trees

Ambertvu:
@sofietoft I've updated the files, I can't preview them locally (couldn't find a vscode plugin for it?)  but theoretically they should work.
Tried to make it a bit more clear which version should be used (Default vs Legacy)

umbraco-cloud/security.md
```diff
@@ -120,14 +120,16 @@ The following rule can be added to your web.config file in the `system.webServer
...
-Since December 8th, 2020 all Umbraco Cloud sites uses Cloudflare for DNS, so new and updated projects should use the Reverse Proxy version.
+Since December 8th, 2020 all Umbraco Cloud sites uses Cloudflare for DNS, so new and updated projects should use the reverse proxy version.
 
 If you are unsure whether your Cloud project uses Cloudflare or not, get in touch with the friendly support team, and they will help you out.
 {% endhint %}
 
-**Reverse Proxy version (eg. Cloudflare)**
+{% tabs %}
+{% tab title="Default (reverse proxy)" %}
+**Default setup (reverse proxy version, eg. Cloudflare)**
```
sofietoft, line 130:
```suggestion
**Reverse proxy version (default)**
```

umbraco-cloud/security.md
```diff
@@ -120,14 +120,16 @@ The following rule can be added to your web.config file in the `system.webServer
...
 
-When using Cloudflare, which is the default setup for all Cloud projects, the project will from behind a reverse proxy get the IPs from the `CF-Connecting-IP` header. In this case, which is most cases, use the first variation here to restrict access to your backoffice using IP filtering.
+For all cloud projects since December 8th, 2020, Cloudflare is used as a reverse proxy service for your project. This means that you'll need to get the IPs from the `CF-Connecting-IP` header. Use the example below to restrict access to your backoffice using IP filtering.
```
sofietoft, line 132: This date is already mentioned in the note above the tabs.
Perhaps we could rephrase this a bit:

"By default, Cloudflare is used as a reverse proxy service on all projects. This means [...]"

umbraco-cloud/security.md
```diff
@@ -150,14 +152,11 @@ You can read more about the HTTP request headers coming from Cloudflare in the [
   <action type="CustomResponse" statusCode="403" />
 </rule>
 ```
+{% endtab %}
+{% tab title="Legacy (non-reverse proxy)" %}
+**Projects from before December 8th, 2020 (Non-Reverse Proxy, eg. non-Cloudflare)**
```
sofietoft, line 157: Tried to shorten this heading as well. Not sure the "create date" part if necessary, as we could mention that in the in the tab instead 🤔 

```suggestion
{% endtab %}

{% tab title="Legacy (non-reverse proxy)" %}
**Non-Reverse Proxy (create date < 2021)**
```

umbraco-cloud/security.md
```diff
@@ -150,14 +152,11 @@ You can read more about the HTTP request headers coming from Cloudflare in the [
...
 
-{% hint style="info" %}
-In the first rule we exclude the Umbraco Deploy endpoints, so that all deployment and content transfers can still work.
-{% endhint %}
-
-**Non-Reverse Proxy (eg. non-Cloudflare)**
-
-When your Cloud project is not using Cloudflare, your site gets the Remote IP address of the website visitor. In this case, you should use the second variation as shown below, when restricting access to your backoffice.
+When your Cloud project is not using Cloudflare, your site needs to use the Remote IP address of the website visitor. Use the example below to restrict access to your backoffice using IP filtering.
```
sofietoft, line 159:
```suggestion
If you created the Cloud project earlier than December 8th, 2020, Cloudflare is not used on your project.

When your Cloud project is not using Cloudflare, your site needs to use the Remote IP address of the website visitor.

Use the example below to restrict access to your backoffice using IP filtering:
```

sofietoft (commented):
Looks great @Ambertvu ! 💪 

I've added a couple of comments and suggestions. Let me know what you think!
There are two reasons why we cannot preview this PR:
- We haven't enabled previews for PRs from external Forks (we are working on fixing that!)
- You are not part of the Umbraco org. on GitBook 🙈 (nothing to do about that, at the moment, I'm afraid)

umbraco-cloud/security.md
Ambertvu: Updated the PR :-) Changed it a little bit, as another by default felt a bit too excessive, hah.

umbraco-cloud/security.md
Ambertvu: Changed it a little bit as to improve readability

sofietoft:
Looks great @Ambertvu ! 💪 

Let's get this merged in!

sofietoft (approved):

sofietoft:
Thanks again for the PR @Ambertvu !

Looks like it checks out really nicely on the docs platform 👏 https://docs.umbraco.com/umbraco-cloud/security#restrict-backoffice-access-using-ip-filtering

Ambertvu:
Thank you too!

PR #5139, opened 2023-04-25
Updated session-timeout.md

Abdjulaziz:
This page Is kindof useless and needed more context. So if people search for the session timeout for user acitivty they can se how to adjust it to their prefrence. Also added the risk for having it too long 😄

### DIFF
--- a/11/umbraco-cms/tutorials/editors-manual/tips-and-tricks/session-timeout.md
+++ b/11/umbraco-cms/tutorials/editors-manual/tips-and-tricks/session-timeout.md
@@ -1,3 +1,18 @@
 # Session Timeout
 
-Umbraco is set up to automatically log a user out if they have been inactive for over 20 minutes. Don't worry if you are logged out of Umbraco. Log back in using your credentials and continue editing.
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @Abdjulaziz !

This is all great information, but I'm afraid it's not really the place for it.
The article you've updated is part of the "Editors Manual" tutorial, which is a series of articles targeting content editors - that's people who will not have access to the file, and no development experience.

We have it documented here: https://docs.umbraco.com/umbraco-cms/reference/configuration/globalsettings#timeout 
Perhaps we should instead expand a little on what is documented there?
What do yout think?

Abdjulaziz:
Hello @sofietoft 

Would be great to mention something about session timeout in the global settings.

sofietoft:
Excellent! Let's try to keep the docs for the editors, editor friendly and without too many technical details.

Are you up for making a PR with the additions to the settings article? 💪

Abdjulaziz:
Done pr: https://github.com/umbraco/UmbracoDocs/pull/5162

sofietoft:
Great, thanks @Abdjulaziz !

I'll close down this PR.

PR #5141, opened 2023-04-26
Fix link to Azure Web Apps in Load Balancing docs

callumbwhyte:
Does what it says on the tin. The link currently points to the file system replication docs.


sofietoft:
Good catch @callumbwhyte ! 💪 
Thanks for updating the links!

I'll merge it right in.

PR #5142, opened 2023-04-27
Fix CustomIndex sample

marcemarc:
The order that you do things in your ExamineComposer seems crucial...

basically, it seems you can't add the ConfigureOptions for the ProductIndex, before you have added the ProductIndex!

If we move that line to after the adding of the index then the 'Validator' rule in the example ConfigureOptions is applied.

If it's before we get the default Options (maybe calling AddExamineLuceneIndex resets the options?)

### DIFF
--- a/11/umbraco-cms/reference/searching/examine/indexing.md
+++ b/11/umbraco-cms/reference/searching/examine/indexing.md
@@ -340,9 +340,9 @@ namespace Umbraco.Docs.Samples.Web.CustomIndexing
     {
         public void Compose(IUmbracoBuilder builder)
         {
-            builder.Services.ConfigureOptions<ConfigureProductIndexOptions>();
-
             builder.Services.AddExamineLuceneIndex<ProductIndex, ConfigurationEnabledDirectoryFactory>("ProductIndex");
+                    
+            builder.Services.ConfigureOptions<ConfigureProductIndexOptions>();
 
             builder.Services.AddSingleton<ProductIndexValueSetBuilder>();
 


sofietoft:
This seems reasonable to me! 💪 Thanks for the PR, @marcemarc !

Do you need a second opinion from the HQ D-team? I see there's a question in there, and I do think we need some additional involvement to get an answer 😅

marcemarc:
@sofietoft I think, with this one, the order is definitely important, the question mark surrounds me nailing 100% why, but with the order in the docs, it doesn't work, so at the moment, it's broken - but there are other issues with the samples, and I've created another PR to the docs here:

https://github.com/umbraco/UmbracoDocs/pull/5144

which at least makes the samples work, but has an overarching 'what are we trying to do here 'ark'' - because if the aim is to create a custom index of Umbraco Content, there are easier ways to do it, and I've created a PR to the core to potentially make that easier: https://github.com/umbraco/Umbraco-CMS/pull/14177 (but we may find out that is undesired). If the sample is mainly to show off how you can populate a custom index, then I reckon we should have a better example that doesn't use Umbraco as the source to populate it...

But at the moment, the samples won't work without these changes...

sofietoft:
Thanks for the additional details @marcemarc !

So both of the Docs PRs are related to the same code sample, right?
Just wondering if they should be merged together instead, then, as it does sound like the sample won't work without both being merged in?

I've passed the other PR along to the Dteam.
Might be able to find some time to run some tests with this next week 🤞

nikolajlauridsen (approved):
Seems like a reasonable change that makes sense to me 😄

alina-tincas:
Thank you for confirming @nikolajlauridsen and thank you for the PR @marcemarc 💪 

Am merging this in and will make sure it is updated in v10 and v12 and well 👍

PR #5145, opened 2023-04-28
Added link for Gitlab on how to setup it up

alina-tincas:
On the  "A source control repository with your own code"


sofietoft:
Thanks for the PR @alina-tincas ! 💪 

Great idea with a link to this article.
Let's put it into a note below this line instead. Then we could also add some more context to the link, like "here's an example of how is can be set up using GitLab..."

What do you think?

alina-tincas:
That sounds good 😊

sofietoft:
@alina-tincas - I've added the link to a note.
Let me know what you think!

alina-tincas:
It looks really good 🙌

PR #5148, opened 2023-05-02
Spelling mistake / misphrased?

HamDerAndrew:
Was browsing the docs and found this.
It's just a suggestion as to what it could say instead 😁

### DIFF
--- a/new-backoffice/README.md
+++ b/new-backoffice/README.md
@@ -8,7 +8,7 @@ description: >-
 
 Version 13 of Umbraco CMS will include a new backoffice. It will look more or less the same but be built on modern technology using a UI Library built for the purpose.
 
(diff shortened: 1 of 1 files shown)


new-backoffice/README.md
```diff
@@ -8,7 +8,7 @@ description: >-
 
 Version 13 of Umbraco CMS will include a new backoffice. It will look more or less the same but be built on modern technology using a UI Library built for the purpose.
```
sofietoft, line 9:
```suggestion
Version 14 of Umbraco CMS will include a new backoffice. It will look more or less the same but be built on modern technology using a UI Library built for the purpose.
```

sofietoft (commented):
Good catch @HamDerAndrew ! 💪 
Looks like we should also update the target version number!

HamDerAndrew:
Yeah good point @sofietoft 😀

PR #5151, opened 2023-05-02
Incorrect information

hnky:
After running the command, and running Umbraco by default is sql lite selected.


sofietoft (approved):
Good catch @hnky ! 💪 
Sure looks like this was not properly updated when we switched to SQL lite.

I'll merge it right in!

PR #5153, opened 2023-05-03
Update cache example with an example of using the cache

jemayn:
It's not super clear on this page how you need to store and retrieve values from the cache, so I extended the example a little bit with a usecase 🙂


sofietoft:
Looks great @jemayn ! Thanks!

Would it make sense to mention in a comment in the code, that this is a sample usecase for retrieving values from the cache? 🤔 
Perhaps around line 32, just above the method?

sofietoft:
I like it! Thanks again @jemayn ! 💪

PR #5157, opened 2023-05-04
Add instructions for enabling Azure Blob Storage ImageSharp cache

OwainJ:
With v11.0.0 of the Umbraco.StorageProviders, the ImageSharp cache was split out into a separate package, which means it now needs to be installed and added to the startup file separately.

I have therefore updated the documentation to include instructions for the new way of enabling the ImageSharp Azure Blob storage cache using the `Umbraco.StorageProviders.AzureBlob.ImageSharp` package.

Updates were based on the instructions found here: https://github.com/umbraco/Umbraco.StorageProviders#umbracostorageprovidersazureblobimagesharp

### DIFF
--- a/11/umbraco-cms/extending/filesystemproviders/azure-blob-storage.md
+++ b/11/umbraco-cms/extending/filesystemproviders/azure-blob-storage.md
@@ -86,7 +86,7 @@ Invoke the `.AddAzureBlobMediaFileSystem()` and the `.AddAzureBlobImageSharpCach
 
 As of [version 11.0.0](https://github.com/umbraco/Umbraco.StorageProviders/releases/tag/release-11.0.0) of `Umbraco.StorageProviders`, the ImageSharp dependency has been separated into its own package.
 
-Therefore, if you're planning on upgrading your site from Umbraco 9/10 to 11+, then don't forget to install and setup the new `Umbraco.StorageProviders.AzureBlob.ImageSharp` package; this will ensure that your ImageSharp cache continues to be stored in your blob storage container.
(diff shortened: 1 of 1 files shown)


11/umbraco-cms/extending/filesystemproviders/azure-blob-storage.md
```diff
@@ -96,9 +98,8 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         .WithMiddleware(u =>
```
OwainJ, line 98: Also, is this hint about Umbraco 9 still needed for the v11 version of the docs?

Should it be replaced with a hint explaining the difference between the v10.0.0 and v11.0.0 versions of `Umbraco.StorageProviders.AzureBlob`? i.e. the fact that ImageSharp cache support has been split out.
As someone upgrading their site, from Umbraco 10 to 11, might be confused as to why their ImageSharp cache suddenly stops uploading to their blob storage.

sofietoft:
Thanks for the PR @OwainJ ! 💪 Much appreciated.

I'll send this to our developers, so we can get everything verified, and also get their thoughts on the note.

I do think though, that the note can be removed.
If it should be replaced with mentioning something that's different between 10 and 11 - I'm not sure. You can already switch the major version for the docs, and get the correct docs for 10. Perhaps a note about how to use the dropdown, to see differences in v10? 🤔

11/umbraco-cms/extending/filesystemproviders/azure-blob-storage.md
Zeegaan: Yes I think we can remove the hint about being in v9, (as this is v10 documentation now), and instead replace it with what you're suggesting 💪 Are you up for that ? 🤞

OwainJ: @Zeegaan I've replaced the hint, let me know if it's any good 😄

Zeegaan (approved):
LGTM 👍

ronaldbarendse (approved):

PR #5160, opened 2023-05-09
Added IPv6 info to the docs

FransdeJong:
I added information about IPv6 because more and more clients are asking for IPv6 support and most DNS providers don't support setting a CNAME on a root domain.

I discussed this with Ronald Barednse and I'm now setting up the sites like this:

webwonders.nl A        104.19.191.28
webwonders.nl AAAA 2606:4700::6813:bf1c
www.webwonders.nl CNAME dns.umbraco.io

This works and enables clients to get full IPv6 support when CNAMEs on root domains aren't allowed.

Keep in mind that if the IPv6 addresses are changed Umbraco Cloud clients must be informed but that process is also needed for the IPv4 addresses.

### DIFF
--- a/umbraco-cloud/set-up/project-settings/manage-hostnames/README.md
+++ b/umbraco-cloud/set-up/project-settings/manage-hostnames/README.md
@@ -27,10 +27,11 @@ You can bind any hostname to your project environments. Ensure that the hostname
 
 Once you add a hostname to one of your environments, ensure to update the hostname DNS entry to resolve to the `umbraco.io` service.
 
-We recommend:
+We recommend setting a CNAME record for your hostname using **dns.umbraco.io**.
+In cases where providers don't offer this option on the root domain you can add the CNAME for the desired subdomain(s) and create the following records for the root domain:
 
-* Setting a CNAME record for your hostname using **dns.umbraco.io** or
-* An A record using an IP, either `104.19.191.28` or `104.19.208.28`.
+* An A record using an IP, either `104.19.191.28` or `104.19.208.28`. 
+* If you need IPv6 support add an AAAA record using one of the following addresses, `2606:4700::6813:bf1c` or `2606:4700::6813:d01c`.
 
 {% hint style="info" %}
 Once you have updated your DNS, we recommend that you check if the correct records are being picked up using a site like [whatsmydns.net](https://www.whatsmydns.net/) before adding the hostname on Umbraco Cloud.


umbraco-cloud/set-up/project-settings/manage-hostnames/README.md
```diff
@@ -27,10 +27,11 @@ You can bind any hostname to your project environments. Ensure that the hostname
 
 Once you add a hostname to one of your environments, ensure to update the hostname DNS entry to resolve to the `umbraco.io` service.
 
-We recommend:
+We recommend setting a CNAME record for your hostname using **dns.umbraco.io**.
+In cases where providers don't offer this option on the root domain you can add the CNAME for the desired subdomain(s) and create the following records for the root domain:
```
ronaldbarendse, line 32:
```suggestion
We recommend setting a CNAME record for your hostname using **dns.umbraco.io**. Although a CNAME cannot be placed at the root domain level, some providers support CNAME flattening to overcome this limitation.

However, if your provider doesn't support setting a CNAME record for your hostname, you can manually set the following records:
```

umbraco-cloud/set-up/project-settings/manage-hostnames/README.md
```diff
@@ -27,10 +27,11 @@ You can bind any hostname to your project environments. Ensure that the hostname
...
 
-* Setting a CNAME record for your hostname using **dns.umbraco.io** or
-* An A record using an IP, either `104.19.191.28` or `104.19.208.28`.
+* An A record using an IP, either `104.19.191.28` or `104.19.208.28`. 
+* If you need IPv6 support add an AAAA record using one of the following addresses, `2606:4700::6813:bf1c` or `2606:4700::6813:d01c`.
```
ronaldbarendse, line 34:
```suggestion
* A records to either or both IPv4 addresses:
  * `104.19.191.28`
  * `104.19.208.28`
* Optionally, AAAA records to either or both IPv6 addresses (to support IPv6 connectivity):
  * `2606:4700::6813:bf1c`
  * `2606:4700::6813:d01c`
```

ronaldbarendse (commented):
I've added suggestions to make it more clear that using CNAME flattening could be an option (if your provider supports it) and otherwise, you can set either **or both** IPv4 and IPv6 addresses.

It's not a strict requirement to configure both IPv4 addresses and neither is adding the IPv6 addresses, but that will help to distribute the traffic between our available endpoints.

We should probably also update the warning below this text, because `The static IP is highly volatile towards changes to the Umbraco Cloud infrastructure and as such, it may change.` seems like a contradiction in terms (static != highly volatile). Maybe there was some confusion with the outbound IP addresses of Cloud, which can indeed change if you don't use the Static Outbound IP feature.

umbraco-cloud/set-up/project-settings/manage-hostnames/README.md
```diff
@@ -27,10 +27,15 @@ You can bind any hostname to your project environments. Ensure that the hostname
 
 Once you add a hostname to one of your environments, ensure to update the hostname DNS entry to resolve to the `umbraco.io` service.
 
-We recommend:
-
-* Setting a CNAME record for your hostname using **dns.umbraco.io** or
-* An A record using an IP, either `104.19.191.28` or `104.19.208.28`.
+We recommend setting a CNAME record for your hostname using **dns.umbraco.io**. Although a CNAME cannot be placed at the root domain level, some providers support CNAME flattening to overcome this limitation.
+
+However, if your provider doesn't support setting a CNAME record for your hostname, you can manually set the following records:
```
jonat123, line 32:
```suggestion
If your provider doesn't support setting a CNAME record for your hostname, you can manually set the following records:
```

FransdeJong:
I accepted the change. Thanks for the input.

Issue #5150, opened 2023-05-02
Deprecated/Oboslete example in EmbedProviders docs


daanbakker1995:
## What type of issue is it? (Choose one - delete the others)
Wrong documentation

## What article/section is this about?
[embedded-media-providers snippet example](https://docs.umbraco.com/umbraco-cms/extending/embedded-media-providers#register-the-provider-with-the-oembedproviderscollection)

## Describe the issue

In this snippet the `OEmbedProviders` collection builder is mentioned to compose the embed providers. When using this in umbraco 10 the obsolete attribute is present. That says use the `EmbedProviders` collection instead. 

Just wanted to  mention this so the documentation could be updated. 

Greetings,
Daan

sofietoft:
Thanks for reporting this @daanbakker1995 ! 💪 

I can see the the OEmbedProviders are mentioned multiple places in the article - not just the sample you're referring to 🙈 
Sounds like we need to get this entire article verified against the latest version!

I'll make sure we get some developers to look into this.

daanbakker1995:
Thank you!

PR #5168, opened 2023-05-12
Update v10 & v11 articles to not use Obsolete EmbedProviderBase

Zeegaan:
Fixes https://github.com/umbraco/UmbracoDocs/issues/5150

# Notes
- Updates both v10 & v11 articles to no longer reference obsolete `EmbedProviderBase`
- Also no longer reference IUserComposer, instead refer to `IComposer`


sofietoft:
Looks great @Zeegaan ! 💪 

Thanks for looking into this.
I'll merge it right away!

PR #5171, opened 2023-05-16
Update getting-started-with-entity-framework-core.md

Zeegaan:
Fix up code-styling for EF core documentation, line was very long, so it caused weird side-scrolling


jonat123 (approved):
@Zeegaan looks good :D 

Merging it in.

PR #5174, opened 2023-05-17
V12: Fix EF Core tutorial code snippets

nikolajlauridsen:
While going through the tutorial I noticed that the included snippets didn't work, so I've made some changes here to remedy that. 

* Changed the migration snippet to not run in a scope, running ef core migrations in a scope is not allowed
	* I also changed the handler to be async so we can await the migrations
* Changed the controller snippet 
	* Refactored the namespace to be consistent with other examples
	* Updated `GetComments` to properly take in `Guid umbracoNodeKey`, since `BlockPostUmbracoKey` is a guid
	* Most importantly, I added `.ToArray();` to the `GetComments` endpoint, this is actually quite important in order to force the Linq query to actually execute, otherwise it won't happen til after the scope is completed and the DB context is disposed, leading to an error

### DIFF
--- a/12/umbraco-cms/tutorials/getting-started-with-entity-framework-core.md
+++ b/12/umbraco-cms/tutorials/getting-started-with-entity-framework-core.md
@@ -209,50 +209,51 @@ using Microsoft.AspNetCore.Mvc;
 using Umbraco.Cms.Persistence.EFCore.Scoping;
 using Umbraco.Cms.Web.Common.Controllers;
 
-namespace Umbraco.Demo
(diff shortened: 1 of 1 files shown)


sofietoft:
Looks good to me @nikolajlauridsen ! 💪 

Do you need @Zeegaan to look at it as well, before merge?

sofietoft (approved):

nikolajlauridsen:
He gave an over the shoulder review yesterday, so should be fine 😄

nikolajlauridsen:
Awesome, thank you 🎉

PR #5175, opened 2023-05-17
Update forgotten property in EFCore docs

Zeegaan:
Last change I swear! 🤣 

While going through the docs one last time, I found this small error 😁


sofietoft:
I'll hold you to that promise then 😉

Merging.

PR #5183, opened 2023-05-19
Add negation of preview Url for rewrite rules

marcemarc:
Currently if you are following the documentation on Umbraco Cloud to redirect the native backoffice io url to your front end url, you will encounter a problem!

You can't preview a page that hasn't been published yet, which might sound quite edge case, but actually most people want to see a preview before they publish something for a first time!

The reason for this is the suggested rewrite rules in this article, show how to put the redirect from the cloud backoffice url to the front end url, and has a list of patterns to negate, eg the Umbraco folder!

But what it doesn't consider is how 'preview' works

when you preview a page in the backoffice it appears to be loading a url from within the Umbraco Folder eg
/umbraco/preview/?id=3195
and therefore it doesn't get redirected

but underneath the hood it uses an iframe to load content from a url in the format
/3195 
where 3195 is the id of the page (pre v9, this would have been /3195.aspx)

and this request is not included in the list of negations, and so the 'redirect' to the front end domain kicks in for the iframe - and returns a 404, cos the page isn't published, of if page is published, returns a page that doesn't include the changes for preview!

Therefore I'm suggesting in your advice to include a negation for /anynumber pattern, and then preview will work again when you have this rule in place.

If you are enforcing a trailing slash in a different rule, you'd need to add a similar negation!

### DIFF
--- a/umbraco-cloud/set-up/project-settings/manage-hostnames/rewrites-on-cloud.md
+++ b/umbraco-cloud/set-up/project-settings/manage-hostnames/rewrites-on-cloud.md
@@ -40,6 +40,7 @@ One approach for this is to add a new rewrite rule to the `<system.webServer><re
     <add input="{REQUEST_URI}" negate="true" pattern="^/DependencyHandler.axd" />
     <add input="{REQUEST_URI}" negate="true" pattern="^/App_Plugins" />
     <add input="{REQUEST_URI}" negate="true" pattern="localhost" />
+    <add input="{REQUEST_URI}" negate="true" pattern="^\/([0-9]*)$" />
   </conditions>
   <action type="Redirect" url="http://<your actual domain here>.com/{R:0}"
         appendQueryString="true" redirectType="Permanent" />


jonat123:
Hey @marcemarc I've tried testing it out with the rewrite rule and using the preview.

It seems to me that it works fine, adding the rewrite, creating a new page with some content, and then saving it, without publishing it, then I preview the page and it loads fine, no 404 or anything 🤔

I am wondering if I am missing something?

marcemarc:
Hi @jonat123 that's super weird...

So if you visit the preview url directly in the browser (and not in the iframe) with the redirect rules from the docs in place...

Does it redirect you to the front end domain?
or keep you on the backoffice domain?

regards

Marc

marcemarc:
Hey @jonat123  - Does your backend and frontend urls have the same domain? (thinking when you have a completely different domain, the preview cookie isn't passed?

jonat123:
@marcemarc I have two domains, the default cloud domain and one custom domain, with which I used the rewrite with:
image

Let me know if that is not way to do it :-)

marcemarc:
@jonat123 I am intrigued now!! maybe I'm not doing this right!

Soooo

If you visit the backoffice on

https://docs-team-cloud-project.euwest01.umbraco.io/umbraco

and login (and how you login might be key here, if you are using an account created for this Umbraco instance or whether you are using an Umbraco.IO account)

if you then create, but don't publish, a brand new page

If you press 'save and preview' then you stay on the backoffice url and have the preview url in the browser window:

https://docs-team-cloud-project.euwest01.umbraco.io/umbraco/preview/?id=1234

and this Iframes in a page using a url from within the Umbraco folder and so is ok:

https://docs-team-cloud-project.euwest01.umbraco.io/umbraco/preview/frame?id=1234

but if in the footer of the preview mode you click the 'Preview Website' button then it takes you out of the iframe and to a Url in the format

https://docs-team-cloud-project.euwest01.umbraco.io/1234

and it sets a cookie, UMB_PREVIEW for the subdomain docs-team-cloud-prouect.euwest01.umbraco.io

which allows you to preview the content, because the cookie is present...

... however without the negation - do you not find that you are redirected to your front end url? 

https://swat-umbraco.com/1234

and here, because the UMB_PREVIEW cookie isn't set for that domain - do you find that because the page isn't published and because you are not in preview mode on the swat-umbraco.com domain, that it errors?

or do you get the previewed content of the new page via the swat-umbraco.com domain? and if so, is it setting an UMB_PREVIEW cookie for that domain? Just thinking if you are logged in via Umbraco.IO account using go to backoffice from the portal, that makes it work for you.

jonat123:
@marcemarc Thanks for the detailed steps!

I've tried to reproduce it following your steps, and I have still not been able to reproduce the issue.

I don't think that which user I am using should matter, as all users on Umbraco Cloud are centralized through Umbraco ID 🤔

As I can't reproduce the issue I will close this PR.

If you in the future think that this is still an issue, do not hesitate to re-open this PR or create an issue on our tracker, then we will be more than happy to take a look at it again.

I hope you have a great day and summer 😊

marcemarc:
Thanks @jonat123, happy to close, doesn't sounds like 'it is a thing' - would be interesting to know when you followed the steps how your auth cookie for umbraco.io worked on umbraco.com! - but ultimately not an issue for us now we have the negation in place, apologies for all the rigmarole - enjoy the summer break!

PR #5188, opened 2023-05-22
Update major-upgrades.md

alina-tincas:
seems most people miss this step so maybe it needs more elaborating that they need to be latest CMS version of the major and minor version they are on currently


jonat123 (approved):
@alina-tincas Looks good to me :) 

Merging in the PR.

PR #5190, opened 2023-05-23
Heartcore - Added Multinode Treepicker Information

cornehoskam:
Hi,

One of the things I was missing in the Heartcore Content Management documentation is the expected structure when working with Multinode Treepickers. After some trial and error I was able to figure out what it was expecting, and thought it would be a worthy addition to the docs!
Feel free to change the wording/examples as necessary! 😊

Kind regards,
Corné


sofietoft:
Thanks for the PR @cornehoskam ! 💪 

I'll make sure we get this reviewed as soon as possible.

jonat123:
@cornehoskam Looks good, one thing I was thinking, would it maybe make sense to put your sentence in a hint? 

As it is only relevant if one is using the multi-node picker.

cornehoskam:
Thank you for the explanation @jonat123! I've pushed another commit adding the changes as requested! 🙂

jonat123 (approved):
Awesome @cornehoskam 😄 Looks good! 

merging it in right away!

PR #5195, opened 2023-05-24
Update current examine documentation

Zeegaan:
# Notes
- The current examine documentation was a bit confusing, so this PR should remedy that
- Examine index should no longer contain the content fields, but instead only the 3 mandatory ones, and the ones we add
- The indexing code should now no longer show bad practice, and should show a better example in indexing all content


11/umbraco-cms/reference/searching/examine/indexing.md
```diff
@@ -305,29 +304,43 @@ namespace Umbraco.Docs.Samples.Web.CustomIndexing
...
+                    while (descendants.Length == pageSize);
                 }
             }
-
         }
     }
 }
 ```
+{% hint style="info" %}
+Note here that this is just an example of how you could do indexing, this means that we're indexing all content, also unpublished & trashed content.
```
kjac, line 336:
```suggestion
Note that this is just an example of how you could do indexing. In this sample we're indexing all content, also unpublished content.
```

Since we're only fetching descendants of content roots, we can't ever be indexing trashed content here?

11/umbraco-cms/reference/searching/examine/indexing.md
```diff
@@ -305,29 +304,43 @@ namespace Umbraco.Docs.Samples.Web.CustomIndexing
...
+If you want to filter those out, you'd have to make your own logic for that.
```
kjac, line 337:
```suggestion
If you want to filter those out, you will have to create your own logic for that.
```

11/umbraco-cms/reference/searching/examine/indexing.md
```diff
@@ -305,29 +304,43 @@ namespace Umbraco.Docs.Samples.Web.CustomIndexing
...
+You could as an example look at published status in the `ProductIndexValueSetBuilder.GetValueSets()` method and filter it there.
```
kjac, line 338: Not entirely correct; the published status can change for sub pages, leaving potentially unroutable published content in the index. Probably better to link to the `IndexPublishedContent` method of `ContentIndexPopulator`, which performs a structural filtering of published content below unpublished.

11/umbraco-cms/reference/searching/examine/indexing.md
```diff
@@ -352,8 +365,15 @@ namespace Umbraco.Docs.Samples.Web.CustomIndexing
 }
 ```
 
+{% hint style="info" %}
+The order of registration may seem arbitrary, but it's important so register your index, before `ConfigureOptions`.
```
kjac, line 369:
```suggestion
The order of these registration matter. It is important to register your index with `AddExamineLuceneIndex` before calling `ConfigureOptions`.
```

11/umbraco-cms/reference/searching/examine/indexing.md
Zeegaan: I have tried rewriting this 👍

kjac (approved):
Looks great to me 👍

10/umbraco-cms/reference/searching/examine/indexing.md
```diff
@@ -304,25 +293,42 @@ namespace Umbraco.Docs.Samples.Web.CustomIndexing
...
+                    while (descendants.Length == pageSize);
                 }
             }
-
         }
     }
 }
 ```
+{% hint style="info" %}
+Note that this is only an example of how you could do indexing. In this example we're indexing all content, both published and unpublished.
```
sofietoft, line 325:
```suggestion
This is only an example of how you could do indexing. In this example, we're indexing all content, both published and unpublished.
```

sofietoft (commented):
I've added some comments and suggestions - let me know if something doesn't make sense!

Once they have been resolved, I'll take care of adding the changes to the v11 article as well 💪

PR #5199, opened 2023-05-26
Update Deploy 12 API key recommendation

ronaldbarendse:
Deploy 12 adds support for API keys that are Base64-encoded, allowing for more secure keys (since the full byte range of 0-255 can be encoded, resulting in greater variation) and will recommend setting this to a larger/longer value.


12/umbraco-deploy/getting-started/install-configure.md
```diff
@@ -110,17 +110,33 @@ In order to deploy content based on certain rich core and community property edi
 
 When Umbraco Deploy has been installed, to be able to use it in the project you will need to create and add configuration for an API key.
 
-The API key should be a **randomly generated string of 64 characters (recommended)**. The minimum requirement is 10 characters.
+The recommended API key value is a **Base64-encoded string of 64 randomly generated bytes** (which will result in a value of 88 characters), although to maintain backwards compatibility the minimum requirement is 10 characters.
```
sofietoft, line 113:
```suggestion
The recommended API key value is a **Base64-encoded string of 64 randomly generated bytes**, resulting in a value of 88 characters. In order to maintain backward compatibility the minimum requirement is 10 characters.
```

sofietoft (commented):
Just a tiny suggestions 🤞

sofietoft:
Thanks for considering the suggestion @ronaldbarendse !

Just one more thing - it doesn't look like the article you're editing here, is added to the SUMMARY file for the 12 Deploy site yet.
Would you mind adding that?

You should add the following around line 7:

```

## Getting Started

* [Setting up Umbraco Deploy](getting-started/install-configure.md)
* 
```

ronaldbarendse:
@sofietoft I've added the page to the summary 👍🏻

I did notice that we haven't yet created all pages for v12 and there's at least one other reference to the API key and length in the current version: https://docs.umbraco.com/umbraco-deploy/deploy-settings#apikey. What's the process for making these available and getting them ready for v12?

sofietoft:
Thanks!

Yeah, so the idea was to not add all documentation to the 12 sites until the final release. So, it would be easier to find the material that has been updated/written new for the version.

To add an article from the 11 version:
- Make a copy of the file
- Place in the same location in the 12 directory
- Make sure it's referenced in the SUMMARY file as well

Hope it makes sense! Otherwise let me know 😄

12/umbraco-deploy/SUMMARY.md
```diff
@@ -26,6 +26,10 @@
   * [Partial Restores](deployment-workflow/restoring-content/partial-restore.md)
 * [Deploy Dashboard](deployment-workflow/deploy-dashboard.md)
 
+## Getting Started
+
+* [Setting up Umbraco Deploy](getting-started/install-configure.md)
+
```
jonat123, line 32: Any reason to add this section again? 🙈

jonat123, line 32:
```suggestion
```

jonat123:
@ronaldbarendse Looks good!

I've made a small suggestion to remove the duplicate section in the SUMMARY file, otherwise everything else looks good 😄

ronaldbarendse:
@jonat123 This shouldn't yet be merged, since we introduced a new `ApiSecret` setting in the latest RC (and now final v12) release... I updated the current changes to account for that, but noticed the API key was mentioned on other pages as well, so I'm still working on updating those!

ronaldbarendse:
This is now ready to review/merge 👍🏻

I've updated the Deploy 12 pages to use/recommend the newly added `ApiSecret` setting, added and improved the examples to generate the API secret/key (using collapsible sections for the similar, but slightly different ways of doing this). The v10/v11 pages are also updated to use the same example for generating the API key...

12/umbraco-deploy/deploy-settings.md
```diff
@@ -69,9 +70,9 @@ For illustration purposes, the following structure represents the full set of op
 }
 ```
 
-## ApiKey
+## ApiKey or ApiSecret
 
-The API key is a 10 character random string applied with the same value to all environments in order to authenticate HTTP requests between them.
+The `ApiKey` is a random string (of at least 10 characters) that is set to the same value on all environments in order to authenticate HTTP requests between them. For improved security, set the `ApiSecret` to a cryptographically random value of 64 bytes instead (using Base64-encoding).
```
jonat123, line 75:
```suggestion
The `ApiKey` is a random string (of at least 10 characters) set to the same value on all environments to authenticate HTTP requests between them. For improved security, set the `ApiSecret` to a cryptographically random value of 64 bytes instead (using Base64-encoding).
```

12/umbraco-deploy/getting-started/install-configure.md
```diff
@@ -105,22 +105,91 @@ To be able to use Umbraco Forms with Umbraco Deploy, you need to install the `Um
...
+In order to deploy content based on certain rich core and community property editors - including Nested Content, Multi URL Picker and Block List/Grid Editor - there is one further NuGet package to install: `Umbraco.Deploy.Contrib`.
 {% endhint %}
 
-When Umbraco Deploy has been installed, to be able to use it in the project you will need to create and add configuration for an API key.
+When Umbraco Deploy has been installed, to be able to use it in the project you will need to create and add configuration for an API key/secret.
 
-The API key should be a **randomly generated string of 64 characters (recommended)**. The minimum requirement is 10 characters.
+For improved security, it is recommended to set the `ApiSecret` (instead of the `ApiKey`) setting to a **cryptographically random value of 64 bytes** (using Base64-encoding to get the string representation, resulting in a value of 88 characters). For versions prior to Deploy 12 or when not using the API secret setting, the recommendation is to set the `ApiKey` to a randomly generated string of 64 characters.
```
jonat123, line 113:
```suggestion
For improved security, it is recommended to set the `ApiSecret` (instead of the `ApiKey`) setting to a **cryptographically random value of 64 bytes**. Using Base64-encoding to get the string representation, will result in a value of 88 characters. For versions prior to Deploy 12 or when not using the API secret setting, the recommendation is to set the `ApiKey` to a randomly generated string of 64 characters.
```

12/umbraco-deploy/getting-started/install-configure.md
```diff
@@ -105,22 +105,91 @@ To be able to use Umbraco Forms with Umbraco Deploy, you need to install the `Um
 {% endhint %}
 
 {% hint style="info" %}
-In order to deploy content based on certain rich core and community property editors - including Nested Content, Multi URL Picker and Block List Editor - there is one further NuGet package to install: `Umbraco.Deploy.Contrib`.
+In order to deploy content based on certain rich core and community property editors - including Nested Content, Multi URL Picker and Block List/Grid Editor - there is one further NuGet package to install: `Umbraco.Deploy.Contrib`.
 {% endhint %}
 
-When Umbraco Deploy has been installed, to be able to use it in the project you will need to create and add configuration for an API key.
+When Umbraco Deploy has been installed, to be able to use it in the project you will need to create and add configuration for an API key/secret.
```
jonat123, line 111:
```suggestion
With Umbraco Deploy installed, to use it in the project you will need to create and add configuration for an API key/secret.
```

jonat123 (changes requested):
@ronaldbarendse Looks good :) 

I noticed some sentences were a bit too long, so I have made some suggestions to shorten them.

Please have a look at them and see if they make sense :-)

jonat123 (approved):

PR #5217, opened 2023-06-08
Update manage-cdn-caching.md

rammi987:
## Description

Missing in documentation that anything under the reserved path /umbraco* is not part of the CDN cache do to the server response incl. cache-control: no-cache reponse header parameter

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ x] Other

### DIFF
--- a/umbraco-cloud/set-up/project-settings/manage-cdn-caching.md
+++ b/umbraco-cloud/set-up/project-settings/manage-cdn-caching.md
@@ -44,6 +44,10 @@ The following example adds a cache-control header with 'no-cache' as the value w
 The webpage itself is not cached when CDN Caching is enabled.
 {% endhint %}
 
+{% hint style="info" %}
+Everything under the path /umbraco, is not being cached
+{% endhint %}
+
 ## Cache Everything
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @rammi987 ! 💪 

I'll make sure we get this reviewed as soon as possible.

jonat123:
@rammi987 Just want to give you an update on this one :-) 

I have talked with some of the developers in the Cloud team and they say that this is in the CMS itself where the  /umbraco is not part of the cache.

Also, they are saying that there might also be more paths that are not being cached in the CMS, which we then would also need to list, which we are not sure that we want to, as it might potentially end up being quite long and harder to maintain.

In any case, our cloud team is taking it up with the CMS team, and when i hear back from them, i will of course let you know 😊

jonat123:
@rammi987 I hope you have had a great summer 😊

I have just talked with our developers, and for now, we have decided that it is not something that we want to add to the documentation, as it is not cloud-specific, but rather CMS-specific.


This means that I will unfortunately close this PR down for now.

I hope you have a good day and the rest of the summer 🌞

PR #5222, opened 2023-06-15
Member Registration and Login Doc fix

netcamo:
Solves error "Type or namespace "UmbAlternativeRegirsterController" could not be found" .

## Description

Error "Type or namespace "UmbAlternativeRegirsterController" could not be found" is shown when following the doc. To solve that problem I have updated last code snippet that adds custom register controller.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

### DIFF
--- a/11/umbraco-cms/tutorials/members-registration-and-login.md
+++ b/11/umbraco-cms/tutorials/members-registration-and-login.md
@@ -323,8 +323,8 @@ using (Html.BeginUmbracoForm<UmbRegisterController>(
 we have to instead use the custom controller we added, as well as include an anti-forgery token:
 
 ```csharp
-    using (Html.BeginUmbracoForm<UmbAlternativeRegisterController>(
-            "HandleRegisterMember",
+    using (Html.BeginUmbracoForm(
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @netcamo ! 💪 

I'll make sure we take a look at this as soon as possible!

jonat123:
@netcamo Thanks for creating this PR 😄

I've just taking a look at your PR and our documentation, and as it states: _"Since the member saving form is processed in a controller, we can copy the default UmbRegisterController and add a function to assign the newly created member to a group."_ Meaning that in the tutorial, we copy the content from the [UmbRegisterController](https://github.com/umbraco/Umbraco-CMS/blob/contrib/src/Umbraco.Web.Website/Controllers/UmbRegisterController.cs) and create our own controller in your project, which you can call what you want, we decided to call it `UmbAlternativeRegisterController`.

However, if you called it something else and try to register it in the partial view with the name from the tutorial it will give you the error that you see, as the names do not match.

We do also mention that in the note under the last code examples 😄

I hope this makes sense 😊

I will close the PR for now.

Have a great summer 😄

PR #5236, opened 2023-06-21
Added docs for API versioning and OpenAPI

kjac:
## Description

As V12 ships with Swagger and applies both OpenAPI and API versioning configuration in a global scope, we need to document how site implementers can work around these global configurations to achieve an API versioning and/or OpenAPI (Swagger) setup that fits their liking.

We're assigning a D-team dev to review the code samples of this new article ASAP.

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

Umbraco CMS v12.

## Deadline (if relevant)

As soon as possible. This applies to the 12 RC, and although we're not explicitly asking for people to test it, the information might still be useful for people performing early upgrades.

At the latest, this should be included in the docs when v12 ships on June 29th.

### DIFF
--- a/12/umbraco-cms/reference/api-versioning-and-openapi.md
+++ b/12/umbraco-cms/reference/api-versioning-and-openapi.md
@@ -24,7 +24,7 @@ The Umbraco APIs rely on having the requested API version as part of the URL. If
 
 The following code sample illustrates how you can use a custom header to pass the requested API version to your own APIs.
 
-{% code title="Changing the API versioning" %}
+{% code title="MyConfigureApiVersioningOptions.cs" %}
 ```csharp
 using Asp.Versioning;
 using Microsoft.Extensions.Options;
@@ -63,7 +63,7 @@ As mentioned in the beginning of this article, Umbraco exposes Swagger and the S
 
 The code sample below shows how to change the Swagger route and availability.
 
-{% code title="Changing the Swagger route and availability" %}
+{% code title="MySwaggerRouteTemplatePipelineFilter.cs" %}
 ```csharp
 using Umbraco.Cms.Api.Common.OpenApi;
 using Umbraco.Cms.Web.Common.ApplicationBuilder;
@@ -124,7 +124,7 @@ The Umbraco APIs use custom operation IDs for that exact reason. In order to rem
 
 If you want to apply custom operation IDs to your APIs, you must ensure that the Umbraco APIs retain their custom operation IDs. The following code sample illustrates how this can be done.
 
-{% code title="Adding custom operation IDs" %}
+{% code title="MyOperationIdSelector.cs" %}
 ```csharp
 using Asp.Versioning;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
@@ -176,7 +176,7 @@ Custom schema IDs can also make it easier for your API consumers to understand a
 
 If you want to create custom schema IDs for your APIs, you must ensure that the Umbraco APIs retain their custom schema IDs. The following code sample illustrates how that can be done.
 
-{% code title="Adding custom schema IDs" %}
+{% code title="MySchemaIdSelector.cs" %}
 ```csharp
 using Umbraco.Cms.Api.Common.OpenApi;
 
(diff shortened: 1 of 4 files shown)


nikolajlauridsen (approved):
LGTM

12/umbraco-cms/reference/api-versioning-and-openapi.md
```diff
@@ -0,0 +1,346 @@
+---
+description: How to use API versioning and OpenAPI (Swagger) for your own APIs.
+---
+
+# API versioning and OpenAPI (Swagger)
+
+Umbraco 12 ships with Swagger to document the Content Delivery API. Swagger and the Swagger UI is available at `{yourdomain}/umbraco/swagger`. For security reasons, both are disabled in non-production environments.
```
sofietoft, line 7: Removed "12" as this will quickly become redundant.

Also, in case someone doesn't know what Swagger is, should we consider linking off to something that could explain this? Or perhaps add a few more words to it here?

```suggestion
Umbraco ships with Swagger to document the Content Delivery API. Swagger and the Swagger UI is available at `{yourdomain}/umbraco/swagger`. For security reasons, both are disabled in non-production environments.
```

12/umbraco-cms/reference/api-versioning-and-openapi.md
```diff
@@ -0,0 +1,346 @@
...
+That being said, it would be sensible to consider migrating your API documentation to Swashbuckle.AspNetCore. This way you can avoid having multiple dependencies that perform the same tasks.
+{% endhint %}
+
+## API versioning
+
+The Umbraco APIs rely on having the requested API version as part of the URL. If you prefer a different versioning for your own APIs, you can setup alternatives while still preserving the functionality of the Umbraco API.
+
+The following code sample illustrates how you can use a custom header to pass the requested API version to your own APIs.
+
+{% code title="Changing the API versioning" %}
```
sofietoft, line 27: When we use the Code Title, we use it for the file name.
Would that make sense to do in this case as well?

12/umbraco-cms/reference/api-versioning-and-openapi.md
```diff
@@ -0,0 +1,346 @@
...
+        });
+        return builder;
+    }
+}
+```
+{% endcode %}
+
+## Adding custom operation IDs
+
+Custom operation IDs can be a great way to make your API easier to use. Specially for consumers that generate API contracts from your Swagger documents.
```
sofietoft, line 121:
```suggestion
Custom operation IDs can be a great way to make your API easier to use. Especially for consumers that generate API contracts from your Swagger documents.
```

12/umbraco-cms/reference/api-versioning-and-openapi.md
```diff
@@ -0,0 +1,346 @@
...
+        //         .Build();
+        builder.Services.ConfigureOptions<MyConfigureSwaggerGenOptions>();
+        return builder;
+    }
+}
+```
+{% endcode %}
+
+With these Swagger documents in place, you can now assign the different versions of your API controllers to their respective documents using the `MapToApi` annotation.
+
```
sofietoft, line 280: For the following code snippets - would it make sense to put them into tabs instead?
That way you can quickly switch between them while avoiding having to scroll up and down to compare,

sofietoft (commented):
Looks great overall @kjac ! 💪 

I've made a couple of comments.
Mostly regarding the code snippets.

12/umbraco-cms/reference/api-versioning-and-openapi.md
kjac: @sofietoft this article is only relevant for Umbraco 12 and above - previous versions do not ship with Swagger. But I suppose since this article will only be part of the 12+ docs, it doesn't matter much if we remove the version notation here...? If it ends up confusing people on older versions, we can always amend it to "Umbraco 12+". 

Also - people reading this article will know what Swagger is 😄

12/umbraco-cms/reference/api-versioning-and-openapi.md
kjac: I'll update all the code titles where it makes sense, or remove them where it does not.

12/umbraco-cms/reference/api-versioning-and-openapi.md
kjac: Not really. We need to ensure that the readers include both versions of the API when testing. I reckon hiding one of them in tabs will increase the risk of that one being left out when testing, thus not achieving the desired result.

12/umbraco-cms/reference/api-versioning-and-openapi.md
kjac: Done 👍

12/umbraco-cms/reference/api-versioning-and-openapi.md
sofietoft: Sure! Makes sense.

sofietoft:
Thanks for applying the suggestions 💪 

I'll merge it.

PR #5244, opened 2023-06-22
Surely this is supposed to be actions

lars-erik:
## Description

A word

## Type of suggestion

Fix freudean slip

* [x] Typo/grammar fix


sofietoft:
Thanks for the PR @lars-erik !

This sure seems like an oversight!

PR #5254, opened 2023-06-26
readability improvements - imagingsettings

erikjanwestendorp:
## Description

Remove 'very'. 
There is also an error related to using the CPU abbreviation, but is this something to add to the acronyms list?

## Type of suggestion

* [x] Readability improvements

### DIFF
--- a/.github/styles/UmbracoDocs/Acronyms.yml
+++ b/.github/styles/UmbracoDocs/Acronyms.yml
@@ -59,3 +59,4 @@ exceptions:
   - UTC
   - HEX
   - RGBA
+  - CPU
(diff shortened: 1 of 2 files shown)


sofietoft:

I think we can get around using CPU here, by only using memory.
Isn't it basically the same thing, just two different words? 🤔 
At least, I don't think the distinction between the two really matters here 😅

sofietoft:
Sounds like you know more about this than me! 😄 

Let's get CPU added to the list of approved acronyms 💪

erikjanwestendorp:
@sofietoft Thanks for your reply 😄. Just updated acronyms.yml

sofietoft:
Excellent! Thanks! 🙌 

I'll get it merged!

PR #5258, opened 2023-06-27
Update rewrites-on-cloud.md

jcrittenden:
Need to exclude /umbidlocallogin from trailing slash, otherwise the 'Sign in with Umbraco ID' will not work on localhost.

## Description

Added "/umbidlocallogin" to the trailing slash redirect rule.

## Type of suggestion

Updated outdated content

Umbraco Cloud 10.4.2


sofietoft:
Thanks for the PR @jcrittenden ! I'll make sure we review this as soon as possible.

Do you by any chance know if this is also an issue on the latest version of Umbraco CMS?

jonat123:
Hey @jcrittenden I've just tried to see whether I could reproduce your issue.

However, I have been unable to do so, It works fine on my end when I try to log in with the rewrite in place without excluding /umbidlocallogin 🤔

I have tested it on both v10 latest and v11 latest, and both work without any issues on my end.

I am a bit curious about how you have added the rewrite.

Did you locally create a web.config for the project or are you using a config transform?

jonat123:
@jcrittenden I have not heard back from you in a while.

I hope that everything works as it should 😊

I will for now close this issue down, however you are welcome to open it up again if you have more information about the issue you where having 😊

Have a good summer 😁

PR #5266, opened 2023-06-28
Add namespace 

erikjanwestendorp:
## Description

Add namespace to code sample.

## Type of suggestion

* [x] Update code sample

v10 and v11

### DIFF
--- a/10/umbraco-cms/extending/section-trees/trees/README.md
+++ b/10/umbraco-cms/extending/section-trees/trees/README.md
@@ -63,8 +63,6 @@ The first node in the tree is referred to as the **Root Node**. You can customis
 ### Implementing the Tree
 
 ```csharp
-using System;
-using System.Collections.Generic;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Umbraco.Cms.Core;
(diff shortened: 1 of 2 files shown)


sofietoft:

We're experiencing some issues with GitBook at the moment.
I'll make sure we pick this one up when everything is back to normal.

erikjanwestendorp:
@sofietoft  That's totally fine, there's no rush. 😄

sofietoft:
GitBook issues all sorted 🤞 

Let's take a look at this one!
When looking at the file changed, it looks like a lot of changes.
But, is it correct to assume that you "only" added an extra line + brackets, to wrap the sample in a namespace `namespace My.Website.Tress` ? 🤔

erikjanwestendorp:
@sofietoft Thanks for your review! It does seem like a lot has changed, but the only change is indeed that a namespace has been added.

sofietoft:
Great 😅 
Just wanted to double-check before sending it to the developers for now apparent reason.

I'll get it merged then 💪 
And thanks again for the PR!

Have a great weekend!

sofietoft:
ACtually!
I was a little too quick there.
We just released Umbraco 12 yesterday, meaning that 12 is not also the default version of the docs.

Should the changes from this PR be added to version 12 as well? 🤔

erikjanwestendorp:
@sofietoft You are totally right this also have to be added to version 12 as well. I'll create another PR😄

PR #5270, opened 2023-06-28
Update 'Content' properties to remove ContentSchedule

marcemarc:
## Description

I think this Pull Request: https://github.com/umbraco/Umbraco-CMS/commit/14e1f7f7fcbb13ad83dd6215d1eab8d9a2efcdd8

Removes the property 'ContentSchedule' from the IContent object

But it's still documented here as if it does exist!

--

I think the new performant way to get the details of any published content schedule is to use the ContentService and GetContentScheduleByContentId

eg

`ContentScheduleCollection schedule = contentService.GetContentScheduleByContentId(id);`

But this is no longer a 'property' of IContent, so I don't think we put this information into the IContent page...

... in fact, the docs as a rule, doesn't document all the methods if they already exist in the autogenerated docs ...

... and this does exist here:

https://apidocs.umbraco.com/v10/csharp/api/Umbraco.Cms.Core.Services.IContentService.html#Umbraco_Cms_Core_Services_IContentService_GetContentScheduleByContentId_System_Int32_

So I think it should just be a case of removing the ContentSchedule property, unless we feel it should have a note at the bottom saying, ContentSchedule property was removed in V10, and link to the autogenerated, but I can't imagine it's that common an issue!

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

v10+

## Deadline (if relevant)

It can be published now


sofietoft:

I'll make sure this is reviewed as soon as possible.
I can see that the PR targets Umbraco 10 currently. Is it a change that applies to newer versions as well?

marcemarc:
It does indeed, it has been removed as a thing in all later versions tooo, was going to ask how best to update all the docs for thus kind of change?

sofietoft:
Ah, of course! 💪 

If you make the change in all the 3 available versions in this PR that would be great!
You're also welcome to do one PR for each version of course - that's really up to you :)

marcemarc:
@sofietoft hopefully have done that correctly!

sofietoft:
Looks great @marcemarc ! 💪 Thanks!

I'll get it merged!

PR #5299, opened 2023-07-13
Update major-upgrades.md

aguyfromdenmark:
Adds Umbraco.Cloud.StorageProviders.AzureBlob to list of packagess to upgrade.


jonat123 (approved):
@aguyfromdenmark Thanks for the PR and good catch that we where missing that one 🙈

Merging it in right away!

PR #5302, opened 2023-07-17
Update maximumuploadsizesettings

erikjanwestendorp:
## Description

Small PR to improve readability. 

- [x] Write shorter sentence.

## Type of suggestion

* [x] Styleguide

v10, 11 and 12

### DIFF
--- a/10/umbraco-cms/reference/configuration/maximumuploadsizesettings.md
+++ b/10/umbraco-cms/reference/configuration/maximumuploadsizesettings.md
@@ -50,7 +50,7 @@ An example of a configuration could look something like this:
 }
 ```
 
-`MaxRequestLength` is specified in kilobytes. This configuration would limit requests, and therefore uploaded files, to 2 megabytes, and a maximum query string length of 90 characters.
+`MaxRequestLength` is specified in kilobytes. This configuration will limit requests, and therefore uploaded files, to 2 megabytes, and a maximum query string length of 90 characters.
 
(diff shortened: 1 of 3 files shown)


10/umbraco-cms/reference/configuration/maximumuploadsizesettings.md
```diff
@@ -50,7 +50,7 @@ An example of a configuration could look something like this:
 }
 ```
 
-`MaxRequestLength` is specified in kilobytes, so this configuration would limit requests, and therefore uploaded files, to 2 megabytes, and a maximum query string length of 90 characters.
+`MaxRequestLength` is specified in kilobytes. This configuration would limit requests, and therefore uploaded files, to 2 megabytes, and a maximum query string length of 90 characters.
```
sofietoft, line 53: Does it make sense to say "This configuration **will** limit requests ... ", instead?
At least that sounds a little better to me 🤔 What do you think?

sofietoft (commented):

I've made a comment - let me know what you think!

10/umbraco-cms/reference/configuration/maximumuploadsizesettings.md
erikjanwestendorp: @sofietoft Definitely agree!

erikjanwestendorp:
@sofietoft Thaks four your review, just updated the PR with your suggestion 😄

sofietoft:
Thanks for considering the suggestion @erikjanwestendorp ! 💪 

I'll get it merged then 😄

PR #5314, opened 2023-07-20
Update upgrade-from-8-to-latest.md

jesperlange:
Changed the wording to specify it's latest version of v10, instead of just latest version, as that would imply current latest version (v12) which does not work.

## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/12/umbraco-cms/fundamentals/setup/upgrading/version-specific/upgrade-from-8-to-latest.md
+++ b/12/umbraco-cms/fundamentals/setup/upgrading/version-specific/upgrade-from-8-to-latest.md
@@ -1,20 +1,20 @@
 ---
 description: >-
-    Learn how to upgrade your Umbraco 8 project to the latest version of Umbraco CMS.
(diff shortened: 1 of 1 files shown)


sofietoft:
Ah, shoot! Didn't get back to you before you created the PR 😅 my bad!

I was gonna suggest that we add a note to the article instead.
It should be possible to go to the latest version, so we've added some tasks to do some more testing.
Also, if we say that this is just to 10, then we don't really provide options for people to go from 10 to 11, and then from 11 to 12 😅 

What do you think?
Would you be up for adding a note to the start of the article instead, saying that currently you need to upgrade to 10 first, before going through to 11 and 12?

jesperlange:
No worries, thought about that as well.
Yes I'll look into adding it as a node instead to start with 👍 
Will do it tomorrow :D

sofietoft:
Yo 👋 
You had some time to give this a second look?

jesperlange (commented):
Looks awesome VAMOS

PR #5320, opened 2023-07-27
Update rendering-scripts.md

skartknet:
TempData doesn't work on Azure if we don't do this. It still works in local even without marking the cookie as essential, I'm not sure why :/

## Description

I added an info box with this details.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)


skartknet:
I just realised that it still doesn't work on Azure even after doing this changes. Sorry, I thought it did. Well, not sure why it doesn't work then. I will close the PR.

PR #5322, opened 2023-07-27
Document "Timestamp" constraint for minification

kjac:
## Description

The runtime minification configuration has some undocumented constraints. Hopefully the PR changes make those constraints abundantly clear, otherwise this PR should not be accepted 😄 

I have updated the docs for v10, v11 and v12.

Cross-linking the origin for future reference: https://github.com/umbraco/Umbraco-CMS/issues/14608

## Type of suggestion

* [x] Other

v10+

## Deadline (if relevant)

Not relevant.

### DIFF
--- a/12/umbraco-cms/reference/configuration/runtimeminificationsettings.md
+++ b/12/umbraco-cms/reference/configuration/runtimeminificationsettings.md
@@ -16,9 +16,13 @@ This section allows you to configure the runtime minifications (defaults shown),
   }
 }
 ```
-## Use 'in memory' cache 
+## Use 'in memory' cache
 
 This setting determines whether Smidge should save it's cached output in memory, or in a file on disk. If set to false, then the folder will be created at the wwwroot of your Umbraco site in a folder called 'Smidge'/
+
+{% hint style="info" %}
+It is not possible to disable in memory caching when `Timestamp` is chosen as `CacheBuster` method.
+{% endhint %}
 
 ## Cache buster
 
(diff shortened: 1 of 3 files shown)


12/umbraco-cms/reference/configuration/runtimeminificationsettings.md
```diff
@@ -16,10 +16,14 @@ This section allows you to configure the runtime minifications (defaults shown),
   }
 }
 ```
-## Use 'in memory' cache 
+## Use 'in memory' cache
 
 This setting determines whether Smidge should save it's cached output in memory, or in a file on disk. If set to false, then the folder will be created at the wwwroot of your Umbraco site in a folder called 'Smidge'/
 
+{% hint style="info" %}
+It is not possible to disable in memory caching when `Timestamp` is chosen as `CacheBuster` method.
```
sofietoft, line 24: How about we do something like this instead?
It would possibly also make more sense to have this warning along with the `CacheBuster` method, as that's the value that affects the in-memory cache.

What do you think?

```suggestion
It is not possible to disable in-memory caching when the `CacheBuster` method has `Timestamp` as the value.
```

sofietoft (commented):
Thanks for the PR @kjac !

Just a couple of thoughts / comments.
Let me know what you think!

kjac:
All good with me 💯

alina-tincas:
Looks good @kjac so am merging it in 💪 🙌

PR #5329, opened 2023-08-02
Fixed wrong webhook URL

mattbrailsford:
## Description

Looks like we had the wrong callback URL set so updated to be correct

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other


alina-tincas:
Hi @mattbrailsford, that´s a good catch there! 💪 

All looks good so I am merging this in 👍

PR #5335, opened 2023-08-04
Update code examples for external authentication

iOvergaard:
## Description

Updated the code examples in the article concerning external login to reflect the latest interfaces as tested with Umbraco 12. These changes are needed to make the examples work.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

12+

### DIFF
--- a/12/umbraco-cms/tutorials/add-google-authentication.md
+++ b/12/umbraco-cms/tutorials/add-google-authentication.md
@@ -128,8 +128,12 @@ A package manager console appears at the bottom where you can install packages w
 3. Type the following in the console:
 
 ```js
-Install-Package Microsoft.AspNetCore.Authentication.Google -Version 7.0.5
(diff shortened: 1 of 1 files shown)


alina-tincas (approved):
Hi @iOvergaard 👋 

Thank you for updating this #h5YR

All looks good so I will just merge it in. 👍 

And just to confirm, can you let me know if this change is only for 12 or does this change need to be done on previous CMS versions as well?

iOvergaard:
Thanks, @alina-tincas. It looks like the docs were written for v10, so should be good for that. Not sure if this change was introduced in v11, so I put it for v12. I think you can leave it be. The most important thing is that this gets put into v13 (where we will introduce other changes as well, when we get closer!).

alina-tincas:
Awesome! Merging it in 💪

PR #5346, opened 2023-08-10
Update deploy-settings.md

Jette94:
Added a comment on the values. 
the higher the value, the higher the batches.


alina-tincas:
Heya @Jette94 👋 

Thank you for the PR! It looks good, however I was wondering if you would be up to making this change to the v11 and v10 of deploy? 🙌

alina-tincas:
I will merge this in and make sure it is updated in v11 and v10 of deploy as well 😊

PR #5354, opened 2023-08-18
Define minimum SQL Server version

mattbrailsford:
Umbraco supports SQL 2012+, but Umbraco Commerce requires 2015+ so we should be explicit


jonat123 (approved):
@mattbrailsford The change makes sense :-) 

Merging it in right away!

mattbrailsford:
That was super quick 🙌

PR #5364, opened 2023-08-23
added magic string dependency instructions

garpunkal:
## Description

I've added instructions on adding the dependency for another nuget package. 

## Type of suggestion

* [ ] Typo/grammar fix
* [X ] New content
* [ ] Other

Umbraco 10, 11, 12.

## Deadline (if relevant)

ASAP

### DIFF
--- a/10/umbraco-forms/developer/extending/adding-a-magic-string-format-function.md
+++ b/10/umbraco-forms/developer/extending/adding-a-magic-string-format-function.md
@@ -10,6 +10,14 @@ These values can be formatted using [filter functions](../magic-strings.md#forma
 
 Filter functions for common operations such as truncating a string or formatting a date or number are provided.  It's also possible to create custom ones in code.
 
+## Dependency
+
+You need to install the following nuget package for Umbraco.Forms.Core.Providers: 
+
+```
+https://www.nuget.org/packages/Umbraco.Forms.Core.Providers
+```
+
 ## Creating a custom format function
 
 To create a custom format function, create a class that implements `IParsedPlaceholderFormatter`.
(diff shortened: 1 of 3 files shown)


alina-tincas:
Hi @garpunkal 👋 
Thank you for the PR! #h5YR

We will review this as soon as possible and once confirmed that it looks good it will be merged in 💪

AndyButland:
This looks OK to me, but I'm not sure we need it.  We have this dependency referenced on the first page for this section, [here](https://docs.umbraco.com/umbraco-forms/developer/extending/adding-a-type).  So I'm not sure it needs repeating here.

Of course you could land on this page directly, but if we said to include it, we should also include it on all the other pages in this section.  And that would get rather repetitive.

If you look for example [here](https://docs.umbraco.com/umbraco-forms/developer/extending/adding-a-fieldtype), we have a line at the top saying _"This builds on the adding a type to the provider model chapter"_.

So perhaps it would be better to add that to the page in question, so it's clear there's some introductory material you should ensure that you are familiar with first.

The `using` statements added certainly do make sense to add and we should at least keep them.

garpunkal:
I understand Andy. 

I landed on this page directly looking for a custom magic string and I didn't have the reference to the using. 

So I thought I'd add this to make it clear.

alina-tincas:
Thank you for confirming @AndyButland! 🙌 

@garpunkal would you be up to making the change as Andy mentioned, to add a note such as below:
image

to the magic string format documentation instead of adding a new section called deployment and sharing link to nuget (as Andy mentioned that it is all ready documented [here ](https://docs.umbraco.com/umbraco-forms/developer/extending/adding-a-type#preparations))?

alina-tincas:
Hi @garpunkal 👋 

I was wondering if you have had some time to look into my last message, or would you prefer for us to make this change instead? 💡

garpunkal:
If you could make the change, that would be great! :)

alina-tincas:
Sure thing! I have made the change in this PR #5426, therefore I am closing this one 👍 

Thank you again for bringing this to our attention 🙌

PR #5378, opened 2023-08-30
Update major-upgrades.md

aguyfromdenmark:
## Description

I am not sure if I have worded it in the right way. But to me, it seems a bit wierd that one of the prerequisites for upgrading to the latest version, is that you are running on the latest version. 

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [X] Other


jonat123:
@aguyfromdenmark Thanks for the PR :) 

The change makes sense to me :D

PR #5380, opened 2023-08-31
Add note about Umbraco application URL

bjarnef:
## Description

In Umbraco 7 and 8 we had the following:

```
<web.routing
  trySkipIisCustomErrors="true"
  internalRedirectPreservesTemplate="false" disableAlternativeTemplates="false" validateAlternativeTemplates="false" disableFindContentByIdPath="false"
  umbracoApplicationUrl="">
</web.routing>
```

where the comment says:

```
@umbracoApplicationUrl
  The url of the Umbraco application. By default, Umbraco will figure it out from the first request.
  Configure it here if you need anything specific. Needs to be a complete url with scheme and umbraco
  path, eg http://mysite.com/umbraco. NOT just "mysite.com" or "mysite.com/umbraco" or "http://mysite.com".
```

However @nul800sebastiaan pointed out it isn't required as the BackOffice path on no longer configurable in Umbraco 9+
https://discord.com/channels/869656431308189746/1146380570398568478

Not sure what happens I including `/umbraco` though - maybe it is just ignored, when Umbraco parse the URL and only use the domain and append `/umbraco` even though `http://mysite.com/admin` was set in this setting?

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/12/umbraco-cms/reference/configuration/webroutingsettings.md
+++ b/12/umbraco-cms/reference/configuration/webroutingsettings.md
@@ -101,4 +101,4 @@ Will set the URL provider mode, options are:
 
 Defines the Umbraco application URL that the server should reach itself. By default, Umbraco will guess that URL from the first request made to the server. Use this setting if the guess is not correct (because you are behind a load-balancer, for example). Format is: `http://www.mysite.com/`, ensure to contain the scheme (http/https) and complete hostname.
 
-Note that previously it was also to specify BackOffice path as this was customizable (`/umbraco` by default), but this is no longer possible in Umbraco 9+ so using the full URL including scheme and hostname is sufficient.
+Note that previously it was also required to specify BackOffice path as this was customizable (`/umbraco` by default), but this is no longer possible in Umbraco 9+ so using the full URL including scheme and hostname is sufficient.
--- a/12/umbraco-cms/reference/configuration/webroutingsettings.md
+++ b/12/umbraco-cms/reference/configuration/webroutingsettings.md
@@ -101,4 +101,6 @@ Will set the URL provider mode, options are:
 
 Defines the Umbraco application URL that the server should reach itself. By default, Umbraco will guess that URL from the first request made to the server. Use this setting if the guess is not correct (because you are behind a load-balancer, for example). Format is: `http://www.mysite.com/`, ensure to contain the scheme (http/https) and complete hostname.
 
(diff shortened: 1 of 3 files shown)


12/umbraco-cms/reference/configuration/webroutingsettings.md
```diff
@@ -100,3 +100,5 @@ Will set the URL provider mode, options are:
 ## Umbraco application URL
 
 Defines the Umbraco application URL that the server should reach itself. By default, Umbraco will guess that URL from the first request made to the server. Use this setting if the guess is not correct (because you are behind a load-balancer, for example). Format is: `http://www.mysite.com/`, ensure to contain the scheme (http/https) and complete hostname.
+
+Note that previously it was also required to specify BackOffice path as this was customizable (`/umbraco` by default), but this is no longer possible in Umbraco 9+ so using the full URL including scheme and hostname is sufficient.
```
alina-tincas, line 104: We have a styleguide in place to make the documentation more readable, and one is to write sentences that are[ less than 25 words](https://docs.umbraco.com/welcome/contribute/style-guide#long-sentences). Therefore I am suggesting the below, with the addition of an info warning similar to [this](https://docs.umbraco.com/umbraco-cms/reference/configuration/webroutingsettings#:~:text=Are%20you%20on%20Umbraco%20Cloud%3F). I have suggested a new sentence on 2nd part to match the same wording as [here](https://our.umbraco.com/Documentation/Reference/Configuration-for-Umbraco-7-and-8/umbracoSettings/#:~:text=the%20scheme%20(http/https)%2C%20complete%20hostname)
```suggestion
{% hint style="info" %}
Previously before v9, it was required to specify **backofffice** path as this was customizable (`/umbraco` by default). However, from v9+ this is no longer possible, so it's sufficient to use the URL that contains the scheme (http/https) and complete hostname.
{% endhint %}
```

alina-tincas (changes requested):
Hi @bjarnef 👋 

Thank you for the PR! I have made a few suggestions, can you please take a look at them and let me know what you think?

Also, I was wondering if you would be up to making the same change for the v11 and v10 documentation as well? 😊

12/umbraco-cms/reference/configuration/webroutingsettings.md
bjarnef: I have updated this :)

bjarnef:
@alina-tincas I have applied the suggestions and added it to v10 and v11 docs as well.

alina-tincas:
@bjarnef thank you for the adjustments, all looks awesome now 😁 

Merging it in 💪

PR #5391, opened 2023-09-06
Update major-upgrades.md

aguyfromdenmark:
Change to make it more clear that you are only to make sure the .net version is available, and not actually to enable it.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ C] Other

### DIFF
--- a/umbraco-cloud/product-upgrades/major-upgrades.md
+++ b/umbraco-cloud/product-upgrades/major-upgrades.md
@@ -53,7 +53,7 @@ Upgrade tutorial Umbraco 9 to 10 on Cloud
 * Go to the project in the Umbraco Cloud portal.
 * Navigate to **Settings** -> **Advanced**.
 * Scroll down to the **Runtime Settings** section.
-* **Ensure that the latest version of .NET is enabled** for each environment on your Cloud project.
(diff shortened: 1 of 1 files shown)


aguyfromdenmark:
This might be wrong. Someone who can clear this up for me?
Do you really have to change the runtime version before upgrading? Isn't that risky?
But the original wording just says to make sure it's enabled, but in step 3, it tells you to select the same version as on the cloud project. So, should you change it on your cloud project or not? :)

jonat123:
Hey @aguyfromdenmark :) 
Thanks for the PR!

So, when you change the .NET version in the cloud portal it changes it for the environment (The dev, stage or live environment) However it does not change .NET version for the Umbraco project.

What we are doing in step 3 is that before upgrading to the latest major locally we need to set the target framework to the same .NET version as the one that was set on the cloud environment :-) 

I hope this clears it up, otherwise let me know :-)

aguyfromdenmark:
Thanks @jonat123 , that makes sense. I have always been afraid of chancing the .NET version on the cloud portal, because to me that sould like a recipe for disaster. But every upgrade has still been working, as the .NET version adjusted itself. So I assumed I was doing it right. I would still like to make some adjustments to the wording, but I will do it in another PR, and close this one.

PR #5398, opened 2023-09-07
Update Migrate from Vendr to Umbraco Commerce README.md

Wiggee11:
I think this is a typo in the docs. I'm guessing the correct interface name.


alina-tincas:
Hi @Wiggee11, thank you for the PR 🙌 

That is a good point you have there! Before merging this in, I was wondering if you would be up to making this change in the v12 of umbraco commerce on the same article as well? 😊

Wiggee11:
> Hi @Wiggee11, thank you for the PR 🙌
> 
> That is a good point you have there! Before merging this in, I was wondering if you would be up to making this change in the v12 of umbraco commerce on the same article as well? 😊

Done https://github.com/umbraco/UmbracoDocs/pull/5408 :)

alina-tincas:
Thank you! #h5YR. Merging both PRs 😁

PR #5433, opened 2023-09-18
Update activecampaign with umbraco forms

erikjanwestendorp:
## Description

Fix definition of CRM 

## Type of suggestion

* [x] Styleguide

Integrations


alina-tincas:
Hi @erikjanwestendorp good catch! All looks good so am merging it in 💪

PR #5436, opened 2023-09-19
Updated forms docs about removing built-in fields

karltynan:
## Description

Updated forms docs about removing built-in fields

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

Umbraco Forms 10.5+

### DIFF
--- a/10/umbraco-forms/developer/extending/customize-default-workflows.md
+++ b/10/umbraco-forms/developer/extending/customize-default-workflows.md
@@ -258,7 +258,7 @@ builder.Services.AddUnique<IApplyDefaultFieldsBehavior, CustomApplyDefaultFields
 
 ### Example - Excluding a Built-in Field
 
-The following class shows how to exclude a built-in field types using a custom composer. The `Password`, `Recaptcha2` and `RichText` field types (or "answers") will no longer be available for selection when creating a form in the backoffice.
+The following class shows how to exclude built-in field types using a custom composer. The `Password`, `Recaptcha2` and `RichText` field types (or "answers") will no longer be available for selection when creating a form in the backoffice.
 
 ```
 using Umbraco.Forms.Core.Providers;
--- a/10/umbraco-forms/developer/extending/customize-default-workflows.md
+++ b/10/umbraco-forms/developer/extending/customize-default-workflows.md
@@ -1,5 +1,5 @@
 ---
-description: "How to amend the built-in behavior of adding fields and associating workflows with new forms"
+description: "How to amend the built-in behavior of adding/excluding fields and associating workflows with new forms"
 ---
 
 # Customize Default Fields and Workflows For a Form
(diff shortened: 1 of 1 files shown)


alina-tincas:
Hi @karltynan hope you´re well! Thank you again for the amazing CodeCabin 😁 

Thank you for the PR as well! I will just need to run this over with our forms developers to confirm if all is good 👍

10/umbraco-forms/developer/extending/customize-default-workflows.md
```diff
@@ -254,4 +254,27 @@ Again, you will need to register your custom class, for example, in a composer w
...
+namespace MyNamespace
+{
+    public class MyFormFieldsComposer : IComposer
+    {
+        public void Compose(IUmbracoBuilder builder)
+        {
+            builder.WithCollectionBuilder<FieldCollectionBuilder>()
+               .Exclude<Password>()
+                .Exclude<Recaptcha2>()
+                .Exclude<RichText>();
```
AndyButland, line 276:
```suggestion
            builder.FormsFields()()
                .Exclude<Password>()
                .Exclude<Recaptcha2>()
                .Exclude<RichText>();
```

AndyButland (commented):
The content is great to add - I've just made a small suggestion that adds some syntactic sugar to make the code a tiny bit shorter.

The question is more to find the right place for it.  I don't think this page is quite right, as it's about customising what fields and workflows you get on a new "empty" form, rather than what fields are available to use across the installation.

We have a reference to this already at the bottom of [this section](https://docs.umbraco.com/umbraco-forms/developer/extending/adding-a-type#registering-the-class-with-umbraco-and-forms).  But again that's not ideal, as you would most likely be reading this page only if you want to add fields, not remove them.

Overall I think probably better is that it's moved to a new page within the [Extending section](https://docs.umbraco.com/umbraco-forms/developer/extending).  I suspect it's a common requirement so if we a page dedicated for how to remove fields, even if it's quite short, that's probably best.

It should probably reference removing workflows too.

We would want this under the Umbraco 11 and 12 versions as well.

10/umbraco-forms/developer/extending/customize-default-workflows.md
karltynan: This syntax improvement is great to know - I will use this approach in my code!

I was struggling to find how to do this, so it is going to be very useful. I am not sure of support for v11/12 (since we only use LTS) so perhaps someone could confirm this for v11/12 and move this content to wherever you think is most appropriate?

AndyButland: Yes, this would be the same across 11 and 12.  Over to you I think @alina-tincas to see if you can migrate this to an appropriate place (you can consider my suggestions above).

alina-tincas: @AndyButland do you think this example can be added here: https://docs.umbraco.com/umbraco-forms/developer/extending/adding-a-fieldtype as this is about field types? Because adding a new article we need to have a bit more content 🤔

AndyButland: It could, but actually I would suggest that even though the page would be a bit light on content, it's probably better as a new page.  The reason being that removing existing fields and adding new fields, whilst very related in a technical sense, are fairly different when it comes to the use case of why you'd want to do one or other of these things.

In the former case, it's because for example you have no use for file upload and don't want editors using them.

In the latter, it's because there's some new functionality you need you want to build yourself.

I could see people looking to do either of these things, but not generally both at the same time.

alina-tincas: Thank you for confirming Andy! I have created a new PR for this here: #5486

alina-tincas:
As I have created a new PR for this here: #5486 I am closing this PR. 👍 

Thank you again for bringing this up @karltynan 😊 🙌

PR #5452, opened 2023-09-25
Update extending-the-dashboard-using-umbraco-ui-library - the new backoffice

erikjanwestendorp:
## Description

Update extending-the-dashboard-using-umbraco-ui-library:

- [x] Define UUI 

Or is this something we want to add to the list of acronyms?

extending-the-dashboard-using-umbraco-ui-library
## Type of suggestion

* [x] Styleguide

The new backoffice


nielslyngsoe (approved):
Good to specify this

alina-tincas:
Thank you for the PR @erikjanwestendorp and for confirming @nielslyngsoe 🙌 

I agree as well that it should be defined instead of adding it to the list of acronyms, as it is not that self-explanatory what UUI means by default 😊

PR #5453, opened 2023-09-25
Release notes for Workflow 10.3.1, 11.3.1 and 12.1.1

nathanwoulfe:
## Description

Adds release notes for the latest patch releases of Workflow. Wondering too if it's really necessary (or useful) for later versions to include release notes for all previous versions (ie v12 release notes include v11 and v10), given almost all changes are the same across all three versions?

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

Workflow 10.3.1, 11.3.1 and 12.1.1

### DIFF
--- a/10/umbraco-workflow/release-notes.md
+++ b/10/umbraco-workflow/release-notes.md
@@ -12,6 +12,14 @@ In this section, you can find the release notes for each version of Umbraco Work
 
 <summary>Version 10</summary>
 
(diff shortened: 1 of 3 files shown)


alina-tincas:
Thank you for the PR @nathanwoulfe 🙌 

That is a good question you have there! Jonathan has reached out to Andy to discuss with the team what they think about not including older versions in the newer versions  (because if this is done for Workflow then it needs to be done for all DXP products as well to keep the docs consistent)  💡 

Meanwhile, can you please let me know when should this be merged in/published? 🙏

PR #5511, opened 2023-10-13
Update two-factor-authentication.md

emmagarland:
I noticed that the namespace was Umbraco9, so I changed it to something more generic and as per https://github.com/umbraco/UmbracoDocs/blob/main/12/umbraco-cms/implementation/composing.md.

I don't know if there is a best practice for what namespace to use? But I thought it best not to be version specific, anyway.

## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

Any .NET Core Umbraco version currently

## Deadline (if relevant)

### DIFF
--- a/10/umbraco-cms/reference/security/two-factor-authentication.md
+++ b/10/umbraco-cms/reference/security/two-factor-authentication.md
@@ -27,7 +27,7 @@ using Google.Authenticator;
 using Umbraco.Cms.Core.Security;
 using Umbraco.Cms.Core.Services;
 
-namespace Umbraco9
+namespace My.Website
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


alina-tincas:
Hi @emmagarland thank you for the PR 🙌 

That is a good point you have there, and I agree! The namespace can be anything and should not be the version name 💪 

All looks good so am merging it in 😊

emmagarland:
> Hi @emmagarland thank you for the PR 🙌
> 
> That is a good point you have there, and I agree! The namespace can be anything and should not be the version name 💪
> 
> All looks good so am merging it in 😊

Thanks @alina-tincas!

PR #5523, opened 2023-10-17
Adding new AllowConcurrentLogins setting (for 12.3)

elit0451:
## Description
- Adding documentation about a new configuration setting that will be introduced with v12.3;

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [X] Updates related to a new version
* [ ] Other

CMS v12.3

## Deadline (if relevant)
Between 19th Oct and 2nd Nov 🙂 - 12.3RC is out on 19th Oct and the final release is out on 2nd Nov. So, it is fine to merge it in this period (at latest on 2nd Nov).

---

> **Note**
> In v13.0.0 we will change the default value of `Umbraco:CMS:Security:AllowConcurrentLogins` from `true` to `false` - related to https://github.com/umbraco/Umbraco-CMS/pull/14989. So after we merge this PR we need to make sure that the following changes are made and available for 1st Nov:

Indicate that the default value of `AllowConcurrentLogins` is changed to `false`:
```json
      "MemberDefaultLockoutTimeInMinutes": 43200,
      "AllowConcurrentLogins": false
```

And remove the hint and keep the description:
```md
## Allow concurrent logins

When set to `false`, any user account is prevented from having multiple simultaneous sessions. In this mode, only one session per user can be active at any given time. This enhances security and prevents concurrent logins with the same user credentials.
```

### DIFF
--- a/12/umbraco-cms/reference/configuration/securitysettings.md
+++ b/12/umbraco-cms/reference/configuration/securitysettings.md
@@ -37,7 +37,8 @@ A full configuration with all default values can be seen here:
         "MaxFailedAccessAttemptsBeforeLockout": 5
       },
       "UserDefaultLockoutTimeInMinutes": 43200,
-      "MemberDefaultLockoutTimeInMinutes": 43200
+      "MemberDefaultLockoutTimeInMinutes": 43200,
+      "AllowConcurrentLogins": true
     }
   }
 }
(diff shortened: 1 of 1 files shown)


bjarnef:
@elit0451 I wonder if this setting can be misunderstood as I guess concurrent logins is still allowed?
Would something like `AllowConcurrentLoginsAsUser` be a bit more clear?

bjarnef:
After reading it again.

So it means only one session is allowed, so either a single login with different or same user?

elit0451:
Consider signing in from two different browsers with the same account. When the setting is set to `false`, concurrent logins are not allowed, meaning that only the newest user session is active and all the others will be prompted to login again.

alina-tincas:
Thank you for the PR @elit0451 and thank you for your feedback @bjarnef 🙌 #h5YR 

I will make sure this is merged in tomorrow 💪

PR #5528, opened 2023-10-17
Block Grid custom view with image sample code bug

snerpton:
## Description

Sample code to display an image from a Media Picker throws some javascript errors:

  1. `https://localhost:44309/umbraco/backoffice/umbracoapi/media/GetById?id=%5Bobject%20Object%5D 404 (Not Found)`

  2. `Possibly unhandled rejection: {"errorMsg":"Failed to retrieve data for media id [object Object]","data":"","status":404,"xhrStatus":"complete"}`

The sample code `$scope.block.data.image` expects a single image to be returned whilst it returns a collection.

Additionally, the image ID is accessed via the `mediaKey` property.

## Possible solution

Appending `image[0].mediaKey` returns the media GUID which can be passed to the `mediaResource.getById` function, and resolves the error.

## My setup

Umbraco: 10.7.0 / .Net6
Rider: 2023.2.2
OS: MacOS 14.0 / Arm
Browser: Chrome 118.0.5993.70

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)

No panic.

### DIFF
--- a/10/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/build-custom-view-for-blocks.md
+++ b/10/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/build-custom-view-for-blocks.md
@@ -87,7 +87,7 @@ With the setup of files above, you need to amend the `customBlock.controller.js`
 angular.module("umbraco").controller("customBlockController", function ($scope, mediaResource) {
 
     //your property is called image so the following will contain the udi:
-    var imageUdi = $scope.block.data.image;
+    var imageUdi = $scope.block.data.image[0].mediaKey;
     //the mediaResource has a getById method:
     mediaResource.getById(imageUdi).then(function (media) {
         console.log(media);


alina-tincas:
Hi @snerpton thank you for the PR 🙌 

We will look into it as soon as we can 💪

alina-tincas:
Hi @snerpton thank you for your patience! 🙌 

I was able to reproduce the issue and by adding image[0].mediaKey it does indeed resolve it 💪 

I have tested this on v12 as well and it happens there as well, so I was wondering if you would be able to make this same change on v11, v12 and v13 of CMS as well? Otherwise, I can add this to our backlog to make the same change so it is consistent on all versions of CMS. Let me know what you would prefer! 😊

alina-tincas:
I also found out that it also works if on the next line for `mediaResource `if `[0].mediaKey` is added such as `mediaResource.getById(imageUdi[0].mediaKey).then(function (media)` then it works as well 👍

alina-tincas:
Hi @snerpton as I haven´t heard from you for a while, I will merge this PR in and make the same change to all the other versions as well 💪 Thank you again for bringing this to our attention 😊

PR #5638, opened 2023-11-20
Update Heartcore GraphQL schema generation

rasmusjp:
Adds Block Grid reserved types and removes unneeded directives from types


jonat123 (approved):
@rasmusjp Looks good 😄

Merging in right away 😄

PR #5662, opened 2023-11-28
Ignore rewrite of Smidge Bundle assets

bjarnef:
## Description

The recommended redirect rule for project Umbraco Cloud domain caused some font icons missing and different fonts due to rewrite of backoffice CSS, which caused the relative URL's to fonts to load from redirected domain and not requested domain. This caused CORS errors and thus the fonts wasn't loaded in backoffice.

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/12/umbraco-cms/reference/routing/iisrewriterules.md
+++ b/12/umbraco-cms/reference/routing/iisrewriterules.md
@@ -30,6 +30,7 @@ To use rewrites with Umbraco 9 you have to register the middleware in your `Star
         <add input="{HTTP_HOST}" pattern="\.umbraco\.io$" />
         <add input="{REQUEST_URI}" pattern="^/App_Plugins/" negate="true" />
         <add input="{REQUEST_URI}" pattern="^/umbraco" negate="true" />
+        <add input="{REQUEST_URI}" pattern="^/sb" negate="true" /> <!-- Don't redirect Smidge Bundle -->
       </conditions>
       <action type="Redirect" url="https://example.com/{R:0}" />
     </rule>


bjarnef:
Alternative this may also work to ignore redirect of static assets in case the path is not within `/umbraco` like in the case.

```
<add input="{REQUEST_FILENAME}" matchType="IsFile" negate="true" />
<add input="{REQUEST_FILENAME}" matchType="IsDirectory" negate="true" />
```

bjarnef:
I also noticed the preview doesn't work from 
https://project-alias.euwest01.umbraco.io/umbraco/preview/?id=1152&culture=en

because I redirect the frontend to https://mydomain.com/

image

https://project-alias.euwest01.umbraco.io/1152?culture=en -> https://www.mydomain.com/1152?culture=en

and it `Refused to display 'https://www.mydomain.com/' in a frame because it set 'X-Frame-Options' to 'sameorigin'.`

However preview does work from https://mydomain.com/umbraco

Not sure if there is any way to fix this?

bjarnef:
Hi @alina-tincas 

Oh, I didn't noticed there was a separate IIS rewrite rule for Umbraco Cloud (but actually I did know about it).

I think it should be in both places since Smidge also is using on project outside Umbraco Cloud - at least by default.

alina-tincas:
Thank you for feedback Bjarne! We will take a look at this PR as soon as we can 🙏

bjarnef:
I think by default .NET Core is configured to allow same origin.

Maybe something like this?
https://stackoverflow.com/a/31942128

https://www.bytehide.com/blog/cors-aspnet-core

Not sure if it is possible to handle it dynamically based on domains configured on Umbraco or maybe a config in `appsettings.json` in future?

alina-tincas:
Hi @bjarnef! As you are currently in touch with support about these issues, I suggest we wait until the case is resolved and then update this PR as per the findings 🙏

bjarnef:
@alina-tincas I updated the redirect rule not to redirect other environment Umbraco Cloud URLs like Development and Staging.

Otherwise it would require to handle this via web.config transform or deliver a different IISUrlRewrite.xml depending on environment.

jonat123:
@bjarnef I have had a look at reviewing your PR.

I am just wondering whether   adding the:  `<add input="{REQUEST_URI}" pattern="^/sb" negate="true" />  
`
It may be too specific since it is based on your setup. If you are someone else you might call your folder something other than `sb`.

I am thinking that maybe it could make more sense, to add a note on the pages under the specific rules to mention that you should add a condition to your rewrite if you are using smidge so that it do not redirect the smidge bundle :-)

Let me know what you think!

bjarnef:
@jonat123 it is based on default Umbraco Cloud structure `sb` is not a folder name, it's a dynamic path in bundle name, which is generated by default on Umbraco Cloud projects and used in backoffice.

So it the recommended redirect for Umbraco Cloud URL is added the

jonat123:
@bjarnef Ahhh, thanks for elaborating! 

Just had a look myself and you are right, so I see no problem in adding it to the rewrites :) 

Regarding the other rule, the recommended approach to adding rewrites on the cloud is to create config transforms which we also mention in the cloud article. 

So not sure that it makes too much sense to add the  `<add input="{HTTP_HOST}" pattern="^(dev|stage)(.*)?.euwest01.umbraco.io$" ignoreCase="true" negate="true" />
`

bjarnef:
@jonat123 yeah, transform can be used, but requires moving it to a `web.config` and only works on Windows. On Kestrel it would need to handle it different, e.g. deliver a specific XML based on environment.

I just though this was a simple way to handle the rule on multiple environment in general.

jonat123:
ahh thats fair :) It just needs to be removed from the cloud docs, as that should just be done with the config.transforms 👍 For the CMS docs we can keep it 😄

jonat123:
@bjarnef Hey, I was wondering if you have seen my previous message. 😊

bjarnef:
@jonat123 although config transforms can be used on Umbraco Cloud, I don't seen anything wrong using pattern matching on Umbraco Cloud. Sometimes we prefer implementing code that works on both Umbraco Cloud, Azure or somewhere else which may not be a Windows server.

Jeavon:
I was going to make this change too 😉

There's some cross over with https://github.com/umbraco/UmbracoDocs/pull/5791 but I would recommend removing the data centre all the rules. 

Also as you mentioned `DependencyHandler.axd` is v7/8 and should be removed.

We've been using the following conditions
```
	<conditions>
		<add input="{HTTP_HOST}" pattern="\.umbraco\.io$" />
		<add input="{REQUEST_URI}" pattern="^/umbraco" negate="true" />
		<add input="{REQUEST_URI}" pattern="^/App_Plugins/" negate="true" />
		<add input="{REQUEST_URI}" pattern="^/sb" negate="true" />
		<add input="{HTTP_HOST}" pattern="^localhost(:[0-9]+)?$" negate="true" />
	</conditions>
```

jonat123:
@bjarnef Sorry for the confusion! 

Just did a bit of testing and I believe that you are correct, and it should be fine to use pattern matching on Cloud 😊

Would it be possible to add it so that it can work for all regions and not just the euwest? 

 A bit like Jeavon's example above?

Jeavon:
This rule also stops the preview from working properly as the preview window makes a request to the nodeid at the root, e.g `https://example.com/1232` which will get redirected
 
Checking the `UMB_PREVIEW` seems like a good solution
`<add input="{HTTP_COOKIE}" pattern="^(.+; )?UMB_PREVIEW=([^;]*)(;.+)?$" negate="true" />`
However if the website loads fonts you might still get CORS issues in preview so we have found checking the `UMB_UCONTEXT` cookie is the best overall solution as this cookie is present in all preview requests.
`<add input="{HTTP_COOKIE}" pattern="^(.+; )?UMB_UCONTEXT=([^;]*)(;.+)?$" negate="true" />`

I can send a PR for fixing that after this is merged if you rather

jonat123:
@Jeavon Just to be sure, when you say _"This rule also stops the preview from working properly as the preview window makes a request to the nodeid at the root, e.g https://example.com/1232 which will get redirected"_ Which rule are you referring to? 😊

In any case, you are more than welcome to send a PR ahead, once this has been merged 😄

Jeavon:
@jonat123 it's the hide umbraco.io rule

Try to use preview with:
```
<rule name="Redirects umbraco.io to actual domain" stopProcessing="true">
  <match url=".*" />
  <conditions>
    <add input="{HTTP_HOST}" pattern="\.umbraco\.io$" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/umbraco" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/App_Plugins" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/sb" /> <!-- Don't redirect Smidge Bundle -->
    <add input="{REQUEST_URI}" negate="true" pattern="localhost" />
  </conditions>
  <action type="Redirect" url="http://<your actual domain here>.com/{R:0}"
        appendQueryString="true" redirectType="Permanent" />
</rule>
```

Then try to add the below as a condition
`<add input="{HTTP_COOKIE}" pattern="^(.+; )?UMB_UCONTEXT=([^;]*)(;.+)?$" negate="true" />`

umbraco-cloud/set-up/project-settings/manage-hostnames/rewrites-on-cloud.md
```diff
@@ -39,6 +39,7 @@ One approach for this is to add a new rewrite rule to the `<system.webServer><re
     <add input="{REQUEST_URI}" negate="true" pattern="^/umbraco" />
     <add input="{REQUEST_URI}" negate="true" pattern="^/DependencyHandler.axd" />
     <add input="{REQUEST_URI}" negate="true" pattern="^/App_Plugins" />
+    <add input="{REQUEST_URI}" negate="true" pattern="^/sb" /> <!-- Don't redirect Smidge Bundle -->
     <add input="{REQUEST_URI}" negate="true" pattern="localhost" />
```
jonat123, line 43:
```suggestion
    <add input="{HTTP_HOST}" pattern="^(dev|stage)(.*)?.euwest01.umbraco.io$" ignoreCase="true" negate="true" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/umbraco" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/DependencyHandler.axd" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/App_Plugins" />
    <add input="{REQUEST_URI}" negate="true" pattern="^/sb" /> <!-- Don't redirect Smidge Bundle -->
    <add input="{REQUEST_URI}" negate="true" pattern="localhost" />
```

jonat123:
@Jeavon Thanks! Just tested it out and the preview does indeed not work properly without adding the `<add input="{HTTP_COOKIE}" pattern="^(.+; )?UMB_UCONTEXT=([^;]*)(;.+)?$" negate="true" />` rule.

You are more than welcome to create a PR, otherwise, I can also just add it to the rule, let me know what you prefer 😄

@bjarnef I have added the `<add input="{HTTP_HOST}" pattern="^(dev|stage)(.*)?.euwest01.umbraco.io$" ignoreCase="true" negate="true" />` rule back again as what you said about makes sense, and that it is possible to add rules to the web.config on cloud without the need of creating a config transform 😄

I will go ahead and commit the change and then merge this PR so we can get it out 😄

Again, thanks for the help @bjarnef #H5YR

Jeavon:
> @Jeavon Thanks! Just tested it out and the preview does indeed not work properly without adding the `<add input="{HTTP_COOKIE}" pattern="^(.+; )?UMB_UCONTEXT=([^;]*)(;.+)?$" negate="true" />` rule.
> 
> You are more than welcome to create a PR, otherwise, I can also just add it to the rule, let me know what you prefer 😄
> 
> @bjarnef I have added the `<add input="{HTTP_HOST}" pattern="^(dev|stage)(.*)?.euwest01.umbraco.io$" ignoreCase="true" negate="true" />` rule back again as what you said about makes sense, and that it is possible to add rules to the web.config on cloud without the need of creating a config transform 😄
> 
> I will go ahead and commit the change and then merge this PR so we can get it out 😄
> 
> Again, thanks for the help @bjarnef #H5YR

@jonat123 Go ahead and add it in!

PR #5684, opened 2023-12-05
Update README.md

Ric-Carey:
Added note around Upgrade from 9 to 10, if Umbraco Forms is present, which causes a conflict with the backoffice JavaScript.

## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/12/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
+++ b/12/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
@@ -774,10 +774,10 @@ When the upgrade is completed and tested, and prior to deploying to any publicly
    * `/umbraco/UmbracoWebsite`
    * `/umbraco/config/lang`
    * `/umbraco/config/appsettings-schema.json`
-   * If upgrading Umbraco Forms (from 9.* to 10.*) delete ‘/App_Plugins/UmbracoForms’
-4. Deploy the site how you normally would to your public facing environment.
-5. Start the site. At this point it will launch and upgrade the database, after which the site should become accessible and your upgrade is complete.
-6. Check the logs for any errors which may have occurred during the upgrade process.
+4. If you are using Umbraco Forms, update your files and folders according to the [Upgrading - version specific](https://docs.umbraco.com/umbraco-forms/installation/version-specific) for version 10 article.
+5. Deploy the site how you normally would to your public facing environment.
+6. Start the site. At this point it will launch and upgrade the database, after which the site should become accessible and your upgrade is complete.
+7. Check the logs for any errors which may have occurred during the upgrade process.
 
 </details>
 
--- a/12/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
+++ b/12/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
@@ -774,6 +774,7 @@ When the upgrade is completed and tested, and prior to deploying to any publicly
    * `/umbraco/UmbracoWebsite`
    * `/umbraco/config/lang`
    * `/umbraco/config/appsettings-schema.json`
+   * If upgrading Umbraco Forms (from 9.* to 10.*) delete ‘/App_Plugins/UmbracoForms’
 4. Deploy the site how you normally would to your public facing environment.
 5. Start the site. At this point it will launch and upgrade the database, after which the site should become accessible and your upgrade is complete.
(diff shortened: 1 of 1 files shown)


alina-tincas:
Hi @Ric-Carey thank you for the PR! 

After reviewing, the part that you are adding is actually all ready added above, on step 11 https://docs.umbraco.com/umbraco-cms/fundamentals/setup/upgrading/version-specific#9.latest-to-10: 
image

The step 11 is part of "Steps to upgrade using Visual Studio" and it seems you are adding it for the "When the upgrade is completed and tested, and prior to deploying to any publicly accessible environment, you should consider the following". 

If you follow the link from step 11 " [Upgrading - version specific](https://docs.umbraco.com/umbraco-forms/installation/version-specific) for version 10 article." there is more information on what steps are needed to do:
image

The reference link is redirected to the forms documentation from step 11, as it is a different documentation where we document Umbraco Forms related article. The current documentation you have created the PR for is for Umbraco CMS, therefore we do not document what is all ready documented on other products. 

I hope this makes sense, otherwise if you have any further questions, please let us know! 🙏 

So in this case as we already have this step mentioned in the documentation, I am closing this PR. 👍

Ric-Carey:
Hi @alina-tincas the issue being is that because a lot of that happens automatically via NuGet etc, and so will work when you do the upgrade and test locally, BUT when you deploy to the live environment e.g. onto Azure or a FileSystem deploy which requires manual intervention unless you explicitly know to actually do that, you wont. Hence why I've lost over half a day of time trying to fix this in the live environment (whereby the bundled back office JS was conflicting and aspects of nested content and other features within the backoffice to be broken), and I only knew to fix it because of by chance watching this official video https://youtu.be/AN5OOKLHmPE?t=527 see 8:43 and its not detailed above. The video and the docs should be the same, otherwise the same issue will happen to other people (or myself in the future! haha). 

With your reason not to merge my change, you should therefore add another step the section I added to, saying to revisit the link in step 11 in the prior steps. As these are two separate tasks in their own right.

Ric-Carey:
Cheers @alina-tincas. So this particular instance was an Azure Web App (not Umbraco Cloud), but it would also be applicable to any self-hosted website (server agnostic) where it doesn’t have some sort of automated deployment which cleans these files. If the files are present, the error will be too (probably its not been flagged by users, as it appears to have been connected to a doc type with a Form Picker, inside a Nested Content which is a particular scenario, but also uSync and other backoffice functions appeared broken).

12/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
```diff
@@ -774,6 +774,7 @@ When the upgrade is completed and tested, and prior to deploying to any publicly
    * `/umbraco/UmbracoWebsite`
    * `/umbraco/config/lang`
    * `/umbraco/config/appsettings-schema.json`
+   * If upgrading Umbraco Forms (from 9.* to 10.*) delete ‘/App_Plugins/UmbracoForms’
 4. Deploy the site how you normally would to your public facing environment.
 5. Start the site. At this point it will launch and upgrade the database, after which the site should become accessible and your upgrade is complete.
 6. Check the logs for any errors which may have occurred during the upgrade process.
```
alina-tincas, line 780:
```suggestion
4. If you are using Umbraco Forms, update your files and folders according to the [Upgrading - version specific](https://docs.umbraco.com/umbraco-forms/installation/version-specific) for version 10 article.
5. Deploy the site how you normally would to your public facing environment.
6. Start the site. At this point it will launch and upgrade the database, after which the site should become accessible and your upgrade is complete.
7. Check the logs for any errors which may have occurred during the upgrade process.
```

alina-tincas (changes requested):
Thank you for confirming and elaborating @Ric-Carey 🙌 

In this case I have made a suggestion on where I believe is the best place to add this 💪 

Can you please take a look at it and let me know what you think? Also, I was wondering if you would be up for making this same change in v12 documentation of CMS as well (to keep the documentation consists on all versions)?😊

alina-tincas:
Hi @Ric-Carey as I haven´t heard from you in a while, I will go ahead and commit my change and merge this in 👍 

If you have other opinions, feel free to create another PR or issue and we can take if from there 😊

alina-tincas (approved):

PR #5697, opened 2023-12-09
Color Picker - add explicit comment defining "Color" property

jacksorjacksor:
## Description

In Color Picker docs I'd suggest adding an explicit comment specifying that "Color" is a user-created property on the model.

Other editor examples use naming such as "myColorPicker" (example - [Toggle](https://docs.umbraco.com/umbraco-cms/v/10.latest-lts/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/true-false) uses "myCheckBox") to imply that the property was created by the user. This naming schema might also be more appropriate to use.

In my opinion the problem with just having Model.Color is that "Color" is quite generic, so it might not be obvious that there's the prerequisite steps of creating the property called "Color".

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

10-12

## Deadline (if relevant)
N/A, just a suggestion

### DIFF
--- a/10/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/color-picker.md
+++ b/10/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/color-picker.md
@@ -22,6 +22,7 @@ It's possible to add a label to use with the color.
 
 ```csharp
 @{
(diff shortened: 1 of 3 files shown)


alina-tincas (approved):
Hi @jacksorjacksor these are some awesome suggestions 🙌 

Thank you for this! Merging it in 😁 #h5YR

PR #5714, opened 2023-12-16
Update README.md

steiniofur:
Latest version of Umbraco.Cms.SqlCe is 9.5.4. to get Umbraco 13.0.1 to work I had to use Umbraco.Cms.Persistence.Sqlite instead.

Another thing unrelated, I had to include Umbraco.Cms.Core in the project when nuget restoring, as I got some errors when not included.

## Description

updated a package reference

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

### DIFF
--- a/13/umbraco-cms/fundamentals/setup/upgrading/README.md
+++ b/13/umbraco-cms/fundamentals/setup/upgrading/README.md
@@ -102,7 +102,7 @@ Add a package reference to your project by executing the `dotnet add package Umb
 Run `dotnet restore` to install the package.
 
 {% hint style="warning" %}
-If you are using SQL CE in your project you will need to run `dotnet add package Umbraco.Cms.SqlCe --version <VERSION>` too before running the `dotnet restore` command.
+If you are using SQL CE in your project you will need to run `dotnet add package Umbraco.Cms.Persistence.Sqlite --version <VERSION>` too before running the `dotnet restore` command.
 {% endhint %}
 
 When the command completes, open the **.csproj** file to make sure the package reference was updated:


alina-tincas:
Hi @steiniofur thank you for the PR 🙌 

We will take a look into it as soon as we can 💪

13/umbraco-cms/fundamentals/setup/upgrading/README.md
```diff
@@ -102,7 +102,7 @@ Add a package reference to your project by executing the `dotnet add package Umb
 Run `dotnet restore` to install the package.
 
 {% hint style="warning" %}
-If you are using SQL CE in your project you will need to run `dotnet add package Umbraco.Cms.SqlCe --version <VERSION>` too before running the `dotnet restore` command.
+If you are using SQL CE in your project you will need to run `dotnet add package Umbraco.Cms.Persistence.Sqlite --version <VERSION>` too before running the `dotnet restore` command.
 {% endhint %}
```
alina-tincas, line 106: This whole hint should be removed in this case, as it refers specifically to SQL CE. From v10 SQL CE is not used anymore: https://umbraco.com/products/umbraco-cms/umbraco-10/#:~:text=This%20replaces%20SQL%20CE%20which%20has%20reached%20End%20Of%20Life.%C2%A0

alina-tincas (changes requested):
Hi @steiniofur thank you for the PR 🙌 

After reviewing this PR, it is correct that since v10 SQL CE has been replaced with SQLite as per the following [article](https://umbraco.com/products/umbraco-cms/umbraco-10/#:~:text=This%20replaces%20SQL%20CE%20which%20has%20reached%20End%20Of%20Life.%C2%A0) 

In this case, the info note is specifically for SQL CE so in this case it would be best to just remove this instead not only for v13 of CMS but for v10 and V12 as well. Or another idea is to replace the text with something like "SQL CE has been replaced with SQLite since v10" 

Let me know what you think of the above and if you would be up for making this change to v10 and v12 as well? 😊

alina-tincas:
HI @steiniofur I was wondering if you have seen my latest message and if you are able to make the necessary changes? 😊

alina-tincas:
Hi @steiniofur as I have not heard from you in a while, I have updated the hint to the following:

> For v9: If you are using SQL CE in your project you will need to run `dotnet add package Umbraco.Cms.SqlCe --version <VERSION>` before running the `dotnet restore` command.
> From v10 SQL CE has been replaced with SQLite so a `dotnet restore` should be sufficient. If this is not working then you will need to run `dotnet add package Umbraco.Cms.Persistence.Sqlite --version <VERSION>` and then `dotnet restore`.

If you have any other suggestions, feel free to create another PR 😊 🙌

PR #5718, opened 2023-12-19
Update package-manifest.md

AaronSadlerUK:
## Description

Removed ~ from file path as it doesn't need it

## Type of suggestion

Fixed example

10, could be later too


alina-tincas:
Hi @AaronSadlerUK thank you for the PR 🙌 

We will take a look at it as soon as we can 😊

alina-tincas:
After reviewing this PR I can confirm the ~ is not needed 💪 I will make the same change for v12 and v13 as well 👍 

Thank you again for bringing this to our attention 🙌

PR #5720, opened 2023-12-19
Remove non breaking change

erikjanwestendorp:
## Description

In the Umbraco 13 documentation, there is still mention of the release candidate. I think this can now be changed to Umbraco 13? I also removed an item from the list because it is not a breaking change when upgrading from v12 to 13 (only from Umbraco 13 rc1 to 13 rc2). Maybe the list needs further adjustments?

https://docs.umbraco.com/umbraco-cms/fundamentals/setup/upgrading/version-specific#umbraco-13

See also comments in https://github.com/umbraco/Umbraco-CMS/pull/15113#issuecomment-1862681128

## Type of suggestion

Umbraco CMS

### DIFF
--- a/13/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
+++ b/13/umbraco-cms/fundamentals/setup/upgrading/version-specific/README.md
@@ -19,7 +19,7 @@ Use the [general upgrade guide](../) to complete the upgrade of your project.
 
 <summary>Umbraco 13</summary>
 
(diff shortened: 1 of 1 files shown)


alina-tincas (approved):

alina-tincas:
Hi @erikjanwestendorp thank you for the PR and good catch as well 💪 
I appreciate that you have linked to the CMS thread as well! #h5YR 

Merging this in 😊

Issue #5709, opened 2023-12-14
Delete code examples that update Program.cs or Startup.cs


erikjanwestendorp:
## What type of issue is it? (Choose one - delete the others)
Discussion/Wrong documentation

## What article/section is this about?
Several articles like: 
https://docs.umbraco.com/umbraco-cms/extending/section-trees/sections 
https://docs.umbraco.com/umbraco-cms/extending/content-apps 
etc.

## Describe the issue
In a discussion about [extension methods](https://github.com/umbraco/Umbraco-CMS/pull/15432#issuecomment-1852091417), the advice is given to make no changes to the `Startup.cs` or `Program.cs` files. I believe it is a good idea to remove the code examples from the docs that make changes to the `Startup.cs` or `Program.cs`. Instead, examples using an `IComposer` can be added.

alina-tincas:
Hi @erikjanwestendorp thank you for opening this issue! We are currently aware of it and have this is a task in our backlog 🙏

erikjanwestendorp:
@alina-tincas Thanks for your reply! 😄  As far as I know the advice in general is: Do not touch the 'Program.cs/Startup.cs' files and use an `ICompser` instead. My PRs aren't so much about changing `Startup.cs` to `Program.cs` or updating to the minimal hosting model. My PRs removes code samples that update `Startup.cs` or `Program.cs` in general, since that is the advice (These articles also provide examples with an `IComposer` already). I think it is good to remove these examples, otherwise the documentation goes against this advice 😄.

alina-tincas:
I understand! We will discuss this with our CMS developers on what is the best approach in this case, if those examples should indeed be removed. I will get back to you on this 👍

jonat123:
Hey Erik and happy new year 🎉

I want to let you know, that I have talked with the Lead developer for the CMS team, and he is fine with removing the examples that updates make changes to the `startup.cs` and `program.cs` files and replacing them with examples using composers instead(If they don't already have composer examples) 😄

There is only one thing to be aware of and that is examples that are using middleware. For Middleware it can also be done in a Composer, using the pipeline filter 😊

erikjanwestendorp:
@jonat123 Happy new year! Thanks for the update 😄! Good to hear that the examples can indeed be removed/replaced

sofietoft:
Hi @erikjanwestendorp ! 👋 
Long time to see! Hope you're doing well, and congratz on getting your MVP status renewed last week 😎 

[...]

PR #5735, opened 2023-12-25
Update content app (remove Program.cs or Startup.cs examples)

erikjanwestendorp:
## Description

Update content app (remove Program.cs or Startup.cs examples). I only removed the `Program.cs`/`Startup.cs` example and did not add any new explanation because there is already an example with an `IComposer` in this article.

Belongs to: 
https://github.com/umbraco/UmbracoDocs/issues/5709 

## Type of suggestion
* [x] Updated 'wrong' content

Umbraco 10, 11, 12 & 13


alina-tincas:
Hi @erikjanwestendorp thank you for the PR! 

We need to discuss about this PR with the CMS developers and when I get an answer on the conclusion I will get back to you here 💪

Reference/Notifications/SendingAllowedChildren-Notification/index.md
```diff
@@ -0,0 +1,5 @@
+---
+versionFrom: 9.0.0
+versionTo: 10.0.0
+---
```
alina-tincas, line 4: Is there a reason why this was added? 🙈 These versions to and from where only used on our old legacy documentation so it is not needed at all on these current documentations. This file is also part of our old documentation repository 😓

alina-tincas:
Hi @erikjanwestendorp I was wondering if you have seen my latest message and if perhaps it would be best to create a new PR?

erikjanwestendorp:
@alina-tincas Sorry for the delay in this one! 😅  Something went wrong here and I think it is better to create a new PR indeed. 👍

PR #5749, opened 2024-01-02
Correct installer syntax

sussexrick:
## Description

The switch should be `--install`. Using just `install` as the docs currently say results in "Error: Invalid option(s):  Umbraco.Templates"

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other


alina-tincas:
Hi @sussexrick thank you for the PR, however after testing this out on my CMD on Windows the correct syntax is as in our docs: **`dotnet new install Umbraco.Templates`**. If you add `--install `there is a warning that this is deprecated and to use without --.

Can you please confirm this as well?
image

sussexrick:
You're right, it's changed and I had an old version.

PR #5770, opened 2024-01-11
Update label-property-configuration.md

greystate:
Add missing tilde (~) in path to JS file, which was causing the example to not work.

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other

### DIFF
--- a/13/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/label-property-configuration.md
+++ b/13/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/label-property-configuration.md
@@ -57,7 +57,7 @@ An example `package.manifest` file is:
     "version": "1.0.0",
     "allowPackageTelemetry": false,
     "javascript": [
-        "/App_Plugins/MyFilters/myFilter.filter.js"
+        "~/App_Plugins/MyFilters/myFilter.filter.js"
     ]
 }
 ```


alina-tincas:
Hi @greystate thank you for the PR 🙌 

We will take a look at it as soon as we can 💪

alina-tincas:
Hi @greystate I have tried to reproduce this example and in my case it works both with and without ~ in the path. 

When you say it is not working for you, do you get any errors in console, or can you please elaborate how it is not working for you?

alina-tincas:
Hi @greystate I was wondering if you have seen my latest message and if you are able to elaborate on the setup in this case? 😊

greystate:
Hi @alina-tincas  - I've tried to see if I could replicate it, but haven't been able to either... when I initially had the problem, my filter didn't work, so I had a look in the Dev tools where there was a message along the lines of *"couldn't find the path XXX"*, so I added the tilde (~) and rebuilt the site, at which point it started working.

The site is currently on Umbraco 12.3.6, but as far as I can see it was on v12.3.5 at the time I added the PR (though I highly doubt that I'd hit a bug that would be fixed merely days after).

Thank you for testing the PR — feel free to close it as "mysterious fluke" 😁

/Chriztian

alina-tincas:
Thank you for elaborating Chriztian 🙌 

It is indeed mysterious 😁 

Then I am closing this one and wish you a nice week!

PR #5883, opened 2024-02-24
Updated examine-manager.md to reflect up-to-date steps for getting a specific searcher

abjerner:
## Description

The documentation for getting a specific searcher has been outdated for a while. The current example somewhat correctly illustrates how to get a searcher by a specific name.

The problem is that since Umbraco 9 (I think), there hasn't been an `ExternalSearcher` searcher, as you'd have to get the searcher via the `ExternalIndex` index instead.

The example would however still explain the correct way to get a custom searcher (aka not `ExternalSearcher`). I've updated the examples to reflect this.

I've also taken a few liberties that should hopefully improve the understanding of the examples:

- added comments to better explain the different parts
- used exact types instead of `var`. In Visual Studio or another IDE with intellisense, you can see what type `var` really is, but you can't see this in an example in the browser
- added null check on the `IPublishedContent` instance. Examine indexes can get out of sync, so a node in the index doesn't necessarily still exist in the content cache.
- updated some of the `if` statements to avoid nesting

Why reason for updating the examples was based on this issue for the CMS: https://github.com/umbraco/Umbraco-CMS/issues/14749

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

Umbraco 10-13

## Deadline (if relevant)

No deadline


alina-tincas:
Hi @abjerner thank you for the PR 🙌 

We will take a look at it as soon as we can 🙏

jonat123 (commented):
@abjerner Thanks for creating this PR 😄

The changes look good to me.

You are mentioning that the example has not really been working since Umbraco 9, should this then also be added to Umbraco 10 and 12? 🙂

jonat123:
@abjerner I see that you wrote that it is for 10-13, so I will merge in your PR and make sure that it is added to the other versions 😄

Once again thank for your PR #H5YR 😎

PR #5895, opened 2024-03-01
Update IIS rewrite rule examples

Jeavon:
Following on from discussions on #5662 I have update the rewrite rule example to remove the data centre specific match and add a check on the UMB_UCONTEXT cookie so that backoffice preview works when this rule is in place

### DIFF
--- a/13/umbraco-cms/reference/routing/iisrewriterules.md
+++ b/13/umbraco-cms/reference/routing/iisrewriterules.md
@@ -1,20 +1,20 @@
 # URL Rewrites in Umbraco
 
-With the release of Umbraco 9 and the change of the underlying web framework that is decoupled from the webserver, the way that you configure rewrites has changed as well.
+Since the release of Umbraco v9 and the change of the underlying web framework that is decoupled from the webserver, the way that you configure rewrites has changed as well.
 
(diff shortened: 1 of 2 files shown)


alina-tincas:
Hi @Jeavon thank you for the PR 🙌 

We will take a look at the PR as soon as we can 😊

jonat123:
@Jeavon Hey! 

I merged in another PR a bit ago, which has caused a conflict with your changes 🙈

Would you mind having a quick look at it and resolve it? 

To me, it looks like the only major difference is that you are adding this:    `<add input="{HTTP_HOST}" pattern="^(dev-|stage-)(.*)?\.umbraco\.io$" ignoreCase="true" negate="true" />
`
And in the PR i merged in it was changed to this: `<add input="{HTTP_HOST}" negate="true" pattern="^(dev|stage)\.(.*)\.umbraco\.io$" />
`

And to be fair, I am not sure which one it should be 🤔

Jeavon:
@jonat123 I would keep the hyphens, Umbraco cloud dev/stage domains are always prefixed with `dev- `or `stage-` e.g.  `https://dev-mysite.uksouth01.umbraco.io/`
I have merged up

jonat123 (approved):

PR #5896, opened 2024-03-01
Update routing:

erikjanwestendorp:
## Description

Update routing:

* [x] Remove 'various'
* [x] Remove 'very'
* [x] ~~Add 'AKA' to exceptions list~~

## Type of suggestion

* [x] Styleguide

Umbraco CMS v10, 12 & 13


alina-tincas:
Hi @erikjanwestendorp thank you for the PR! The CMS changes looks good, however I have talked over with the team if we should accept AKA as acronym and we have agreed that it is not obvious/clear what it means and should not be used ideally. Therefore this is something that we will not accept 🙏 

I have also looked it up on google and it is recommended for a documentation either 🙈 
image

erikjanwestendorp:
@alina-tincas Thanks for your review! Just updated the PR 😄

alina-tincas (approved):

alina-tincas:
Thank you for understanding and for making the necessary changes! All looks good so am merging this in 😊 💪

PR #6022, opened 2024-04-12
adding raiseEvents info

### LINKED ISSUE #5947: Outdated information for ContentService Notifications
Wrong documentation

[docs.umbraco.com/umbraco-cms/reference/notifications/contentservice-notifications](https://docs.umbraco.com/umbraco-cms/reference/notifications/contentservice-notifications)

## Describe the issue
The description documentation for some of the notification events refers to using the `raiseEvents` parameter if you need to suppress a specific notification. The Content Service not longer appears to have this parameter, for example doing something like `_contentService.Save(contentNode)`, doesn't have a `raiseEvents` parameter.

I'm not actually sure how you are meant to do this any more so can't update this documentation myself.

alina-tincas:
## Description

adding info what happened to raiseEvents  from PR https://github.com/umbraco/Umbraco-CMS/pull/10655 

also, it fixes #5947 

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

cms v10, 12, 13 and v14

## Deadline (if relevant)

anytime

### DIFF
--- a/10/umbraco-cms/reference/notifications/contentservice-notifications.md
+++ b/10/umbraco-cms/reference/notifications/contentservice-notifications.md
@@ -173,12 +173,54 @@ bool IsCultureEdited(string culture);
 bool IsCulturePublished(string culture);
 ```
 
-### What happened to Creating and Created events?
+<details>
+
+<summary>What happened to Creating and Created events?</summary>
 
 Both the ContentService.Creating and ContentService.Created events were removed, and therefore never moved to notifications. Why? Because these events were not guaranteed to trigger and therefore should not be used. This is because these events would only trigger when the ContentService.CreateContent method was used which is an entirely optional way to create content entities. It is also possible to construct a new content item - which is generally the preferred and consistent way - and therefore the Creating/Created events would not execute when constructing content that way.
 
 Furthermore, there was no reason to listen to the Creating/Created events. They were misleading since they didn't trigger before and after the entity persisted. They are triggered inside the CreateContent method which never persists the entity, it constructs a new content object.
 
-#### What do we use instead?
+**What do we use instead?**
 
 The ContentSavingNotification and ContentSavedNotification will always be published before and after an entity has been persisted. You can determine if an entity is brand new in either of those notifications. In the Saving notification - before the entity is persisted - you can check the entity's HasIdentity property which will be 'false' if it is brand new. In the Saved notification you can [check to see if the entity 'remembers being dirty'](determining-new-entity.md)
+
+</details>
+
+<details>
+
+<summary>What happened to `raiseEvent` method parameters?</summary>
+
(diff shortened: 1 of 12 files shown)


10/umbraco-cms/reference/notifications/contentservice-notifications.md
```diff
@@ -173,12 +173,54 @@ bool IsCultureEdited(string culture);
...
+
+```csharp
+using (IScope scope = ScopeProvider.CreateScope(autoComplete: true))
+using (IDisposable _ = scope.Notifications.Supress())
+{
+    // TODO: Calls to service methods here will not have notifications
+}
+```
+
+Child scope will inherit the parent Scope's notification object which means if a parent scope has notifications suppressed, then so does the child scope. You cannot call Supress() more than once for the same outer scope instance else an exception will be thrown. This is to ensure that you cannot un-suppress notifications at a child level for an outer scope and ensures that suppressing events is a very explicit thing to do.
```
alina-tincas, line 220:
```suggestion
Child scope will inherit the parent Scope's notification object which means if a parent scope has notifications suppressed, then so does the child scope. You cannot call Supress() more than once for the same outer scope instance else an exception will be thrown. This ensures that you cannot un-suppress notifications at a child level for an outer scope. It also ensures that suppressing events is an explicit thing to do.
```

10/umbraco-cms/reference/notifications/contentservice-notifications.md
```diff
@@ -173,12 +173,54 @@ bool IsCultureEdited(string culture);
...
+using (IScope scope = ScopeProvider.CreateScope(autoComplete: true))
+using (IDisposable _ = scope.Notifications.Supress())
+{
+    // TODO: Calls to service methods here will not have notifications
+}
+```
+
+Child scope will inherit the parent Scope's notification object which means if a parent scope has notifications suppressed, then so does the child scope. You cannot call Supress() more than once for the same outer scope instance else an exception will be thrown. This ensures that you cannot un-suppress notifications at a child level for an outer scope. It also ensures that suppressing events is an explicit thing to do.
+
+Why would you want to suppress events? The main reason for ever doing this would be performance for bulk operations. The caller will need to be aware that suppressing events will lead to an inconsistent content cache state (if notifications are suppressed for content or media services). This is because notifications are used by things such as NuCache to populate the cmsContentNu table and populate the content caches. They are also used to populate the Examine indexes.
```
alina-tincas, line 222:
```suggestion
**Why would one want to suppress events?** The main reason for ever doing this would be performance for bulk operations. The callers hould be aware that suppressing events will lead to an inconsistent content cache state (if notifications are suppressed for content or media services). This is because notifications are used by NuCache to populate the cmsContentNu table and populate the content caches. They are also used to populate the Examine indexes.
```

PR #6101, opened 2024-05-21
Adding the DefaultUILanguage in appsettings.json file

eshanrnh:
## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)

### DIFF
--- a/10/umbraco-cms/tutorials/multilanguage-setup.md
+++ b/10/umbraco-cms/tutorials/multilanguage-setup.md
@@ -36,7 +36,7 @@ We can add multiple languages depending on our website requirements. In the prev
 2. For English Variant:
    * Select **English (United States)** from the drop-down list.
    * Click **Save**.
-3.  For German Variant:
+3. For German Variant:
 
     * Select **German** from the drop-down list.
     * Toggle **Mandatory Language** option.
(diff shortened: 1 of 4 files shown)


10/umbraco-cms/tutorials/multilanguage-setup.md
```diff
@@ -51,10 +51,28 @@ To change the default language of a website:
...
+        "Global": {
+            "DefaultUILanguage": "fr"
+            }
+          }
+        }
+    }
+
+    ```
+
+7. Save the `appsettings.json` file.
```
jonat123, line 75: To be honest I don't think this is necessary to have here 🤔

This section is just about adding multiple languages in the settings section, which is used for having different languages displayed on the frontend.

However, the DefaultUILanguage is for changing the language used in the backoffice, so adding that setting should not be relevant in this context 😊

I hope it makes sense!

eshanrnh: @jonat123 I agree, it's slightly a misfit for this article. 🙈

Would it make sense to add the info in the [Language Variants](https://docs.umbraco.com/umbraco-cms/fundamentals/backoffice/variants) article as a section at the end? 🤔

I think its important to have this info somewhere in the docs, though..

jonat123: I think the problem is that this is not related to Language variants 😊

Also this is already documented here: https://docs.umbraco.com/umbraco-cms/reference/configuration/globalsettings#default-ui-language :-)

eshanrnh: Ummm okay.. If it's not needed, I'll close this PR then 💪

PR #6164, opened 2024-06-05
Updating the wording

jesperlange:
It's not intuitive to say to disable this feature, then set it to false, when it by default is set to false.

If they want to enable "DisableRelationTracking" they should set this to true

## Description

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)


eshanrnh (commented):
Thanks for the PR, @jesperlange 💪 It indeed makes sense to write it this way 🙌

PR #6169, opened 2024-06-06
Static outbound IPs vary per region

pijemcolu:
## Description

Static outbound IPs vary per region and are not the same across all regions.

## Type of suggestion

Umbraco Cloud


eshanrnh:
Thanks, @pijemcolu for the PR. We will review it as soon as we can. 🙌

jonat123 (approved):
@pijemcolu LGTM 💪

PR #6190, opened 2024-06-12
Update limit-orderline-quantity.md

Wiggee11:
## Description

`_productService.GetProductStock()` does not have a method accepting a single argument, so updated to use `_productService.GetProductStock(evt.Order.StoreId, productReference)` - include `storeId`.

AddProduct handlers was also checking orderlines for quantity but orderlines are empty, and checking stock value incorrectly.

Umbraco 13.3.0
Umbraco.Commerce 13.1.1


Wiggee11:
When adding a product to a new order/cart, the OrderLines are empty so you will need to use the `evt.Quantity` value.

If an order/cart already exists and you are adding a product then the OrderLines do exist.

eshanrnh:
Thanks, @Wiggee11 for the PR 🙌 We will take a look as soon as we can 🙂

eshanrnh (approved):

PR #6197, opened 2024-06-17
Update configuring-umbraco.md

MMasey:
## Description

I've added a note regarding the requirements of a trailing slash if using a remove trailing slash rewrite rule. I've used a footnote to add the info, but if it should be displayed differently that's no problem.

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

### DIFF
--- a/commerce-add-ons/payment-providers/stripe/configuring-umbraco.md
+++ b/commerce-add-ons/payment-providers/stripe/configuring-umbraco.md
@@ -26,9 +26,9 @@ The following steps are handled within the payment method editor in the Umbraco 
 
 | Name                                    | Description                                                                                                                                                                    |
 | --------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| Continue URL                            | The URL of the page to navigate to after payment is successful - e.g. `/confirmation/`                                                                                         |
-| Cancel URL                              | The URL of the page to navigate to if the customer cancels the payment - e.g. `/cart/`                                                                                         |
-| Error URL                               | The URL of the page to navigate to if there is an error with the payment - e.g. `/error/`                                                                                      |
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @MMasey !
I'll forward it to Matt for a review.

Also, I'm not familiar with the footnote format, `url-note`. I'm also not seeing anything from your PR when previewing it on the docs.
How does it work exactly? 🤔 😅

MMasey:
Hey @sofietoft, I wonder if the footnote feature doesn't work in the docs system. It should be adding a little bit of text to the bottom of the article, which is referenced further up, which is what the `url-note` is for. It's the same as an anchor tag I guess.

I may need to refactor this so it's displayed differently, although speaking with Matt, it may be that it's actually a bug that needs to be fixed.

sofietoft:
It might very well be something that doesn't work on the GitBook platform.
Let me run some tests and get back to you.

GitBook has another way of handling information like this: https://docs.gitbook.com/content-editor/editing-content/inline#annotations
Perhaps annotations could be used here instead?

sofietoft:
Hi again, Mike 👋 

I've done some testing with this, and the "normal" URL note format doesn't show up on the GitBook platform or on the live docs site I'm afraid.

The annotations I mentioned, though, are added in a very similar way:

```
* [Chrome (Latest)](#user-content-fn-1)[^1]
* Edge (Chromium)
* Firefox (Latest)
* [Safari (Latest)](#user-content-fn-2)[^2]

...
...

[^1]: Annotation tests
[^2]: This is just a tiny small test.
```

Let me know if this could work for the info you want to add here!

sofietoft:
Hi @MMasey ! 👋 

It's been a while since we've heard back from you, so I'll be closing this for now.
Please feel free to get back in touch, if you still think this is something we should consider for the docs 💪

PR #6206, opened 2024-06-21
Update multisite-setup.md

Jette94:
Made an additional comment, as it is important to configure hostnames and cultures, or the user might experience 404 errors on the pages not configured.

## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/14/umbraco-cms/tutorials/multisite-setup.md
+++ b/14/umbraco-cms/tutorials/multisite-setup.md
@@ -30,7 +30,7 @@ Keep in mind the [hostnames have to be configured in a specific way.](https://do
 
 ## Mapping the hostnames to individual websites/root nodes
 
-At this point, you should have several root nodes, where each is a separate website. To map the hostnames to individual root nodes:
+At this point, you should have several root nodes, where each is a separate website. By design, hostnames and cultures should be configured for each page, or you will experience 404 errors on the website. To map the hostnames to individual root nodes:
 
 1. Go to the **Content** section.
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @Jette94 ! 💪 

I do believe it used to be enough to set this on the Root node of the site.
That is not longer the case then?

Also, could you let me know which version of Umbraco this is related?

14/umbraco-cms/tutorials/multisite-setup.md
```diff
@@ -30,7 +30,7 @@ Keep in mind the [hostnames have to be configured in a specific way.](https://do
 
 ## Mapping the hostnames to individual websites/root nodes
 
-At this point, you should have several root nodes, where each is a separate website. To map the hostnames to individual root nodes:
+At this point, you should have several root nodes, where each is a separate website. By design, hostnames and cultures should be configured for each page, or you will experience 404 errors on the website. To map the hostnames to individual root nodes:
```
sofietoft, line 33: Instead of adding a note here, I have another suggestion:

How about we add a step at the bottom of this list saying something like "Repeat steps 2-6" for each of your root nodes"?
What do you think?

sofietoft (changes requested):

Jette94:
Sharp, Sofie - looks good 💪

sofietoft:
I updated the article based on our discussion.
I'll merge it in as soon as the checks clear.

PR #6226, opened 2024-06-28
V14: Extend the Package Manifest article with all available types

iOvergaard:
## Description

Added all extension types and other properties to the reference article for Package Manifest.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

14

## Deadline (if relevant)

ASAP

### DIFF
--- a/.github/styles/UmbracoDocs/Acronyms.yml
+++ b/.github/styles/UmbracoDocs/Acronyms.yml
@@ -83,5 +83,4 @@ exceptions:
     - SVG
     - DHL
     - UUI
-    - RCL
     - MDN
--- a/.github/styles/UmbracoDocs/Acronyms.yml
+++ b/.github/styles/UmbracoDocs/Acronyms.yml
@@ -83,3 +83,5 @@ exceptions:
     - SVG
     - DHL
     - UUI
+    - RCL
+    - MDN
--- a/14/umbraco-cms/extending/package-manifest.md
+++ b/14/umbraco-cms/extending/package-manifest.md
@@ -158,7 +158,7 @@ These are the current types of UI Extensions:
 | healthCheck | A health check is a check that can be added to the health check dashboard. Read more about the backend side of [Health Checks](../reference/configuration/healthchecks.md). |
 | icons | An icon is a set of icons that can be added to the icon picker. |
 | localization | A localization is a set of translations that can be added to the localization service. Read more about [Localization](./language-files/ui-localization.md) in the UI. |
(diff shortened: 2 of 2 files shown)


.github/styles/UmbracoDocs/Acronyms.yml
```diff
@@ -83,3 +83,5 @@ exceptions:
     - SVG
     - DHL
     - UUI
+    - RCL
```
sofietoft, line 86: I think it makes sense to add "MDN", as it sounds to me like this is the name now, and not an actual acronym.

But let's take out RCL here as well!
```suggestion
```

14/umbraco-cms/extending/package-manifest.md
```diff
@@ -44,40 +38,173 @@ Before Umbraco 14, the manifest was declared in a `package.manifest` file instea
...
+
+```javascript
+import { MyService } from 'mypackage/services';
+```
+
+{% endcode %}
+
+Umbraco supports the current specification of the property as outlined on MDN Web Docs: [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap).
+
+### Extensions
```
sofietoft, line 127: As there is so much text for each of these extensions, do you think it would make sense to add it to a table instead?

I think it's a little hard to get an overview of these lists: 
image

At least in a table, the extensions would be in one column, and it would be a bit easier to find what you're looking for.

What do you think?

sofietoft (commented):
Looks good @iOvergaard ! 💪 

Just a couple of comments - one reg the RCL acronym and one reg the text-heavy lists!

14/umbraco-cms/extending/package-manifest.md
iOvergaard: Yes, it makes sense to use a table instead. As I only saw the raw markdown, a table looked more confusing than a list, but it makes sense on GitBook to do a table. I'll see if I can be clever with some search/replace.

.github/styles/UmbracoDocs/Acronyms.yml
iOvergaard: Yes absolutely. MDN used to be "Mozilla Developer Network" or something to that effect, but now it has become their de facto name "MDN Web Docs".

iOvergaard:
Converted it to a table now. What do you think, @sofietoft?

sofietoft (approved):
The tables look great @iOvergaard !

image

Except for a small bug, but I'll see about fixing that in GitBook when we've merged this one!

Green lights from me. Ready to merge?

iOvergaard:
Go ahead!

PR #6228, opened 2024-07-01
Create property dataset README.md

nielslyngsoe:
## Description

Example of Property Datasets

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)


sofietoft:
Thanks for the PR @nielslyngsoe ! 💪 

In order for new articles to be part of the docs on GitBook, the articles need to be references in the SUMMARY.md file.
You can find it in the root of the 14/umbraco-cms folder.

Also, would it make sense to add some more context to the article, regarding which file this code would be added into, in which situations you will be using, etc.
What do you think?

Issue #6204, opened 2024-06-20
crosssitescriptingprotection is still using startup file


SSmale:
## What type of issue is it? (Choose one - delete the others)
Wrong documentation

## What article/section is this about?
13/umbraco-cms/extending/health-check/guides/crosssitescriptingprotection.md

## Describe the issue
This is still using the the startup.cs file which is not present in v13 or 14

sofietoft:
Thanks for reporting this @SSmale ! 💪 

I'll make sure we have someone look at it as soon as possible!

PR #6232, opened 2024-07-02
Update healthchecks for v13

nikolajlauridsen:
## Description

Updated the health checks examples to use the new `Program.cs`

I opted to only update the v13 versions, since v13/14 should be identical, this way you only have to fix-up once 😄 

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

v13/14


sofietoft (approved):
Thanks for helping out with this @nikolajlauridsen ! 💪🥳 

I'll make sure the same articles will be updated for Umbraco 14.

This fixes: #6204

PR #6279, opened 2024-07-19
Update loggingsettings 

erikjanwestendorp:
## Description

* [x]  Write shorter sentences

Is there a reason why the v12 & 13 version of the article doesn't provide an example of the JSON?

## Type of suggestion

* [x] Styleguide

Umbraco CMS v10


eshanrnh (approved):
Thanks for the PR, @erikjanwestendorp 🙌 The changes look good 💪

PR #6340, opened 2024-08-12
Update weight description for dashboard as highest comes first in v14

Jeavon:
This is the opposite of how it works in versions <14


sofietoft:
Thanks for updating this @Jeavon ! 
I'll make sure it's merged 💪

PR #6361, opened 2024-08-19
Update custom extension conditions

benstigsen:
## Description
After having tried to create a custom condition for my extension, I ran into several issues following the documentation.

There were minor errors such as `true` being written as `strue`. But there were also some major issues, like things being imported which weren't used, extending the wrong classes and in general unclear ways to do stuff.

I ran into an issue requiring many hours of debugging, reading documentation and trial-and-error, where the custom condition was extending the wrong class, leading to something like `args.host` being `undefined`, resulting in the entire extension never being permitted, never reaching the timeout.

It also wasn't being shown how to make use of `match`, even though a condition config type was created to allow for this.

## Type of suggestion

* [x] Typo/grammar fix
* [x] New content
* [ ] Other

Umbraco 14

### DIFF
--- a/14/umbraco-cms/customizing/section-trees/sections/README.md
+++ b/14/umbraco-cms/customizing/section-trees/sections/README.md
@@ -20,7 +20,7 @@ You can create your own sections to extend Umbraco with room for more features.
 
 ### **Manifests**
 
-When creating a new section it's recommended to use a [Entry Point](../../../extending/backoffice-setup/extension-registry/entry-point.md)-extension in your [Umbraco Package Manifest](../../package-manifest.md). This is to get better control over all the additional extensions required for the new section.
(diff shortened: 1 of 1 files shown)


benstigsen:
Hmm. I'm not a big fan of using `any` for the `args` parameter.
I'll come back to it later.

eshanrnh:
Thanks for the PR, @benstigsen 🙌 We will review it as soon as we can. 💪

sofietoft:
Hi @benstigsen !

Apologies for the wait on this one.
I've run it through our developers, and everything seems to be checking out 💪 

I see though, that there are some merge conflicts - most likely related to us restructuring the backoffice articles a couple of weeks back 😅 
The article has been moved to here: https://github.com/umbraco/UmbracoDocs/blob/main/14/umbraco-cms/customizing/extending-overview/extension-types/condition.md 

Are you up for making a new PR with the same changes? 🙈 
I fully understand if you're not - just let me know, and I'll add the changes directly!

benstigsen:
@sofietoft the new pull-request can be found here: https://github.com/umbraco/UmbracoDocs/pull/6424

PR #6367, opened 2024-08-21
Update import-with-migrations.md

peaxe:
## Description

The using seems to perhaps have gone a little off which is natural with time when refactoring of class libraries  happens.

## Type of suggestion

* [ ] Typo/grammar fix
* [x ] Updated outdated content
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/13/umbraco-deploy/deployment-workflow/import-with-migrations.md
+++ b/13/umbraco-deploy/deployment-workflow/import-with-migrations.md
@@ -68,8 +68,9 @@ You can do this via a composer, as in the following example. Here we register tw
 
 ```csharp
 using Umbraco.Cms.Core.Composing;
-using Umbraco.Deploy.Core.Migrators;
+using Umbraco.Cms.Core.DependencyInjection;
 using Umbraco.Deploy.Infrastructure.Migrators;
+using Umbraco.Extensions;
 
 internal class ArtifactMigratorsComposer : IComposer
 {


sofietoft:
Thanks for the PR @peaxe ! 💪 

I'll make sure we have it reviewed as soon as possible.

13/umbraco-deploy/deployment-workflow/import-with-migrations.md
```diff
@@ -68,8 +68,9 @@ You can do this via a composer, as in the following example. Here we register tw
 
 ```csharp
 using Umbraco.Cms.Core.Composing;
-using Umbraco.Deploy.Core.Migrators;
+using Umbraco.Cms.Core.DependencyInjection;
```
Frost117, line 71: When I create the class and using this example, they are set as not being used.

@peaxe am I doing this correctly or do we even need to use the Umbraco.Cms.Core.DependencyInjection Umbraco.Extensions for the composer?

peaxe: Hi @Frost117 

Thank you for the reply.

Not sure that I answer your question correctly.

I have created the composer in a class library.

This is what happens if I comment out 
`using Umbraco.Extensions;`
image

This it what happens if I comment out
`using Umbraco.Cms.Core.DependencyInjection;`
image

Frost117: Ah gotcha - I missed the part that it was in a class library. Let me check in with the deploy team and I'll get back to ya :D

peaxe: No worries.
Further more I did not provide sufficient data to cover this scope properly for you to understand that I have been testing this with a class library. Sorry for that. 😊

ronaldbarendse: These 2 using directives are added as implicit usings in the `Umbraco.Cms.Targets` package (see [CMS source](https://github.com/umbraco/Umbraco-CMS/blob/385a5345b1dcd1fbc20afb854cfeff1c36f52210/src/Umbraco.Cms.Targets/buildTransitive/Umbraco.Cms.Targets.props#L8-L11)).

If a project (like a class library) doesn't reference this package, you need to explicitly add them. This is similar to the .NET implicit usings, which depend on the SDK the project is using (see [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/overview#implicit-using-directives)).

This applies to all examples in our documentation and it might indeed be better to ensure all code explicitly lists the required usings by opting-out of implicit usings by adding [`<ImplicitUsings>disable</ImplicitUsings>` to the project](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#implicitusings), before copying the code. Something for the docs team to consider I guess!

sofietoft: Thanks for your comments on this.
I'm not entirely sure I understand what it is we need to consider here.

Are we talking about adding more usings to the snippets than needed, or are we talking about documenting the "ImplicitUsings"? 🤔

peaxe: Sorry if I'm captain obvious, nevertheless I will mention @ronaldbarendse as I think that he is best suited to answer that question. (And thank you all for your time in trying to figure out if this PR makes sense or not.)

ronaldbarendse: Let me clarify 😄 The changes in this PR are valid and do make sense, as:
- `Umbraco.Deploy.Core.Migrators` wasn't used, so it can indeed be removed;
- `Umbraco.Cms.Core.DependencyInjection` and `Umbraco.Extensions` are implicitly added when using the Umbraco CMS template (that has `<ImplicitUsings>enable</ImplicitUsings>` and references `Umbraco.Cms.Targets`), so explicitly adding them also makes the example valid for when this isn't the case (like when it's added to a class library).

However, we do assume the examples are used in a 'default setup', so that's why most (if not all) examples don't contain these 2 usings... It doesn't hurt having them either, so that's why I tagged @sofietoft to see if we need to consider adding them or otherwise maybe have some general documentation about this (mentioning Umbraco adds these 2 usings by default and you need to manually add them if you don't have implicit usings enabled)...

sofietoft:
Thanks a lot for helping us review this one @ronaldbarendse and @Frost117 💪 

Let's keep the 2 "extra" using, as Ronald says, that will make the code valid even when the "ImplicitUsings" is disabled!
It's better to add too much info, than too little, right? 😉 

With that, I also think we can say that this PR is ready to be merged.

Thanks again @peaxe, for taking the time to help us improve the docs - MUCH appreciated ⭐

PR #6369, opened 2024-08-22
Fix legacy ProviderName in sqlite examples

glitchedmob:
## Description

While upgrading an Umbraco site using Sqlite from v13 to v14 I kept getting the following error: `System.ArgumentException: The specified invariant name 'Microsoft.Data.SQLite' wasn't found in the list of registered .NET Data Providers.`

I was using the exact values from the docs for `umbracoDbDSN_ProviderName` which made this error very confusing. 

After some source diving and looking at the git history, I found umbraco/Umbraco-CMS#15999 which removed support for the format mentioned in the docs. 

This PR updates the examples for version 14 to use the correct format. It may make sense to update the docs for versions 12 and 13 as well as it seems this format has been deprecated for a while.

It also may be worth updating the Breaking Changes section of the [Version Specific Upgrades page](https://docs.umbraco.com/umbraco-cms/fundamentals/setup/upgrading/version-specific). 

I'm happy to add those suggested changes to this PR if the maintainers would prefer that.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

Umbraco.Cms 14

## Deadline (if relevant)

No Deadline

### DIFF
--- a/14/umbraco-cms/fundamentals/setup/install/unattended-install.md
+++ b/14/umbraco-cms/fundamentals/setup/install/unattended-install.md
@@ -48,7 +48,7 @@ It is recommended that you make use of the values shown below for the `Cache`, `
 {
   "ConnectionStrings": {
     "umbracoDbDSN": "Data Source=|DataDirectory|/Umbraco.sqlite.db;Cache=Private;Foreign Keys=True;Pooling=True",
-    "umbracoDbDSN_ProviderName": "Microsoft.Data.SQLite"
+    "umbracoDbDSN_ProviderName": "Microsoft.Data.Sqlite"
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


sofietoft:
Thanks for the PR @glitchedmob ! Good catch 💪 

It would be great to have this added as a bullet to the version-specific upgrades.
If you're up for it, please don't hold back adding it to the PR 😄 

Otherwise, I'll make sure we add it to our internal backlog in the docs team 🙌

glitchedmob:
@sofietoft Added this to the breaking changes in v14. Let me know if there are any adjustments I should make! (also feel free to take over and make the changes yourself if that's easier)

sofietoft (approved):
Looks great @glitchedmob ! 💪 Thanks!
I took the libery to make some minor adjustments directly.

Regarding adding info about it to versions 12 and 13, I don't think it necessary.
It's only in Umbraco 14 that the obsolete provider name was removed and becomes a problem.

PR #6373, opened 2024-08-23
V14: Update the example for adding settings to a Custom View

nielslyngsoe:
## Description

Ive updated the example of using settings data in a custom view.
And removed adding stylesheet as that does not make sense any more. With Web Components that is a web component feature that I would say is out of scope for our responsibilities. Any one can search "How to append stylesheet to web component" and find their solution.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/14/umbraco-cms/tutorials/creating-custom-views-for-blocklist.md
+++ b/14/umbraco-cms/tutorials/creating-custom-views-for-blocklist.md
@@ -150,7 +150,7 @@ To add content to the blocks:
 4. Enter the **Name** and **Details** in the Feature window. You will notice you can view the content as you type. This is because we have enabled the **Live editing** mode.
 5. Click **Confirm**.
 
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @nielslyngsoe !

I understand that since appending a stylesheet is now a web component thing, we shouldn't necessarily document it.
Will it be possible to use the custom view without appending a stylesheet?
Otherwise, we should mention that this needs to be done.

nielslyngsoe:
Hi @sofietoft 

Yes, Web Components can have embed styles, which for some makes it irrelevant to append a stylesheet.
But some might want to implement a separate stylesheet in their Web Component. But how they do so is not our business any longer. Cause we do not have a picker, or an uploader field or anything. Its entirely up to their code.

:-)

sofietoft:
Excellent, makes sense. Thanks for making this more clear.

PR #6393, opened 2024-09-04
Update stricttransportsecurityheader.md

Kull1z:
Added info so that you can set a timespan for HSTS.

## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [x ] Updated outdated content
* [x ] New content
* [x ] Updates related to a new version
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/14/umbraco-cms/extending/health-check/guides/stricttransportsecurityheader.md
+++ b/14/umbraco-cms/extending/health-check/guides/stricttransportsecurityheader.md
@@ -32,7 +32,8 @@ else
     //...
 }
 ```
-Add this to Program.cs to be able to set a timespan for HSTS, preferbly six months.
+
+It is possible to configure a timespan for the HSTS, preferbly six months. This can be done by adding the following to the `Program.cs` file:
 
 ```csharp
 builder.Services.AddHsts(options =>
--- a/14/umbraco-cms/extending/health-check/guides/stricttransportsecurityheader.md
+++ b/14/umbraco-cms/extending/health-check/guides/stricttransportsecurityheader.md
@@ -32,6 +32,17 @@ else
     //...
 }
 ```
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @Kull1z ! 💪 

I'll make sure we get it reviewed as soon as possible!

Kull1z (commented):
Looks good!

sofietoft:
Hi again @Kull1z ! 👋 

After talking this over with one of our developers, we agreed that this is more a general ASP.Core thing, and therefore not something we want to add to the Umbraco docs.
The main reason for this, is that we want to avoid having to maintain documentation that is already documented (and maintained) elsewhere.

So, as this is still valuable information, how about we link to the Microsoft docs for this instead: https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-8.0&tabs=visual-studio%2Clinux-ubuntu#http-strict-transport-security-protocol-hsts 

And then write something like this instead:

It is possible to configure a timespan for the HSTS, preferably six months. This can be done by adding and new builder to the `Program.cs` file. Learn more in the [official Microsoft Documentation](https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-8.0&tabs=visual-studio%2Clinux-ubuntu#http-strict-transport-security-protocol-hsts ).

What do you think?

sofietoft:
Hi @Kull1z !

It's been a while since I've back from you, so I've taken the liberty to update the PR myself.
Instead of the code snippet, I added a link to the official ASP.NET docs about this topic.
That way we don't risk adding code to our docs, that might become obsolete over time 💪 

Thanks again for helping up making the docs better! MUCH appreciated.

I'll get this merged.

PR #6406, opened 2024-09-10
Update an image in manualupgrade.md

georgebid:
## Description

I noticed that the image was small and on the same line as step 4, so quickly updated it to fix this. I swapped it to use `HTML <figure> with a  tag `as this seems to be done in many other places.

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)

### DIFF
--- a/14/umbraco-forms/upgrading/manualupgrade.md
+++ b/14/umbraco-forms/upgrading/manualupgrade.md
@@ -22,7 +22,8 @@ To get the latest version of Umbraco Forms, you can upgrade using:
 2. Select **Umbraco.Forms**.
 3. Select the latest version from the **Version** drop-down and click **Install**.
 
-  ![NuGet Package Manager](../installation/images/Manage_packages_v14.png)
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks for the PR @georgebid ! 💪 
That image did look a bit misplaced, so thanks a bunch for taking the time to fix it.

Whether the markdown or the HTML format is used, does not really matter actually.
A little history, if you're interested:
The markdown format is what we've always used in the docs, so this is kinda what we default to doing in the docs team.
The HTML format with the <figure> and  tags, is what's added to the files, when we add images directly through the GitBook platform 😄 

So yeah, both methods are valid. We should probably at some point make this consistent throughout the docs.. but you know, recourses and all!

Anywho! Sorry for the wall of text!
Thanks again for the PR - I'll get it merged 💪

georgebid:
Ah, interesting. That's good to know as I haven't contributed to the docs before! Thanks for merging 😄

PR #6413, opened 2024-09-12
Updated migration IDs for Forms 10+

AndyButland:
## Description

Updated migration IDs for Forms 10+.

Please note I modified the layout to use markdown rather than HTML (as it's much easier to edit in a text file).  I've included migrations from 10+, which seems reasonable given Forms 8 is out of support.

You will want to tidy up and remove the WIP note if you are happy with it.  And probably copy into the 10 and 13 folders, removing the migrations that are for versions subsequent to the one you copy into.

## Type of suggestion

* [ ] Typo/grammar fix
* [X] New content
* [ ] Other

Forms 10+

## Deadline (if relevant)

Any time.

### DIFF
--- a/14/umbraco-forms/upgrading/version-specific/migration-ids.md
+++ b/14/umbraco-forms/upgrading/version-specific/migration-ids.md
@@ -10,30 +10,21 @@ A unique **migration ID** is generated for each Umbraco Forms upgrade that requi
 This article is a work in progress. Please do not use the information in the article as long as this note is present.
 {% endhint %}
 
-## Version 14
+| Migration ID                         | Introduced In Version | Description                                                                    |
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks @AndyButland !
Using the Markdown format is WAY better - it's just GitBook that disagrees 🙈 

I'll merge it, and add to versions 10, 13 and 15 as well 💪 

Should we continue to keep this article hidden in the navigation structure?

AndyButland:
I think you could add it as the last section under "Upgrading".    Don't think it clutters too much and I expect some people who may find it useful will discover it that way.

sofietoft:
Yes, I think that makes sense.
I'll unhide the articles :)

PR #6421, opened 2024-09-16
Update configuration.md

bjarnef:
## Description

Adding Umbraco UI Builder should be added to `Startup.cs` - not `Program.cs`.

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/13/umbraco-ui-builder/getting-started/configuration.md
+++ b/13/umbraco-ui-builder/getting-started/configuration.md
@@ -8,7 +8,9 @@ Umbraco UI Builder can be configured directly via the `AddUIBuilder` extension m
 
 ## AddUIBuilder
 
-To configure Umbraco UI Builder via the `AddUIBuilder` extension method, You can look in the `Startup.cs` file in the root of your web project. From within this file, before the call to `AddComposers()` we can add our `AddUIBuilder` configuration.
+To configure Umbraco UI Builder via the `AddUIBuilder` extension method, You can look in the `Program.cs` file at the root of your web project. From within this file, before the call to `AddComposers()` we can add our `AddUIBuilder` configuration.
+
+Alternatively, if your project is upgraded from earlier versions, you can add the configuration to the `Startup.cs` file.
 
 ```csharp
 builder.CreateUmbracoBuilder()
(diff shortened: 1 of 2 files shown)


sofietoft:
Thanks for the PR @bjarnef ! 💪 

If I'm not mistaken, the Startup.cs file is not part of Umbraco 14. So, perhaps this should be added to a different version of the docs for this product? 🤔

bjarnef:
@sofietoft yes, it seems you are correct.

I updated it in v13 instead, in v11 and v12 it did use `Startup.cs`.
However as `Startup.cs` can be used in existing/upgraded projects, I am in double if a clean v13 install use the new minimal API including only `Program.cs`.

sofietoft:
Yes, I believe you're right that the Startup.cs file can still be used in 13, so it make sense to update that article!
Instead of replacing `Program.cs` here, perhaps instead we could add a line saying something like: 

> Alternatively, if your project is upgraded from earlier versions, you can add the configuration to the `Startup.cs` file.

13/umbraco-ui-builder/getting-started/configuration.md
```diff
@@ -8,7 +8,7 @@ Umbraco UI Builder can be configured directly via the `AddUIBuilder` extension m
 
 ## AddUIBuilder
 
-To configure Umbraco UI Builder via the `AddUIBuilder` extension method, You can look in the `Program.cs` file in the root of your web project. From within this file, before the call to `AddComposers()` we can add our `AddUIBuilder` configuration.
+To configure Umbraco UI Builder via the `AddUIBuilder` extension method, You can look in the `Startup.cs` file in the root of your web project. From within this file, before the call to `AddComposers()` we can add our `AddUIBuilder` configuration.
```
sofietoft, line 11: It's been a while since I've seen any activity here, so here's a suggestion to move this one along a bit.

Let me know what you think!
I'll merge it within the next week, if I haven't heard back from you 💪 

```suggestion
To configure Umbraco UI Builder via the `AddUIBuilder` extension method, You can look in the `Program.cs` file at the root of your web project. From within this file, before the call to `AddComposers()` we can add our `AddUIBuilder` configuration.

Alternatively, if your project is upgraded from earlier versions, you can add the configuration to the `Startup.cs` file.
```

sofietoft:
Sorry for all the commits @bjarnef ! 😅 

I'm trying to get all the GitBook checks to give green light.
It's a little tricky when so many spaces needs to be checked 🙈

PR #6444, opened 2024-09-23
Add JPG & JPEG to the list of acronyms

erikjanwestendorp:
## Description

Add JPG & JPEG to the list of acronyms since PNG is also on the list. 

## Type of suggestion

* [x] Styleguide


sofietoft:
Definitely makes sense to add these two to the accepted acronyms!

I'll get it merged.

PR #6462, opened 2024-09-25
Remove mentions of Macros to render Umbraco Forms

sofietoft:
## Description

I've updated the "Render Forms" article by removing the macro section.
I've updated the "Themes article" by linking to the "Rendering Forms" article instead of the example using Macros.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

Forms - 14

### DIFF
--- a/13/umbraco-ums/security-and-privacy/gdpr/how-to-become-gdpr-compliant-using-cookiebot.md
+++ b/13/umbraco-ums/security-and-privacy/gdpr/how-to-become-gdpr-compliant-using-cookiebot.md
@@ -1,12 +1,12 @@
-# how To Become GDPR compliant using cookiebot
+# How To Become GDPR Compliant Using Cookiebot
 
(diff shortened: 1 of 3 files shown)


AndyButland (commented):
Looks great, but I'm just wondering if the `additionalData` feature (relatively recently added, and only needed in a minority of use cases) might confuse things a bit.

So perhaps:

- Remove the line "This example also adds magic strings replacements using the `additionalData` parameter."
- Update the first code sample to be:

```
@await Component.InvokeAsync("RenderForm", new { formId = @Model.Form, theme = @Model.Theme, includeScripts = false })
```

- And then at the end of the section, after the list presented under "Six parameters can be provided" where the `additionalData` parameter is the last to be discussed, add one more example:

---

The following example shows how the `additionalData` parameter is used:


```csharp
var additionalData = new Dictionary<string, string> { { "foo", "bar" }, { "buzz", "baz" } };
@await Component.InvokeAsync("RenderForm", new { formId = @Model.Form, theme = @Model.Theme, includeScripts = false, additionalData })
```

---

sofietoft:
I think that sounds great @AndyButland 💪 

I've updated the article based on your suggestion.
I'll merge it once the checks clears 🙌

PR #6476, opened 2024-09-27
Remove references to images in v12 directory

sofietoft:
## Description

Before we unpublish the documentation for Umbraco 12 and move the directory for the version out of the main branch, some image references needed to be cleaned up.


jonat123 (approved):
@sofietoft LGTM, the images 💪

PR #6503, opened 2024-10-02
V15: Updated and additional articles to Extension Manifests and sub topics

nielslyngsoe:
## Description

Notice this is only correct for v.15

Update and additional articles to the topic of Extension Manifests and sub topics.

Writting your own extension type is changed slitghly in v.15 — this lead me to update a write a bit more about Extension Manifest.

And then related topics of Conditions and Kinds have been updated with additional articles from an overview perspective. But notice the structural challenge of both kinds and conditions begin an extension type of them self but also a Extension Manifest Feature. The point to understand here is that its a Core feature, which can be extended with additional opportunities. hence the two perspectives.

Please see changes to get a better overview of this PR.

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/14/umbraco-cms/customizing/extending-overview/custom-extension-type/custom-extension-type.md
+++ b/14/umbraco-cms/customizing/extending-overview/custom-extension-type/custom-extension-type.md
@@ -1,0 +1,23 @@
+# Custom Extension Type
+
+The extension registry is an open system, which can hold any Extension Manifest Type. This article describes how you can declare your own types.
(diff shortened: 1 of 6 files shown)


eshanrnh (commented):
Thanks @nielslyngsoe for the PR 🙌 We will review as soon as we can.

sofietoft (approved):

PR #6505, opened 2024-10-02
Update readme for extending section

alina-tincas:
## Description

Update readme for extending section + fixed reference links for all other articles in the section
## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [x] Other

cms v14 and v15 as the structure for extending section is different for v13 and below

## Deadline (if relevant)

anytime


eshanrnh (commented):
Thanks, @alina-tincas for the PR 🙌 We will review as soon as we can.

sofietoft:
Thanks for the PR @alina-tincas !

I think it's a great idea to make the link to the legacy docs on GitHub more generic, calling it EOL.
It's an acronym though, so we'll need to define it.

How about we add some more text to the start of this article?
Explaining a bit more about the EOL strategy, linking to the article on .com, and stuff like that? 🤔 

What do you think?

PR #6508, opened 2024-10-03
Restructure Customize Section

nielslyngsoe:
## Description

To align our content with new structure many of the root articles has been moved into their respective location.

This means some articles have been moved into extension-types, as we should cover all extension-types in this folder.

Some topics then might miss a entry article(in the root of customize), but such should be introductionary and link into the extension type article when the reader is ready to write their own extensions.

this pr contains one edit, I removed mentioning Vite in the general articles. It is not relevant for this reader, it should be opaque to them.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)


eshanrnh (commented):
Thanks @nielslyngsoe for the PR 🙌 We will review as soon as we can 💪

sofietoft:
I've completed the restructure directly on the GitBook platform.
The change are live.

I will close this PR.

PR #6511, opened 2024-10-03
Updated the RTE configuration

naepalm:
Changed Plugins and ValidElements to display all items for easier copy-paste

## Description

Updated the RTE configuration documentation to include the entire settings for plugins and valid elements. Because you have to copy the valid elements to append items to them, to get the full list you currently have to dig into the source code. I think it's much more user friendly to be able to copy-paste right from the documentation, even though it makes the list very long.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

v10 & v13

## Deadline (if relevant)

As soon as possible :)


sofietoft:
Thanks for the PR @naepalm ! 🥳 

I'll make sure we review this as soon as possible.

jonat123:
@naepalm THe only thing about this is that the text above the code example should be changed if we display all the elements: 
image

Could I get you to change it? 😊

naepalm:
Hi @jonat123! I didn't change it because there's actually a spot further down where I left the text truncated - I don't think it's as important for the entire commands list to be copied in since it's easier to change with the way it's set up.

But maybe we want to modify that too? (It IS a huge list...)

image

jonat123:
Ah okay, I missed that 🙈

I think it is fine to just let it be then 😄

PR #6515, opened 2024-10-04
Update requirements.md

alina-tincas:
update from 13 to 14 as it is v14 docs

## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/14/umbraco-cms/fundamentals/setup/requirements.md
+++ b/14/umbraco-cms/fundamentals/setup/requirements.md
@@ -39,7 +39,7 @@ As Umbraco releases are aligned to the .NET release cadence, it's also aligned w
 * [.NET Supported releases](https://dotnet.microsoft.com/en-us/platform/support/policy)
 
 {% hint style="info" %}
-For the above, as Umbraco version 13 is based on .NET 8 you will need to follow the .NET 8 supported versions.
+For the above, as Umbraco version 14 is based on .NET 8 you will need to follow the .NET 8 supported versions.
 {% endhint %}
 
 * [IIS Supported releases](https://learn.microsoft.com/en-us/lifecycle/products/internet-information-services-iis)


sofietoft (commented):
Thanks for the PR @alina-tincas !

Instead of mentioning the specific .NET version, which we already do in the requirements above, why not just link to the generic Windows installation docs: https://learn.microsoft.com/en-us/dotnet/core/install/windows 
Then we can also get rid of the note!

What do you think?

alina-tincas:
I think in this case it might be best to just remove the note and leave the link as it is, because the one at the top was requirements for local solution while the one I tried to change was for hosting umbraco.

15/umbraco-cms/fundamentals/setup/requirements.md
```diff
@@ -11,7 +11,7 @@ The Umbraco UI works in all modern browsers:
 
 ## Local Development
 
-Below you can find the minimum requirements to run Umbraco 13 on your machine:
+Below you can find the minimum requirements to run Umbraco 15 on your machine:
 
 * [.NET 8.0 and higher](https://dotnet.microsoft.com/en-us/download/dotnet/8.0)
```
sofietoft, line 16: This should be 9 then, right? 😁 

```suggestion
* [.NET 8.0 and higher](https://dotnet.microsoft.com/en-us/download/dotnet/9.0)
```

15/umbraco-cms/fundamentals/setup/requirements.md
```diff
@@ -11,9 +11,9 @@ The Umbraco UI works in all modern browsers:
 
 ## Local Development
 
-Below you can find the minimum requirements to run Umbraco 13 on your machine:
+Below you can find the minimum requirements to run Umbraco 15 on your machine:
 
-* [.NET 8.0 and higher](https://dotnet.microsoft.com/en-us/download/dotnet/8.0)
+* [.NET 8.0 and higher](https://dotnet.microsoft.com/en-us/download/dotnet/9.0)
```
sofietoft, line 16:
```suggestion
* [.NET 9.0 and higher](https://dotnet.microsoft.com/en-us/download/dotnet/9.0)
```

sofietoft (commented):
Final change 😅

PR #6543, opened 2024-10-09
Removed duplicate code and add note for obsolete method

sofietoft:
## Description

Removed duplicate code.
Added a note about the obsolete `GetAll()` method on the `_domainService`.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

15.0.0

### DIFF
--- a/14/umbraco-cms/reference/routing/request-pipeline/icontentfinder.md
+++ b/14/umbraco-cms/reference/routing/request-pipeline/icontentfinder.md
@@ -164,7 +164,7 @@ public class My404ContentFinder : IContentLastChanceFinder
     public Task<bool> TryFindContent(IPublishedRequestBuilder contentRequest)
     {
         // Find the root node with a matching domain to the incoming request
-        var allDomains = _domainService.GetAll(true).ToList(); // <---- IDomainService.GetAll(bool)' is obsolete: 'Please use GetAllAsync. Will be removed in V15'
+        var allDomains = _domainService.GetAll(true).ToList();
         var domain = allDomains?
             .FirstOrDefault(f => f.DomainName == contentRequest.Uri.Authority
                                     || f.DomainName == $"https://{contentRequest.Uri.Authority}"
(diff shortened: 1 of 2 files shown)


bergmania:
That entire example should be updated to something like

```cs
using Umbraco.Cms.Core.PublishedCache;
using Umbraco.Cms.Core.Routing;
using Umbraco.Cms.Core.Services;

namespace RoutingDocs.ContentFinders;

public class My404ContentFinder : IContentLastChanceFinder
{
    private readonly IDomainService _domainService;
    private readonly IPublishedContentCache _publishedContentCache;

    public My404ContentFinder(
        IDomainService domainService,
        IPublishedContentCache publishedContentCache)
    {
        _domainService = domainService;
        _publishedContentCache = publishedContentCache;
    }

    public async Task<bool> TryFindContent(IPublishedRequestBuilder contentRequest)
    {
        // Find the root node with a matching domain to the incoming request
        var allDomains = (await _domainService.GetAllAsync(true)).ToList();
        var domain = allDomains
            .FirstOrDefault(f => f.DomainName == contentRequest.Uri.Authority
                                    || f.DomainName == $"https://{contentRequest.Uri.Authority}"
                                    || f.DomainName == $"http://{contentRequest.Uri.Authority}");

        var siteId = domain != null ? domain.RootContentId : allDomains.Any() ? allDomains.FirstOrDefault()?.RootContentId : null;


        var siteRoot = _publishedContentCache.GetById(false, siteId ?? -1);

        if (siteRoot is null)
        {
            return false;
        }

        // Assuming the 404 page is in the root of the language site with alias fourOhFourPageAlias
        var notFoundNode = siteRoot.Children()?.FirstOrDefault(f => f.ContentType.Alias == "fourOhFourPageAlias");

        if (notFoundNode is not null)
        {
            contentRequest.SetPublishedContent(notFoundNode);
        }

        // Return true or false depending on whether our custom 404 page was found
        return contentRequest.PublishedContent is not null;
    }
}

```

PR #6552, opened 2024-10-12
Clarifying management API authentication

richarth:
Added a note with the client_id, that you need to leave the client_secret empty for non-production environments as it's easy to miss the hint below the screen shot that lets you know that you need to leave the client_secret empty.

### DIFF
--- a/14/umbraco-cms/reference/management-api/README.md
+++ b/14/umbraco-cms/reference/management-api/README.md
@@ -48,7 +48,7 @@ You can see an example of how to connect a backoffice user via OAuth2 in Postman
 
 To test a Management API endpoint, follow these steps:
 
-1. Authenticate via the **Authorize** button. Make sure you use `umbraco-swagger` as the `client_id` and in non-production environments you need to leave the `client_secret` blank:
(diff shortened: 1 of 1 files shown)


14/umbraco-cms/reference/management-api/README.md
```diff
@@ -48,7 +48,7 @@ You can see an example of how to connect a backoffice user via OAuth2 in Postman
 
 To test a Management API endpoint, follow these steps:
 
-1. Authenticate via the **Authorize** button. Make sure you use `umbraco-swagger` as the `client_id`:
+1. Authenticate via the **Authorize** button. Make sure you use `umbraco-swagger` as the `client_id` and in non-production environments you need to leave the `client_secret` blank:
```
sofietoft, line 51:
```suggestion
1. Authenticate via the **Authorize** button. Make sure you use `umbraco-swagger` as the `client_id` and leave the `client_secret` blank in non-production environments:
```

sofietoft (commented):
Thanks for the PR @richarth ! 💪 

I've made a single suggestion to make it a little more readable. Let me know what you think!

richarth:
Hi @sofietoft,

Thanks for the feedback.  I agree that reads a lot better so I've accepted the change.

Thanks again.

PR #6607, opened 2024-10-30
Added key and default value information to most sections

abjerner:
## Description

Every time I end up at the **Global Settings** page, I feel like each section on the page misses a bit of context. You can see the description about a given option, but you can't see the actual key unless you scroll back up to the top of the page. Also if you enter the page searching for a specific key, you can only find it in the big JSON example.

With this commit, I've added an extra line for each option stating the name of the associated key, and for most of the options, also specified a default value. I feel that this improving the descriptions for the various options. And keep scrolling frantically up and down to minimum.

There are a few options where I'm unsure about what to enter, so I've skipped these. Some of them I've skipped adding the key entirely, why others I've skipped the default key. E.g. if this is mentioned in the text already. Or if the description says one, but the option via `IOptions<GlobalSettings>` is really an empty string by default.

I hope this makes sense 😉 

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)

No deadline

### DIFF
--- a/14/umbraco-cms/reference/configuration/globalsettings.md
+++ b/14/umbraco-cms/reference/configuration/globalsettings.md
@@ -65,13 +65,15 @@ In the root level section, that is those without a seperate sub section like SMT
 
 ### Reserved urls
 
-Key: `ReservedUrls` (default: `~/.well-known,`)
+Key: `ReservedUrls`  
+Type: `string` (default: `~/.well-known,`)
 
 A comma-seperated list of files to be left alone by Umbraco, these files will be served, and the Umbraco request pipeline will not be triggered.
 
 ### Reserved paths
 
-Key: `ReservedPaths` (default: `~/app_plugins/,~/install/,~/mini-profiler-resources/,~/umbraco/,`)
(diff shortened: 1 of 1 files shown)


sofietoft:
Thanks @abjerner !
This is a great addition to the article 👏 Anything to make these articles more readable! 😁 

I do have one suggestion though after looking at the preview:
image

Could we try adding the default value to a separate line?
So, something like:

Key: `UmbracoMediaPath`
Default: `~/media`

What do you think? 🤔

sofietoft:
Yeah! Adding the value type could be excellent 🙌 

Then perhaps we do:

Key: `UmbracoMediaPath`
Type: `string` (default: `~/media`)

Because, I don't think we want to go for 3 lines, right? 😅

abjerner:
@sofietoft I agree. Should be updated now 😉

sofietoft (approved):
Looks great @abjerner !

Let's go with it 😁👏🏻
Thanks again for suggesting and implementing this change!

PR #6635, opened 2024-11-08
Update localization.md

PerplexDaniel:
Use the correct property, it should be `"localizations"` and not `"translations"`.

Proof that `umbraco-package-schema` also suggests `"localizations"`:
image


sofietoft:
Good catch @PerplexDaniel ! 💪 

Thanks for updating this!

PR #6675, opened 2024-11-15
Sample code is using v13 namespace and doesn't compile

stevetemple:
## Description

The sample code is using an old namespace so will not compile

## Type of suggestion

* [ ] Typo/grammar fix
* [X] Updated outdated content
* [ ] New content
* [ ] Other


sofietoft:
Good catch @stevetemple ! 🙌 
Thanks.

I'll get it merged.

PR #6686, opened 2024-11-21
Restructured some Deploy pages to align with links from product and oved some getting started docs into a separate section.

AndyButland:
## Description

Restructured some Deploy pages to align with links from product and moved some getting started docs into a separate section.

The main reason for this is I noticed today that the links we have in the product to the docs page about configuration were broken (after [this update](https://github.com/umbraco/UmbracoDocs/pull/5757) to align the summary and navigation for the products).  The page had been moved into a folder, so the link no longer worked.  As it's easier to update the docs (hosted in a single place) than the product (hosted in many places) I think it makes sense to revert that part of the update which I've done here.

Whilst doing so I also moved a couple of more advanced articles out of the "getting started" section and into a new "extending" one.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [X] Other

Deploy 10+

## Deadline (if relevant)

Can be reviewed and published at any time.

CC // @ronaldbarendse


sofietoft:
Thanks for the PR @AndyButland !

I think it's great moving the extending related articles to a group on their own 💪 
We'll need to add rewrite for these though.

Also, the configuration article - if you want to keep it in the same spot in the SUMMARY/navigation, I think we should add a redirect instead.
Currently, the path in the SUMMARY file doesn't match the file structure, and I'm concerned how GitBook will handle this change.
It's pretty strict when it comes to the structure, and changes might occur the next time changes are pushed from GitBook to GitHub.

I have a different suggestion for handling the changes in this PR:

I'll move the Extending articles into a new group - I'll do this on GitBook, which will ensure the redirects are handled automatically.

Regarding the Configuration article: Do you wanna keep that in the Getting Started group? We could move it to Installation or something instead perhaps? I'm just looking for a way to handle this on GitBook, as this will give us the automatic redirects 😅

AndyButland:
Very happy if you have a better way to solve this Sofie, so please make whatever changes you see fit.

The only key thing is the broken link that's coming from the Deploy products at the moment.

We link to: `https://docs.umbraco.com/umbraco-deploy/deploy-settings`

But the page is now at: `https://docs.umbraco.com/umbraco-deploy/getting-started/deploy-settings`

Might be you can just solve that with a redirect and you don't really need the the other changes... which I think make sense, but if they causes more problems than they give benefit we can leave as is.

sofietoft:
Sounds good!
I'll see if I can get around having to make a manual redirect for the Deploy Settings (configuration) article.
Will run some tests on v10 first, and apply to the rest of the versions if it works.
Otherwise, I'll make a manual redirect for it.

Should we still create the new "Extending" group in the navigation?
If yes, I'll do that directly on GitBook for all versions - that will give us the automatic redirects! 😁

AndyButland:
> Should we still create the new "Extending" group in the navigation?

I think it would be an improvement if you can make it work without causing other issues - the articles I've put in there are more advanced topics, so it doesn't feel right in "getting started".

sofietoft:
Hi @AndyButland !

I've moved around the articles directly in the GitBook platform.
The links you've provided should work again - or at least they are now redirecting correctly! 💪 
I've also put the two more advanced article into an "Extending" section - redirects are working.

Could you give it a few tests, so we can ensure the redirects are in fact working as they should? 🤞

AndyButland:
Seems to be working fine - thanks very much.

sofietoft:
Excellent! Then I'll close this PR again 😁

PR #6690, opened 2024-11-21
Change code example from Implements to Extends

warrenbuckley:
## Description

A code sample for the UFM component example, using implements means I needed to do more than what I needed to do.
Looked at the source code of the existing UFM-Components and they use the `extends` approach instead so we only need to implement the `render` method

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

## Deadline (if relevant)


sofietoft:
Thanks @warrenbuckley !
This makes a lot of sense 💪 

Let's get it merged!

PR #6718, opened 2024-12-03
Simplified steps in Major upgrades Cloud article

eshanrnh:
## Description

Removed screenshots, made the steps concise, fixed vale error of long sentence in Note

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [x] Other

Cloud

## Deadline (if relevant)

Anytime


umbraco-cloud/product-upgrades/major-upgrades.md
```diff
@@ -76,20 +76,12 @@ Look for the "**Upgrade from/to Umbraco xx"** boxes. These boxes contain importa
...
-   1. Choose the version you set on your Cloud environment in [Step 1](major-upgrades.md#step-1-enable-.net).
-
-![Target Framework](images/Target-Framework.png)
-
-5. Go to **Tools** > **NuGet Package Manager** > **Manage NuGet Packages for Solution.**
-6. Navigate to the **Updates** tab.
-7. Select the version you are updated to, and follow the instructions:
+1. Open the `csproj` file located in the `/src/UmbracoProject` folder.
+2. Right-click your project in **Solution Explorer** and select **Properties**.
+3. Change the **Target framework** under the **Application** tab to match the version set in your Cloud environment from [Step 1](major-upgrades.md#step-1-enable-.net).
```
sofietoft, line 81: I'm wondering if it makes sense, to change these steps a bit.
So, instead of opening the properties, we open the .csproj file in Visual Studio.
There you can also change the .NET version.

If we do this, we can move the part about the package references that needs to be deleted in 15 - we can move these into an expandable under these bullets, as it's changes that needs to be made in the same file 🤔 

Currently, the steps for deleting, needs to be done after a step that says you need to open the NuGet package manager - which I guess is not really a good workflow 🤔

sofietoft (commented):
I just realized that we have a challenge here.

In the guide, we mention that you need to update the .NET versions. This is not necessary for ALL major upgrades - only every other Major upgrade.
So, it was necessary for 15, but it wont be for 16 and will again be necessary for 17 🤔 

If you have any ideas for how we can tackle this, please let me know!

eshanrnh:
Thanks, @sofietoft for your review. I've made some changes based on your suggestion. 

Regarding updating the .Net version, I added a decision-making step to the list. Depending on the user's target upgrade they can decide either to update or skip the step. Since Umbraco 16 and 17 are not yet released I have only added 15. Once the other versions are released, we can change it to something like this:

Determine if you need to update the .NET version:

- For Umbraco 15 or 17: Update the <TargetFramework> to match the version set in your Cloud environment from [Step 1](https://github.com/umbraco/UmbracoDocs/pull/major-upgrades.md#step-1-enable-.net).

- For Umbraco 16 or other versions that do not require a .NET update: Skip this step.

Optionally, we can add a Note instead - something like:

Note: Updating the .NET version is necessary only for every other major upgrade. For example, .NET updates were required for version 15 but are not needed for version 16. The next .NET update will be required for version 17.

sofietoft:
Thanks for putting some thought into this!
I like the idea about having to paths to follow, but I'm just worried it becomes a little too messy - looking at the preview, it's not super clear what needs to happen 🤔 
We also have this issue already in Step 1, so I think this is a general thing for the entire guide.

Also, since it's Cloud, we need to consider all versions going back to Umbraco 10, I think. At least all LTS versions and STS versions still supported. But, I also think we can get around this, without actually mentioning any specific versions regarding it.
Perhaps we link to the CMS article that lists Umbraco versions vs .NET versions: https://docs.umbraco.com/umbraco-cms/fundamentals/setup/upgrading#choose-the-correct-.net-version 

How about we do this:
- In Step 1 we add a step to determine which .NET version to use - and depending on that, the .NET version might need to be bumped.
- Then, in Step 3, the task is to only opdate the .NET version in the TargetFramework if it was also changed in Step 1 🤔 

Do you think that could work, perhaps?

umbraco-cloud/product-upgrades/major-upgrades.md
```diff
@@ -63,7 +63,8 @@ Look for the "**Upgrade from/to Umbraco xx"** boxes. These boxes contain importa
 1. Go to the project in the Umbraco Cloud portal.
 2. Navigate to **Configuration** -> **Advanced**.
 3. Scroll down to the **Runtime Settings** section.
-4. **Ensure that the latest version of .NET is enabled** for each environment on your Cloud project, by selecting it from the dropdown.
+4. From the **Change .NET framework runtime for your Umbraco install** dropdown, select the appropriate .NET version for each environment in your Cloud project.
```
sofietoft, line 66: Should start with the action word 💪

umbraco-cloud/product-upgrades/major-upgrades.md
```diff
@@ -63,7 +63,8 @@ Look for the "**Upgrade from/to Umbraco xx"** boxes. These boxes contain importa
...
+   * Refer to the [Choose the correct .NET version](https://docs.umbraco.com/umbraco-cms/fundamentals/setup/upgrading#choose-the-correct-.net-version) section for guidance on selecting the appropriate .NET version for your upgrade.
```
sofietoft, line 67: I think this information would be better before this list of steps start.
If you don't need to change the .NET version, then the entire Step 1 is not necessary 🤔 
So perhaps, then first step in this list should be about determining whether the .NET framework needs to be updated?

eshanrnh:
I think the steps should be clear now 😀

umbraco-cloud/product-upgrades/major-upgrades.md
```diff
@@ -60,10 +60,14 @@ Look for the "**Upgrade from/to Umbraco xx"** boxes. These boxes contain importa
...
 
+Before proceeding, you must determine whether the .NET framework version needs to be updated for your project. If no changes to the .NET version are required, you can skip this step and proceed with Step 2.
+
+Refer to the [Choose the correct .NET version](https://docs.umbraco.com/umbraco-cms/fundamentals/setup/upgrading#choose-the-correct-.net-version) section to identify whether a .NET version update is necessary for your upgrade.
+
 1. Go to the project in the Umbraco Cloud portal.
 2. Navigate to **Configuration** -> **Advanced**.
 3. Scroll down to the **Runtime Settings** section.
-4. **Ensure that the latest version of .NET is enabled** for each environment on your Cloud project, by selecting it from the dropdown.
+4. Select the appropriate .NET version from the **Change .NET framework runtime for your Umbraco install** dropdown for each environment in your Cloud project.
```
sofietoft, line 70: This actually shouldn't actually mention all environments - it should only be the first environment / development environment 🤔 

In a step 5, Step 1 is referenced, mentioning that you need to bump the .NET version on the next environment before deploying the upgrade to it. Hm!

Perhaps we could say:
"Select the correct .NET version from the [...] for the development environment, or the environment you are deploying the upgrade to."

Not sure if that becomes too confusing if you're only starting the upgrade 😅 
I think I definitely might be overcomplicating things here!!

eshanrnh: Umm, I don't think it matters cz we aren't providing separate steps for different environments. If I change it specifically to development I will have to add another line somewhere to repeat all these steps for another environment. In that case, it will just be making the guide too long. With this statement, there are chances the person doesn't miss the step for another environment 😅

As for step 5 list, the step to enable .net is mentioned twice in point 5 and point 11. Ideally, step 1 covers all versions, so those steps can be removed to reduce the number of steps while deploying the upgrade.

sofietoft: Yeah, I did think about that as well.. But I'm not sure what impact changing this version actually has on the environment on Cloud. Like, would the environment continue to work even though the .NET version is changed without it being upgraded 🤔 I've asked the Cloud developers how this works .. if it doesn't impact the running environment, let's get rid of those extra steps in Step 5!

sofietoft: OK, I finally got some answers.
Let's get the .NET version updated on all environments in Step 1, and then remove the references to this in Step 5 💪

eshanrnh:
Awesome 🙌 I've removed references to step 1 from step 5. Hopefully 🤞its good to merge now 🙈

sofietoft (approved):

PR #6733, opened 2024-12-10
Removed and updated label description for blocks

Ambertvu:
## Description

Updated the block-grid-editor hint as it still contained old angularJS description. This is a bit odd but it is also referenced in the link a bit above: https://docs.umbraco.com/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/label-property-configuration. So this should either be completely removed, or still easily accessible for findings like it is with my PR.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

v14/v15 (new backoffice)

### DIFF
--- a/14/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/block-grid-editor.md
+++ b/14/umbraco-cms/fundamentals/backoffice/property-editors/built-in-umbraco-property-editors/block-editor/block-grid-editor.md
@@ -74,11 +74,7 @@ Customize the user experience for your content editors when they work with the B
 * **Label** - Defines a label for the appearance of the Block in the editor. The label can use AngularJS template-string-syntax to display values of properties.
 
 {% hint style="info" %}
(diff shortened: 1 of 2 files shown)


Ambertvu:
> Thanks for the PR @Ambertvu ! Just to make sure I understand, you're removing some mentions of AnuglarJS here, right? Also, just to let you know, this article has yet to be verified for Umbraco 14/15.

Correct! I've removed the AngularJS syntax and description and added a tiny example of how it can be done currently.
I've been 'out' for a while, so not sure what the implications are for the 'yet to be verified' and my pr.

sofietoft:
Perfect. Sounds good! Let's definitely get rid of that.

Yeah, that's fair! 😄 
We have the article on a list here, and we hope to get it verified for the latest versions of Umbraco soon.

For now, I'll merge this PR.
Thanks again!

sofietoft:
Oh, just read the description again - you're referencing another article, the one about the label properties.
Let's definitely get that removed!

I'll add a task to get it done as soon as possible 💪 
Thanks for flagging it!

PR #6744, opened 2024-12-12
Update UI Builder Licence Key Key

marcemarc:
Add 'Products' to the key

In v14/v15 Umbraco have added 'Products' to the structure of the JSON for adding licences to the products, the docs have been updated with the new JSON but not the handy cut and paste configuration key

we did umbraco commerce, yesterday, this is UI builder, but likely they are all affected...


sofietoft:
Thanks again @marcemarc ! 🙌

PR #6747, opened 2024-12-12
Described option for using underscores in product keys when configuring product licenses with environment variables.

AndyButland:
## Description

Described option for using underscores in product keys when configuring product licenses with environment variables.

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [X] Other

Commerce, UI Builder, Engage and Deploy (on-premise)

## Deadline (if relevant)

Any time.

### DIFF
--- a/10/umbraco-commerce/installation/licensing-model.md
+++ b/10/umbraco-commerce/installation/licensing-model.md
@@ -67,6 +67,14 @@ Once you have received your license code it needs to be installed on your site.
 }
 ```
 
+{% hint style="info" %}
+If using environment variables rather than a JSON file for your configuration you may run into issues with the periods used in the product key.  If so, an underscore can be used instead:
+
+```json
+"Umbraco_Commerce": "YOUR_LICENSE_KEY"
+```
+{% endhint %}
+
 ### Verify the license installation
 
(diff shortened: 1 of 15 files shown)


10/umbraco-commerce/installation/licensing-model.md
```diff
@@ -67,6 +67,14 @@ Once you have received your license code it needs to be installed on your site.
 }
 ```
 
+{% hint style="info" %}
+If using environment variables rather than a JSON file for configuration you may run into issues with the periods used in the product key.  If so, an underscore can be used instead:
```
sofietoft, line 71: Could this, an potentially the rest, be stripped down to this? Or is it lacking too much detail? 🤔 

```suggestion
Use an underscore in the product name, to avoid issues when using environment variables.
```

AndyButland: It would be OK, but the detail is kind of useful too for background, and also as we've documented above using the ".".

sofietoft: Yeah, I see 🤔 

How about:

"When using environment variables you might run into issues when using a period in the product name. Use an underscore in the product name instead, to avoid issues."

AndyButland: Sounds fine, do you want me to make that update or will you do so?

Issue #6728, opened 2024-12-10
Incorrect reference for Creating a backoffice API


russellshome:
## What type of issue is it? (Choose one - delete the others)
Wrong documentation

## What article/section is this about?
https://docs.umbraco.com/umbraco-cms/14.latest/tutorials/creating-a-backoffice-api
https://docs.umbraco.com/umbraco-cms/tutorials/creating-a-backoffice-api

## Describe the issue
The documentation says 
To create a custom API, you need to create a class that inherits from Umbraco.Cms.Web.BackOffice.Controllers.ManagementApiControllerBase

But Umbraco.Cms.Web.BackOffice no longer exists in Umbraco 14+
 
For Umbraco 14+ I think the following should be replaced

Umbraco.Cms.Web.BackOffice.Controllers.ManagementApiControllerBase
=>
Umbraco.Cms.Api.Management.Controllers.ManagementApiControllerBase

sofietoft:
Thanks for raising this @russellshome ! 🙌 Much appreciated!

I'll make sure we get this verified so we can update the documentation accordingly!

sofietoft:
Hi @russellshome !

The changes you're suggestions are indeed correct, and our docs should be updated.
Would you be up for making the change?
Let me know, otherwise I'll make sure we update it as soon as possible.

russellshome:
Done https://github.com/umbraco/UmbracoDocs/pull/6758

russellshome:
Just a suggestion about the "Edit on GitHub"  link on the right of the article. It is dark grey on black rendering it almost invisible to me. This is also a problem for code in the articles. I end up having to highlight the content to get white on dark blue so I can read it comfortably.

You can use https://webaim.org/resources/contrastchecker/ to work out a better colour combination that makes the content easier to read. (ie. passes WCAG AAA)

sofietoft:
Thanks for the suggestions @russellshome !

I'll make sure we look into this in the new year.
I've merged your PR, so will mark this issue as resolve!

PR #6758, opened 2024-12-19
Fix for https://github.com/umbraco/UmbracoDocs/issues/6728

russellshome:
For Umbraco 14+ I think the following should be replaced

Umbraco.Cms.Web.BackOffice.Controllers.ManagementApiControllerBase =>
Umbraco.Cms.Api.Management.Controllers.ManagementApiControllerBase

## Description

See above

## Type of suggestion

14

## Deadline (if relevant)


sofietoft (approved):
Looks great @russellshome ! 🙌 
Thanks for making the updates!

PR #6765, opened 2025-01-02
Update repositories

erikjanwestendorp:
## Description

- [x] Write shorter sentences

## Type of suggestion

* [x] Styleguide

Umbraco CMS v14 and 15

### DIFF
--- a/14/umbraco-cms/customizing/foundation/working-with-data/repositories.md
+++ b/14/umbraco-cms/customizing/foundation/working-with-data/repositories.md
@@ -20,7 +20,7 @@ const repositoryManifest = {
 };
 ```
 
-With a repository we can have different data sources depending on the state of the app. It can be from a server, an offline database, a store, a Signal-R connection, etc. That means that the consumer will not have to be concerned how to access the data, add or remove items from a collection, etc. This means we get a loose connection between the consumer and the data-storing procedures hiding all complex implementation.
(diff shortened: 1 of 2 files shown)


14/umbraco-cms/customizing/foundation/working-with-data/repositories.md
```diff
@@ -20,12 +20,12 @@ const repositoryManifest = {
 };
 ```
 
-With a repository we can have different data sources depending on the state of the app. It can be from a server, an offline database, a store, a Signal-R connection, etc. That means that the consumer will not have to be concerned how to access the data, add or remove items from a collection of items, etc. This means we get a loose connection between the consumer and the data-storing procedures hiding all complex implementation.
+With a repository we can have different data sources depending on the state of the app. It can be from a server, an offline database, a store, a Signal-R connection, etc. That means that the consumer will not have to be concerned how to access the data, add or remove items from a collection, etc. This means we get a loose connection between the consumer and the data-storing procedures hiding all complex implementation.
```
sofietoft, line 23:
```suggestion
With a repository we can have different data sources depending on the state of the app. The data sources can come from places like a server, an offline database, a store, or a Signal-R connection. That means that the consumer will not have to be concerned how to access the data, add or remove items from a collection, etc. This means we get a loose connection between the consumer and the data-storing procedures hiding all complex implementation.
```

sofietoft (commented):
Thanks for the PR @erikjanwestendorp ! 😄 

I added a single comment, in an attempt to avoid using "etc".
What do you think? 🤔

erikjanwestendorp:
@sofietoft Thanks for the review, just updated the PR 😄

PR #6769, opened 2025-01-07
Update Defaults for Color and Look on External Providers config

marcemarc:
## Description

When setting either the 'color' or the 'look' of the button for the external providers login button , although the settings come from the UUI link provided.. we are specifying the default values for the wrong thing in the list... so 'Color' can either be Default, Positive, Warning, Danger and
'Look' can be either Primary, Secondary, Outline or Placeholder

But we say here the default for color is 'secondary' and default for Look is 'default'

but they should be the other way around!!!

This PR flips them so they match the implementation

I've also added examples of what the colors are because, they are sort of historically bootstrappy names and not 'actual' colors :-P

Someone setting a property called 'color' might expect to be able to set it to a #hex reference or a color eg blue, but actually the options are not colors at all just words! I think although the info is 'there' in the UUI Colour link, it's not clear which one is which, bourne out by the fact that we've documented it wrong here, so maybe just listing the options as I've done here is helpful? but maybe less is more, and the link to the UUI library tells the full story!

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

Umbraco 15

## Deadline (if relevant)

Now


15/umbraco-cms/reference/security/external-login-providers.md
```diff
@@ -387,8 +387,8 @@ You have a few options to configure the button:
...
 * `meta.label` - The label to display on the button. The user will see this text. For example: "Sign in with Generic".
 * `meta.defaultView.icon` - The icon to display on the button. You can use any of the icons from the Umbraco Icon Picker. If you want to use a custom icon, you need to first register it to the [`icons` extension point](../../customizing/extending-overview/extension-types/icons.md).
-* `meta.defaultView.color` - (Default: "secondary") The color of the button. You can use any of the colors from the [Umbraco UI Library](https://uui.umbraco.com/?path=/story/uui-button--looks-and-colors).
-* `meta.defaultView.look` - (Default: "default") The look of the button. You can use any of the looks from the [Umbraco UI Library](https://uui.umbraco.com/?path=/story/uui-button--looks-and-colors).
+* `meta.defaultView.color` - (Default: "default") The color style of the button. You can use any of the color styles from the [Umbraco UI Library](https://uui.umbraco.com/?path=/story/uui-button--looks-and-colors). eg default (blue), positive (green), warning (yellow) or danger (red)
```
sofietoft, line 390:
```suggestion
* `meta.defaultView.color` - (Default: "default") The color style of the button. You can use any color style from the [Umbraco UI Library](https://uui.umbraco.com/?path=/story/uui-button--looks-and-colors).
  * Default (blue)
  * Positive (green)
  * Warning (yellow)
  * Danger (red)
```

sofietoft (commented):
Thanks for PR @marcemarc !
Great to get this corrected, and it makes sense to add the different options as well.

I've made a few suggestions to the structure of it though - it might be a little easier to read when put into sub-bullets.
What do you think?

marcemarc:
Thanks @sofietoft Thanks and Happy New Year... you know how I like one long sentence that never never ends, sometimes with commas but often just changing context in the middle, hope you had a great Christmas, but on reflection, I think: 

yes, I like the bullets, so much easier to look up at a glance!

sofietoft:
Happy New Years to you as well @marcemarc - hope you had a wonderful holiday with the family!

And only 41 words? You can do better than that 😄 😉 

Thanks for applying the suggestions here - I'll make sure the PR is merged.

PR #6820, opened 2025-01-24
Changed "Is is" to "It is"

JBoye:
## Type of suggestion

Typo/grammar fix


sofietoft:
Good catch @JBoye ! 😄 
Thanks for correcting this.

I'll merge it right away.

PR #6849, opened 2025-02-04
Document `userinfo` endpoint

bjarnef:
## Description

Documentation of new `userinfo` endpoint from Umbraco 13.6+
https://github.com/umbraco/Umbraco-CMS/pull/17719

Also document of this part would be great, e.g. a custom properties from member like `firstName` and `lastName`:
> The implementation is build to be extendable, so custom claims can be added to these claims - and the core claims can be removed, too.

Perhaps it should be mentioned `sub` contains Umbraco member guid key, but not sure if it always guid if integration with external providers.

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/13/umbraco-cms/reference/content-delivery-api/protected-content-in-the-delivery-api.md
+++ b/13/umbraco-cms/reference/content-delivery-api/protected-content-in-the-delivery-api.md
@@ -382,6 +382,28 @@ To terminate the active session for any given member, you must redirect the brow
 GET /umbraco/delivery/api/v1/security/member/signout?post_logout_redirect_uri={valid URL from LogoutRedirectUrls}
 ```
 
+### User info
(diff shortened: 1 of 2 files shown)


kjac (approved):
Good stuff @bjarnef ❤️

bjarnef:
> Perhaps it should be mentioned `sub` contains Umbraco member guid key, but not sure if it always guid if integration with external providers.

@kjac it seems `sub` is member key/guid in this this, but I guess it could be something else with external provider or if overwriting key.

It would also be great somewhere to place to .NET part:

> The implementation is build to be extendable, so custom claims can be added to these claims - and the core claims can be removed, too.

PR #6858, opened 2025-02-06
Fixing some wording

filipbech:
## Description

Since we are creating a workspace view, we should call it a workspace-view, and not a dashboard (I'm thinking this is a copy/paste mistake :-D)

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other

14+ I think

## Deadline (if relevant)

ASAP


eshanrnh (commented):
That's a good catch, @filipbech Indeed, it looks like an oversight 💪

eshanrnh (approved):

PR #6902, opened 2025-02-20
Added a Note...

eshanrnh:
## Description

... and shortened a long sentence

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [x] Other

CMS v15

## Deadline (if relevant)

Anytime


sofietoft (commented):
I can see I should have been more clear when setting this task 😅 
I don't think we should mention the specific step here, as they should really just be using the Cloud upgrade guide, if they're upgrading a Cloud project.

Instead, I think higher up in the article somewhere, it should be noted that you should use the Cloud upgrade guide, if it's a Cloud project that you're upgrading.
So, a more generic message, sort of.

Hope it makes sense!

PR #6906, opened 2025-02-21
Update umbraco-flavored-markdown.md

JeffreyPerplex:
Pipes are not escaped with a slash and this broke the table

Before it looked like this:

afbeelding

After it looked like this (at least in preview): 

afbeelding


eshanrnh (approved):
Thanks for fixing the contents of the table, @JeffreyPerplex 🙌 Appreciate your help in keeping the docs up to date 💪

PR #6910, opened 2025-02-25
UI Builder: Updated format of Areas section

eshanrnh:
## Description

- Updated the layout of the Sections article making it easier to nabigate and less overwhelming (hopefully 🤞).
- Removed the code from the heading tags and added it separately as a **Code Reference** below the example.
- Added some extra content/descriptions where required.
- 
## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

UI Builder v15

## Deadline (if relevant)

Anytime

### DIFF
--- a/15/umbraco-ui-builder/areas/overview.md
+++ b/15/umbraco-ui-builder/areas/overview.md
@@ -1,19 +1,17 @@
 ---
-description: Choosing an area to connect Umbraco UI Builder, the backoffice UI builder for Umbraco.
+description: Learn how to choose and configure the appropriate area for connecting Umbraco UI builder for Umbraco.
 ---
 
 # Areas
 
-There are different areas of the Umbraco UI that Umbraco UI Builder can be injected into. Before you get to managing your actual content you need to choose which area makes the most sense to present that data in. Then you can review how to go about configuring that particular type of area.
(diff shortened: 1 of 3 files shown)


sofietoft:
PR looks great @eshanrnh !

I like that the code reference isn't used as the headings. This was done a lot also in the CMS docs in the past, but I agree on moving away from this 💪 
Could we make the headings even shorter by removing the "Example" bit? 🤔 
So, an instead of _Example: Using the AddSectionBefore() Method_ we do _Using the AddSectionBefore() Method_? 

Regarding the "Code reference" bit. It looks a little strange as the code extends more than one line 🤔 
Would it help to keep "Code reference" on one line, and then move the actual code to a line below?

Overall great changes to the article 💪 My suggestions are just to see if we can make it even cleaner 😄

eshanrnh:
Thanks for the feedback, @sofietoft 🙌

I’m happy to hear you like the changes, especially moving away from using the "Code reference" as headings. I agree that we can aim for cleaner and shorter headings.

Regarding the code reference, it definitely makes sense to move it to the next line. While doing that, it clicked me to rearrange the code example and method syntax for more clarity. For now, I have applied the change only to the **Using the AddSection() Method** method. If you are fine with it, I'll update the rest of the methods with the same format 🤞

All suggestions and comments are appreciated 🤩

sofietoft (approved):
Changes looks great 💪 
I like moving the code reference up, and putting it into a block 👏🏻

PR #6944, opened 2025-03-17
First draft of the Macro migration inspirate tutorial

Migaroez:
## Description

## Type of suggestion
* [ ] New content
CMS v15

## Deadline (if relevant)


15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
```diff
@@ -0,0 +1,889 @@
...
+```
+
+Next is to get all (relevant) macro tags out of the markup. One way of doing this is trough regex. Do not that the sample regex does not take into account that the order of parameters might be different from tag to tag. One way of dealing with this is to not take out the paramaters in the first match, but to move each paramater to a seperate regex that runs on the first match.
+
+###### Example regex
+```csharp
+// this regex does not take into account that the paramaters might be in a different order.
+private static readonly Regex MacroTagRegex = new(
+    @"<\?UMBRACO_MACRO\s+macroAlias=['""](?<macroAlias>.+)['""]\s+title=['""](?<title>.+)['""]\s+youtubeVideoId=['""](?<youtubeVideoId>.+)['""]\s*/>",
+    RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace | RegexOptions.Multiline);
```
ronaldbarendse, line 151: Apparently there's an 'old' macro format, which starts with `<umbraco:macro` and uses a different attribute for the alias. Deploy already parses macros to convert local database integer IDs to UDIs and adds dependencies, so I'd suggest using the following (battle tested) regex instead:

```c#
new Regex(@"\<(?<tag>\??umbraco(?:_|:)macro) (?<attributes>.*?(macro)?Alias=""(?<alias>.+?)"".*?)?\/?\>", RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace | RegexOptions.CultureInvariant);
```

You can also take a look at the `MacroParser` class in the Deploy v13 source code.

15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
```diff
@@ -0,0 +1,889 @@
...
+# Putting the updated value back
+Once we have transformed the data we need to store it. One again we have chosen to use some custom sql to perform this. If you however need to perform validation on the updated value, you will either have to use some higher level services (`IContentValidationService`/`IContentEditingService`) or use the `RichTextPropertyValueEditor.Validate()` method. Because this examples fetches the current data and overwrites the value it, the old value will not show up in the version history of the affected node. If you do need this to happen then we advice you to use the `IContentValidationService` or `IContentService` instead.
+
+# Code time!
+Now that we have a converter we need a way to call the correct one depending on the macros found in an RTE value. For this we will Create a `MacroMigrationService` that holds the following method.
+
+```csharp
+public void Migrate(IEnumerable<int> PropertyDataIds)
+    {
+        using IScope scope = _scopeProvider.CreateScope(autoComplete: true);
```
ronaldbarendse, line 347: Do not use `autoComplete: true` when writing to the database, as exceptions withing the using block won't cause the transaction to be aborted/rolled back: remove this parameter and explicitly call `scope.Complete()` below.

15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
```diff
@@ -0,0 +1,889 @@
...
+                ? richTextEditorValue.Blocks.Layout[Constants.PropertyEditors.Aliases.RichText].ToList()
+                : new List<IBlockLayoutItem>();
+            layoutList.Add(layoutItem);
+            richTextEditorValue.Blocks.Layout[Constants.PropertyEditors.Aliases.RichText] = layoutList;
+            
+            // now that we have converted the data into a block, replace the macro tag by a block tag
+            richTextEditorValue.Markup = richTextEditorValue.Markup.ReplaceFirst(macroMatch.Value, $"<umb-rte-block-inline data-content-key=\"{blockKey}\"></umb-rte-block-inline>");
+        }
+        
+        return _jsonSerializer.Serialize(richTextEditorValue);
```
ronaldbarendse, line 318: Use `RichTextPropertyEditorHelper.SerializeRichTextEditorValue()` to serialize the RTE content, so we're doing exactly the same as the CMS (and Deploy for that matter).

ronaldbarendse (commented):
I've added my initial comments and would suggest another proof-read to fix all the typos...

15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
```diff
@@ -0,0 +1,889 @@
+---
+description: >-
+  Get started with developing a custom migration path for Macros to Blocks in the Rich Text Editors.
+---
+
+# Introduction
+As with many things in Umbraco, there are a multitude of ways on how to migrate away from macros and use the blocks in the RTE instead. In this article we will be showcasing a solution that lets you scan and then fix each macro one by one (or in batch). This solution will most likely not be used in a production setup, but we hope it canserve as an inspiration on how you can build your own solution specific your current setup. At the end of the article we will discuss a few other ways of running a larger migration.
```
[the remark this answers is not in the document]
Migaroez: The clarification is in the last sentence. I will see if I can reword things.

15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
```diff
@@ -0,0 +1,889 @@
...
+
+In this article we will be doing a one on one conversion from macro to block with each paramater matching the same named properties on an element document type. We are also keeping it quite simple and are using pure text as values. If your migration deals with more complex types, we advice you to create an instance of the new data format and compare the old value against the new as there might have been more differences between the paramater type on the macro and the propertyEditor/datatype on the element document type.
+
+Because most people will be dealing with this kind of migrations when they move between LTS version from 13 to 15, we chose to do just that in this article. More precisly from 13.7.2 to 15.2.3
```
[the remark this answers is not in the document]
Migaroez: It's fine that they get outdated. It is just there to mark that the code worked from/to those versions. It is not a step by step guide and more of a getting inspired article. I will add a clarification about this at the top of the page.

15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
Migaroez: Not important, it's up to the implementor to figure out the best way of determining macro values.

15/umbraco-cms/tutorials/inspiration/migrating-macros/migrating-macros.md
```diff
@@ -0,0 +1,891 @@
...
+
+In this article we will be doing a one on one conversion from macro to block with each paramater matching the same named properties on an element document type. We are also keeping it relatively simple and are using pure text as values. If your migration deals with more complex types, we advice you to create an instance of the new data format and compare the old value against the new as there might have been more differences between the paramater type on the macro and the propertyEditor/datatype on the element document type.
+
+Because most people will be dealing with this kind of migrations when they move between LTS version from 13 to 15, we chose to do just that in this article. More precisly from 13.7.2 to 15.2.3
```
nielslyngsoe, line 11: Could we mention something in this line:

"Notice this example would also be compatible if you where to migrate from 13 to 17".

In this way a future read do not think they have to go to the v.15 version, but they need to understand that they can go from LTS to LTS. :-)

AndyButland (approved):
I've reviewed and made a few updates to the text, just to align headings, fix spelling errors and resolve some of the linting warnings about long sentences.

I've marked this as ready for review, but @Migaroez or @nielslyngsoe, please switch that back if actually you have more amends to make here.

Thanks very much putting this together @Migaroez - it's a tricky one as, as you note, there are always going to be solution specific things to deal with.  But it's good to have this available as guidance as to how it can be handled.

PR #7037, opened 2025-04-21
Commerce 15.3.0: Refund and partial refund

umbracotrd:
## Description

- Add new pages explains Umbraco Commerce Refund and Partial Refund feature.
- Update Commerce Payment Providers page about the new partial refund feature.

## Type of suggestion

* [ ] Typo/grammar fix
* [x] New content
* [ ] Other

Umbraco Commerce 15.3.0

## Deadline (if relevant)
When Umbraco Commerce 15.3.0 is released, which is in 3 or 4 weeks.


umbracotrd:
@mattbrailsford : this is the PR for Commerce refund feature. If you have any suggestions about new pages or content, please let me know.

mattbrailsford:
I think this is enough content wise, but I would suggest pulling out any text in the images and have it within the written text. Text in images is not accessible. By all means put boxes round things to highlight elements, but any descriptive text should be in the written docs, not embedded in the images.

umbracotrd:
Thanks @mattbrailsford , I've paraphrased the text in images and put them to the written docs. I still keep the text inside the images though.

sofietoft:
Thanks for the PR @umbracotrd !

In order for us to review this properly, please add the new articles to the `SUMMARY.md` file.
Otherwise the articles will not show in the preview or on the published site once merged.

Also, please reconsider the text within the images.
As Matt mentions, this is not great for accessibility and if the images for some reason will not load, the user will miss out on important information, if the same is not added in the text.

umbracotrd:
@sofietoft : I've added new pages to the `SUMMARY.md`.

About the text within images, they're not exclusive information. All the information is copied as written text to the article, the texts inside images are just there to help those who can see. Removing texts from the images is just crippling majority of the audience.

15/umbraco-commerce/reference/payments/README.md
sofietoft, line 1: Is it intentional that only one of the sub-articles is linked to from here?

sofietoft (changes requested):
Thanks again for the PR!

I've updated formatting and structure of the article to follow our [styleguide](https://docs.umbraco.com/contributing/documentation/style-guide).
Please remember to follow that, when making PR to the documentation 🙏 

With the text from the images being added to the content as well, there is no reason to keep the text in the images. 
As I see it, no information will be lost without it. 
Keeping the text in the images will make the images harder to maintain should changes be made to the feature or UI in the future.
Adding an arrow to highlight a button or section is fine, as long as the color used matches out brand guidelines 🙌 

Furthermore, the images should be added to an `/images` directory next to the articles where they are used. Learn more about the structure in the [Styleguide](https://docs.umbraco.com/contributing/documentation/style-guide/structure).

I'd be happy to make the structural changes here and update the screenshots as well. However, that might slow down the timeline for when this PR will be merged.

15/umbraco-commerce/reference/payments/README.md
umbracotrd: I've added both articles to the readme

umbracotrd:
About the text inside images, the argument about making the images harder to maintain can be said the same even if there're no text inside them. When we update a page's content, we need to update both written text and the images, it doesn't matter if the images contain texts or not. If we worry about the effort to maintain the images, how about we don't have them in the first place. No images = no maintenance.

Would you rather read the whole point `7. Fill in the fields in the Order Refund modal` or see an image with red circles and arrows pointing exactly where you need to look and tell you which section is which on a single image? Like I said, removing the descriptive texts will just slow down the majority of the audience.

I'm not aware of our brand guidelines for arrow and square box. Please share me the article if you have them. Or, I can replace all images by plain screenshot with nothing added to avoid the hassles.

umbracotrd:
@sofietoft : I've removed texts and drawings from images. They're just screenshots now.

Can docs team reconsider about the text inside images? Being inclusive and supporting accessibility shouldn't mean making it equally hard for everyone.

sofietoft (approved):
Hi @umbracotrd !

Thanks for applying the suggestions and updating the images.
I see your point about it being quicker to understand and connect the information with the buttons/UI elements, when the text is in the images. But as mentioned, it is not good for accessibility and general maintenance of the docs.

I will discuss with the rest of team, whether we want to allow text within screenshots in the documentation going forward, or if there are alternative ways to go about adding information like this.

Thanks again for the PR.
I'll make sure it's merged when 15.4 is released 💪

mattbrailsford:
This is good to merge in now 👍

PR #7038, opened 2025-04-23
Updates relating to 16.0.0-rc

AndyButland:
## Description

Includes the following:

- Version specific upgrade guide breaking changes.
- Details of additional logging settings provided via https://github.com/umbraco/Umbraco-CMS/pull/19074
- Details of asynchronous package migrations provided via https://github.com/umbraco/Umbraco-CMS/pull/17057

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [X] Updates related to a new version
* [ ] Other

Umbraco CMS 16.0.0-rc

## Deadline (if relevant)

We should publish this when the RC is released.  For now, keeping in draft.

### DIFF
--- a/16/umbraco-cms/reference/configuration/loggingsettings.md
+++ b/16/umbraco-cms/reference/configuration/loggingsettings.md
@@ -13,7 +13,9 @@ The following configuration is available in the Logging settings:
   "CMS": {
     "Logging": {
       "MaxLogAge": "2.00:00:00",
-      "Directory": "~/CustomLogFileLocation"
+      "Directory": "~/CustomLogFileLocation",
+      "FileNameFormat": "UmbracoTraceLog.{0}..json",
+      "FileNameFormatArguments": "MachineName"
     }
   }
 }
@@ -30,3 +32,22 @@ To increase the maximum age of the entries in the audit log to 48 hours (2 days)
 By default, all log files are saved to the `umbraco/Logs` directory. You can define a custom directory for your log files by using the `Directory` key in the Logging settings.
 
 Set the value to `~/LogFiles` to add all log files to a `LogFiles` directory in the root of the file structure.
+
+## FileNameFormat
+
+The default file name format for the Umbraco log file is `UmbracoTraceLog.{0}..json`. The single argument is replaced at runtime with the server's machine name.
+
+If you want to change the file name or include additional arguments, you can amend the format with this setting.
+
+## FileNameFormatArguments
+
+By default the single argument for the log file format name is the server's machine name.
+
+Other or additional arguments can be provided via this setting as a comma delimited string:
+
+- `MachineName` - the server's name.
+- `EnvironmentName` - the ASP.NET environment name such as "Development" or "Production.
+
+So for example, to provide both supported arguments you would configure `MachineName,EnvironmentName`.
+
+The number of arguments provided should match the placeholders in the configured `FileNameFormat`.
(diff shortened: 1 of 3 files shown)


sofietoft:
Hi @AndyButland !

Will this include all Docs updates for Umbraco 16?

Also, the 16 docs is not published yet, and won't be until we publish the RC docs.
So, this can easily be merged before, as it will only be available here on GitHub until the RC release.

AndyButland:
I've suggested to the team they can add small things here - little updates as and where needed, like the one I've added.  Though likely anything more substantial will have their own PRs to make them easier to review.

sofietoft:
Great! I see we have a meeting tomorrow, so we should coordinate who writes what - just so we don't accidentally do double-work.

16/umbraco-cms/reference/configuration/loggingsettings.md
```diff
@@ -30,3 +32,22 @@ To increase the maximum age of the entries in the audit log to 48 hours (2 days)
 By default, all log files are saved to the `umbraco/Logs` directory. You can define a custom directory for your log files by using the `Directory` key in the Logging settings.
 
 Set the value to `~/LogFiles` to add all log files to a `LogFiles` directory in the root of the file structure.
+
+## FileNameFormat
+
+The default file name format for the Umbraco log file is `UmbracoTraceLog.{0}..json`. The single argument is replaced at runtime with the server's machine name.
+
+If you want to change the file name or include additional arguments, you can amend the format with this setting.
```
sofietoft, line 40:
```suggestion
If you want to change the file name or include additional arguments, you can amend the format with the `FileNameFormat` setting.
```

16/umbraco-cms/reference/configuration/loggingsettings.md
```diff
@@ -30,3 +32,22 @@ To increase the maximum age of the entries in the audit log to 48 hours (2 days)
...
+
+## FileNameFormatArguments
+
+By default the single argument for the log file format name is the server's machine name.
+
+Other or additional arguments can be provided via this setting as a comma delimited string:
```
sofietoft, line 46:
```suggestion
Other or additional arguments can be provided via the `FileNameFormatArguments` setting using a comma-delimited string:
```

sofietoft (commented):
I know this is still in draft mode, but since the RC is going on tomorrow, I thought it made sense to start the review 😅 

In addition, I have a question:
Should we remove the breaking changes for the previous versions?
Or perhaps only keep the supported versions there? 🤔 So, that would be 10, 13 and 15 in this case..
This goes for both breaking changes and version specific upgrade notes.

AndyButland:
Good point @sofietoft!  Thanks, and sorry for forgetting to mark this as ready for review.

On the breaking changes point, I think it's probably best to keep in actually.  Just as if you are going from say 13 to 16, it makes sense you read the breaking changes between 13 to 14 and 14 to 15, even if you are upgrading in one step.  You could go and find the 14 and 15 docs to see these, but it's an extra step and it may not be obvious that you should review them.

sofietoft:
Yes, good points about the breaking changes docs 🤔 
We could remove the RC and Beta ones for 14 though, right?
image

And perhaps 12 and 11, as they are no longer supported, so you wouldn't be updated to those - and I don't believe you even need to, in order to go from 10 to 13 🤔

sofietoft:
Yes, you are right. The breaking changes remain relevant, so we'll keep them 😄 
Just thought it'd be a good opportunity to clean up a bit.

I'll merge this, and make sure the RC docs are published tomorrow.

PR #7042, opened 2025-04-25
Update backoffice-entry-point.md

BatJan:
Fixed typos in code examples where the parameter for "extensionRegistry" reads "extensionsRegistry", which is not correct.

Also there was an example where it said "extension.unregister" instead of "extensionRegistry.unregister"

## Description

## Type of suggestion

* [x] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)


sofietoft (approved):
Great catch @BatJan ! Thanks for fixing these 🙌

PR #7102, opened 2025-05-21
Update README.md

jacklawry:
Add example for the Templates as per the release page https://our.umbraco.com/download/releases/1381

## Description

Use correct Bash and giving an example as it isn't clear that:

`dotnet new install Umbraco.Templates::13.*`

actually means:

`dotnet new install Umbraco.Templates::13.8.1`

to anyone who doesn't create new sites all the time.  
 
## Type of suggestion

* [* ] Typo/grammar fix
* [ ] New content
* [ *] Other

N/A

## Deadline (if relevant)

N/A

### DIFF
--- a/13/umbraco-cms/fundamentals/setup/install/README.md
+++ b/13/umbraco-cms/fundamentals/setup/install/README.md
@@ -11,7 +11,15 @@ The fastest way to get the latest version of Umbraco up and running is using the
 1. Open your command line.
 2. Install the Umbraco templates:
 
-`dotnet new install Umbraco.Templates::13.*`
+```bash
+dotnet new install Umbraco.Templates::13.*`
+```
+
+e.g.
+
+```bash
+dotnet new install Umbraco.Templates::13.8.1`
+```
 
 3. Create a new project:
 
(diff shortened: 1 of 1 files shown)


13/umbraco-cms/fundamentals/setup/install/README.md
```diff
@@ -11,7 +11,15 @@ The fastest way to get the latest version of Umbraco up and running is using the
...
-`dotnet new install Umbraco.Templates::13.*`
+```bash
+dotnet new install Umbraco.Templates::13.*`
+```
+
+e.g.
+
+```bash
+dotnet new install Umbraco.Templates::13.8.1`
+```
```
sofietoft, line 22: How about we add a bit more explanation to it as well:

```suggestion
{% hint style="info" %}

Replace `13.*` with the specific version you want to install.

Example:
```bash
dotnet new install Umbraco.Templates::13.8.1`
```

{% endhint %}
```

sofietoft (commented):
Thanks for the suggestion @jacklawry ! 🙌 

We don't add specific version numbers, as we would then have to go an update that number whenever a new patch or minor was released!
But I like mentioning how to actually use the versioning here.

I've made a suggestion - let me know what you think!

jacklawry:
Yes that looks good and will be a helpful reminder when people need to do a new install.

sofietoft:
Agreed @jacklawry ! 😄 

I've applied the changes. Will get it merged 💪 
Thanks again for the PR and for helping out with the docs!

PR #7115, opened 2025-05-23
New PR template

eshanrnh:
## Description

## Type of suggestion

* [ ] Typo/grammar fix
* [ ] New content
* [ ] Other

## Deadline (if relevant)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,34 @@
-## Description
+## 📋 Description
 
-_What did you add/update/change?_
+<!-- A clear and concise description of what this PR changes or adds. Include context if necessary. -->
 
-## Type of suggestion
+## 📎 Related Issues (if applicable)
 
-* [ ] Typo/grammar fix
-* [ ] Updated outdated content
-* [ ] New content
-* [ ] Updates related to a new version
-* [ ] Other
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,19 +1,34 @@
...
 
+* [ ] My changes follow the [Umbraco Documentation Style Guide](https://docs.umbraco.com/contributing/documentation/style-guide).
+* [ ] Code blocks are formatted correctly.
+* [ ] Sentences are short and clear (preferably under 25 words).
+* [ ] Avoided passive voice and first-person language (“we”, “I”).
+* [ ] Relevant pages are linked.
+* [ ] All links work and point to the correct resources.
+* [ ] Screenshots or diagrams are included (if useful).
+* [ ] I’ve tested any code examples or instructions where relevant.
+* [ ] Fixed typo/broken links/broken images
```
sofietoft, line 21: It think the language used for these bullets should be the same.

Right now, it's a mix of different formats.
Perhaps, the line before the list should be: "I've followed our style guidelines and can confirm that:"
And then each point in the list can be put directly after that sentence 🤔 

What do you think?

sofietoft (changes requested):

PR #7140, opened 2025-06-04
Update vite-package-setup.md

snappercrumpleddog:
The type declaration has changed in v15, but the documentation still has the old reference from v14.

## 📋 Description

The type declaration has changed in v15, but the documentation still has the old reference from v14.


eshanrnh (approved):
Looks good to me, @snappercrumpleddog 🙌 Thanks for updating it 💪

PR #7183, opened 2025-06-23
Updating CICD docs

AnnicaF:
## 📋 Description

In the V2 script, you need to copy 4 YAML files into the devops folder and not 3, as in the old V1 scripts.


jesp209i:
Great catch! Definitely an oversight when updating 🙈
Can I get you to also fix the same mistake on the GitHub article - it should have the same two mentions

jesp209i (approved):
Nice clarification 👍

PR #7220, opened 2025-07-04
Update key-vault.md

MMasey:
## 📋 Description

The documentation around where the connection string can be found was incorrect in the Azure Key Vault documentation. I've updated it to point to the correct location.


eshanrnh (approved):
The PR looks great, @MMasey 🙌

Issue #7313, opened 2025-08-15
Settings Objects - User Feedbacks


eshanrnh:
## What type of issue is it? (Choose one - delete the others)
Feedbacks
Missing documentation

## What article/section is this about?
[Settings Objects](https://docs.umbraco.com/umbraco-commerce/key-concepts/settings-objects)

## Describe the issue
**Feedback received for Settings Objects article:**

- Where's the actual reference documentation??!?! The article does not include reference documentation for available editor UIs.
- There is no clear overview of configuration options or settings related to editor UI elements.
- Additional guidance is needed to help users understand how to create Properties using the available editor UIs.

Note: Original feedback has been edited for tone. Please reach out to the Docs team to read the original feedback.

eshanrnh:
Thanks Matt 🙌

PR #7419, opened 2025-09-23
Add a reference to the CMS docs for a list of available property editor ui elements

mattbrailsford:
## 📋 Description

Add a reference to the CMS docs for a list of available property editor ui elements

#7313

### DIFF
--- a/16/umbraco-commerce/key-concepts/settings-objects.md
+++ b/16/umbraco-commerce/key-concepts/settings-objects.md
@@ -33,7 +33,7 @@ public class MyDiscountRewardProviderSettings
 Attributes define an optional `Key` parameter to override the default setting alias which would otherwise be the property name in camel case. An optional `EditorUiAlias` and `EditorConfig` options can also be defined to control the Umbraco property editor used to edit the given property. If no view is defined, one will attempt to be automatically chosen based on the property's value type.
 
 {% hint style="info" %}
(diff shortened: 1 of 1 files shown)


16/umbraco-commerce/key-concepts/settings-objects.md
```diff
@@ -30,7 +30,11 @@ public class MyDiscountRewardProviderSettings
 }
 ```
 
-Attributes define an optional `Key` parameter to override the default setting alias which would otherwise be the property name in camel case. An optional `EditorUiAlias` and `EdiutorConfig` options can also be defined to control the Umbraco property editor used to edit the given property. If no view is defined, one will attempt to be automatically chosen based on the property's value type.
+Attributes define an optional `Key` parameter to override the default setting alias which would otherwise be the property name in camel case. An optional `EditorUiAlias` and `EditorConfig` options can also be defined to control the Umbraco property editor used to edit the given property. If no view is defined, one will attempt to be automatically chosen based on the property's value type.
+
+{% hint style="info" %}
+See the CMS docs for a list of [available Property Editor UI elements and their config properties](/umbraco-cms/reference/property-editor-uis).
```
eshanrnh:
```suggestion
See the CMS documentation for a list of [available Property Editor UI elements and their config properties](/umbraco-cms/reference/property-editor-uis).
```

PR #7474, opened 2025-10-07
Global Contexts

bszyman:
## 📋 Description

Revised Global Contexts section for clarity and added best practices guidance.

## ✅ Contributor Checklist

* [ ] Code blocks are correctly formatted.
* [ ] Sentences are short and clear (preferably under 25 words).
* [ ] Passive voice and first-person language (“we”, “I”) are avoided.
* [ ] Typos, broken links, and broken images are fixed.

v16

## Deadline (if relevant)

N/A

## 📚 Helpful Resources

* 🧾 [Umbraco Contribution Guidelines](https://docs.umbraco.com/contributing)
* ✍️ [Umbraco Documentation Style Guide](https://docs.umbraco.com/contributing/documentation/style-guide)

### DIFF
--- a/16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
+++ b/16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
@@ -1,20 +1,20 @@
 ---
-description: Establish the bond for extensions to communication across the application
+description: Establish the bond for extensions to communication between packages and across the application
 ---
 
 # Global Context
 
-{% hint style="warning" %}
-This page is a work in progress and may undergo further revisions, updates, or amendments. The information contained herein is subject to change without notice.
-{% endhint %}
+Global Context extension types allow extension authors to extract out shared logic code and make it available across the entire backoffice. These contexts are kept alive through the entire backoffice session.
 
-A global context manages the logic code from your Lit Element controllers.
+Global Contexts are useful when a package needs to share functionality with other package authors.
+
+Extension authors should prefer to use other context types [Workspace Contexts](workspaces/workspace-context.md) over Global Contexts. Umbraco itself uses Global Contexts sparingly, for clipboard, current user, and icons.
 
 ## Registration of a Global Context
 
 You can register a global context like so:
 
-```typescript
+```json
 {
     type: 'globalContext',
     alias: 'My.GlobalContext.Products',


16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
```diff
@@ -1,24 +1,158 @@
...
+    setPreference(key: string, value: any): void {
+        this.#preferences.set(key, value);
+        console.log(`Preference set: ${key} = ${value}`);
+    }
+    
+    getPreference(key: string): any {
+        return this.#preferences.get(key);
+    }
+
+    getHostElement(): Element {
```
bszyman: `getHostElement()` is required by UmbContextMinimal, but I'm not sure what the intent of this is? To return a DOM element? The DOM element that the context is attached to? That's kind of how it seems to be getting used in the Umbraco backoffice samples I've found, but it seems odd.

https://apidocs.umbraco.com/v16/ui-api/interfaces/libs_context-api.UmbContextMinimal.html#gethostelement-1

iOvergaard: Agreed, it seems odd. We need @nielslyngsoe's perspective on this. We could provide any value as a context, including simple types such as string, number, and array, I thought.

bszyman: In my Umbraco project I'm using for testing all of this docs code, I just did `return new Element();` inside of there to keep the project compiling, but I feel dirty publishing that without further comment or without understanding what might blow up.

nielslyngsoe: Hi everyone :-)

The UmbContextMinimal abstraction/interface might confuse matters, so for understanding, I would say ignore that one. It only exists to avoid circular dependencies and separation of concerns.

So the particular need for the `getHostElement` is that this is the only part of a `UmbControllerHost` that the Context implementation must have.
But it could have been simpler if the example interface extended `UmbControllerHost`. But for someone not basing their class on UmbContextBase(Which is an extension of UmbControllerBase) will then have a more complex interface to adhere to — so that could be the downside to it, but maybe only relevant for complex use cases, like someone implementing a different framework, so maybe an aspect to ignore.

So you could avoid declaring the getHostElement by making the interface extend UmbControllerHost — this is all handled, from an implementation perspective, as your class extends UmbContextBase.

nielslyngsoe: ... and I may add that the `getHostElement` is supposed to return the element that hosts the controllers. For a Context Consumer or Provider, this is the element used to fire or listen for DOM Events. So this cannot be a virtual node, it needs to be the Element that the code belongs to. But @bszyman maybe in your case it was not needed, so it worked from a compiler perspective. But it would eventually result in things not working. :-)

16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
```diff
@@ -1,24 +1,164 @@
...
+    implements MyGlobalContextInterface
 {
-    type: 'globalContext',
-    alias: 'My.GlobalContext.Products',
-    name: 'My Products Context',
-    api: 'my-products.context.js',
+    #preferences: Map<string, any> = new Map();
+
+    constructor(host: UmbControllerHost) {
+        super(host, SERVICE_STATUS_CONTEXT);
```
nielslyngsoe: `SERVICE_STATUS_CONTEXT` does not exist in this example.. 

It maye be an idea to just change the name of `MY_GLOBAL_CONTEXT` and its string value `GlobalContext`. :-)
to SERVICE_STATUS_CONTEXT, so the examples are cohrent.

nielslyngsoe: Or maybe keep them seperate, then you just need to define the SERVICE_STATUS_CONTEXT token somewhere in this section as well :-)

16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
```diff
@@ -1,24 +1,164 @@
...
-{% endhint %}
+The Global Context extension type allows extension authors to create a custom context of data and functions and make it available across the entire backoffice. Unlike other contexts that might be scoped to a specific workspace or view, a global context is available everywhere. These contexts are kept alive through the entire backoffice session.
 
-A global context manages the logic code from your Lit Element controllers.
+Consider using a global context when you need to:
+
+- **Share state across multiple extensions** - Track user preferences or application-wide settings
+- **Provide utility functions** - Common operations that multiple parts of an extension
+- **Manage centralized services** - API clients, notification handlers, or data repositories
+- **Coordinate between different extensions** - Communication layer between dashboards, property editors, and other components
```
nielslyngsoe: It may be relevant to mention the Workspace Context for Property Editor communication. Cause I have a hard time understanding a common scenario for using the Global Context for Property Editors :-)

16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
```diff
@@ -1,24 +1,164 @@
...
-{% hint style="warning" %}
-This page is a work in progress and may undergo further revisions, updates, or amendments. The information contained herein is subject to change without notice.
-{% endhint %}
+The Global Context extension type allows extension authors to create a custom context of data and functions and make it available across the entire backoffice. Unlike other contexts that might be scoped to a specific workspace or view, a global context is available everywhere. These contexts are kept alive through the entire backoffice session.
 
-A global context manages the logic code from your Lit Element controllers.
+Consider using a global context when you need to:
+
+- **Share state across multiple extensions** - Track user preferences or application-wide settings
+- **Provide utility functions** - Common operations that multiple parts of an extension
```
nielslyngsoe: I don't agree with this one. Utilities should be shared via import maps or other ways.

16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
```diff
@@ -1,24 +1,164 @@
...
+- **Manage centralized services** - API clients, notification handlers, or data repositories
+- **Coordinate between different extensions** - Communication layer between dashboards, property editors, and other components
+- **Share functionality with other extensions** - Allow other package developers to access features and functionality within the package
+
```
nielslyngsoe, line 16: Could we add another option:
- **Fallback context** - Work as the fallback context if no other contexts matched for the scope of the consumer.

16/umbraco-cms/customizing/extending-overview/extension-types/global-context.md
bszyman: Probably best to just remove it entirely. Anyone looking for a context to use with property editors would probably be looking at the property editor doc page anyway.

PR #7486, opened 2025-10-09
Update cron expression

### LINKED ISSUE #7472: ScheduledRecordDeletion FirstTimeRun Notation
Wrong documentation?

[ScheduledRecordDeletion](https://docs.umbraco.com/umbraco-forms/developer/configuration#firstruntime)

## Describe the issue
It mentions a “first time run” with the cron expression `* 4 * * *`. I’m not sure about the exact implementation in Umbraco Forms, but this seems a bit confusing. The article states that the first run will be at 4 a.m., but as far as I know, `* 4 * * *` means every minute during the fourth hour (04:00–04:59).

So if my Umbraco instance restarts at 4:05 a.m., the first run would actually happen at the next minute (04:06), not at 4 a.m. the following day, right?

erikjanwestendorp:
## 📋 Description

See discussion in #7472

Fixes #7472

### DIFF
--- a/13/umbraco-cms/reference/configuration/healthchecks.md
+++ b/13/umbraco-cms/reference/configuration/healthchecks.md
@@ -19,7 +19,7 @@ An example of a HealthChecks settings can look something like this:
       ],
       "Notification": {
         "Enabled": true,
-        "FirstRunTime": "* 4 * * *",
+        "FirstRunTime": "0 4 * * *",
         "Period": "1.00:00:00",
         "NotificationMethods": {
           "email": {
--- a/15/umbraco-cms/reference/configuration/healthchecks.md
+++ b/15/umbraco-cms/reference/configuration/healthchecks.md
@@ -19,7 +19,7 @@ An example of a HealthChecks settings can look something like this:
       ],
       "Notification": {
         "Enabled": true,
-        "FirstRunTime": "* 4 * * *",
+        "FirstRunTime": "0 4 * * *",
         "Period": "1.00:00:00",
         "NotificationMethods": {
           "email": {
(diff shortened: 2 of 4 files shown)


13/umbraco-forms/developer/configuration/README.md
```diff
@@ -392,7 +392,7 @@ Set to `true` to enabled the background task.
 
 #### FirstRunTime
 
-This will configure when the record deletion process will run for the first time. If the value is not configured the health checks will run after a short delay following the website start. The value is specified as a string in crontab format. For example, a value of `"* 4 * * *"` will first run the operation at 4 a.m.
+This will configure when the record deletion process will run for the first time. If the value is not configured the health checks will run after a short delay following the website start. The value is specified as a string in crontab format. For example, a value of `"0 4 * * *"` will first run the operation at 4 a.m.
```
eshanrnh:
```suggestion
This setting configures when the record deletion process will run for the first time. If the value is not configured, the health checks will run 3 minutes after the website starts (this delay is hardcoded). The value is specified as a string in crontab format. For example, a value of `"0 4 * * *"` schedules the operation to run daily at 04:00.
```

ronaldbarendse: This text was probably copied from the Health Checks documentation and not updated correctly (hence still referring to the health checks). And the 3 minute delay is a default value, which inherently is hardcoded. Also, the crontab value doesn't specify the recurring part, only the first time it runs. After the initial run, the `Period` configuration specifies when the process runs next.

```suggestion
This setting configures when the record deletion process will run for the first time. If the value is not configured, the process will run 3 minutes after the website starts. The value is specified as a string in crontab format. For example, a value of `"0 4 * * *"` schedules the operation to start at 04:00.
```

PR #7492, opened 2025-10-10
Removal of Dashboard Configuration for V17

NillasKA:
## 📋 Description

We have not had any use for this dashboard configuration in an official capacity since V13. With the re-introduction of the news dashboard in V17 we won't be using any json based documentation like this one. Therefore we wont be needing it.

CMS V17


NillasKA:
> @NillasKA Can you also remove the entry from the `SUMMARY.md` file?

 @eshanrnh Yes - I have removed this now, slight oversight but good spot 😄

eshanrnh (approved):
Looks great 🙌

PR #7696, opened 2025-12-08
Update api rate limiting

### LINKED ISSUE #7095: Update the API rate limiting article
Discussion

[API rate limiting](https://docs.umbraco.com/umbraco-cms/reference/security/api-rate-limiting)

## Describe the issue
When implementing API rate limiting as described [here](https://docs.umbraco.com/umbraco-cms/reference/security/api-rate-limiting), the default setup relies on IP-based limiting via IpRateLimiting. However, in many production environments, Umbraco is hosted behind a Web Application Firewall (WAF) or reverse proxy (such as Cloudflare, Azure Front Door, NGINX, etc.). In these cases, the client IP address is typically replaced by the proxy’s IP, causing the rate limiting to apply to the WAF instead of the actual user.

Suggestion:

It would be helpful if the documentation included a clear note that:

When running Umbraco behind a WAF or reverse proxy, it's necessary to configure the RealIpHeader setting in `appsettings.json` to correctly extract the original client IP.

Somthing like: 

```json
"IpRateLimiting": {
  "RealIpHeader": "X-Forwarded-For",
  "HttpStatusCode": 429,
  ...
}

```

erikjanwestendorp:
## 📋 Description

Add a note as suggested here: 

Fixes #7095

### DIFF
--- a/16/umbraco-cms/reference/security/api-rate-limiting.md
+++ b/16/umbraco-cms/reference/security/api-rate-limiting.md
@@ -258,5 +258,5 @@ public class ApiRateLimiterComposer : IComposer
 {% endcode %}
 
 {% hint style="info" %}
-When Umbraco runs behind a WAF or reverse proxy, rate-limiting may fail if the client IP address is not forwarded correctly. Configure your proxy or WAF to send the original client IP using headers like X-Forwarded-For. Ensure Umbraco reads these forwarded headers, or all requests may appear from one IP and cause incorrect rate-limit enforcement.
+When Umbraco runs behind a WAF or reverse proxy, rate-limiting may fail if the client IP address is not forwarded correctly. Configure your proxy or WAF to send the original client IP using headers like X-Forwarded-For. This will prevent all requests appearing to come from one IP address which would cause incorrect rate-limit enforcement.
 {% endhint %}
--- a/17/umbraco-cms/reference/security/api-rate-limiting.md
+++ b/17/umbraco-cms/reference/security/api-rate-limiting.md
@@ -258,5 +258,5 @@ public class ApiRateLimiterComposer : IComposer
 {% endcode %}
 
 {% hint style="info" %}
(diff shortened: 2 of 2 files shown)


sofietoft:

13/umbraco-cms/reference/security/api-rate-limiting.md
```diff
@@ -255,4 +255,8 @@ public class ApiRateLimiterComposer : IComposer
     }
 }
 ```
-{% endcode %}
\ No newline at end of file
+{% endcode %}
+
+{% hint style="info" %}
+When Umbraco runs behind a WAF or reverse proxy, rate-limiting may fail if the client IP address is not forwarded correctly. Configure your proxy or WAF to send the original client IP using headers like X-Forwarded-For. Ensure Umbraco reads these forwarded headers, or all requests may appear from one IP and cause incorrect rate-limit enforcement.
```
sofietoft: As Andy suggested, rewritten the last sentence of the hint. What do you think @erikjanwestendorp ? Does the note still make sense to you, with this change?

```suggestion
When Umbraco runs behind a WAF or reverse proxy, rate-limiting may fail if the client IP address is not forwarded correctly. Configure your proxy or WAF to send the original client IP using headers like X-Forwarded-For. This will prevent all requests appearing to come from one IP address which would cause incorrect rate-limit enforcement.
```

sofietoft:
Thanks @erikjanwestendorp ! 💪 

I'll get this merged.
Should we close the issue along with it? Not sure, if there was some plans or ideas to extend this at some point? 🤔

erikjanwestendorp:
@sofietoft  We can close the issue, but ideally it would be better to expand the documentation with code examples. However, I still need to find the time to properly test this before I can create a PR 🙈

sofietoft:
No worries @erikjanwestendorp ! 😄 

I'll close the issue along with this PR for now. We can always re-open or create a new one down the line.

Thanks again for pitching in with this one.
Happy Holidays!

PR #7884, opened 2026-03-11
Fix link for MFA enforcement documentation

DitteKKoustrup:
Updated the link for Multi-Factor Authentication enforcement to point to the correct documentation, since it lead to a 404 page before.

Not sure if how I added the link is correct, so feel free to change that 🤗


sofietoft (approved):
Thanks @DitteKKoustrup 👏

PR #7939, opened 2026-04-07
Updates info about default Block Grid templates

aguyfromdenmark:
## 📋 Description

Updates info about default Block Grid templates. Fixes invalid link and changes repository path to correct path.


sofietoft:
Good catch @aguyfromdenmark ! 🙏 

I'll merge this right away.

PR #7956, opened 2026-04-13
Restructuring the AI docs

sofietoft:
## 📋 Description

I'll set up a test space on GitBook to verify that the new structure works.

I've kept the original SUMMARY files in each section, to avoid breaking anything when this is merged.
That way, we control when the new changes are live.


eshanrnh (approved):
Looks great🙌

PR #8011, opened 2026-04-26
Fix "occours" typo across Heartcore API docs

ZLeventer:
The Errors section of every Heartcore API article opens with "If an error occours you will receive a HTTP status code...". `occours` should be `occurs`. Fixed in all 12 affected files (Content Delivery + Content Management).


sofietoft (approved):
Another one 😄 
Thanks @ZLeventer ! 💪

PR #8012, opened 2026-04-26
Fix "accross" typo in Adding Language Variants tutorial

ZLeventer:
Step 6 of the "Adding Language Variants" tutorial says "disable **Shared accross cultures**" — should be "across". Fixed in v16, v17, and v18.


sofietoft (approved):
I feel like I keep repeating myself here 😄 
I don't mind it though. Thanks again @ZLeventer ! 💪

PR #8057, opened 2026-05-13
Add hint for UMBRACO-CLOUD license key for Deploy and Forms on v14+

manutdkid77:
## 📋 Description

Adds a hint block to the "Configure Licenses" article clarifying that Umbraco Deploy and Umbraco Forms on Umbraco Cloud (v14+) can use `UMBRACO-CLOUD` as the license key.

If not configured for Cloud projects we get this message in the Umbraco Forms section for example:
image

### DIFF
--- a/marketplace-and-integrations/configure-licenses.md
+++ b/marketplace-and-integrations/configure-licenses.md
@@ -84,6 +84,24 @@ You might run into issues when using a period in the product name when using env
 
 {% endhint %}
 
+{% hint style="info" %}
+For version 14 and above, if you are using Umbraco Deploy and Umbraco Forms on Umbraco Cloud, you can use `UMBRACO-CLOUD` as the license key.
+
+```json
+{
+  "Umbraco": {
+    "Licenses": {
+      "Products": {
+        "Umbraco.Deploy": "UMBRACO-CLOUD",
+        "Umbraco.Forms": "UMBRACO-CLOUD"
+      }
+    }
+  }
+}
+```
+
+{% endhint %}
+
 ## Verifying the License Installation
 
 You can verify that your license is successfully installed by logging into your project's backoffice:


marketplace-and-integrations/configure-licenses.md
```diff
@@ -84,6 +84,24 @@ You might run into issues when using a period in the product name when using env
 
 {% endhint %}
 
+{% hint style="info" %}
+For version 14 and above, if you are using Umbraco Deploy and Umbraco Forms on Umbraco Cloud, you can use `UMBRACO-CLOUD` as the license key.
```
ronaldbarendse, line 88: This is only supported for versions 17 and up (see [announcement](https://github.com/umbraco/Announcements/issues/25):

```suggestion
For version 17 and above, if you are using Umbraco Deploy and Umbraco Forms on Umbraco Cloud, you can use `UMBRACO-CLOUD` as the license key.
```

ronaldbarendse (commented):
I'm not sure whether we need to explicitly mention this here, since the correct license keys are already pre-configured on new Umbraco Cloud projects and if you're upgrading, mentioned in the [Cloud-specific upgrade steps](https://docs.umbraco.com/umbraco-cloud/optimize-and-maintain-your-site/manage-product-upgrades/product-upgrades/major-upgrades#umbraco-17).

Versions below 17 still use the license files on Umbraco Cloud. So if we do add this, we might want to include that in this message as well. But again: these license files are already present on new projects, so I don't see why we should document that here either...

sofietoft:
Hi @manutdkid77 👋 Thank you for taking the time to contribute and for sharing the screenshot — that's really helpful context!

After reviewing this, we've decided not to include it in the documentation. Since Forms and Deploy are already included in every Umbraco Cloud project, the licensing should be handled automatically and isn't something Cloud users need to configure manually. Adding documentation for it could actually cause more confusion than it resolves.

That said, I'm still curious — how did you end up in a situation where you needed to set this manually?
It sounds like something may not have been set up correctly on the Cloud project side, and I'd like to understand the root cause so we can ensure others don't run into the same thing.

Thanks again for the contribution! 🙏

manutdkid77:
@sofietoft @ronaldbarendse Thank you for looking into this 🙏

I came across this when upgrading a v13 project to v17.

sofietoft:
Thanks @manutdkid77 !

We are currently working on updating the upgrade articles for Umbraco Cloud.
I'll make sure this is highlighted better 🙏

PR #8085, opened 2026-05-22
Added todays release notes

cornehoskam:
Hi Sofie! Some more release notes for you 😇⭐


sofietoft (commented):
Thanks @cornehoskam !

I see you're using a different spelling for the word "personalization/personalisation".
Am I correct to assume that it's because the spelling in the UI uses the UK way (personalisation)? 🤔 
If that's the case, we'll ignore the warnings 🙏

PR #8160, opened 2026-06-16
Documented sort children notification settings (due in CMS 17.6 and 18.1)

AndyButland:
## 📋 Description

Adds details of the `SortChildrenByFieldFiresNotifications` setting added in CMS 17.6 and 18.1.

For now I've only referenced calling this from the management API, but we should update to reference the backoffice feature that calls the API when it's ready.

https://github.com/umbraco/Umbraco-CMS/pull/23077

## ✅ Contributor Checklist

* [ ] Code blocks are correctly formatted.
* [X] Sentences are short and clear (preferably under 25 words).
* [X] Passive voice and first-person language (“we”, “I”) are avoided.
* [X] Typos, broken links, and broken images are fixed.

CMS 17.6 and 18.1

## Deadline (if relevant)

Should be published on 23rd July, along with the related release candidates.

### DIFF
--- a/17/umbraco-cms/develop-with-umbraco/configuration/contentsettings.md
+++ b/17/umbraco-cms/develop-with-umbraco/configuration/contentsettings.md
@@ -47,6 +47,7 @@ The following snippet will give an overview of the keys and values in the conten
       },
       "PreviewBadge": "<![CDATA[<b>My HTML here</b>]]>",
       "ResolveUrlsFromTextString": false,
+      "SortChildrenByFieldFiresNotifications": false,
       "ShowDeprecatedPropertyEditors": false,
       "ShowDomainWarnings": true,
       "ShowUnroutableContentWarnings": true,
@@ -159,6 +160,14 @@ This allows you to customize the preview badge being shown when you're previewin
 ### Resolve urls from text string
 
 This setting is used when you're running Umbraco in virtual directories. Setting this to true can increase render time for pages with a large number of links. However, this is required if Umbraco is running in a virtual directory.
+
+### Sort children by field fires notifications
+
+This setting controls how sorting the children of a node by a field is persisted. Sorting by a field is available through the Management API, which can reorder a node's entire set of children in a single operation.
+
+By default (`false`), the children are reordered with a single set-based database update and a single branch cache refresh. A single audit entry is written, but no per-item save/sort notifications are fired. As a result, webhooks that depend on those notifications are not triggered. This keeps the operation efficient on nodes with many children.
+
+Set this to `true` to restore per-item save/sort notifications (and the webhooks that depend on them). This routes the field sort through the standard per-item sort path, accepting the additional performance cost on nodes with many children.
 
 ### Show deprecated property editors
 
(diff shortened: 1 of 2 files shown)


17/umbraco-cms/develop-with-umbraco/configuration/contentsettings.md
```diff
@@ -160,6 +161,14 @@ This allows you to customize the preview badge being shown when you're previewin
 
 This setting is used when you're running Umbraco in virtual directories. Setting this to true can increase render time for pages with a large number of links. However, this is required if Umbraco is running in a virtual directory.
 
+### Sort children by field fires notifications
+
+This setting controls how sorting the children of a node by a field is persisted. Sorting by a field is available through the Management API, which can reorder a node's entire set of children in a single operation.
+
+By default (`false`), the children are reordered with a single set-based database update and a single branch cache refresh. A single audit entry is written, but no per-item save/sort notifications are fired. As a result, webhooks that depend on those notifications are not triggered. This keeps the operation efficient on nodes with many children.
+
+Set this to `true` to restore per-item save/sort notifications (and the webhooks that depend on them). This routes the field sort through the standard per-item sort path, accepting the additional performance cost on nodes with many children.
```
kjac: These paragraphs are very code-near (mentions set-based DB updates, per-item sort path etc.).

The points are of course valid, but they're also noise to the reader. I'd like it to be more clear and concise - more to the point of "what are the direct implications of this setting" (notifications and webhooks firing), and then maybe a closing remark about how the default value benefits performance.

17/umbraco-cms/develop-with-umbraco/configuration/contentsettings.md
```diff
@@ -160,6 +161,14 @@ This allows you to customize the preview badge being shown when you're previewin
 
 This setting is used when you're running Umbraco in virtual directories. Setting this to true can increase render time for pages with a large number of links. However, this is required if Umbraco is running in a virtual directory.
 
+### Sort children by field fires notifications
+
+This setting controls how sorting the children of a node by a field is persisted. Sorting by a field is available through the Management API, which can reorder a node's entire set of children in a single operation.
```
kjac: This setting does not _really_ control how the sorting of a single node is persisted (we still use SortOrder), but rather the mechanics around calculating and persisting the sort order of a set of nodes.

I rather think this paragraph could be removed entirely without sacrificing any important information.

Perhaps the leading paragraph could instead be an intro to sort-by-field and/or an explanation of when this is invoked - and when "regular" sort is invoked.

AndyButland:
Good points @kjac - though I'm balancing here a couple of things:

- We don't have UI for this yet, so it's hard to talk about different options for sort.  That should come later once the backoffice implementation is done.  For now I'm just documenting an API operation.
- This setting isn't going to be very prominent, so I did want to get into the trade off of notification fidelity and performance so it's clear why you'd choose one option or the other.

Rewriting though, I've come up with this:

> Sorting by a field is available through the Management API, which can reorder a node's entire set of children in a single operation. This is an alternative mechanism to the per-item sort currently supported via the backoffice. 

> By default (`false`), the children are reordered in a performant "set-based" operation. A single audit entry is written, but no per-item save/sort notifications are fired. As a result, webhooks that depend on notifications are not triggered.

> Set this to `true` to restore per-item save/sort notifications (and the webhooks that depend on them). This follows the per-item sort path, accepting the additional performance cost on nodes with many children.

How does that sound to you?

kjac (approved):

kjac:
Sounds good to me 👍

eshanrnh (approved):
Looks good 🙌

PR #8202, opened 2026-06-30
v17 Developer MCP docs: target the lts-17 dist-tag

hifi-phil:
## What
Updates the **v17** CMS Developer MCP docs to install `@umbraco-cms/mcp-dev@lts-17` instead of a specific pinned version.

- Replaces the previously pinned `@17.4.2` (added so v17 readers didn't pick up v18) across the local MCP setup guides and the Developer MCP README.
- Also updates older `@17.1` references — the mcp-chaining and creating-media install snippets, and the version compatibility table — to `@lts-17`.

## Why
The v17 line is published on npm under the **`lts-17`** dist-tag (per the release pipeline, which scopes the v17 line to `lts-17`). v17 users should track the latest 17.x release via that tag rather than a fixed version. The npm `latest` tag now points at `18.0.0`, so an unpinned install would otherwise jump to v18. Targeting `@lts-17` keeps v17 users on the latest compatible 17.x build.

Verified locally with Vale (`0 errors, 0 warnings`) on all changed files.

### DIFF
--- a/17/umbraco-in-ai/mcp/base-mcp/sdk/mcp-chaining.md
+++ b/17/umbraco-in-ai/mcp/base-mcp/sdk/mcp-chaining.md
@@ -32,7 +32,7 @@ const servers: McpServerConfig[] = [
   {
     name: "cms",
     command: "npx",
-    args: ["-y", "@umbraco-cms/mcp-dev@17.1"],
+    args: ["-y", "@umbraco-cms/mcp-dev@lts-17"],
     // Environment variables from the parent process are inherited automatically.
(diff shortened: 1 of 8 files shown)


sofietoft (approved):

sofietoft:
When should these changes be published, @hifi-phil ?

hifi-phil:
@sofietoft we can publish now. I've just released 17.5 which has the new tag in place and means that we don't have to update the docs for the version number going forward. It will just use the latest 17-lts version that is available

PR #8277, opened 2026-07-24
Clarify Checkout view overrides require a rebuild on Umbraco 17+

mattbrailsford:
## What & why

The [Customize Checkout](https://docs.umbraco.com/umbraco-commerce-packages/checkout/customize-checkout) guide tells users to drop overriding Views into `Views/UmbracoCommerceCheckout`, make a change, and verify it appears. On a default **Umbraco 17/18** project this silently does nothing — reported in [Umbraco.Commerce.Issues#865](https://github.com/umbraco/Umbraco.Commerce.Issues/issues/865).

### Root cause (verified end-to-end)

The Checkout package ships its Views **precompiled** (`AddRazorSupportForMvc`). An override only wins if the **host app also compiles a View at that path at build time** — the app's compiled View then out-ranks the package's precompiled one (application-part ordering).

Default projects are created in `BackofficeDevelopment` development mode, whose `.csproj` sets `RazorCompileOnBuild=false` / `RazorCompileOnPublish=false`. So the app never compiles the dropped-in View, and the package's precompiled View keeps being used — even after a rebuild.

Razor **runtime compilation does not work around this.** It compiles on-disk Views only at paths with *no* precompiled View (that's what live-edits your own templates); it does **not** displace a package's precompiled View. So `Umbraco.Cms.DevelopmentMode.Backoffice` doesn't help.

Verified in a real **Umbraco 17.5.3 + Commerce 17.1.6 + Checkout 17.0.0** site (the reporter's exact versions):

| Scenario | Result |
|---|---|
| Default (`RazorCompileOnBuild=false`, `DevelopmentMode.Backoffice` active) | Override **ignored** — `assembly = Umbraco.Commerce.Checkout` |
| `RazorCompileOnBuild=true` | Override **wins** — renders the custom markup |
| Fresh path with no precompiled competitor | Runtime-compiled from disk, live-edits without rebuild |
| `RazorCompileOnBuild=true` + a View referencing an `InMemoryAuto` model | Build **fails** (`CS0234`) — why the source-code Models Builder step is required |

### Change

Rewrites the page to explain: overriding needs **build-time** compilation; the default `BackofficeDevelopment` mode disables it (`RazorCompileOnBuild=false`); runtime compilation can't displace a package's precompiled View; the fix is a source-code Models Builder mode + build-time compilation (which also aligns with Production readiness); and the **Template** alternative (distinct path, renders normally, but receives `IPublishedContent` rather than the checkout view model).

Single unversioned page under `commerce-add-ons/`, so one edit covers both v17 and v18 readers.

> Kept as **draft** pending final review. Supersedes the two earlier commits on this branch, whose guidance was incorrect.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/commerce-add-ons/packages/checkout/customize-checkout.md
+++ b/commerce-add-ons/packages/checkout/customize-checkout.md
@@ -17,7 +17,19 @@ To do this follow these steps:
 1. Copy the equivalent [files and partials](https://github.com/umbraco/Umbraco.Commerce.Checkout/tree/main/src/Umbraco.Commerce.Checkout/Views/UmbracoCommerceCheckout).
 2. Add them to `Views/UmbracoCommerceCheckout` in your project directory. It might be necessary to create the folder first.
 3. Make a small text change to one of the Views to verify that the files are in use.
-4. Verify that the changes are carried out and displayed correctly.
+4. Rebuild and restart your site, then verify that the changes are carried out and displayed correctly.
+
+The Checkout package ships its Views precompiled. When you add a matching View to your project's `Views/UmbracoCommerceCheckout` folder, your copy takes precedence over the packaged one, letting you override a single step without having to redefine them all.
+
+{% hint style="info" %}
+**Umbraco 17 and later (including Umbraco 18): view changes require a rebuild**
+
+From Umbraco CMS 17 onwards, Razor runtime compilation is no longer enabled by default. It has been removed from the CMS core and moved to the optional [`Umbraco.Cms.DevelopmentMode.Backoffice`](https://docs.umbraco.com/umbraco-cms/reference/configuration/runtimesettings) package.
+
+Overriding Views still works, but your overrides are compiled into the site at **build time**. After adding or editing a View, you must **rebuild and restart** the site for the changes to appear - editing a `.cshtml` file and only refreshing the browser will have no effect.
+
+To restore the previous behavior, where View changes are picked up without a rebuild during local development, add the `Umbraco.Cms.DevelopmentMode.Backoffice` package to your project. For background, see the [Umbraco CMS version-specific upgrade notes](https://docs.umbraco.com/umbraco-cms/get-started/upgrading-and-migrating/version-specific).
+{% endhint %}
 
 You are now ready to start customizing the Checkout page to fit the design of your website.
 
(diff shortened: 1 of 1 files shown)


commerce-add-ons/packages/checkout/customize-checkout.md
```diff
@@ -8,18 +8,38 @@ It is assumed that you already have an Umbraco website configured Umbraco Commer
 
 Umbraco Commerce Checkout is a free and open-source add-on package for Umbraco Commerce. It is possible to amend the default behavior to customize the checkout to your needs.
 
-## Setup
+## Overriding the checkout Views
 
-To allow customization you must first 'override' the existing files for the step required to be modified.
+Checkout's Views are precompiled into the package. To customize one, add your own copy at the same path under `Views/UmbracoCommerceCheckout`, and make sure your site compiles Views at build time so your copy is used instead of the packaged one.
```
sofietoft:
```suggestion
To customize one, add your copy to the same path under Views/UmbracoCommerceCheckout. Ensure your site compiles Views at build time so your copy is used instead of the packaged one.
```

commerce-add-ons/packages/checkout/customize-checkout.md
```diff
@@ -8,18 +8,38 @@ It is assumed that you already have an Umbraco website configured Umbraco Commer
...
-4. Verify that the changes are carried out and displayed correctly.
+3. Make your changes.
+4. Rebuild and restart your site, then verify that the change is displayed correctly.
+
+## Alternative: use a Template
+
+Instead of overriding the Views, you can assign a **Template** to the checkout page in the backoffice and place your markup there. Checkout will use the Template instead of its built-in View.
 
-You are now ready to start customizing the Checkout page to fit the design of your website.
+With this approach the Template receives the checkout page's `IPublishedContent` as its model - not the view model the packaged Views use - so you build that step's markup yourself against the Umbraco Commerce APIs.
```
sofietoft:
```suggestion
With this approach, the Template receives the checkout page's IPublishedContent as its model—not the packaged Views' view model.

You then build that step's markup yourself against the Umbraco Commerce APIs.
```

sofietoft (commented):
I've made a couple of suggestions, to fix the long sentences. Let me know if something doesn't sound quite right.

mattbrailsford:
LGTM

sofietoft:
Can this be published right away?

mattbrailsford:
it can yes 👍

PR #8285, opened 2026-07-31
Forms: Guard the UTC timestamp correction script against re-runs

rickbutterfield:
The `correct-utc-timestamps.sql` script published with the Umbraco Forms 17.3.0 release notes has no upper bound on any of its statements. Every one reads `WHERE Created > @UpgradeDate`.

Forms has written correct UTC timestamps since 17.3.0, so a site running the script today shifts those already-correct rows a second time. This came up on https://github.com/umbraco/Umbraco.Forms.Issues/issues/1759, where the site needs the historical fix but is now on 17.4.6.

## Changes

Script:

- New `@FixDate` variable, applied as an upper bound to all 14 statements.
- A marker row in `umbracoKeyValue`. A second run reports the marker and exits without changing rows.
- Wrapped in a single transaction with `SET XACT_ABORT ON`.
- A preview `SELECT` reporting the affected entry count and date span.
- Deletes the affected days from `UFAnalyticsDailySummary` and `UFAnalyticsProcessedDates`, so the background task rebuilds the Analytics charts from the corrected entries. Both tables hold only derived values.
- Guard on `@FixDate` being later than `@UpgradeDate`, and a note on the boundary comparison being approximate.

Release notes:

- The v17.3.0 hint lists all three variables, the backup step, the marker behaviour, and how to confirm the analytics rebuild.

Applied to all four copies of the script (17 and 18, each under `scripts/` and `.gitbook/assets/`).

## Verification

- The script parses against SQL Server with `SET PARSEONLY ON` (exit 0, no errors).
- Vale reports no new issues. The one remaining warning on the `{% file %}` caption line predates this branch.

### DIFF
--- a/17/umbraco-forms/.gitbook/assets/correct-utc-timestamps.sql
+++ b/17/umbraco-forms/.gitbook/assets/correct-utc-timestamps.sql
@@ -1,25 +1,116 @@
--- Correct post-v17 data that was stored with DateTime.Now instead of DateTime.UtcNow.
--- @TimeZone: your server Windows timezone name (e.g. Romance Standard Time)
--- @UpgradeDate: approximate date you first upgraded to v17.0.0
---               (rows BEFORE this date were already correctly migrated to UTC)
+/*
+    Umbraco Forms: correct system dates that were stored as local server time instead of UTC.
 
+    This applies to sites that ran Umbraco Forms 17.0.0 to 17.2.x. In those versions the v17.0.0
+    migration converted existing dates to UTC, but new rows were still written with local server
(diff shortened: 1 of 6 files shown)


sofietoft (approved):

sofietoft:
Should the release notes also be applied for version 18, or is that left out intentionally?

rickbutterfield:
Great point @sofietoft! It makes most sense to go in the v18 version specific notes, rather than release notes. I'll get it updated.

sofietoft (approved):

sofietoft:
This can be merged and published now, right?

rickbutterfield:
Yes please @sofietoft!

PR #8294, opened 2026-08-05
Release notes for Engage 17.4 and 18.1

nathanwoulfe:
Adds release notes for Engage 17.4 and 18.1
Also back-fills notes for 17.3 and 17.3.1. 18.1 is a roll-up of all changes from 17.3-17.4

### DIFF
--- a/17/umbraco-engage/release-notes.md
+++ b/17/umbraco-engage/release-notes.md
@@ -16,6 +16,67 @@ If you are upgrading to a new major version, check the breaking changes in the [
 
 Below are the release notes for Umbraco Engage 17, detailing all changes in this version.
 
+#### [17.4.0](https://www.nuget.org/packages/Umbraco.Engage/17.4.0) (August 2026)
+
+This release is dominated by a rework of how A/B test and personalization variants are previewed, together with a batch of fixes for multilingual and invariant content.
+
+**Previewing**
+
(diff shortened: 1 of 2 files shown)


nathanwoulfe:
@cornehoskam and @andy-ub - let me know if anything is missing or should be removed.

cornehoskam:
@nathanwoulfe I made a handful of changes, and dropped the (internal) additions for AI integration from the public facing changelog.

andy-ub:
Thanks @nathanwoulfe , traced my side.

Server-side paging: hold the line in both versions. The table rewire is merged but the server-side search work is not, and the two were coupled on purpose: paging on its own leaves a search box that only filters the page on screen. 18.1 is further off, because the v18 port hasn't started, so nothing is behind that line yet.

Percentage-change: remove from both. On v17 the differences are still computed across the whole merged row set and then indexed with the page-local row index, so page 2 renders rows 10-19 against differences 0-9. I ran into it during the table rewire and left it as pre-existing.

The unresolved node key line: 18.1 lost it in the 25-word lint pass (a8f54c4e), which took one more line out of the 18 file than out of the 17 file. That fix is in both versions so it should come back.

Lift vs control: is the v17 side documented anywhere? I can't find it at any v17 version.

Heatmap: my understanding of the preview fix is that the main half was preview never exiting after the heatmap closed, not only ending early on a variant switch.

"Every test type can now be previewed" contradicts the Split URL exception two bullets down.

Everything else traced to a tracked change or shipped code.

Issue #8301, opened 2026-08-06
Vale sentence-length is greedy and should include common formatting (hyphen, semver)


nathanwoulfe:
The configuration for sentence-length linting is a bit greedy and returns false-positives.

```Bumped the **Umbraco.Licenses** dependency to 18.0.2. License product IDs are now matched case-insensitively, so a license configured under a differently-cased key still resolves.```

is flagged as a long sentence, but is two separate sentences of six and 17 words each. Reviewer treats this as 28 words:

- `Umbraco.Licenses` => Umbraco, Licenses (2 terms)
- `18.0.2` => 18, 0, 2 (3 terms)
- `case-insensitively` => case, insensitively (2 terms)
- `differently-cased` => differently, cased (2 terms)

Regex used for matching really needs to include `-` and `.`, and potentially additional punctuation when enclosed by alpha-numeric characters.

sofietoft:
Yes, I agree 100%.

I'll have Claude give it a look later today 🙏 
We'll ignore the "long sentence" warnings in the Engage release notes PR.

nathanwoulfe:
You're wonderful, thanks @sofietoft

PR #8303, opened 2026-08-06
Updated the SentenceLenght rule to be more accepting

sofietoft:
## 📋 Description

Updated the rule that checks PR for long sentences.
It now treats words separated by `.` and `-` as one word.
Examples: 17.2.2 is one word, Umbraco.License is one word and case-sensitive is one word.

#8301


eshanrnh (approved):
Tested locally and works great💪

PR #8325, opened 2026-08-12
Clarify organization-level GitHub workflow permissions

erikjanwestendorp:
## 📋 Description

While implementing the GitHub Actions workflow for Umbraco Cloud, I found that the **Read and write permissions** option was disabled at the repository level because workflow permissions were managed at the organization level.

This PR adds a note explaining this possibility and indicates where the setting can be reviewed or updated in the organization settings.

## Product & Version 
Umbraco Cloud


sofietoft (approved):
Thanks for adding this note @erikjanwestendorp ! 🙏

PR #8344, opened 2026-08-18
Keep the default variant separate in the Engage dedup script

nathanwoulfe:
## 📋 Description

The `DeduplicatePageVariants.sql` script consolidates duplicate page variant rows, and the schema alignment guide instructs customers to run it. In its published form the script also destroys the record of which A/B test variant a pageview belonged to. This change corrects the two segment comparisons responsible. The deduplication itself is unchanged.

### What was wrong

Umbraco Engage now records which variant was served on each pageview. An empty `segment` means the original variant. `NULL` means the row predates that tracking. The two values are deliberately different.

Both comparisons in the script collapsed them together:

```sql
ISNULL(dupVariant.[segment], '') = ISNULL(canonical.[segment], '')
```

The `canonical` subquery groups `NULL` and `''` as separate rows, because `segment` sits in the `GROUP BY`. The join then treats them as equal. Every variant row therefore matched both canonical rows.

The result is not a merge. It is a swap. Pre-fix pageviews move onto the row that reports as the original, and original pageviews move onto the row that reports as pre-fix. Step 2 then deletes the row left without pageviews. Bucket totals stay plausible afterwards, so the damage is not visible in reporting.

A/B test variant rows are unaffected, because their segment is a real alias that matches nothing else. Only the original is lost. The script repoints pageviews rather than copying them, so the change cannot be reversed.

### What changed

Both segment comparisons now use a null-safe form that keeps `NULL` and `''` apart:

```sql
(dupVariant.[segment] = canonical.[segment]
 OR (dupVariant.[segment] IS NULL AND canonical.[segment] IS NULL))
```

The same correction applies to the `EXISTS` clause in Step 2, which guards the `DELETE`.

The `contentTypeId` comparison keeps its `ISNULL(..., -1)` form. That column is never `-1`, so no valid value collapses.

A note in the header records why the two values must stay separate. The trap is invisible from the SQL alone.

### One caveat this fix cannot remove

Engage began recording the served segment in 17.4. Earlier releases also wrote an empty segment, but it carried no meaning then. Traffic collected on 17.0, 17.1, or 16.0 to 16.2 therefore sits in rows that look identical to genuine original-variant rows.

Nothing distinguishes the two, so the script groups them together. Original-variant figures on installations with that history include some traffic predating variant tracking. This was already true before this change, and no script can separate them retrospectively.

The schema alignment guide now states this in the deduplication step, and the script header repeats it for anyone reading the SQL on its own. Figures for traffic collected on 17.4 and later are unaffected.

### Testing

The script was run against SQL Server 2025 on a scratch database. The two tables were rebuilt from their real column definitions. The scenario seeds one node with:

* one pre-fix row, `segment` `NULL`, holding 7 pageviews
* two original rows, `segment` `''`, holding 3 pageviews between them
* two A/B variant rows, holding 3 pageviews between them

Bucket sizes differ on purpose, so a swap cannot be mistaken for a correct result. Each pageview carries a tag recording where it started, so it stays traceable through the reassignment.

| Script | `NULL` bucket | `''` bucket | variant bucket |
|---|---|---|---|
| seeded | 7 pre-fix | 3 original | 3 ab-variant |
| before this change | 7 **original** | 3 **pre-fix** | 3 ab-variant |
| after this change | 7 pre-fix | 3 original | 3 ab-variant |

After the change, duplicates collapse correctly, three variant rows remain, no pageview is orphaned, and no bucket holds pageviews from more than one origin.

The published file itself was executed, not a copy of it.

Tracked in the Umbraco Engage internal tracker. There is no public issue.

## ✅ Contributor Checklist

* [x] Code blocks are correctly formatted.
* [x] Sentences are short and clear (preferably under 25 words).
* [x] Passive voice and first-person language (“we”, “I”) are avoided.
* [x] Relevant pages are linked.
* [ ] All links work and point to the correct resources. — no links added or changed
* [ ] Screenshots or diagrams are included if useful. — not applicable
* [x] Any code examples or instructions have been tested.
* [ ] Typos, broken links, and broken images are fixed. — not applicable

Umbraco Engage. Documentation versions 17 and 18, where this script is published. All four changed files remain identical across the two versions.

The script is referenced from `upgrading/schema-alignment-guide.md`, which instructs readers to run it and embeds the file. That guide gains the caveat described above.

## Deadline (if relevant)

Before the next Engage release that records the served segment on a pageview.

Until that release, the `segment` column is always `NULL` in shipped builds, so the script cannot do any harm. Once it ships, anyone following the schema alignment guide destroys their own original-variant data.

### DIFF
--- a/17/umbraco-engage/scripts/DeduplicatePageVariants.sql
+++ b/17/umbraco-engage/scripts/DeduplicatePageVariants.sql
@@ -9,6 +9,11 @@
 -- NOTE: Run this during a maintenance window or low-traffic period.
 -- The UPDATE on umbracoEngageAnalyticsPageview can take a while on
 -- installations with a large number of pageviews.
+--
+-- NOTE: An empty segment is the default (original) variant, and is deliberately
+-- distinct from NULL, which marks rows written before the served segment was
+-- recorded. The comparisons below must not collapse the two together: this
+-- script repoints pageviews and deletes rows, and cannot be undone.
 -- =============================================================================
 
 -- Step 1: Reassign pageviews from duplicate variants to the canonical variant
@@ -29,7 +34,8 @@ INNER JOIN (
 ) canonical
     ON dupVariant.[nodeId] = canonical.[nodeId]
     AND dupVariant.[culture] = canonical.[culture]
-    AND ISNULL(dupVariant.[segment], '') = ISNULL(canonical.[segment], '')
+    AND (dupVariant.[segment] = canonical.[segment]
+         OR (dupVariant.[segment] IS NULL AND canonical.[segment] IS NULL))
     AND ISNULL(dupVariant.[contentTypeId], -1) = ISNULL(canonical.[contentTypeId], -1)
 WHERE dupVariant.[id] <> canonical.[id];
 
@@ -44,7 +50,8 @@ AND EXISTS (
     SELECT 1 FROM [umbracoEngageAnalyticsUmbracoPageVariant] canonical
     WHERE canonical.[nodeId] = v.[nodeId]
     AND canonical.[culture] = v.[culture]
-    AND ISNULL(canonical.[segment], '') = ISNULL(v.[segment], '')
+    AND (canonical.[segment] = v.[segment]
+         OR (canonical.[segment] IS NULL AND v.[segment] IS NULL))
     AND ISNULL(canonical.[contentTypeId], -1) = ISNULL(v.[contentTypeId], -1)
     AND canonical.[id] < v.[id]
 );
--- a/17/umbraco-engage/scripts/DeduplicatePageVariants.sql
+++ b/17/umbraco-engage/scripts/DeduplicatePageVariants.sql
@@ -14,6 +14,10 @@
 -- distinct from NULL, which marks rows written before the served segment was
 -- recorded. The comparisons below must not collapse the two together: this
 -- script repoints pageviews and deletes rows, and cannot be undone.
+--
+-- NOTE: Traffic collected on 17.0, 17.1, or 16.0 to 16.2 also carries an empty
+-- segment, written before variant tracking existed. Those rows cannot be told
+-- apart from genuine original-variant rows, so this script groups them together.
 -- =============================================================================
 
 -- Step 1: Reassign pageviews from duplicate variants to the canonical variant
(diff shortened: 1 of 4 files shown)


nathanwoulfe:
Running the fixed script on a pre-17.4 database is harmless. It still collapses the duplicates it was written to collapse. On a database with 17.0/17.1 history you end up with two "untracked" rows instead of one merged row where pageview totals preserved, nothing orphaned, nothing deleted that shouldn't be. If anyone later decides those two should be one, they can still merge them; the old script's version of that decision couldn't be undone.

So there's no version gate on the fix itself. It's safe to ship to 17 and 18 now, and safe for customers to run at any version. The urgency was only ever about landing it before 17.4 makes the old script destructive. That still holds, but there's no downside to it going in early.

The one thing neither fix can repair: on a site that ran 17.0 or 17.1, the '' rows from that era are indistinguishable from genuine original-variant rows written on 17.4+. Those sites' "original variant" figures permanently include some pre-tracking traffic.

sofietoft (approved):

sofietoft:
Thanks @nathanwoulfe !
Just to confirm: This can be published anytime, right?

nathanwoulfe:
@sofietoft I'd like @cornehoskam to cast an eye over it first, given the nature of the change. If he's happy, then yes, can be merged.

sofietoft:
Ah, excellent! I'll await Corné's approval then 💪

17/umbraco-engage/scripts/DeduplicatePageVariants.sql
```diff
@@ -9,6 +9,15 @@
 -- NOTE: Run this during a maintenance window or low-traffic period.
 -- The UPDATE on umbracoEngageAnalyticsPageview can take a while on
 -- installations with a large number of pageviews.
+--
+-- NOTE: An empty segment is the default (original) variant, and is deliberately
+-- distinct from NULL, which marks rows written before the served segment was
+-- recorded. The comparisons below must not collapse the two together: this
+-- script repoints pageviews and deletes rows, and cannot be undone.
+--
+-- NOTE: Traffic collected on 17.0, 17.1, or 16.0 to 16.2 also carries an empty
```
cornehoskam, line 18: @nathanwoulfe Wasn't it the case that we were collecting NULL and not an empty segment before? Other than that, looks good to me!

nathanwoulfe: Nah, trace says we were storing an empty segment, which means no distinction between empty and null, which is what this change fixes.

nathanwoulfe:
@sofietoft when you can, this is approved and merge-ready.

PR #8350, opened 2026-08-18
Update clickjackingprotection.md

### LINKED ISSUE #8334: Use header assignment instead of Add in the clickjacking protection example
Discussion

The [article](https://docs.umbraco.com/umbraco-cms/run-in-production/infrastructure-and-ops/health-check/guides/clickjackingprotection) currently uses `Add` to set the `X-Frame-Options` response header:

```csharp
context.Response.Headers.Add("X-Frame-Options", "SAMEORIGIN");
```

Using `Add` can throw an exception when the `X-Frame-Options` header has already been added by another middleware or component.

I suggest using assignment instead:

```csharp
context.Response.Headers["X-Frame-Options"] = "SAMEORIGIN";
```

Alternatively, the strongly typed property could be used:

```csharp
context.Response.Headers.XFrameOptions = "SAMEORIGIN";
```

Both alternatives replace an existing value, ensuring that the header has one unambiguous value without throwing an exception when it has already been configured.

If this is considered a desirable change, I would be happy to submit a PR to update the documentation.

erikjanwestendorp:
Fixes #8334

### DIFF
--- a/17/umbraco-cms/run-in-production/infrastructure-and-ops/health-check/guides/clickjackingprotection.md
+++ b/17/umbraco-cms/run-in-production/infrastructure-and-ops/health-check/guides/clickjackingprotection.md
@@ -26,10 +26,22 @@ app.UseXfo(options => options.SameOrigin());
 
 Avoid third-party library dependency by using custom middleware added to the request pipeline.
 
+The strongly typed `XFrameOptions` property can be used to set the header:
+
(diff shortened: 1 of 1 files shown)


sofietoft (approved):
Looks great @erikjanwestendorp ! 🙏 

Thanks for helping us keep the docs up to speed ⭐

PR #8351, opened 2026-08-19
Document ApplicationUrlDetection setting introduced in Umbraco CMS 17.4

sofietoft:
## 📋 Description

Adds a new sub-section covering the `ApplicationUrlDetection` configuration option and updates the existing `UmbracoApplicationUrl` guidance to reflect that auto-detection from HTTP requests is disabled by default from v17.4 (umbraco/Umbraco-CMS#22307).                                                                                    
                                                                                                                                                                                                           
  - Updates the UmbracoApplicationUrl guidance to clarify that auto-detection from HTTP requests is disabled by default from v17.4 (previously it would fall back to detecting the domain from the HTTP    
  request object)                                                                                                                                                                                          
  - Adds a new "Configuring ApplicationUrlDetection" sub-section documenting the three available modes: None (default), FirstRequest, and EveryRequest           

umbraco/Umbraco-CMS#22307

CMS 17.4

## Deadline (if relevant)

ASAP.

### DIFF
--- a/marketplace-and-integrations/configure-licenses.md
+++ b/marketplace-and-integrations/configure-licenses.md
@@ -104,7 +104,7 @@ The website domain used for validating the license is determined from your Umbra
 
 If you are running on a single domain for both your frontend and backend environments, it's not necessary to configure a `UmbracoApplicationUrl`.
 
-If you have different domains for your frontend and backend, then it's advised that you configure an `UmbracoApplicationUrl` set to your backoffice URL. This helps the licensing engine know which URL should be used for validation checks. Without this configuration setting, the licensing engine cannot reliably determine the domain to use for validation. In versions prior to 17.4, it would attempt to detect the domain from the HTTP request object. Detecting the domain from the HTTP request object could lead to errors when switching between domains. From version 17.4, auto-detection is disabled by default, so the `UmbracoApplicationUrl` must be set explicitly.
+If you have different domains for your frontend and backend, configure an `UmbracoApplicationUrl` set to your backoffice URL. This helps the licensing engine know which URL should be used for validation checks. Without this configuration setting, the licensing engine cannot reliably determine the domain to validate. In earlier versions of Umbraco, the domain was auto-detected from the incoming request. From version 17.4, auto-detection is opt-in.
 
 ### How to Configure UmbracoApplicationUrl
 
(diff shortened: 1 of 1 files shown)


marketplace-and-integrations/configure-licenses.md
```diff
@@ -124,6 +124,36 @@ An `UmbracoApplicationUrl` can be configured in your `appsettings.json` file:
 
 See the [Fixed Application URL](https://docs.umbraco.com/umbraco-cms/extending/health-check/guides/fixedapplicationurl) documentation for more details about this setting.
 
+### Configuring ApplicationUrlDetection
```
AndyButland, line 127: This is already documented here: https://docs.umbraco.com/umbraco-cms/develop-with-umbraco/configuration/webroutingsettings#application-url-detection - so I don't think this repetition is necessary.  Should cross-link instead.

marketplace-and-integrations/configure-licenses.md
```diff
@@ -104,7 +104,7 @@ The website domain used for validating the license is determined from your Umbra
...
 
-If you have different domains for your frontend and backend, then it's advised that you configure an `UmbracoApplicationUrl` set to your backoffice URL. This helps the licensing engine know which URL should be used for validation checks. Without this configuration setting, the licensing engine will try and work out the domain to validate from the HTTP request object. This can lead to errors when switching between domains.
+If you have different domains for your frontend and backend, then it's advised that you configure an `UmbracoApplicationUrl` set to your backoffice URL. This helps the licensing engine know which URL should be used for validation checks. Without this configuration setting, the licensing engine cannot reliably determine the domain to use for validation. In versions prior to 17.4, it would attempt to detect the domain from the HTTP request object. Detecting the domain from the HTTP request object could lead to errors when switching between domains. From version 17.4, auto-detection is disabled by default, so the `UmbracoApplicationUrl` must be set explicitly.
```
AndyButland: I suggest here that it would be sufficient to say:

... Without this configuration setting, the licensing engine cannot reliably determine the domain to validate. Earlier versions of Umbraco would attempt to detect the domain from the incoming request, but that behaviour became an optional configuration after 17.4.

sofietoft:
Thanks for the review @AndyButland ! 💪 
I've referenced the CMS configuration doc instead of duplicating it here and applied your other suggestion as well.

Let me know what you think!

AndyButland (approved):
All looks good to me @sofietoft - thanks.

PR #8421, opened 2026-09-03
Engage: updates data consistency SQL 

nathanwoulfe:
Updates the EnsureDataConsistency SQL script to add guards on `projectId` and `goalId` columns, as these are dropped in new versions of Engage. Adding the guard keeps the SQL backwards compatible rather than having to manage different versions.

This change is NOT dependent on the corresponding updates in Engage being released.

@cornehoskam FYI - this is the same change as made in the corresponding files within Engage.


sofietoft (approved):

sofietoft:
Is there a deadline on this PR, @nathanwoulfe?

nathanwoulfe:
Hey @sofietoft no deadline, but given Corné's thumb-of-approval on the comment, it's fine to merge whenever suits you.

cornehoskam:
@nathanwoulfe @sofietoft Oh actually, I just spotted that the EnsureDataConsistency.sql file has been moved by @sofietoft two days ago, and now lives in 17/umbraco-engage/.gitbook/assets/EnsureDataConsistency.sql (and 18/). I think that's the one that we're referencing for download now. Can you confirm Sofie? If so then this PR should update those files instead (and probably remove the old files) 🙂

sofietoft:
Hej! I think this is related to GitBook moving all assets (images, etc.) to the shared gitbook/assets directory for each docs site.
If you could update the files in the assets directory instead, that would be great!
And the scripts folder will most likely not be used no, so feel free to remove that.

I assume there are also some references that need to be updated then, right?

